PR #89, opened 2019-02-28
Use pkg_tar bazel rules instead of genrules.

BenBirt:


BenBirt:
I've spent a bit of time on this and have no idea what is going on here. The tar files have the same contents, same names, etc.

The only substantive difference I can find:
bens-mbp:dataform benbirt$ tar -tf bazel-bin/crossdb/package_tar.tgz
./
./package/
./package/index.d.ts
./package/index.js
./package/package.json
bens-mbp:dataform benbirt$ tar -tf bazel-genfiles/crossdb/package.tgz
package/
package/index.js
package/package.json
package/index.d.ts

(i.e. the genrule version somehow doesn't have ./ prefixed on the contents, but that shouldn't make any difference, indeed if I untar them:)

bens-mbp:Desktop benbirt$ ls -la package 
total 24
dr-xr-xr-x  5 benbirt  staff   160  1 Jan  1970 .
drwx------+ 8 benbirt  staff   256  1 Mar 10:01 ..
-r-xr-xr-x  1 benbirt  staff   117  1 Jan  1970 index.d.ts
-r-xr-xr-x  1 benbirt  staff  1866  1 Jan  1970 index.js
-r-xr-xr-x  1 benbirt  staff   281  1 Jan  1970 package.json
bens-mbp:Desktop benbirt$ ls -la "package 2"
total 24
drwxr-xr-x  5 benbirt  staff   160  1 Mar 10:00 .
drwx------+ 8 benbirt  staff   256  1 Mar 10:01 ..
-r-xr-xr-x  1 benbirt  staff   117  1 Mar 09:24 index.d.ts
-r-xr-xr-x  1 benbirt  staff  1866  1 Mar 09:24 index.js
-r-xr-xr-x  1 benbirt  staff   281  1 Mar 09:24 package.json

(i.e. the only difference is that the pkg_tar rule sets the correct mtime.)

In the interest of getting things working I'll send you a separate PR just to fix the genrule command.

Issue #59, opened 2019-01-11
Move validation into a standalone operation


lewish:
Currently validation happens in a number of different places, during compilation, but also during API calls to the core framework.

As a result, it's hard/impossible for the other parts of the library to validate a graph (for example, if we want to run validation before we build or run a graph).

We should untangle the current validation code into a single methods that takes a CompiledGraph proto and adds returns a copy with validation errors added appropriately.

ShevtsovY:
summarize:

1.move compilationErrors from Graph to GraphErrors (GraphErrors will store in Graph)
2.remove validationErrors from Graph
3.remove validationErrors from Table
4.during the compilation process we will create only compilationErrors
5.during the compilation process we will NOT do any validation checks
6.add validate() function to utils, which will return GraphErrors:
```{
  compilationErrors, // from Graph
  validationErrors   // check Graph and Tables
}```

PR #94, opened 2019-03-05
#59: Move validation into a standalone operation

ShevtsovY:


protos/core.proto
```diff
@@ -36,12 +36,17 @@ message RedshiftOptions {
...
+
 message ValidationError {
-  string file_name = 1;
+  string node_name = 1;
   string message = 2;
 }
 
-message CompileError {
+message CompilationError {
   string file_name = 1;
```
BenBirt: should this be node_name too?

ShevtsovY: > should this be node_name too?
no, this is filename - link to file with error!

Issue #90, opened 2019-02-28
Validate contents of profile json files


lewish:
When we read connection profiles from json files we don't do any validation on them, we just run JSON.parse and cast it to the proto type. This can lead to confusing and hard to debug issues.

We should validate the profile when we read it, checking that one of the supported warehouse types is set, and that for that warehouse, the minimum set of fields are provided.

---
Moved from dataform-co/dataform-co#170
/cc @lewish

PR #100, opened 2019-03-06
#90: Validate contents of profile json files

ShevtsovY:


cli/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+import * as protos from "@dataform/protos";
+
+export enum WarehouseTypes {
+  BIGQUERY = "bigquery",
+  REDSHIFT = "redshift",
+  SNOWFLAKE = "snowflake"
+}
+
+export function readProfile(profilePath: string): protos.IProfile {
+  return fs.existsSync(profilePath) ? JSON.parse(fs.readFileSync(profilePath, "utf8")) : null;
```
lewish: Think it would be good to split out the missing file error here.

cli/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export function validateProfile(profile: protos.IProfile): void {
+  // profile shouldn't be empty
+  if (!profile || !Object.keys(profile).length) {
+    throw new Error("Missing profile JSON file or file is empty.");
```
lewish: Then here we can just say - Profile file is empty.

lewish: (And in both these errors, let's say what the file path we are trying to read is)

cli/utils.ts
```diff
@@ -0,0 +1,40 @@
+import * as fs from "fs";
+import * as protos from "@dataform/protos";
+
+export enum WarehouseTypes {
```
lewish: Can we put this somewhere else, as we are probably going to use it elsewhere? I imagine in core/adapters/index.js is probably a good place to put it

cli/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  // warehouse check
+  const warehouses = Object.keys(WarehouseTypes).map(key => WarehouseTypes[key]);
+  if (!Object.keys(profile).every(key => warehouses.indexOf(key) !== -1)) {
+    const predefinedW = warehouses.map(item => `"${item}"`).join(" | ");
+    throw new Error(`Unsupported warehouse detected. Should only use predefined warehouses: ${predefinedW}`);
+  }
+
+  // props check
+  const err = protos.Profile.verify(profile);
```
lewish: Nice! Didn't know verify existed, this is quite useful.

protos/profiles.proto
```diff
@@ -20,24 +20,24 @@ message Profile {
 }
 
 message JDBC {
-  string host = 2;
-  int64 port = 3;
-  string user = 4;
-  string password = 5;
-  string database = 6;
+  required string host = 2;
```
lewish: As we are using proto3 syntax, I don't think this is valid. On the typescript side, Protobuf.js compiler doesn't care I imagine, but for other languages using the proper proto compiler I suspect this will fail.

You can check by running:
```
bazel build //protos:java
```
The above should fail.

cli/utils.ts
```diff
@@ -0,0 +1,40 @@
+import * as fs from "fs";
```
lewish: These utils should probably go in the API, as we will use them from places where we only depend on the API and not the CLI.
Similarly above - the actual verification of the profiles should happen in the API package, not just the CLI.

ShevtsovY: but in the API package we have already a profile instance, so the validation has already been passed.
or do you mean that we need to pass  only the path in the API package, and then read the file and perform the verification inside the API package?

lewish: we can still pass in a profile instance rather than a filename - but it being correct depends on us being good Typescript developers! As a JS developer, I could of course just pass in any object I wanted, or just cast any old object in Typescript even.

api/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export function validateProfile(profile: protos.IProfile): void {
+  // profile shouldn't be empty
+  if (!profile || !Object.keys(profile).length) {
+    throw new Error("Profile JSON file is empty.");
+  }
+
+  // warehouse check
+  const supurtedWarehouses = Object.keys(WarehouseTypes).map(key => WarehouseTypes[key]);
```
lewish: supurted > supported

api/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+  const warehouse = warehouses[0];
+  const props = Object.keys(profile[warehouse]);
+  const missingProps = requiredWarehouseProps[warehouse].filter(key => props.indexOf(key) === -1);
+
+  if (missingProps.length > 0) {
+    throw new Error(`Missing required properties: ${missingProps.join(", ")}`);
+  }
+}
+
+export function getProfile(profilePath: string): protos.IProfile {
```
lewish: nit: `getProfile` is a strange function name, I think this should probably called `readProfile`, and you can just merge this with the existing `readProfile` function

PR #101, opened 2019-03-07
Fix node module fixtures

lewish:
Yarn was doing strange things. Go back to NPM which behaves as we want it to (flat node_modules) and a ton of hacks to make it work. This *should* be consistent across platforms/builds.

@BenBirt if you could test this on MacOS, that would be great!

### DIFF
--- a/tools/node_modules.bzl
+++ b/tools/node_modules.bzl
@@ -4,24 +4,44 @@ def _impl(ctx):
     tree = ctx.actions.declare_directory(ctx.attr.name)
     folder_path = tree.path.replace("/node_modules", "")
 
-    # Set a unique yarn cache folder for this rule as sharing a folder between
-    # different calls causes errors.
-    # In the future, we should ideally rm -rf this directory first, or disable yarn cache entirely.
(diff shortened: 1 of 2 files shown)


tools/node_modules.bzl
```diff
@@ -4,24 +4,44 @@ def _impl(ctx):
...
-    # In the future, we should ideally rm -rf this directory first, or disable yarn cache entirely.
-    arguments = ["--cwd", folder_path, "--cache-folder", "/tmp/.yarn-bazel-cache/" + folder_path, "add"]
-
-    # Running with cwd means we need to go back up dirs for paths to resolve.
-    root_path = "/".join([".." for f in folder_path.split("/")])
+    #   NPM refuses to install into node_modules directly as it's a strange
+    #       symlinky virtual file system.
+    #   Similarly, it won't use a cache folder in this directory, so we stage
+    #       things in /tmp folders then copy back across.
+    #   We override the install dirwith --prefix, changing directory
```
BenBirt: "dir with"

PR #104, opened 2019-03-12
Fixes bigquery db adapter to have cancellable queries and use the promise pool

lewish:
…ancellation properly, adds a related test but doesn't actually test the bigquery adapter properly yet. #dataform-co/dataform-tracking#122


api/dbadapters/bigquery.ts
```diff
@@ -27,33 +27,32 @@ export class BigQueryDbAdapter implements DbAdapter {
...
 
   execute(statement: string) {
-    return Bluebird.resolve().then(() =>
+    return new Bluebird<any[]>((resolve, reject, onCancel) => {
+      let isCancelled = false;
+      const eEmitter = new EventEmitter();
+
+      onCancel(() => {
+        isCancelled = true;
+        eEmitter.emit("jobCancel");
```
BenBirt: pull out this constant "jobCancel" string since want to ensure the same value is used everywhere for this event?

tests/api/api.spec.ts
```diff
@@ -745,5 +744,60 @@ describe("@dataform/api", () => {
...
+          onCancel(() => {
+            wasCancelled = true;
+          });
+        })
+      );
+
+      const runner = new Runner(instance(mockedDbAdapter), TEST_GRAPH);
+      const promise = runner.execute();
+      // Push this to the back of the "queue" so we can await the promise first.
+      setTimeout(() => runner.cancel(), 10);
```
BenBirt: i don't think i understand what this is doing, can you clarify? it seems like the intention is to allow execution to be running (in the background, triggered by runner.execute()), give it "some time" (10ms), and then cancel? what's the reasoning for waiting 10ms?

tests/api/api.spec.ts
```diff
@@ -745,5 +744,60 @@ describe("@dataform/api", () => {
...
+      when(mockedDbAdapter.execute(TEST_GRAPH.nodes[0].tasks[0].statement)).thenReturn(
+        new Bluebird((resolve, reject, onCancel) => {
+          onCancel(() => {
+            wasCancelled = true;
+          });
+        })
+      );
+
+      const runner = new Runner(instance(mockedDbAdapter), TEST_GRAPH);
+      const promise = runner.execute();
```
BenBirt: nit: call this "execution"?

tests/api/api.spec.ts
```diff
@@ -745,5 +744,60 @@ describe("@dataform/api", () => {
...
+      // Push this to the back of the "queue" so we can await the promise first.
+      setTimeout(() => runner.cancel(), 10);
+      try {
+        const result = await promise;
+      } catch (e) {
+        // Expected.
+        expect(wasCancelled).is.true;
+        return;
+      }
+      throw new Error("Cancelled run did not throw");
```
BenBirt: i think this'd be cleaner as:

expect(() => { await promise; }).to.throw(Error);

tests/api/api.spec.ts
lewish: I've added a better comment here hopefully.
If we don't wait the promise never actually starts, the onCancel handler above never get's set up so we don't test what we actually want to.

tests/api/api.spec.ts
lewish: Refactored all of this now.

api/commands/run.ts
```diff
@@ -101,11 +100,12 @@ export class Runner {
...
       let finishedDeps = node.dependencies.filter(d => allFinishedDeps.indexOf(d) >= 0);
       let successfulDeps = node.dependencies.filter(d => allSuccessfulDeps.indexOf(d) >= 0);
-      if (successfulDeps.length == node.dependencies.length) {
+      if (!this.cancelled && successfulDeps.length == node.dependencies.length) {
         // All required deps are completed, start this node.
         this.executeNode(node);
-      } else if (finishedDeps.length == node.dependencies.length) {
-        // All deps are finished but they weren't all successful, skip this node.
+      } else if (this.cancelled || finishedDeps.length == node.dependencies.length) {
+        // All deps are finished but they weren't all successful, or the run was cancelled.
```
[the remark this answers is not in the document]
lewish: Yeaahhh, maybe - cancellation status is more of a state of the run than a node. Think it's fine to mark them as skipped, the node that actually get's interrupted (if there is any) will show as state failed. We could mark that as cancelled instead.

Issue #88, opened 2019-02-22
Implement inline tables


lewish:
Inline tables shouldn't get created in your data warehouse but instead when referenced, just inject the SQL inline into the calling query.

This makes managing particularly large queries a little easier by breaking them up into multiple SQL files.

For example:

`definitions/inline_table.sql`:
```
--js type("inline");
select * from ${ref("source_table")}
```

Can be used inside a normal table:

`definitions/table.sql`:
```
select * from ${ref("inline_table")}
where true
```

The resulting compiled SQL should be:

```
select * from (
select * from "default_schema"."source_table"
)
where true
```

Any dependencies in the inline table should be passed along to the calling table, in this case `source_table`.

The inline tables should get compiled, and are present in the `CompiledGraph` output, but will never be created and should be filtered out during the `build` step.

- [x] Implement inline tables
- [x] Write tests
- [ ] Update documentation

PR #105, opened 2019-03-14
#88: Implement inline tables

ShevtsovY:


package.json
```diff
@@ -65,6 +65,7 @@
   },
   "scripts": {
     "build": "./scripts/build",
+    "clean": "rm -rf /tmp/.yarn-bazel-cache/ && bazel clean",
```
lewish: You shouldn't need this anymore, I fixed the underlying issue

ShevtsovY: ok, I will remove it

core/utils.ts
```diff
@@ -220,6 +232,18 @@ export function validate(compiledGraph: protos.ICompiledGraph): protos.IGraphErr
         });
       });
     }
+
+    // ignored properties in tables
+    if (!!ignoredProps[node.type]) {
+      ignoredProps[node.type].forEach(key => {
+        if (isPropertyExist(node[key])) {
```
BenBirt: is this `node.hasOwnProperty(key)`?

core/utils.ts
```diff
@@ -220,6 +232,18 @@ export function validate(compiledGraph: protos.ICompiledGraph): protos.IGraphErr
         });
       });
     }
+
+    // ignored properties in tables
+    if (!!ignoredProps[node.type]) {
+      ignoredProps[node.type].forEach(key => {
```
BenBirt: `ignoredProp` instead of `key` please

ShevtsovY: ok

core/utils.ts
ShevtsovY: not quite clear what you mean?! in this function we check for empty arrays and empty objects...

PR #106, opened 2019-03-14
Fixes #98. Support overriding the project's default/assertion schema …

### LINKED ISSUE #98: Support overriding the dataform schemas via the dataform API and CLI
Add the ability to override both the default_schema and assertion_schema project settings defined in the ProjectConfig proto message: https://github.com/dataform-co/dataform/blob/master/protos/core.proto#L7

Support overriding these settings through the following API methods:
- `compile()`
- `build()`
- `run()`

Also add support to the CLI for overriding these settings via the CLI through new flags:
`--default-schema` and `--assertion-schema`

---
Moved from dataform-co/dataform-co#192
/cc @lewish

BenBirt:
…names.

### DIFF
--- a/api/commands/compile.ts
+++ b/api/commands/compile.ts
@@ -1,40 +1,69 @@
 import * as protos from "@dataform/protos";
 
-import { fork } from "child_process";
+import { ChildProcess, fork } from "child_process";
 import * as fs from "fs";
 import { promisify } from "util";
 import * as path from "path";
+import { ICompileIPCParameters, ICompileIPCResult } from "../vm/compile";
 
-export function compile(projectDir: string): Promise<protos.CompiledGraph> {
+export async function compile(
+  projectDir: string,
+  defaultSchemaOverride?: string,
+  assertionSchemaOverride?: string
+): Promise<protos.CompiledGraph> {
   // Resolve the path in case it hasn't been resolved already.
   projectDir = path.resolve(projectDir);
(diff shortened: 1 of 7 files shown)


cli/index.ts
```diff
@@ -6,6 +6,15 @@ import * as chokidar from "chokidar";
 import * as protos from "@dataform/protos";
 import { init, compile, build, run, table, query, utils } from "@dataform/api";
 
+const addCompileYargs = (yargs: yargs.Argv) =>
+  yargs
+    .option("default-schema-override", {
```
lewish: I don't think we need -override suffix, adds a lot of length, feel like it's fairly clear what this is going to do?

api/commands/compile.ts
```diff
@@ -1,40 +1,68 @@
...
+import { ChildProcess, fork } from "child_process";
 import * as fs from "fs";
 import { promisify } from "util";
 import * as path from "path";
+import { ICompileIPCParameters, ICompileIPCResult } from "../vm/compile";
 
-export function compile(projectDir: string): Promise<protos.CompiledGraph> {
+export async function compile(
+  projectDir: string,
+  defaultSchemaOverride?: string,
```
lewish: It's inevitably a short matter of time until we need to add more things here. Would be tempted to but them into some CompileOptions interface and pass that as the second param. This is going to have to be serialized imminently too, might want a proto type to hold these

PR #111, opened 2019-03-19
Fix cancellable promise implementation and add tests

lewish:


api/utils/cancellable_promise.ts
```diff
@@ -1,23 +1,35 @@
...
-  
-    public cancel(): void {
-      this.emitter.emit(CancellablePromise.CANCEL_EVENT);
-    }
-  }
\ No newline at end of file
+export class CancellablePromise<T> implements PromiseLike<T> {
+  private static CANCEL_EVENT = "cancel";
+
+  private emitter = new EventEmitter();
```
BenBirt: all three of these members could presumably be readonly

tests/api/utils/cancellable_promise.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { CancellablePromise } from "@dataform/api/utils/cancellable_promise";
+import { expect } from "chai";
+
+describe("@dataform/api/utils/cancellable_promise", () => {
```
BenBirt: just cancellable_promise is probably enough in this string!

tests/api/utils/cancellable_promise.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+  it("cancel_is_called", () => {
```
BenBirt: these don't need underscores, the idea is that they form some meaningful phrase when all glued together

Issue #76, opened 2019-01-24
Assertions should be wrapped in a sum(1) query instead of selecting actual rows


lewish:
If an assertion returns many rows, they will be returned to the query runner. If there are enough rows, this will be either slow or cause OOMs.

Currently we run a query like:

`select 1 as detect from ... `

Instead we should run:

`select sum(1) as rows from ... `

Finally, instead of the test looking at the number of returned rows, it should inspect the `rows` field of the first and only row, and see if it is > 0.

lewish:
Moved to dataform-co/dataform-tracking#132

PR #112, opened 2019-03-19
#76: Assertions wrapped in a count(*) query instead of selecting actual rows

ShevtsovY:


api/commands/run.ts
```diff
@@ -144,8 +144,9 @@ export class Runner {
             const rows = await this.adapter.execute(task.statement, handleCancel =>
               this.eEmitter.on(CANCEL_EVENT, handleCancel)
             );
+            const rowsNumber = !!rows.length ? rows[0].rowsNumber : 0;
```
BenBirt: rowCount?

core/adapters/bigquery.ts
```diff
@@ -43,7 +43,7 @@ export class BigQueryAdapter extends Adapter implements IAdapter {
       name: a.name
     });
     tasks.add(Task.statement(this.createOrReplaceView(assertionTarget, a.query)));
-    tasks.add(Task.assertion(`select 1 as detect from ${this.resolveTarget(assertionTarget)}`));
+    tasks.add(Task.assertion(`select count(*) as rows_number from ${this.resolveTarget(assertionTarget)}`));
```
BenBirt: rowCount (in the other queries too please)

api/commands/run.ts
ShevtsovY: ok

core/adapters/bigquery.ts
ShevtsovY: ok

PR #114, opened 2019-03-20
Clean up dataform API slightly to make skipping install() consistent with how it's skipped in init().

BenBirt:


api/commands/install.ts
```diff
@@ -2,6 +2,9 @@ import * as path from "path";
 import * as childProcess from "child_process";
 import { promisify } from "util";
 
-export function install(projectDir: string) {
+export function install(projectDir: string, skipInstall?: boolean) {
+  if (skipInstall) {
+    return;
```
lewish: `return Promise.resolve();` ?

BenBirt: Done except with await/async.

PR #119, opened 2019-03-25
Fix type-safety around redshift connection profiles. The code tried to pass a Long to the redshift API with no type errors being thrown until inside that API.

BenBirt:


api/dbadapters/index.ts
```diff
@@ -32,7 +32,7 @@ export function create(profile: protos.IProfile, warehouseType?: string): DbAdap
     return new registry[warehouseType](profile);
   }
   if (!!profile.bigquery) {
-    return new registry["bigquery"](profile);
+    throw new Error(profile.toString());
```
lewish: This doesn't look right...

protos/profiles.proto
```diff
@@ -19,10 +19,16 @@ message Profile {
 
 message JDBC {
   string host = 2;
-  int64 port = 3;
```
lewish: My understanding was these are backwards compatible, so you could just change this and save quite a bit of work here? https://developers.google.com/protocol-buffers/docs/proto#updating

No one has set a non 32-bit port number so we don't need to worry about that :)

api/dbadapters/index.ts
BenBirt: Oops, sorry, debugging stuff, didn't intend it to make it here.

protos/profiles.proto
BenBirt: Interesting, I didn't know this was safely change-compatible, good to know.

PR #145, opened 2019-03-29
Fixes #124. Loudly print compile/run errors to the console.

### LINKED ISSUE #124: Clean up CLI output
* Remove all console logs that aren't in CLI package
* Make sure that all CLI commands have meaningful preamble/postamble
* Make sure that any errors hidden in JSON graphs are loudly printed out

BenBirt:
Also: 
- Suggest commands when the user requests a non-existent one.
- Use nice output colours where useful.
- Require strict command setting.
- Set script name so that help output reads cleanly.

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -6,6 +6,7 @@ import * as chokidar from "chokidar";
 import * as fs from "fs";
 import * as path from "path";
 import * as yargs from "yargs";
+import { dataform } from "df/protos";
 
 const RECOMPILE_DELAY = 2000;
 
@@ -212,7 +213,7 @@ createYargsCli({
     {
       format: "build [project-dir]",
       description:
-        "Build the dataform project. Produces JSON output describing the execution graph.",
+        "Build the dataform project, using the current state of the configured warehouse to compute final SQL.",
       positionalOptions: [projectDirMustExistOption],
       options: [
         fullRefreshOption,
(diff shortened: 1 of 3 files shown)


cli/index.ts
```diff
@@ -110,21 +122,30 @@ createYargsCli({
...
+        result.filesWritten.forEach(file => console.log(file));
+        if (result.installedNpmPackages) {
+          console.log(commandOutputFmtString, "NPM packages successfully installed.");
+        }
+      }
     },
     {
       format: "compile [project-dir]",
       description:
         "Compile the dataform project. Produces JSON output describing the non-executable graph.",
```
G2H: - Where is the JSON output?
- What do we mean by non-executable graph?

cli/index.ts
```diff
@@ -137,63 +158,63 @@ createYargsCli({
...
+            });
+        }
       }
     },
     {
       format: "build [project-dir]",
       description:
-        "Build the dataform project. Produces JSON output describing the execution graph.",
-      positionalOptions: [projectDirOption],
+        "Build the dataform project, using the current state of the configured warehouse to compute final SQL.",
```
G2H: Could we say: "Build the dataform project using warehouse configuration to compute final SQL" ?

cli/index.ts
```diff
@@ -202,131 +223,131 @@ createYargsCli({
...
+          runner.cancel();
+        });
+        const executedGraph = await runner.resultPromise();
+        const prettyPrintedExecutedGraph = prettyJsonStringify(executedGraph);
+        console.log(prettyPrintedExecutedGraph);
+        if (!executedGraph.ok) {
+          executedGraph.nodes
+            .filter(node => node.status === dataform.NodeExecutionStatus.FAILED)
+            .forEach(node => {
+              console.log(errorOutputFmtString, `Execution failed on node "${node.name}":`);
```
G2H: Do we have a similar error message for skipped and disabled nodes?

BenBirt:
That merge shouldn't have happened so quickly, since you have outstanding comments. In the interests of leaving a trail: https://github.com/bobvanderlinden/probot-auto-merge/issues/210

cli/index.ts
BenBirt: * The output gets printed to the console.
* The non-executable graph is the compiled graph, i.e. all of the various references and so on expanded, but no wrapping SQL that would depend on the current state of the warehouse.

cli/index.ts
BenBirt: Added in PR I'm about to send to you

cli/index.ts
BenBirt: No - my opinion is that for skipped or disabled nodes, those were explicitly set by the user, so it makes no sense to error out in those cases.

PR #147, opened 2019-04-01
Add init-creds command to the CLI, a wizard guiding the user through creation of a credentials file.

BenBirt:
Also clean up lots of linter errors (now that we have more tslint checks).


cli/index.ts
```diff
@@ -126,21 +131,50 @@ createYargsCli({
...
+        "Creates a .df-credentials.json file for dataform to use when accessing your warehouse.",
+      positionalOptions: [warehouseOption, projectDirMustExistOption],
+      options: [],
+      processFn: argv => {
+        const credentials = () => {
+          switch (argv.warehouse) {
+            case "bigquery": {
+              return getBigQueryCredentials();
+            }
+            case "redshift": {
```
G2H: The redshift connector works with postgres as well. Since some users may want to use PG as well, would be great to add that option as well.

BenBirt: So generally I agree, though right now the dataform package doesn't really support (in the true meaning of the word) postgres - we should really add a WarehouseType entry, register the redshift dbadapter to support that case, etc. Would prefer to do that separately, as I'm not sure what the knock-on effects might be.

G2H: agreed

cli/index.ts
```diff
@@ -126,21 +131,50 @@ createYargsCli({
...
+        writeStdOut(commandOutput("Files written:"));
+        result.filesWritten.forEach(file => writeStdOut(file));
         if (result.installedNpmPackages) {
-          console.log(commandOutputFmtString, "NPM packages successfully installed.");
+          writeStdOut(commandOutput("NPM packages successfully installed."));
         }
       }
     },
+    {
+      format: "init-creds <warehouse> [project-dir]",
```
lewish: Why do we need to re-specify the warehouse at all if we can get it from the project dir? Seems like we could easily avoid having the user specify the warehouse twice here.

It looks like the project-dir is required, in which case i think it should be `<project-dir>` in the format string.

BenBirt: So, `project-dir` is not actually required, it defaults to the current directory if unset.

I suppose we could make `warehouse` optional too, and if it's unset, try to look in `project-dir/dataform.json`, would you prefer that?

lewish: Ahh ok, that argument is really more of an output folder for the file. Makes sense! I thought you were reading properties from the project but you aren't, I think I prefer the explicit approach for now, unless we decide to tie this command more strongly to a project dir.

BenBirt:
https://github.com/dataform-co/dataform/pull/149

PR #157, opened 2019-04-04
Updates to CLI, mostly to do with help invocation.

BenBirt:


cli/index.ts
```diff
@@ -526,3 +532,18 @@ function createOptionsChain(yargsChain: yargs.Argv, command: ICommand) {
...
+
+function fixArgvForHelp() {
+  // Obviously this is a massive hack.
+  // The outcome of this is that the following commands are interchangeable:
+  // $ dataform help run
+  // $ dataform --help run
+  // The problem is that yargs.help() only allows us to specify an alias for the "--help" built-in option (by default that alias is "help").
+  // But because "--help" is only an option, not a command, it appears to be impossible (?) to configure yargs to respond to "help" correctly
+  // (or at least, to correctly print help strings for commands; it happily prints a top-level help string).
+  const argvCopy = process.argv.slice(2);
```
lewish: Can this ever have length < 3?

BenBirt: No, see https://nodejs.org/docs/latest/api/process.html#process_process_argv for details (0th element is the path, 1th is the script, command line arguments then follow)

PR #159, opened 2019-04-05
Schema suffixes

lewish:
Change override schema behaviour to use a schema suffix option instead.


cli/index.ts
```diff
@@ -21,7 +21,7 @@ const errorOutput = (output: string) => coloredOutput(output, 91);
 const writeStdOut = (output: string) => process.stdout.write(output + "\n");
 const writeStdErr = (output: string) => process.stderr.write(output + "\n");
 
-const projectDirOption = {
+const projectDirOption: INamedOption<yargs.PositionalOptions> = {
```
BenBirt: i would argue that in the general case, we shouldn't specify types on variable declarations. the great thing about typescript's type system (in particular interfaces) is that instead of having to declare tight dependencies (e.g. objects and the interfaces they implement), the binding is much looser so that it's much simpler for functions to specify the shape of the parameters they receive.

in this case, projectDirOption shouldn't have to care that it is any specific type - the place that it matters is the function that accepts the whole ICli object.

this is particularly powerful because a single object can implement multiple interfaces trivially without having to declare that it does so *until the place that it actually matters*.

cli/index.ts
```diff
@@ -66,21 +66,15 @@ const includeDepsOption = {
   }
 };
 
-const defaultSchemaOption = {
-  name: "default-schema",
+const schemaSuffixOption: INamedOption<yargs.Options> = {
+  name: "schema-suffix",
   option: {
-    describe: "An optional default schema name override."
+    describe: "A suffix to add to any output schema names.",
```
BenBirt: suggest "A suffix to be appended to output schema names."

cli/index.ts
```diff
@@ -66,21 +66,15 @@ const includeDepsOption = {
   }
 };
 
-const defaultSchemaOption = {
-  name: "default-schema",
+const schemaSuffixOption: INamedOption<yargs.Options> = {
```
BenBirt: should this have a check() function checking that the provided string only contains alphanumerics and underscores?

protos/core.proto
```diff
@@ -11,6 +11,8 @@ message ProjectConfig {
   string default_schema = 2;
   string assertion_schema = 5;
 
+  string schema_suffix = 7;
```
BenBirt: do we expect users to use this in dataform.json? i'm worried it's a bit confusing. (why would the user set this, given the `default_schema` and `assertion_schema` settings?)

perhaps the answer is that we imagine allowing the user to have multiple dataform.json files at different levels in the directory hierarchy one day?

examples/bigquery/definitions/override_schema.example.sql
```diff
@@ -0,0 +1 @@
+select 1 as test
```
BenBirt: newline at EOF plz

protos/core.proto
```diff
@@ -26,9 +28,8 @@ message CompileConfig {
   // Required.
   string project_dir = 1;
 
-  // Optional overrides for project-level settings.
-  string default_schema_override = 2;
-  string assertion_schema_override = 3;
+  // Compile time configuration options.
+  string schema_suffix = 2;
```
BenBirt: i think we should either call this `schema_suffix_override` (if we plan to keep `schema_suffix` in `ProjectConfig`) or remove `schema_suffix` from `ProjectConfig`.

tests/api/api.spec.ts
```diff
@@ -497,13 +497,11 @@ describe("@dataform/api", () => {
...
-        defaultSchemaOverride: "overridden_default_schema",
-        assertionSchemaOverride: "overridden_assertion_schema"
+        schemaSuffix: "suffix",
       });
-      expect(graph.projectConfig.defaultSchema).to.equal("overridden_default_schema");
-      expect(graph.projectConfig.assertionSchema).to.equal("overridden_assertion_schema");
+      expect(graph.projectConfig.schemaSuffix).to.equal("suffix");
       graph.tables.forEach(table =>
-        expect(table.target.schema).to.equal("overridden_default_schema")
+        expect(table.target.schema).to.match(/.*_suffix$/)
```
BenBirt: match the whole string plz? (that way we test the appending works as we expect it to)

protos/core.proto
lewish: It was more that as we add more settings and overrides, everything is stored in a single object. I don't need to now separately pass a a compileOptions object all the way into the core runtime through genIndex etc.

protos/core.proto
lewish: Changed to `_override`

tests/api/api.spec.ts
lewish: Bit fiddly now as we have one overridden schema, but this basically does the trick..

cli/index.ts
```diff
@@ -66,11 +66,17 @@ const includeDepsOption: INamedOption<yargs.Options> = {
...
   name: "schema-suffix",
   option: {
-    describe: "A suffix to add to any output schema names.",
-    
+    describe: "A suffix to be appended to output schema names."
+  },
+  check: (argv: yargs.Arguments) => {
+    if (argv.schemaSuffix && !/^[a-zA-Z_0-9]+$/.test(argv.schemaSuffix)) {
+      throw new Error(
+        "--schema-suffix must containly only lowercase, uppercase, underscore and number characters."
```
BenBirt: "contain"

suggest:

--schema-suffix should contain only alphanumeric characters and/or underscores.

protos/core.proto
BenBirt: yep understood, just would rather not have this here if we can help it (for API usability)...

FWIW if this were just passed in to compile and not part of ProjectConfig, we could just add the suffixes as part of the compile process, and wouldn't need to touch dataform/core.

lewish: Let's discuss, as I think we are going to very soon add a number of new CLI options that will override projectConfig settings (even the warehouse property which would be very useful for testing) and would like a way to do that without not too much hassle.

PR #160, opened 2019-04-05
Better integration tests (and a number of fixes)

lewish:
Re-wrote the integration tests so they are faster, less flaky, and more explicit in what they are doing.
- Moved test projects into a dedicated data folder rather than using examples.
- Removed a lot of extra test cases, just checking the basics now: views, tables, incremental tables.
- Made the tests more explicit in what they are doing at the cost of some duplicate code.

Issues fixed:
- Format of snowflake credentials was wrong (userName should be username)
- Added a `target` field to the assertions type so we can easily reference and delete them in tests.
- Added tests for assertions and fixed an issue with assertions not actually being evaluated properly in snowflake.


api/commands/run.ts
```diff
@@ -149,20 +149,22 @@ export class Runner {
...
+              // We expect that an assertion query returns 1 row, with 1 field that is the row count.
+              // We don't really care what that field/column is called.
+              const rowCount = rows[0][Object.keys(rows[0])[0]];
+              if (rowCount > 0) {
+                throw [
+                  ...chainResults,
+                  {
+                    ok: false,
+                    task,
+                    error: `Test failed: query returned ${rowCount} row(s).`
```
BenBirt: s/Test/Assertion/g

core/adapters/index.ts
```diff
@@ -36,7 +37,7 @@ export const requiredWarehouseProps = {
   [WarehouseType.REDSHIFT]: requiredJdbcWarehouseProps,
   [WarehouseType.SNOWFLAKE]: [
     "accountId",
-    "userName",
+    "username",
```
BenBirt: whoops, this is my bad. can you please update the required JDBC properties that i changed too?

https://github.com/dataform-co/dataform/commit/86b8058497000f9926aed6ea2f0ed62f341946b2#diff-7821586da59ad263d03aff6431a570a6

tests/integration/utils.ts
```diff
@@ -1,165 +1,38 @@
...
-function deleteTables(tables: ITableInfo[], testConfig: ITestConfig) {
-  return Promise.all(
-    tables.map(item => {
-      const target = { schema: item.schema, name: item.table };
-      const query = testConfig.adapter.dropIfExists(target, item.type);
-      const sqlDelete = query.replace(/`/g, "\\`");
-
-      return queryRun(sqlDelete, testConfig);
-    })
+export function toMap<V>(values: V[], keyFn: (value: V) => string): { [key: string]: V } {
```
BenBirt: `keyBy`?

core/adapters/index.ts
lewish: Oops - we could really do with a way to get the ACTUAL proto field names at run time and check them here as this was pretty broken.

PR #161, opened 2019-04-06
Use the next_site bazel rules to build the docs site

lewish:
Migrate's the next bazel rules into the open-source repo and fixes up the next config so that everything works.

This should make it easy for us to build and deploy the docs site with our release process.


fe/components/footer.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from "react";
 
 import * as styles from "@dataform/fe/components/footer.css";
 
-import FooterLogoImage from "df/static/images/new_logo_with_text_white.svg";
+const footerLogoImage = require("df/static/images/new_logo_with_text_white.svg");
```
BenBirt: why are we `import`ing some SVGs and `require()`ing others?

lewish: Oh man, I wish I knew. TLDR you have to require inside ts_library rules, and anywhere else you have to do import. Tried to work out why exactly, couldn't, gave up.

PR #165, opened 2019-04-08
Release candidate for version 0.1.0 with a handful of bug fixes.

lewish:
- Fixes issues with suffixed schemas not being created, and being ignored in some SQL generation code.
- Fixes issue with compilation errors being incorrectly detected when there are none on the CLI.
- Bumps the dataform version to 0.1.0-rc.

### DIFF
--- a/api/commands/build.ts
+++ b/api/commands/build.ts
@@ -134,6 +134,7 @@ export class Builder {
       name: assertion.name,
       dependencies: assertion.dependencies,
       type: "assertion",
+      target: assertion.target,
       tasks: this.adapter.assertTasks(assertion, this.compiledGraph.projectConfig).build()
     });
   }
--- a/api/commands/run.ts
+++ b/api/commands/run.ts
@@ -59,21 +59,22 @@ export class Runner {
       throw Error("Executor already started.");
     }
 
-    // Work out all the schemas we are going to need to create first.
-    const uniqueSchemas = {};
-    this.graph.nodes
-      .filter(node => !!node.target)
-      .map(node => node.target.schema)
-      .filter(schema => !!schema)
-      .forEach(schema => (uniqueSchemas[schema] = true));
-
-    const prepareSchemas = Promise.all(
-      Object.keys(uniqueSchemas).map(schema => this.adapter.prepareSchema(schema))
-    );
-
     this.executionTask = new Promise(async (resolve, reject) => {
       try {
-        await prepareSchemas;
+        // Work out all the schemas we are going to need to create first.
+        const uniqueSchemas = {};
+        this.graph.nodes
+          .filter(node => !!node.target)
+          .map(node => node.target.schema)
+          .filter(schema => !!schema)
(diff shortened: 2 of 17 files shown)


api/commands/build.ts
```diff
@@ -134,6 +134,7 @@ export class Builder {
       name: assertion.name,
```
BenBirt: you can get rid of a bunch of this if you use `...assertion` and just fill the `type` and `tasks` fields

api/commands/run.ts
```diff
@@ -58,15 +58,22 @@ export class Runner {
...
+
+    // Work out all the schemas we are going to need to create first.
+    const uniqueSchemas = {};
+    this.graph.nodes
+      .filter(node => !!node.target)
+      .map(node => node.target.schema)
+      .filter(schema => !!schema)
+      .forEach(schema => (uniqueSchemas[schema] = true));
+
+    const prepareSchemas = Promise.all(
```
BenBirt: since you don't actually use this in this function, suggest pushing `prepareSchemas` into the Promise function below, probably just inline it where it's used.

cli/index.ts
```diff
@@ -445,9 +449,20 @@ function getSnowflakeCredentials(): dataform.ISnowflake {
   };
 }
 
+function graphHasErrors(graph: dataform.ICompiledGraph) {
+  return (
+    graph.graphErrors &&
+    graph.graphErrors.compilationErrors &&
+    graph.graphErrors.compilationErrors.length > 0 &&
```
BenBirt: should be `||` (i.e. might have validation errors but no compilation errors, and vice versa)

core/adapters/bigquery.ts
```diff
@@ -49,14 +49,8 @@ export class BigQueryAdapter extends Adapter implements IAdapter {
 
   public assertTasks(a: dataform.IAssertion, projectConfig: dataform.IProjectConfig): Tasks {
```
BenBirt: let's plz call `a` `assertion`

api/commands/build.ts
lewish: As far as I can tell, the create proto method doesn't remove extra fields, so this actually makes me a bit uncomfortable.

core/adapters/bigquery.ts
lewish: I've done the same for `t` too

api/commands/build.ts
BenBirt: hmm, OK. perhaps academic, but i would argue that if some code is looking up *all* properties on some object (which may happen to be a protobuf), it's that code's responsibility to get it right (e.g. if looping through properties on some protobuf object, should really only fetch properties that exist in the .proto file, i.e. those in `keyOf I<proto_name>`).

(IMO this is equivalent to passing around an implementation of an interface in Java; if some code is doing something sketchy to get access to members of that object that aren't defined on the interface, that code has the responsibility to get it right)

PR #168, opened 2019-04-09
Fix backwards compatibility issue with schema suffixes for assertions.

lewish:
The node.target object doesn't exist for older versions of @dataform/core.

Adds the build step to the backwards compatibility test case.


examples/bigquery_backwards_compatibility/definitions/example_assertion.assert.sql
```diff
@@ -0,0 +1 @@
+select 1 as value where false
```
BenBirt: newline at eof plz

PR #169, opened 2019-04-10
Add rules for pushing docs site docker nginx images

lewish:
So we can actually finally serve the new docs site!


docs/BUILD
```diff
@@ -51,3 +51,23 @@ next_site(
...
+
+load("@io_bazel_rules_docker//container:container.bzl", "container_push")
+
+container_push(
+    name = "push",
+    format = "Docker",
+    image = ":image",
+    registry = "gcr.io",
+    repository = "tada-analytics/static-docs",
+    tag = "latest",
```
BenBirt: ditto re other PR comment & `latest`

scripts/docs/serve
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
```
BenBirt: do we even need this file any longer?

docs/BUILD
lewish: Changed to prod as discussed

scripts/docs/serve
lewish: Good point, nah!

PR #174, opened 2019-04-15
Add a new utility function to test the user's Credentials. Use it in the 'init-creds' command.

BenBirt:


api/utils.ts
```diff
@@ -49,6 +50,28 @@ export function coerceCredentials(warehouse: string, credentials: any): Credenti
   }
 }
 
+export async function testCredentials(credentials: Credentials, warehouse: string) {
```
lewish: if we are going to call this via the api should it not be in it's own command file like the other commands? Seems strange for us to import this from `@dataform/api/utils`

cli/index.ts
```diff
@@ -173,10 +182,19 @@ const builtYargs = createYargsCli({
...
+          try {
+            await utils.testCredentials(credentials, argv.warehouse);
+            writeStdOut(commandOutput("Warehouse test query completed successfully."));
+          } catch (e) {
+            throw new Error(`Warehouse test query failed: ${e.stack || e.message}`);
+          }
+        }
+        const filePath = path.resolve(argv["project-dir"], CREDENTIALS_FILENAME);
+        fs.writeFileSync(filePath, prettyJsonStringify(credentials));
+        writeStdOut(commandOutput("Credentials file written:"));
```
lewish: Want to add a comment saying to change settings that you can edit this file directly?

api/utils.ts
lewish: Either way, you want to export this in the api/index.ts file probably so it can just be used directly from importing `@dataform/api`

BenBirt: As discussed, moved all credentials related stuff to `credentials.ts` and exported that.

cli/index.ts
BenBirt: Done

PR #176, opened 2019-04-16
Always resolve references and add resolve method

lewish:
And a few bugs I found along the way...

- Change the behaviour of ref to always try to resolve even if the dependency doesn't exist.
- This is fairly safe as there is a deterministic mapping from a node name to a target (with the exception of assertions, that should not be referenced).
- Any missing dependencies will be picked up during validation rather than as compile time errors.
- Adds a resolve method that can be used within scripts to simply map a node name to a query-able table name without adding it to dependencies.
- Deprecates hasOutput for ops, updates docs accordingly.
- Moves the dependency wildcard resolution to compilation out of validation and adds a test for it.

Although this is more flexible, the downside of all of this is that referencing an assertion or operation will not throw any warnings. We can fix that in the future by differentiating ref'd dependencies from normal dependencies in the compiled graph, enabling us to check this during validation.


api/vm/query.ts
```diff
@@ -20,7 +20,8 @@ export function compile(query: string, projectDir?: string): string {
       projectDir,
       `(function() {
         const { session } = require("@dataform/core");
-        const ref = session.ref.bind(session);
+        const ref = session.resolve.bind(session);
```
BenBirt: `ref` no?

core/assertion.ts
```diff
@@ -48,7 +48,11 @@ export class AssertionContext {
 
   public ref(name: string) {
     this.assertion.dependencies(name);
-    return this.assertion.session.ref(name);
+    return this.assertion.session.resolve(name);
```
BenBirt: `return this.resolve(name);` ?

core/operation.ts
```diff
@@ -59,7 +57,11 @@ export class OperationContext {
 
   public ref(name: string) {
     this.operation.dependencies(name);
-    return this.operation.session.ref(name);
+    return this.operation.session.resolve(name);
```
BenBirt: ditto?

core/session.ts
```diff
@@ -167,6 +160,27 @@ export class Session {
       graphErrors: this.graphErrors
     });
 
+    // Expand node dependency wilcards.
```
BenBirt: wildcards

core/session.ts
```diff
@@ -167,6 +160,27 @@ export class Session {
...
+      compiledGraph.assertions,
+      compiledGraph.operations
+    );
+    const allNodeNames = allNodes.map(node => node.name);
+
+    allNodes.forEach(node => {
+      const uniqueDeps: { [d: string]: boolean } = {};
+      const deps = node.dependencies || [];
+      // Add non-wildcard deps normally.
+      deps.filter(d => !d.includes("*")).forEach(d => (uniqueDeps[d] = true));
```
BenBirt: s/d/dep/ (throughout)

docs/pages/guides/operations.mdx
```diff
@@ -36,14 +36,12 @@ CREATE OR REPLACE VIEW sometable.someschema AS (SELECT 1 AS TEST)
 
 ## Declaring and referencing outputs
 
-If an operation creates a table or view that you would like to make available to other scripts, then you can use the `hasOutput()` configuration method to tell
-Dataform that the output of this operation should be made available to callers of the `ref()` function.
+If an operation creates a table or view that you would like to make available to other scripts, then you can use reference this operation like you would a normal table using the `ref()` function. Additionally operations may use the `self()` to create a table or view that matches the current file name.
```
BenBirt: If an operation creates a table or view that you would like to make available to other scripts, you can reference this operation as you would a normal table by using the `ref()` function. Note that operations may use `self()` to create a table or view that matches the current file name.

api/vm/query.ts
lewish: I renamed session.ref to session.resolve as that's all it's really doing.

PR #178, opened 2019-04-16
Docs style

G2H:


docs/pages/platform_guides/publish_tables.mdx
```diff
@@ -113,9 +123,11 @@ _*Your compiled script will differ depending on your cloud data warehouse. This
 
 The overview page (linked at the top left of Dataform) gives you an overview of your project. The overview includes a visualization of the dependency tree of your project, containing the two tables created by your scripts.
 
+
 <img
   src="/static/images/platform_guides/publish_tables/dependency_tree.png"
```
GJMcGowan: This link is broken, I see a missing image tag when I load the site.

fe/global.css
```diff
@@ -60,55 +57,77 @@
...
+  font-weight: 500;
   letter-spacing: -0.02em;
-  line-height: 54px;
-  font-size: 40px;
-  margin: 0px;
+  line-height: 22px !important;
+  font-size: 22px !important;
+  color: #2f2d44;
+  margin-top: 60px !important;
+  padding-top: 60px !important;
```
GJMcGowan: In general a lot of `!important` tags is a red flag when it comes to css. 

Looking in the browser, it looks like we don't need these `!important`'s, as this css is just overriding the h2 on blueprint anyway (as it's more specific).

I would prefer we got rid of these.

docs/components/navigation.tsx
```diff
@@ -1,12 +1,14 @@
 import { Menu, MenuDivider, MenuItem } from "@blueprintjs/core";
 import * as React from "react";
+import * as styles from "df/docs/components/navigation.css";
+import * as styles2 from "df/docs/components/navigation.css";
```
lewish: remove this?

fe/global.css
GJMcGowan: Also with blueprint 3.0, the typography styles aren't automatically applied (you have to opt in on a heading) so this definitely won't be a problem when we get around to integrating 3.0 into the docs site.

fe/global.css
```diff
@@ -148,20 +158,20 @@
...
+:global body p > code,
+:global body li > code {
   color: #2f2d44;
 }
 
-:global body pre > code, :global body pre > code * {
-  background-color: rgba(27, 31, 35, 0.05);
+:global body pre > code,
+:global body pre > code * {
+  /* background-color: rgba(27, 31, 35, 0.05); */
```
GJMcGowan: Can we remove if we're not using this?

docs/pages/guides/incremental-tables.mdx.orig
```diff
@@ -0,0 +1,177 @@
+import Documentation from "df/docs/layouts/documentation";
```
lewish: This file shouldn't be here, plz delete

PR #182, opened 2019-04-16
Docs style

G2H:


docs/components/screenshot_wrapper.tsx
```diff
@@ -1,7 +1,11 @@
 import * as React from "react";
 import * as styles from "df/docs/components/screenshot_wrapper.css";
 
-export default class ScreenshotWrapper extends React.Component<any, any> {
+interface IProps {
```
lewish: You don't need to do this FYI, children is there automatically.
You could just do `extends React.Component<{}, {}>` if you aren't actually adding any prop types

G2H: I think George asked me to do that

PR #187, opened 2019-04-17
Various changes to make the CLI more usable.

### LINKED ISSUE #183: Don't print full error stack traces out by default

### LINKED ISSUE #185: Make postgres properly supported

### LINKED ISSUE #186: Exit immediately after dataform commands complete
(i.e. don't wait for node to clear its event loop)

BenBirt:
- Properly register postgres as a CLI option, fixes #185 
- Only ever print the first line of an Error's message, fixes #183 
- Exit immediately after any command completes, fixes #186

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -367,9 +367,8 @@ const builtYargs = createYargsCli({
   .wrap(null)
   .recommendCommands()
   .fail((msg, err) => {
-    writeStdErr(
-      errorOutput(`Dataform encountered an error: ${err ? err.stack || err.message : msg}`)
-    );
+    const message = err.message.split("\n")[0];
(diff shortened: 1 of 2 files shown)


cli/index.ts
```diff
@@ -527,7 +526,10 @@ function createYargsCli(cli: ICli) {
       command.format,
       command.description,
       yargsChainer => createOptionsChain(yargsChainer, command),
-      command.processFn
+      async (argv: any) => {
```
GJMcGowan: ew, `any`

BenBirt: I've given this a slightly better type, but there isn't much we can do here, it's really a map/dict mapping from argument names to arguments (which could be anything)

PR #193, opened 2019-04-17
Make the CLI more usable.

### LINKED ISSUE #190: Non-bigquery CLI runs cannot be cancelled with ctrl+c

### LINKED ISSUE #191: User should be able to easily skip port number selection if using default

### LINKED ISSUE #192: Dataform commands that require a project directory should check that it appears to be a dataform directory

BenBirt:
- Correctly interpret `~` tildes in flag paths
- In all situations that require a project directory to exist, check that it really does appear to be a dataform directory: fixes #192 
- Enable cancelling runs on non-bigquery warehouses with ctrl+C: fixes #190 
- Enable the user to simply use the default port number for redshift/postgres: fixes #191 
- Hide the asterisks that were printed as part of password entry

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -22,9 +22,7 @@ const errorOutput = (output: string) => coloredOutput(output, 91);
 const writeStdOut = (output: string) => process.stdout.write(output + "\n");
 const writeStdErr = (output: string) => process.stderr.write(output + "\n");
 
-const actuallyResolve = filePath => {
-  return path.resolve(untildify(filePath));
-};
+const actuallyResolve = filePath => path.resolve(untildify(filePath));
 
 const projectDirOption: INamedOption<yargs.PositionalOptions> = {
   name: "project-dir",
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 import { build, compile, credentials, init, run, table } from "@dataform/api";
 import { prettyJsonStringify } from "@dataform/api/utils";
-import { WarehouseType, supportsCancel } from "@dataform/core/adapters";
+import { supportsCancel, WarehouseType } from "@dataform/core/adapters";
 import { dataform } from "@dataform/protos";
 import * as chokidar from "chokidar";
 import * as fs from "fs";
@@ -458,7 +458,7 @@ function getJdbcCredentials(hostQuestion: string, defaultPort: number): dataform
   const port = parseInt(
     readlineSync.prompt({
       limit: value => {
-        const intValue = parseInt(value);
+        const intValue = parseInt(value, 10);
         return (
           !isNaN(intValue) && typeof intValue === "number" && intValue >= 0 && intValue <= 65536
         );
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -67,6 +67,7 @@
     "uglify-es": "^3.3.9",
     "uglify-js": "^3.4.9",
     "umd-compat-loader": "^2.1.1",
+    "untildify": "^3.0.3",
```
GJMcGowan: Yes, yes, add more random `npm` libraries 🕺

core/adapters/index.ts
```diff
@@ -29,6 +29,13 @@ export enum WarehouseType {
   SNOWFLAKE = "snowflake"
 }
 
+export function supportsCancel(warehouseType: WarehouseType) {
```
GJMcGowan: Why not `return warehouseType === WarehouseType.BIGQUERY`

cli/index.ts
```diff
@@ -429,12 +452,26 @@ function getPostgresCredentials() {
...
-    commandOutput(`Enter the port that Dataform should connect to (usually ${defaultPort}):\n`)
+  writeStdOut(
+    commandOutput(`Enter the port that Dataform should connect to (leave blank to use default):`)
+  );
+  const port = parseInt(
+    readlineSync.prompt({
+      limit: value => {
+        const intValue = parseInt(value);
+        return (
+          !isNaN(intValue) && typeof intValue === "number" && intValue >= 0 && intValue <= 65536
```
GJMcGowan: perhaps just use `Number.isInteger(intValue)` instead of the first two conditions? (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger)

cli/index.ts
```diff
@@ -21,18 +22,33 @@ const errorOutput = (output: string) => coloredOutput(output, 91);
 const writeStdOut = (output: string) => process.stdout.write(output + "\n");
 const writeStdErr = (output: string) => process.stderr.write(output + "\n");
 
+const actuallyResolve = filePath => {
```
GJMcGowan:
```suggestion
const actuallyResolve = filePath => path.resolve(untildify(filePath));
```

package.json
BenBirt: What it does is literally a one-liner, I feel like I am falling to the dark side

cli/index.ts
BenBirt: Good idea. Sadly: if I pass in `value`, typescript complains I'm not passing in a `number` - so I could pass `intValue`, but we already know that's going to be an integer...

GJMcGowan: `parseInt(null)` returns `NaN` which is not an integer. Any non number string would return `NaN`.

The point I was making was that `Number.isInteger` means you can get rid of the first two conditions.

Investigating more, I actually think the `typeof intValue === "number"` is irrelevant since the only two possible values of `parseInt` are an integer or `NaN`. And `typeof NaN` evaluates to "number" 🙃 

So actually, I would say, just drop the typeof condition and then leave things as they are.

BenBirt: Thanks, done

PR #195, opened 2019-04-18
Docs highlighting improvements

lewish:
- Move to reHYPE prism as the highlighter as it seems a lot better
- Introduce new styles that are much more like Github and light in color

image


lewish:
Good spot!! Have removed the selector styles for now.

PR #207, opened 2019-04-25
Add nginx configs for static sites and single page applications

lewish:
Two new nginx configs and base nginx images built with them.

The base config for both was built by extracting the configs from the nginx docker image we were originally using.

- Remove last modified headers so we rely purely on etags instead.
- Set a 5m default cache time.
- Always serve the index file for 404's on SPAs so they can use browser history.

Unfortunately there is no easy way to compose these configs into multiple files so there is a bit of duplication.

### DIFF
--- a/tools/nginx/static/nginx.conf
+++ b/tools/nginx/static/nginx.conf
@@ -1,3 +1,5 @@
+# Basic nginx config for a static site serving static assets.
+
 # This file is mostly taken from the defaults in the official nginx
 # docker image with a few tweaks as commented below.
 
--- a/tools/nginx/static/nginx.conf
+++ b/tools/nginx/static/nginx.conf
@@ -1,0 +1,53 @@
+# This file is mostly taken from the defaults in the official nginx
+# docker image with a few tweaks as commented below.
+
+user  nginx;
+worker_processes  1;
+
+error_log  /var/log/nginx/error.log warn;
+pid  /var/run/nginx.pid;
+
+
+events {
+    worker_connections  1024;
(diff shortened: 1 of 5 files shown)


tools/nginx/static/nginx.conf
```diff
@@ -0,0 +1,55 @@
...
+            root  /usr/share/nginx/html;
+            index  index.html;
+
+            # Set a reasonble default cache time.
+            expires  5m;
+            add_header  Cache-Control "public";
+
+            # Disable caching based on last modified headers as
+            # with our docker builds file timestamps are always the same.
+            add_header Last-Modified "";
```
BenBirt: FWIW it might be fairly easy to populate this with a meaningful timestamp.

- bazel populates bazel-out/volatile-status.txt with a `BUILD_TIMESTAMP` variable
- we already have your `expand_templates` BUILD rule

we could do something like:
```
expand_templates(
    ...
    substitutions = {
        "${LAST_MODIFIED_TIMESTAMP}": "{BUILD_TIMESTAMP}",
    },
    stamp = True,
    ...
)
```

lewish: I did consider this - but then realized this is in the open source repo and we don't have those env vars available here :(

We can do this once we get the docs site building in the closed repo via bazel though!

BenBirt: That env variable is in bazel by default, without any changes. (that said, open-source repo doesn't have the relevant expand_templates BUILD rule... yet)

Issue #171, opened 2019-04-12
Script comments are not being interpreted correctly


BenBirt:
```
-- --js <stuff>
```

The above code should not be run as JS, but it is.

PR #216, opened 2019-05-03
Change regex to fix the issue #171

faiyaz26:
#171


tests/core/core.spec.ts
```diff
@@ -484,27 +484,40 @@ describe("@dataform/core", () => {
...
         */
         /*js
         var c = 3;
         */
-        --js var b = 2;
         /*
         normal_multiline_comment
         */
+        -- --js var x = 1200;
+        --js var b = 2;
```
BenBirt: nit: did this have to move?

faiyaz26: yes, just to make sure it works when some `js` code is inside.

PR #218, opened 2019-05-21
Make sure all the onChange handlers are finished running

faiyaz26:


api/commands/run.ts
```diff
@@ -109,12 +109,12 @@ export class Runner {
...
       .map(fn => fn.name);
 
-    pendingNodes.forEach(node => {
+    await Promise.all(pendingNodes.map(async (node) => {
       const finishedDeps = node.dependencies.filter(d => allFinishedDeps.indexOf(d) >= 0);
       const successfulDeps = node.dependencies.filter(d => allSuccessfulDeps.indexOf(d) >= 0);
       if (!this.cancelled && successfulDeps.length == node.dependencies.length) {
         // All required deps are completed, start this node.
-        this.executeNode(node);
+        await this.executeNode(node);
```
BenBirt: i'm a bit concerned about this change; i think it reduces the parallelism of Node execution - i think this means that each iteration of loop() waits for all nodes whose execution was started to finish before starting another iteration, which (i think?) means that fewer nodes will run in parallel and overall execution will take longer.

what we want is two things:
1) as much parallelism in *execution* as possible
2) a happens-before relationship on the onChange handlers (or at least, for resultPromise to not resolve until all change handlers have resolved).

one way we could achieve this is by doing a (fairly major, so we should consider it carefully) refactor of this file to invert the dependency graph: once a node completes successfully, call triggerChange(), and then execute all of its children (and so on). that way, all nodes are running that can be running, at all times, and we enforce that triggerChange() always completes before resultPromise does.

i'm not actually sure I see any obvious alternatives to this though...

api/commands/run.ts
```diff
@@ -74,7 +74,7 @@ export class Runner {
         );
 
         // Start the main execution loop.
-        this.loop(() => resolve(this.result), reject);
+        await this.loop(() => resolve(this.result), reject);
```
BenBirt: let's revert this

api/commands/run.ts
```diff
@@ -93,10 +93,10 @@ export class Runner {
   }
 
   private triggerChange() {
-    this.changeListeners.forEach(listener => listener(this.result));
+    return Promise.all(this.changeListeners.map(listener => listener(this.result)))
```
BenBirt: please change `onChange` to be: `public onChange(listener: (graph: dataform.IExecutedGraph) => Promise<void> | void): Runner`

api/commands/run.ts
```diff
@@ -109,28 +109,28 @@ export class Runner {
...
       .map(fn => fn.name);
 
-    pendingNodes.forEach(node => {
+    await Promise.all(pendingNodes.map(async (node) => {
       const finishedDeps = node.dependencies.filter(d => allFinishedDeps.indexOf(d) >= 0);
       const successfulDeps = node.dependencies.filter(d => allSuccessfulDeps.indexOf(d) >= 0);
       if (!this.cancelled && successfulDeps.length == node.dependencies.length) {
         // All required deps are completed, start this node.
-        this.executeNode(node);
+        let executedNode = this.executeNode(node);
```
BenBirt: revert this change please

api/commands/run.ts
```diff
@@ -109,28 +109,28 @@ export class Runner {
...
       } else {
         this.pendingNodes.push(node);
-        this.triggerChange();
       }
-    });
+    }));
+
     if (this.pendingNodes.length > 0 || this.result.nodes.length != this.graph.nodes.length) {
-      setTimeout(() => this.loop(resolve, reject), 100);
+      setTimeout(async () => await this.loop(resolve, reject), 100);
```
BenBirt: revert

api/commands/run.ts
```diff
@@ -145,10 +145,10 @@ export class Runner {
     }
   }
 
-  private executeNode(node: dataform.IExecutionNode) {
+  private async executeNode(node: dataform.IExecutionNode) {
     const startTime = process.hrtime();
     // This creates a promise chain that executes all tasks in order.
-    node.tasks
+    await node.tasks
```
BenBirt: revert this, and the change to `executeNode` (it doesn't need to be `async` once this is removed)

api/commands/run.ts
faiyaz26: I get this error when I remove await. https://tsetse.info/must-use-promises

I added a dummy var to get rid of the build error.

api/commands/run.ts
faiyaz26: added: https://github.com/dataform-co/dataform/pull/218/files#diff-f23e82029447c823a22994252f66fa60R51

api/commands/run.ts
```diff
@@ -74,7 +74,7 @@ export class Runner {
         );
 
         // Start the main execution loop.
-        this.loop(() => resolve(this.result), reject);
+        let runLoop = this.loop(() => resolve(this.result), reject);
```
BenBirt: this is actually unsafe: we should not add a dummy variable, but instead add a .catch handler for any thrown Errors.

api/commands/run.ts
```diff
@@ -109,26 +109,26 @@ export class Runner {
       )
       .map(fn => fn.name);
 
-    pendingNodes.forEach(node => {
```
BenBirt: this can also be reverted TBH

PR #221, opened 2019-05-23
Add readme for CLI npm package

faiyaz26:
We should discuss what will be the keywords. 
I have copy pasted the readme of the github repo to this one. Let me know if any changes required. Will create PR for other packages. Another thing is we need to update the version of the cli I think, its because npm registry doesn't allow to push to npm with same version.


cli/package.json
```diff
@@ -1,7 +1,14 @@
 {
   "name": "@dataform/cli",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "description": "Dataform command line interface.",
+  "homepage": "https://github.com/dataform-co/dataform",
+  "license": "MIT",
+  "keywords": [
```
lewish: additional suggestions:
- big data
- data modelling
- sql
- bigquery
- redshift
- dataops

cli/package.json
```diff
@@ -1,7 +1,14 @@
 {
   "name": "@dataform/cli",
-  "version": "0.1.0",
+  "version": "0.1.1",
```
lewish: We need to increment these elsewhere so can we leave as is for now?

cli/package.json
```diff
@@ -22,4 +29,4 @@
   "publishConfig": {
     "access": "public"
   }
-}
+}
```
lewish: keep the newline plz

PR #223, opened 2019-05-28
Add schedule initialization and validation

faiyaz26:


api/commands/init.ts
```diff
@@ -18,14 +18,21 @@ export interface InitResult {
...
+export interface InitOptions {
+  skipInstall?: boolean,
+  includeSchedules?: boolean
+}
+
 export async function init(
   projectDir: string,
   projectConfig: dataform.IProjectConfig,
-  skipInstall?: boolean
+  options: InitOptions
```
lewish: I'd make this optional, and you can set a default value like `options?: InitOptions = {}`

api/commands/validate.ts
```diff
@@ -0,0 +1,50 @@
+import * as cronParser from "cron-parser";
+import { dataform } from "@dataform/protos";
+
+
+function validateEmail(email: string) {
+  const email_regex = /^(([^<>()\[\]\.,;:\s@\"]+(\.[^<>()\[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})$/i;
+  return email_regex.test(email);
+}
+
+export function validateSchedules(schedules: dataform.schedules.Schedules) {
```
lewish: Just thinking ahead a bit - we are going to want to show this in the UI so we may want to instead build up a list of all the errors and return or throw that instead, so you can actually see multiple issues as the same time, otherwise you'll only see one error, fix it, then another, so on.

protos/schedules.proto
```diff
@@ -0,0 +1,28 @@
...
+
+message Schedule {
+  string name = 1;
+  bool enabled = 2;
+  ScheduleOptions options = 3;
+  string cron = 4;
+  Notification notification = 5;
+}
+
+message Schedules {
```
lewish: nit: Maybe let's call this `SchedulesJSON` just so it's really clear that this refers to a JSON file. Might also be worth a comment at the top of this file to say that field names cannot be changed as this is serialized as JSON.

tests/api/validate.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+      });
+      assert.throws(() => validateSchedules(invalid_schedule), Error, "sdsdf is not a valid cron expression");
+    });
+
+    it("test_invalid_emails", () => {
+      const invalid_schedule_1 = dataform.schedules.Schedules.create({
+        defaultNotification: {
+          emails: ["test.com"]
+        },
+        schedules: [
```
lewish: you can remove these here as we don't test them

api/commands/validate.ts
```diff
@@ -0,0 +1,50 @@
+import * as cronParser from "cron-parser";
+import { dataform } from "@dataform/protos";
+
+
+function validateEmail(email: string) {
+  const email_regex = /^(([^<>()\[\]\.,;:\s@\"]+(\.[^<>()\[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})$/i;
```
lewish: prefer camelCase for variables in general, here and in a few other places in the PR

api/commands/init.ts
```diff
@@ -65,6 +72,23 @@ export async function init(
   fs.writeFileSync(gitignorePath, gitIgnoreContents);
   filesWritten.push(gitignorePath);
 
+  if (options.includeSchedules) {
+    fs.writeFileSync(
+      schedulesJsonPath,
+      prettyJsonStringify(
+        dataform.schedules.Schedules.create({
+          defaultNotification: {
+            emails: [],
```
BenBirt: these fields all have the same values as proto3 default values so these lines *shouldn't* be having any effect. (`emails`, `success`, `failure`, and `schedules`)

api/commands/validate.ts
BenBirt: nit: pull this out as a top-level constant?

api/commands/validate.ts
BenBirt: +1, but please throw the final error (don't return it), the CLI *should* handle that nicely

api/commands/validate.ts
```diff
@@ -0,0 +1,50 @@
...
+  return email_regex.test(email);
+}
+
+export function validateSchedules(schedules: dataform.schedules.Schedules) {
+
+  if (schedules.defaultNotification && schedules.defaultNotification.emails) {
```
BenBirt: pretty sure that `schedules.defaultNotification.emails` is *always* truthy, because it's a proto3 repeated field and thus is always at least an empty array.

api/commands/validate.ts
```diff
@@ -0,0 +1,50 @@
...
+    schedules.defaultNotification.emails.forEach(email => {
+      if (!validateEmail(email)) {
+        throw new Error(`${email} is not a valid email address`);
+      }
+    });
+  }
+
+  const uniqueNameMap = new Map();
```
BenBirt: can you give this Map some generic type parameters? (`new Map<X, Y>()`)

api/commands/validate.ts
```diff
@@ -0,0 +1,50 @@
...
+  schedules.schedules.forEach(schedule => {
+    if (!schedule.name) {
+      throw new Error("schedule name is required");
```
BenBirt: nit: let's turn all of these Errors into sentences please

api/commands/validate.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+    if (uniqueNameMap.has(schedule.name)) {
+      throw new Error(`${schedule.name} is not unique. All the schedules name should be unique.`);
+    }
+    uniqueNameMap.set(schedule.name, true);
```
BenBirt: it looks like you don't actually care about the values of this `Map`, in which case you probably want a `Set` - it's simpler for that case.

api/commands/validate.ts
```diff
@@ -0,0 +1,50 @@
...
+      throw new Error("cron expression is required");
+    }
+
+    try {
+      const _ = cronParser.parseExpression(schedule.cron);
+    } catch (e) {
+      throw new Error(`${schedule.cron} is not a valid cron expression`);
+    }
+
+    if (schedule.notification && schedule.notification.emails) {
```
BenBirt: i think `schedule.notification.emails` is always truthy.

api/commands/validate.ts
```diff
@@ -0,0 +1,50 @@
...
+      schedule.notification.emails.forEach(email => {
+        if (!validateEmail(email)) {
+          throw new Error(`${email} is not a valid email address`);
+        }
+      });
+    }
+  })
+
+  return true;
```
BenBirt: nit: don't bother returning; the caller doesn't get useful returned data here - suggest just throwing in case of invalid data.

cli/index.ts
```diff
@@ -158,6 +158,13 @@ const builtYargs = createYargsCli({
             describe: "Whether to skip installing NPM packages.",
             default: false
           }
+        },
+        {
+          name: "include-schedules",
+          option: {
+            describe: "Whether to include schedules.json.",
```
BenBirt:
```suggestion
            describe: "Whether to initialize a schedules.json file.",
```

cli/index.ts
```diff
@@ -169,7 +176,10 @@ const builtYargs = createYargsCli({
               warehouse: argv.warehouse,
               gcloudProjectId: argv["gcloud-project-id"]
             },
-            argv["skip-install"]
+            {
+              skipInstall: argv["skip-install"],
+              includeSchedules: argv['include-schedules']
```
BenBirt: i'm surprised prettier isn't replacing these `'` with `"` (seems like it should be consistent with the previous line)

protos/schedules.proto
```diff
@@ -0,0 +1,28 @@
...
+  Notification defaultNotification = 1;
```
BenBirt: lower_case_with_underscores everywhere in protobuf please

api/commands/init.ts
faiyaz26: it seems you can't have both initial value and optional flag. So I gave initial value.

api/commands/validate.ts
faiyaz26: I don't have any command from the cli to execute this validation method

faiyaz26: I am returning the errors array, if there is no error it will return a empty array

api/commands/init.ts
faiyaz26: creating an empty json file now

api/commands/validate.ts
faiyaz26: I get `Cannot read property 'emails' of undefined`

api/commands/validate.ts
faiyaz26: if I don't provide the defaultNotification data, it get undefined.

api/commands/validate.ts
faiyaz26: caller will be dataform web, isn't it? if caller gets array with errors will print that out otherwise will ignore.

api/commands/validate.ts
BenBirt: I think there's some confusion here, sorry for that!

previously when I said to remove the proto3 default values, I didn't mean to remove setting a `defaultNotification` object - we'll presumably still need that, if it's a required field. I just meant that the fields *inside* that object don't need to be set explicitly if we're just going to set them to their default values, because that's a no-op.

note that if `defaultNotification` is indeed a required field, we should check that it is set in this validate function.

separately, at this line of code, if you only remove `schedules.defaultNotification.emails`, the code should function exactly the same. because of the way proto3 works, if `schedules.defaultNotification` evaluates to `true` (i.e. it is not `undefined`), `schedules.defaultNotification.emails` will *always* evaluate to `true`, because a proto3 repeated field is always at least an empty array.

api/commands/validate.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export function validateSchedules(schedules: dataform.schedules.SchedulesJSON) {
+
+  const errors = [];
+
+  if (schedules.defaultNotification && schedules.defaultNotification.emails) {
+    schedules.defaultNotification.emails.forEach(email => {
+      if (!validateEmail(email)) {
+        errors.push(`${email} is not a valid email address.`);
```
BenBirt: suggest surrounding these validated protobuf fields (`email` in this case) in quotes, so that the error looks nicer:

```
fo1w1r3 is not a valid email address.
```

vs 

```
"fo1w1r3" is not a valid email address.
```

api/commands/validate.ts
```diff
@@ -0,0 +1,52 @@
...
+      }
+    });
+  }
+
+  const uniqueNames = new Set<string>();
+  schedules.schedules.forEach(schedule => {
+    if (!schedule.name) {
+      errors.push("schedule name is required.");
```
BenBirt: nit: please capitalize these sentences as they'll eventually be shown in the UI.

api/commands/validate.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+    if (uniqueNames.has(schedule.name)) {
+      errors.push(`${schedule.name} is not unique. All the schedules name should be unique.`);
```
BenBirt:
```suggestion
      errors.push(`${schedule.name} is not unique. All schedule names must be unique.`);
```

protos/schedules.proto
```diff
@@ -0,0 +1,29 @@
+syntax="proto3";
```
BenBirt: can you add some comments here about what fields are required (if any)?

faiyaz26:
@BenBirt just discussed with @lewish , as per discussion, `defaultNotifications` and `schedules` are not required fields.

And I have added a bit more code to validate the actions against the compiledGraph. Please take a look.

api/commands/validate.ts
faiyaz26: only checking the `defaultNotification` value now.

api/commands/validate.ts
faiyaz26: change the code, just checking the `schedule.notification` value

protos/schedules.proto
faiyaz26: added here: https://github.com/dataform-co/dataform/pull/223/files#diff-f6013730658e3d5392d8da6698d03b66R18

api/commands/validate.ts
```diff
@@ -0,0 +1,66 @@
...
+    try {
+      const _ = cronParser.parseExpression(schedule.cron);
+    } catch (e) {
+      errors.push(`"${schedule.cron}" is not a valid cron expression.`);
+    }
+
+    if (schedule.options) {
+      schedule.options.actions.forEach(action => {
+        if (allNodeNames.indexOf(action) < 0) {
+          errors.push(`"${action}" doesn't exist in the project.`);
```
faiyaz26: @BenBirt here is the new validation for actions

api/commands/validate.ts
```diff
@@ -7,42 +7,56 @@ function validateEmail(email: string) {
   return EMAIL_REGEX.test(email);
 }
 
-export function validateSchedules(schedules: dataform.schedules.SchedulesJSON) {
+export function validateSchedules(schedules: dataform.schedules.SchedulesJSON, compiledGraph: dataform.ICompiledGraph) {
 
   const errors = [];
+  const allNodes = [].concat(
```
BenBirt: this isn't quite typesafe, suggest:

```
const allNodes = { name: string }[].concat(
```

api/commands/validate.ts
BenBirt: ah, this is a great idea, thanks for adding this!

protos/schedules.proto
```diff
@@ -0,0 +1,29 @@
...
+
+package dataform.schedules;
+
+message Notification {
+  repeated string emails = 1;
+  bool success = 2;
```
BenBirt: these are going to appear in JSON form in a checked in project somewhere in the end, so suggest:
`on_success` and `on_failure`

protos/schedules.proto
```diff
@@ -0,0 +1,29 @@
...
+
+message ScheduleOptions {
+  repeated string actions = 1;
+  bool include_dependencies = 2;
+  bool full_refresh = 3;
+} 
+
+message Schedule {
+  string name = 1; // required field
+  bool enabled = 2;
```
BenBirt: suggest inverting this to `disabled`. that way, the default state for a text schedules.json file is to not state a value for this field.

protos/schedules.proto
```diff
@@ -0,0 +1,29 @@
+syntax="proto3";
+
+package dataform.schedules;
+
+message Notification {
```
BenBirt: `NotificationSettings`?

tests/api/validate.spec.ts
```diff
@@ -0,0 +1,94 @@
+import { validateSchedules } from "@dataform/api";
+import { dataform } from "@dataform/protos";
+import { assert, config, expect } from "chai";
+
+describe("@dataform/api/validate", () => {
+  describe("validateSchedules", () => {
+
+    it("test_valid_schedule", () => {
```
BenBirt: it's good practice to name jasmine testcases with human-readable strings. https://github.com/CareMessagePlatform/jasmine-styleguide#speak-human has some good guidance here.

for example, this one might be "returns no errors for valid schedules object"

tests/api/validate.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+            enabled: false,
+            options: {
+              actions: ["action2"]
+            }
+          },
+          {
+            name: "name2",
+            cron: "*/2 * * * *",
+            notification: {
+              emails: ['tada@test.com'],
```
BenBirt: nit: our prettifier style should be replacing these `'` with `"`, can you double check that you have it installed / that it runs on save?

tests/api/validate.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+              success: true,
+              failure: false
+            }
+          }
+        ]
+      })
+
+      const errors = validateSchedules(valid_schedule, compiledGraph);
+      expect(errors).to.have.members([]);
```
BenBirt: i think you can do this which might be a bit clearer:

```
expect(errors).to.eql([]);
```

protos/schedules.proto
faiyaz26: done

tests/api/validate.spec.ts
faiyaz26: done

tests/api/validate.spec.ts
faiyaz26: done

api/commands/validate.ts
```diff
@@ -10,11 +10,12 @@ function validateEmail(email: string) {
 export function validateSchedules(schedules: dataform.schedules.SchedulesJSON, compiledGraph: dataform.ICompiledGraph) {
 
   const errors = [];
-  const allNodes = [].concat(
+  let allNodes = new Array<{ name?: string; }>().concat(
```
BenBirt: `const` please

tests/api/validate.spec.ts
```diff
@@ -35,16 +35,16 @@ describe("@dataform/api/validate", () => {
...
+              onSuccess: true,
+              onFailure: false
             }
           }
         ]
       })
 
       const errors = validateSchedules(valid_schedule, compiledGraph);
-      expect(errors).to.have.members([]);
+      expect(errors).to.eql([]);
```
BenBirt: please make this change below too?

PR #236, opened 2019-05-30
Handle undefined array

faiyaz26:


api/commands/validate.ts
```diff
@@ -18,7 +18,7 @@ export function validateSchedules(schedules: dataform.schedules.SchedulesJSON, c
 
   const allNodeNames = allNodes.map(node => node.name);
 
-  if (schedules.defaultNotification) {
+  if (schedules.defaultNotification && schedules.defaultNotification.emails) {
```
BenBirt: proto3 default should mean that this shouldn't be required. at the callsite, are you calling dataform.schedules.Schedule.create(...) and then passing that result in here?

faiyaz26: I have added a test case for which I am getting undefined
```
notification: {
              onFailure: true
            }
```

if I provide some of the value, others get the value of undefined, so the `email` field gets undefined value

BenBirt: huh, that is pretty weird, protobufjs should definitely not be working like that! anyways, thanks for the fix.

PR #237, opened 2019-05-31
change the field name from Nodes to actions

faiyaz26:


api/commands/build.ts
```diff
@@ -61,8 +61,8 @@ export class Builder {
 
     // Determine which nodes should be included.
     const includedNodeNames =
```
BenBirt: `includedActionNames` now?

(actually there are a lot of usages of the term 'node' both before and after this code, can you replace them with 'action' to be consistent please?)

cli/index.ts
```diff
@@ -371,7 +371,7 @@ const builtYargs = createYargsCli({
           compiledGraph,
           {
             fullRefresh: argv["full-refresh"],
-            nodes: argv.nodes,
+            actions: argv.nodes,
```
BenBirt: ditto, let's change the name of the `nodes` flag?

api/commands/build.ts
faiyaz26: I have only replaced those place which is related to actions. There are some places where nodes are referred to the execution graph nodes.

PR #239, opened 2019-06-03
Add a common shared package.json and new build rules 

lewish:
Common package information such as version is now in one single place, and can be overriden in individual packages if needed.

- Adds a new tool for doing JSON merges via a CLI call that can be used in bazel
- Makes every package inherit from the top level `common.package.json`
- Dedupes the readme in the CLI package, using a bazel copy rule instead
- Adds some common npm package rules

Still todo: Automatically generate a package.json from bazel build deps.


cli/BUILD
```diff
@@ -33,17 +33,24 @@ nodejs_binary(
...
 )
 
-load("@build_bazel_rules_nodejs//:defs.bzl", "npm_package")
+load("//tools/common:copy.bzl", "copy_file")
 
-npm_package(
-    name = "package",
-    srcs = ["package.json", "README.md"],
-    deps = [":cli"],
+copy_file(
```
BenBirt: do we need this? i understood bazel to support per-file dependencies natively, i.e. that the dependent BUILD rule could already just depend on a specific file.

tools/json-merge/index.ts
```diff
@@ -0,0 +1,14 @@
+import * as fs from "fs";
+import * as yargs from "yargs";
+
+const argv = yargs.option("output", { required: true }).option("layers", { array: true }).argv;
```
BenBirt: nit: `output-path` and `layer_paths`?

tools/json-merge/index.ts
```diff
@@ -0,0 +1,14 @@
...
+
+let result = {};
```
BenBirt: suggest
```
const result = argv.layers.reduce((accumulator, currentValue) => {...accumulator, ...currentValue}, {});
```

cli/BUILD
lewish: It kind of does, but the npm_package rules get upset when the file isn't in the same directory as everything else and I haven't found another workaround :/

PR #240, opened 2019-06-04
Add support for v2 language.

BenBirt:
As requested, am sending the whole change in one PR - happy to split it up if you change your mind.


core/compilers.ts
```diff
@@ -81,6 +85,75 @@ export function extractJsBlocks(code: string): { sql: string; js: string } {
...
+  case "view":
+  case "table":
+  case "incremental":
+  case "inline": {
+    action.query(ctx => {
+      ${getFunctionPropertyNames(SqlxTableContext.prototype)
+        .map(name => `const ${name} = ctx.${name}.bind(ctx);`)
+        .join("\n")}
+      ${results.js}
+      if (finalConfig.type === "incremental") {
```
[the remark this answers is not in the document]
lewish: Also feels like this should have an - if hasIncrementalBlock then { ... }

core/session.ts
```diff
@@ -42,9 +54,105 @@ export class Session {
...
+    }
+
+    const action = (() => {
+      switch (sqlxConfig.type) {
+        case "view":
+        case "table":
+        case "inline":
+        case "incremental": {
+          const dataset = this.publish(sqlxConfig.name);
+          dataset.type(sqlxConfig.type);
```
lewish: You could just call `dataset.config(sqlxConfig)` - that's the idea anyway! Then you remove some duplication here (this method is mostly the same as the config method in dataset.

lewish: It would probably be nice to write a config({}) style method for operations and assertions too. This should make it easier to add or remove properties in the future for different types as this method just becomes routing to the correct action type.

core/sqlx_parser.ts
```diff
@@ -0,0 +1,257 @@
...
+  postOperations: string[];
+}
+
+export function parseSqlx(code: string): ISqlxParseResults {
+  const results = {
+    config: "",
+    js: "",
+    sql: [""],
+    incremental: "",
+    preOperations: [""],
```
lewish: What's the reason for this initialization?

core/sqlx_parser.ts
```diff
@@ -0,0 +1,257 @@
...
+    if (
+      token.type === SQL_LEXER_TOKEN_NAMES.STATEMENT_SEPERATOR ||
+      token.type === INNER_SQL_BLOCK_LEXER_TOKEN_NAMES.STATEMENT_SEPERATOR
+    ) {
+      if (state === "incremental") {
+        throw new Error(
+          "Incremental code blocks may not contain SQL statement separators ('---')."
+        );
+      }
+      (results[state] as string[]).push("");
```
lewish: This doesn't seem right - results.incremental is a string, not an array.

core/sqlx_parser.ts
```diff
@@ -0,0 +1,257 @@
...
+    }
+    if (Array.isArray(results[state])) {
+      (results[state] as string[])[results[state].length - 1] += token.value;
```
lewish: How do these arrays ever get longer than 1 if we just always modify the string at index 0?

core/sqlx_parser.ts
```diff
@@ -0,0 +1,257 @@
...
+    push: LEXER_STATE_NAMES.JS_BLOCK,
+    value: () => "{"
+  };
+  sqlLexer[SQL_LEXER_TOKEN_NAMES.START_JS] = {
+    match: "js {",
+    push: LEXER_STATE_NAMES.JS_BLOCK,
+    value: () => ""
+  };
+  sqlLexer[SQL_LEXER_TOKEN_NAMES.START_INCREMENTAL] = {
+    match: "if_incremental {",
```
lewish: Ok so I think I got ahead of myself here. This block should probably be called `incremental_where` or just `where`.

Longer term, the idea here was that you would write something like:

```
select user_id, sum(1) from test
if_incremental {
  where ts > (select max(ts) from ${self()}
}
group by 1
```

The idea being you can control exactly where the conditional lines go. We need a completely new proto structure to do something like this for blocks of SQL. As it stands we currently just wrap the query in a select statement and apply the where clause, and I don't think the if_incremental name makes that very clear (plus we should save that name for when we do the new thing!).

tests/api/api.spec.ts
```diff
@@ -491,6 +496,163 @@ describe("@dataform/api", () => {
...
+      expect(exampleSampleData.query).equals(
+        "select 1 as sample union all\nselect 2 as sample union all\nselect 3 as sample"
+      );
+      expect(exampleSampleData.dependencies).to.eql([]);
+    });
+
+    it("bigquery using v2 language compiles", async () => {
+      const graph = await compile({ projectDir: path.resolve("df/examples/bigquery_language_v2") });
+
+      // throw new Error(graph.graphErrors.compilationErrors[0].message);
```
lewish: Doesn't look quite right

lewish: On this note, I think we probably need a few test cases for failures as we don't currently have any and added a lot of new checks!
Let's discuss validation/error handling in general though

core/sqlx_parser.ts
BenBirt: discussed on chat, hopefully makes sense now

tests/api/api.spec.ts
BenBirt: removed the commented out code and added a bunch of error test cases!

BenBirt:
I've upgraded the version numbers, and have removed some stuff from the JS `config()` API to make the `...Config` types match as suggested!

PR #248, opened 2019-06-05
Remove axios vulnerability. Make it possible to use this code locally in a dataform-co build again.

### LINKED ISSUE #235: Fix axios package version security vulnerability
https://github.com/dataform-co/dataform/network/alerts

Given that this is only in the CLI and not any services, it is unlikely that we are affected by this. However, we should address it nonetheless!

BenBirt:
fixes #235

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -957,44 +957,107 @@
       }
     },
     "@google-cloud/bigquery": {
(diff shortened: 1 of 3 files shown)


BenBirt:
Yep, //tests/integration:bigquery passes.

PR #252, opened 2019-06-06
Integration test keys

lewish:
- Adds encrypted test keys that can be decrypted with `./scripts/decode_credentials` (only for those with access unfortunately)
- Runs integration tests during cloud build
- Moves to a new bigquery project that only has integration test data


scripts/decode_credentials
```diff
@@ -0,0 +1,17 @@
...
+gcloud kms decrypt \
+  --plaintext-file=test_credentials/redshift.json \
+  --ciphertext-file=test_credentials/redshift.json.enc \
+  --keyring=dataform-co-build --key=dataform-co-build --location=global
+
+
+gcloud kms decrypt \
+  --plaintext-file=test_credentials/snowflake.json \
+  --ciphertext-file=test_credentials/snowflake.json.enc \
+  --keyring=dataform-co-build --key=dataform-co-build --location=global
```
BenBirt: nit: newline plz

scripts/decode_credentials
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+gcloud kms decrypt \
```
BenBirt: so... cloudbuild is able to decrypt these files. presumably it can do so because it sits inside our cloud organization?

if so, doesn't that let just about anybody get access to the raw credentials? all they need to do is add a console.log line to the test files, push to github, and cloudbuild will gladly decode and print out the credentials.

on the other hand, if we're fine with that, then do we need to encrypt them at all?

lewish: As discussed:
- Only we can see the cloud build logs
- Only we can create branches inside the dataform repo and therefore decrypt the keys

If anyone else needs to run the integration tests, I guess we can create a branch for them inside dataform with their code

PR #263, opened 2019-06-13
Add a projectConfig() API to @dataform/core.

### LINKED ISSUE #262: @dataform/core 'session' object needs to expose some project config (dataform.json) level data
(P1 because it blocks true launch of SQLX)

BenBirt:
Fixes #262

### DIFF
--- a/api/api.package.json
+++ b/api/api.package.json
@@ -4,8 +4,8 @@
   "main": "index.js",
   "types": "index.d.ts",
   "dependencies": {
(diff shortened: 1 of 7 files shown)


BenBirt:
Closing this PR after discussion in-person: For now, users can use the `session` private internals by just using the global `session` object. We regard this as reasonable usage but make no guarantees around not breaking that API. We will document this on the docs site once we clean up the `Session` class to an API that we are happier to support longer-term.

PR #271, opened 2019-06-18
Clean dates bigquery

GJMcGowan:


api/dbadapters/bigquery.ts
```diff
@@ -134,12 +134,37 @@ export class BigQueryDbAdapter implements DbAdapter {
...
     }
   }
+
+  private cleanRows(rows: any[]) {
+    if (rows.length === 0) {
+      return rows;
+    }
+
+    const sampleData = rows[0];
+    const BIGQUERY_DATE_CLASS_NAME = "BigQueryDate";
```
BenBirt: extract this to the top of the file plz?

api/dbadapters/bigquery.ts
```diff
@@ -134,12 +134,37 @@ export class BigQueryDbAdapter implements DbAdapter {
...
+    const fieldsWithBigQueryDates = Object.keys(sampleData).filter(
+      key =>
+        sampleData[key] &&
+        sampleData[key].constructor &&
+        sampleData[key].constructor.name === BIGQUERY_DATE_CLASS_NAME
+    );
+    if (fieldsWithBigQueryDates.length === 0) {
```
BenBirt: suggest this which might be a little simpler:

```
fieldsWithBigQueryDates.forEach(dateField => {
  rows.forEach(row => row[dateField] = row[dateField].value);
});
```

GJMcGowan: Fair enough, changed. I was vaguely thinking of not mutating the original variable but I don't think it matters too much.

GJMcGowan:
@BenBirt I've updated the version in common.package.json - is there anywhere else I need to update the version before I bring it into `dataform-co`?

PR #285, opened 2019-06-27
Remove scheduling page from the docs site

faiyaz26:
Fixes https://github.com/dataform-co/dataform-co/issues/920


faiyaz26:
We need this scheduling page for some features on dataform-web. So going to revert this.

@lewish @dwl285

PR #288, opened 2019-07-02
added docs section for built in functions

dwl285:


dwl285:
image
this gets rid of the background in the hover state, but it makes the titles look weird. WDYT @GJMcGowan

PR #291, opened 2019-07-09
Add a testcase which times out during compilation.

### LINKED ISSUE #258: Add a test case for a project that never finishes compiling
A project containing a `while(true)` loop in a JS file should do it.

BenBirt:
fixes #258


tests/api/api.spec.ts
```diff
@@ -948,6 +949,15 @@ describe("@dataform/api", () => {
         "example_view"
       ]);
     });
+
+    it("times out after timeout period during compilation", async () => {
+      try {
+        await compile({ projectDir: "df/examples/redshift_never_finishes_compiling" });
+        fail("Compilation timeout Error expected.");
```
faiyaz26: why you need this `fail` ?

BenBirt: It's to handle the case where `compile` (incorrectly) doesn't throw a timeout Error.

PR #292, opened 2019-07-10
Add initial dataset test case support.

BenBirt:
- does not support any queries that are not datasets
- does not yet support incremental tables (this needs a little thought!)
- TODO: add SQLX support
- TODO: expose a new `test` command in CLI once SQLX support exists

### DIFF
--- a/api/commands/test.ts
+++ b/api/commands/test.ts
@@ -41,7 +41,7 @@ async function runTest(
     return {
       name: testCase.name,
       successful: false,
-      message: `Expected ${expectedColumns.length} columns, but saw ${actualColumns.length} columns.`
+      message: `Expected columns "${expectedColumns}", but saw "${actualColumns}".`
     };
   }
   for (const expectedColumn of expectedColumns) {
@@ -49,7 +49,7 @@ async function runTest(
       return {
         name: testCase.name,
         successful: false,
-        message: `Could not find expected "${expectedColumn}" column.`
+        message: `Expected columns "${expectedColumns}", but saw "${actualColumns}".`
       };
     }
   }
--- a/api/commands/test.ts
+++ b/api/commands/test.ts
@@ -24,9 +24,11 @@ async function runTest(
     return {
       name: testCase.name,
       successful: false,
-      message: `Expected ${expectedResults.length} rows, but saw ${actualResults.length} rows.`
+      messages: [`Expected ${expectedResults.length} rows, but saw ${actualResults.length} rows.`]
     };
   }
+  // If the result set is empty and the number of actual rows is equal to the number of expected rows
+  // (asserted above), this test is therefore successful.
   if (actualResults.length === 0) {
     return {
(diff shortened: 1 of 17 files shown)


api/commands/test.ts
```diff
@@ -0,0 +1,77 @@
...
+  const expectedColumns = Object.keys(expectedResults[0]);
+  if (actualColumns.length !== expectedColumns.length) {
+    return {
+      name: testCase.name,
+      successful: false,
+      message: `Expected columns "${expectedColumns}", but saw "${actualColumns}".`
+    };
+  }
+  for (const expectedColumn of expectedColumns) {
+    if (!actualColumns.some(actualColumn => actualColumn === expectedColumn)) {
```
lewish: Also just possibly a bit confusing, there are two things assumed here, maybe needs comment:
- column order does not matter
- returned column names are unique

api/commands/test.ts
```diff
@@ -0,0 +1,77 @@
...
+  const actualColumns = Object.keys(actualResults[0]);
+  const expectedColumns = Object.keys(expectedResults[0]);
+  if (actualColumns.length !== expectedColumns.length) {
+    return {
+      name: testCase.name,
+      successful: false,
+      message: `Expected columns "${expectedColumns}", but saw "${actualColumns}".`
+    };
+  }
+  for (const expectedColumn of expectedColumns) {
```
lewish: OK so this is fun - different warehouses care differently about case sensitivity. In fact, snowflake ALWAYS returns results in uppercase, as you probably noticed.

I'm pretty sure bigquery is case insensitive, let's check redshift. If it is too, then we should probably try and deal with this.

api/commands/test.ts
```diff
@@ -0,0 +1,77 @@
...
+  for (let i = 0; i < actualResults.length; i++) {
+    const actualResultRow = actualResults[i];
+    const expectedResultRow = expectedResults[i];
+
+    for (const column of actualColumns) {
+      if (actualResultRow[column] !== expectedResultRow[column]) {
+        return {
+          name: testCase.name,
+          successful: false,
+          message: `For row ${i} and column "${column}": expected "${expectedResultRow[column]}", but saw "${actualResultRow[column]}".`
```
lewish: I think it would be very useful to stack this up and print them all out together at once if there are multiple mismatched rows.

core/test.ts
```diff
@@ -0,0 +1,93 @@
...
+  public apply<T>(value: TContextable<T>): T {
+    if (typeof value === "function") {
+      return (value as any)(this);
+    } else {
+      return value;
+    }
+  }
+}
+
+class RefReplacingContext extends table.TableContext {
```
lewish: probably shouldn't extend here - as we still have those old methods around in TableContext that could have side-effects. Unfortunately, I think we need an ITableContext interface and to explicitly implement no-ops for all the other methods.

PR #293, opened 2019-07-10
Add test case support to SQLX.

BenBirt:


core/session.ts
```diff
@@ -13,10 +13,11 @@ interface IActionProto {
 }
 
 interface ISqlxConfig extends TConfig {
-  type: "view" | "table" | "inline" | "incremental" | "assertion" | "operations";
+  type: "view" | "table" | "inline" | "incremental" | "assertion" | "operations" | "test";
   schema?: string;
   name: string;
   hasOutput?: boolean;
+  datasetUnderTest?: string;
```
lewish: What about just `dataset` ? or `datasetName` ?

PR #297, opened 2019-07-11
Add a docs page on unit test support.

BenBirt:


docs/pages/guides/tests.mdx
```diff
@@ -0,0 +1,68 @@
...
+export const meta = {
+  title: "Unit tests for dataset queries",
+  __filename
+};
+
+export default ({ children }) => <Documentation {...meta}>{children}</Documentation>;
+
+## Introduction
+
+Unit tests give you confidence that your code produces the output data you expect.
```
lewish: Can we explain the difference between an assertion and a test?

docs/pages/guides/tests.mdx
```diff
@@ -0,0 +1,68 @@
...
+- the number of output rows must match
+- the number of output columns and their names must match
+- the contents of each row must match
+
+Note that unit tests do not fail if columns are not output in the same order.
+
+## Test hermiticity
+
+As with unit testing in other languages and frameworks, it's considered bad practice for a unit test to be non-hermetic. This means that running your test
+should have no dependencies on any state in your data warehouse. Thus all input datasets should be referenced using `ref()` or `resolve()`, with fake inputs
```
lewish: Completely unrelated, I don't think we check that one actually mocks all the inputs at creation time, and we probably should (and throw a nice error message!).

docs/pages/guides/tests.mdx
```diff
@@ -0,0 +1,68 @@
...
+provided for each input dataset.
```
lewish: Should we add a section here for how to do this with the JS API too?

docs/pages/guides/tests.mdx
BenBirt: actually we do that check! https://github.com/dataform-co/dataform/blob/master/core/test.ts#L84

(also nit: we should stop using the word 'mock' IMO, I think if we use that term in communications we'll confuse folks!)

docs/pages/guides/tests.mdx
BenBirt: I've followed the example of the other action types and have added details to the JS API page.

PR #304, opened 2019-07-15
Issue 229

canetcesc:
- Creating a pull request so as to test integration tests. I don't expect this to be mergeable into master yet but "draft pull requests" seem to be disabled.

- I don't know how to test if the new tags are added. 

- How can I check if the scheduler on dataform web can read those tags?

- I added a new example file `examples/redshift/definitions/example_with_tags.sqlx`.

- The `api` and `core` tests run OK.

- Not sure how to test the JS api.


BenBirt:
Hi Cesc!

A few pointers:
- The best way to test that the new tags are added is by updating the test api.spec.ts testcase. Edit a `.sqlx` file in `examples/bigquery_language_v2/definitions` to add a tag, and add an assertion to `api.spec.ts` (the `bigquery using v2 language...` testcase) to check that they make it into the compiled graph
- You won't be able to change our closed source dataform web code to start reading these tags - that's work we'll have to do (shouldn't be a big job)
- I wouldn't worry too much about it, but if you do want to test the JS API you could add find an existing example `.js` file, call the new `.tags()` function, and assert that the tags end up in the compiled graph in the relevant test case in `api.spec.ts`

canetcesc:
Yesterday's build check is taking so long to run. How can I test the latest commit in here ?

BenBirt:
Sorry, not sure how I missed that message!

Not sure what you mean by yesterday's build check - are you saying that the cloudbuild runs on github take a while? I'm actually not sure why they don't seem to be running for you - do you want to bring the PR out of draft mode, as I'm not sure what effect that has?

I wouldn't worry about the tests acting differently locally vs. in cloudbuild, `bazel test tests/api/... tests/core/...` should run locally exactly the same as it would on cloudbuild.

If you want to chat in realtime, feel free to join our community slack: https://dataform-users.slack.com/

core/assertion.ts
```diff
@@ -27,6 +27,16 @@ export class Assertion {
     return this;
   }
 
+  public tags(value: string | string[]) {
+    const newTags = typeof value === "string" ? [value] : value;
+    newTags.forEach(d => {
```
BenBirt: nit: let's rename `d` to `t`

(this is repeated a couple of places in this PR)

core/table.ts
```diff
@@ -42,12 +42,14 @@ export interface TConfig {
   protected?: boolean;
   redshift?: dataform.IRedshiftOptions;
   bigquery?: dataform.IBigQueryOptions;
+  tags?: string[];
```
BenBirt: nit: want to move this line just below the `dependencies` line?

core/table.ts
```diff
@@ -82,6 +84,11 @@ export class Table {
     if (config.bigquery) {
       this.bigquery(config.bigquery);
     }
+    if (config.tags)
+    {
```
BenBirt: nit: not sure what editor you are using, but fwiw we tend to use VSCode with the Prettier plugin, it should autoformat things like this on save. anyways, move this `{` to the previous line?

core/table.ts
```diff
@@ -313,4 +330,9 @@ export class TableContext implements ITableContext {
       return value;
     }
   }
-}
+  public tags(tags: string[]) {
+    this.table.tags(tags);
+    return "";
+  }
+
+}
```
BenBirt: nit: newline at EOF

core/table.ts
```diff
@@ -217,6 +233,7 @@ export interface ITableContext {
   ) => string;
   describe: (key: string, description?: string) => string;
   apply: <T>(value: TContextable<T>) => T;
+  tags: (name: string[]) => string;
```
BenBirt: this should probably have the same type signature as the other `Context` types, i.e. `(name: string | string[]) => string`

core/table.ts
canetcesc: I use VSCode. Thanks for the tip.

core/table.ts
canetcesc: Yes. BTW, this is off-topic but on line 228 you have: 

` dependencies: (name: string) => string;`

So perhaps that one needs to be changed to this:

` dependencies: (name: string | string[]) => string;`

core/table.ts
canetcesc: Prettier will fix these going forward. Thanks.

core/assertion.ts
canetcesc: Done

core/table.ts
BenBirt: Ah, yes good catch, we should clean that up!

PR #310, opened 2019-07-17
Escape embedded expressions inside SQLX comments so that they aren't interpreted as expressions at compile-time.

### LINKED ISSUE #309: single line commenting of a javascript function in a sqlx file doesn't work as expected
The function gets expanded onto multiple lines, and only the first line gets commented out, leading to errors

---
Moved from dataform-co/dataform-co#1126
/cc @dwl285

BenBirt:
fixes #309

### DIFF
--- a/scripts/publish
+++ b/scripts/publish
@@ -1,6 +1,16 @@
 #!/bin/bash
 set -e
 
+if [ "$(git status --porcelain)" ]; then 
+    echo "There are uncommitted changes; aborting." 1>&2
+    exit 1
+fi
+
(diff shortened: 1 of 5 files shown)


scripts/publish
```diff
@@ -1,6 +1,16 @@
 #!/bin/bash
 set -e
 
+if [ "$(git status --porcelain)" ]; then
```
faiyaz26: this file is intended?

BenBirt: Yup, I'm adding a little cleanup to prevent anybody (particularly myself!) from publishing new NPM versions without being on a clean master branch.

PR #317, opened 2019-07-20
Issue 316 (continuation of Issue 229).

canetcesc:
Creating PR for visibility only.
API tests are still not succeeding and I am investigating why.


api/commands/build.ts
```diff
@@ -55,23 +56,36 @@ export class Builder {
       this.compiledGraph.operations.map(o => this.buildOperation(o)),
       this.compiledGraph.assertions.map(a => this.buildAssertion(a))
     );
-    const allActionNames = allActions.map(n => n.name);
+
     const actionNameMap: { [name: string]: dataform.IExecutionAction } = {};
+
```
BenBirt: nit: rm this new empty line? this is quite tied to the next code line.

api/commands/build.ts
```diff
@@ -55,23 +56,36 @@ export class Builder {
...
     allActions.forEach(action => (actionNameMap[action.name] = action));
 
+    // Include only tagged actions
+    const taggedActions =
+      this.runConfig.tags && this.runConfig.tags.length > 0
+        ? allActions.filter(action => this.runConfig.tags.some(t => action.tags.includes(t)))
```
BenBirt: unless i'm misreading this, i *believe* this logic is equivalent to filtering to only the actions that are in --actions *AND* in --tags. i think that AND should probably be an OR, i.e. if somebody specifies an action in --actions but which does not have a tag in --tags, we should still run it.

protos/core.proto
```diff
@@ -22,6 +22,7 @@ message RunConfig {
   bool include_dependencies = 3;
   bool full_refresh = 2;
   bool carry_on = 4;
+  repeated string tags = 5;
```
BenBirt: would you mind also adding this to the relevant section of `schedules.proto`? (no code should need to change, just the protobuf)

BenBirt: nit: suggest putting this just below `actions` since it is quite related!

protos/core.proto
```diff
@@ -164,14 +165,15 @@ message ExecutionTask {
 message ExecutionAction {
   string name = 1;
   repeated string dependencies = 3;
+  repeated string tags = 7;
```
BenBirt: do we need `tags` in `ExecutionAction`?

api/commands/build.ts
canetcesc: Yes, you're right. That _AND_ should become an _OR_ because it would not make sense otherwise that the user specified that action using the --action option.

protos/core.proto
canetcesc: Sure

api/commands/build.ts
canetcesc: Done

protos/core.proto
canetcesc: Done

protos/core.proto
canetcesc: No, we don't ! Thanks

api/commands/build.ts
```diff
@@ -56,52 +55,77 @@ export class Builder {
...
-
+    const allActionNames = allActions.map(n => n.name);
     const actionNameMap: { [name: string]: dataform.IExecutionAction } = {};
-
     allActions.forEach(action => (actionNameMap[action.name] = action));
 
-    // Include only tagged actions
-    const taggedActions =
+    // Actions selected with --tag option should be included
+    const tagFilteredTables =
```
BenBirt: suggest defining a function so that we can stop repeating this block:

```
const hasAnyTag = (action: { tags: string[] }) => !this.runConfig.tags || this.runConfig.tags.length === 0 || action.tags.some(actionTag => !!this.runConfig.tags.find(actionTag));
```

then do something like:

```
const allTaggedActions: dataform.IExecutionAction[] = [].concat(
  this.compiledGraph.tables.filter(table => hasAnyTag(table)),
  this.compiledGraph.operations.filter(table => hasAnyTag(table)),
  this.compiledGraph.assertions.filter(table => hasAnyTag(table)),
);
```

tests/api/api.spec.ts
```diff
@@ -212,24 +212,57 @@ describe("@dataform/api", () => {
...
+            name: "op_e",
+            dependencies: ["op_d"],
+            tags: ["tag1", "tag2"],
+            queries: ["create or replace view schema.someview as select 1 as test"]
           }
         ]
       });
 
-      const builder = new Builder(
+      const builder_with_tags_and_actions_option = new Builder(
```
BenBirt: nit: we should camelCase this (and other variables)

tslint.json
```diff
@@ -1,3 +1,4 @@
+
```
BenBirt: can we check this out fresh from `master`?

api/commands/build.ts
canetcesc: Done

Issue #312, opened 2019-07-18
Support for Azure DW


lewish:
Just putting this here to keep track of it.

Azure DW AIUI looks like a SQL Server. It seems we probably want to integrate using this package: https://github.com/tediousjs/node-mssql

- [ ] Define protos for connection strings
- [ ] Add a SQLServerDBAdapter class
- [ ] Add a SQLServerAdapter, probably using Postgres as a starting point
- [ ] Extend integration tests to cover SQLServer

PR #318, opened 2019-07-22
initial commit, added sqldatawarehouse support

ichavez:
Hi,

I wanted to try dataform on Azure SQL Data Warehouse so I took a stab at writing the adapter.

Can you please review ?

I read #312 and wanted to name it SQLServerDBAdapter, but decided not to, since SQL Server and SQL Data Warehouse are different products, and although we can use the same driver to connect to both, they differ in features and relevant T-SQL code. 

For instance: https://docs.microsoft.com/en-us/sql/t-sql/statements/create-table-as-select-azure-sql-data-warehouse?view=aps-pdw-2016-au7

I implemented distributions, which are only supported on Data Warehouse and ended up naming it SQLDataWarehouseDBAdapter to avoid confusing people later on.

The "warehouse" type is sqldatawarehouse.

I know, those are horrible names, but ¯\_(ツ)_/¯.


api/dbadapters/sqldatawarehouse.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    return request
+          .query(statement)
+          .then(result => result.recordset);
+  }
+
+  public async evaluate(statement: string) {
+    let pool = await this.pool;
+    return pool.request()
```
BenBirt: as above, let's do async/await where possible

api/dbadapters/sqldatawarehouse.ts
```diff
@@ -0,0 +1,111 @@
...
+  public async tables(): Promise<dataform.ITarget[]> {
+    let result = await this.execute(`select table_name, table_schema from information_schema.tables`)
+    return result.map(row => ({
+      schema: row.table_schema,
+      name: row.table_name
+    }))
+  }
+
+  public table(target: dataform.ITarget): Promise<dataform.ITableMetadata> {
+    return Promise.all([
```
BenBirt: this can be something like:

```
const [columnRows, tableRows] = await Promise.all([ ... ]);
```

api/dbadapters/sqldatawarehouse.ts
```diff
@@ -0,0 +1,111 @@
...
+        return {
+          target,
+          type: results[1][0].table_type == "VIEW" ? "view" : "table",
+          fields: results[0].map(row => ({
+            name: row.column_name,
+            primitive: row.data_type,
+            flags: row.is_nullable && row.is_nullable == "YES" ? ["nullable"] : []
+          }))
+        };
+      } else {
```
BenBirt: suggest inverting this to a guard clause:

```
if (results[1].length === 0) {
  throw new ...
}
return ...
```

api/dbadapters/sqldatawarehouse.ts
```diff
@@ -0,0 +1,111 @@
...
+        throw new Error(`Could not find relation: ${target.schema}.${target.name}`);
+      }
+    });
+  }
+
+  public async prepareSchema(schema: string): Promise<void> {
+    return Promise.resolve().then(() =>
```
BenBirt: this can be simplified a bit:

```
public async prepareSchema(schema: string): Promise<void> {
  await this.execute(...);
}
```

api/vm/legacy_compiler.ts
```diff
@@ -26,6 +26,7 @@ const tableContextPropertyNames = [
   "disabled",
   "redshift",
   "bigquery",
+  "sqldatawarehouse",
```
BenBirt: i'd leave out touching any of the legacy stuff, we want to remove this and definitely no new users should be using it

cli/credentials.ts
```diff
@@ -38,6 +38,22 @@ export function getRedshiftCredentials() {
   );
 }
 
+export function getSQLDataWarehouseCredentials(): dataform.ISQLDataWarehouse {
+  const server = question("Enter your Server name (for example 'name.database.windows.net'):");
```
BenBirt: nit: uncapitalize Server?

core/adapters/index.ts
```diff
@@ -3,6 +3,7 @@ import { Tasks } from "../tasks";
 import { BigQueryAdapter } from "./bigquery";
 import { RedshiftAdapter } from "./redshift";
 import { SnowflakeAdapter } from "./snowflake";
+import { SQLDataWarehouseAdapter } from "./sqldatawarehouse";
```
BenBirt: nit: can you change imports where you find them to be fully-qualified? so in this case, `@dataform/core/adapters/sqldatawarehouse`

core/adapters/index.ts
```diff
@@ -26,11 +27,15 @@ export enum WarehouseType {
...
   POSTGRES = "postgres",
   REDSHIFT = "redshift",
-  SNOWFLAKE = "snowflake"
+  SNOWFLAKE = "snowflake",
+  SQLDATAWAREHOUSE = "sqldatawarehouse"
 }
 
 export function supportsCancel(warehouseType: WarehouseType) {
-  return warehouseType === WarehouseType.BIGQUERY;
+  if (warehouseType === WarehouseType.BIGQUERY || warehouseType === WarehouseType.SQLDATAWAREHOUSE)
```
BenBirt: suggest pulling this out to an array called `supportsCancel` or something, and then doing

`return !!supportsCancel.find(supportsCancel)`

core/adapters/sqldatawarehouse.ts
```diff
@@ -0,0 +1,109 @@
+import { dataform } from "@dataform/protos";
+import { Task, Tasks } from "../tasks";
+import { Adapter } from "./base";
+import { IAdapter } from "./index";
+
+export class SQLDataWarehouseAdapter extends Adapter implements IAdapter {
+  private project: dataform.IProjectConfig;
```
BenBirt: looks like you don't use this? in which case, let's remove it & the constructor.

core/adapters/sqldatawarehouse.ts
```diff
@@ -0,0 +1,109 @@
...
+  constructor(project: dataform.IProjectConfig) {
+    super();
+    this.project = project;
+  }
+
+  private createOrReplace(table: dataform.ITable) {
+    if (table.type === "view") {
+      return (
+        Tasks.create()
+        // Drop the view in case we are changing the number of column(s) (or their types).
```
BenBirt: do you know if this is required for SQL datawarehouse?

core/adapters/sqldatawarehouse.ts
```diff
@@ -0,0 +1,109 @@
...
+    return `"${target.schema}"."${target.name}"`;
+  }
+
+  public publishTasks(
+    table: dataform.ITable,
+    runConfig: dataform.IRunConfig,
+    tableMetadata: dataform.ITableMetadata
+  ): Tasks {
+    const tasks = Tasks.create();
+    // Drop the existing view or table if we are changing it's type.
```
BenBirt: nit: s/it's/its/

core/adapters/sqldatawarehouse.ts
```diff
@@ -0,0 +1,109 @@
...
+      .add(Task.statement(`
+        create view ${this.resolveTarget(target)}
+        as ${assertion.query}`))
+      .add(Task.assertion(`select sum(1) as row_count from ${this.resolveTarget(target)}`));
+  }
+
+  public dropIfExists(target: dataform.ITarget, type: string) {
+    if(type === "view")
+      return `drop ${this.baseTableType(type)} if exists ${this.resolveTarget(target)} `;
+    else
```
BenBirt: nit: rm the else, and just return the next line immediately

core/session.ts
```diff
@@ -153,7 +159,7 @@ export class Session {
     if (action.proto.target) {
       action.proto.target = this.target(
         actionOptions.sqlxConfig.name,
-        actionOptions.sqlxConfig.schema
+        schema
```
BenBirt: not sure what this change is for, can you clarify for me?

core/table.ts
```diff
@@ -225,6 +235,7 @@ export interface ITableContext {
   disabled: () => string;
   redshift: (redshift: dataform.IRedshiftOptions) => string;
   bigquery: (bigquery: dataform.IBigQueryOptions) => string;
+  sqldatawarehouse: (sqldatawarehouse: dataform.ISQLDataWarehouseOptions) => string;
```
BenBirt: i'd leave this out of `ITableContext`, these other methods are here mostly to support non-SQLX usage, as commented elsewhere i don't think we want to add to that legacy stuff. for SQLX and JS it should be enough to include it in `TConfig` and to have your new method defined on `Table`.

core/table.ts
```diff
@@ -121,6 +126,11 @@ export class Table {
     return this;
   }
 
+  public sqldatawarehouse(sqldatawarehouse: dataform.ISQLDataWarehouseOptions) {
```
BenBirt: suggest camel-casing like `sqlDataWarehouse`

core/table.ts
```diff
@@ -43,6 +44,7 @@ export interface TConfig {
   protected?: boolean;
   redshift?: dataform.IRedshiftOptions;
   bigquery?: dataform.IBigQueryOptions;
+  sqldatawarehouse?: dataform.ISQLDataWarehouseOptions;
```
BenBirt: camelcase this? `sqlDataWarehouse`

core/utils.ts
```diff
@@ -184,6 +184,18 @@ export function validate(compiledGraph: dataform.ICompiledGraph): dataform.IGrap
       validationErrors.push(dataform.ValidationError.create({ message, actionName }));
     }
 
+    // sqlserver config
+    if(!!action.sqldatawarehouse) {
```
BenBirt: nit: this can just be `(action.sqldatawarehouse && action.sqldatawarehouse.distribution)`

core/utils.ts
```diff
@@ -184,6 +184,18 @@ export function validate(compiledGraph: dataform.ICompiledGraph): dataform.IGrap
...
+      if(action.sqldatawarehouse.distribution){
+        let dist = action.sqldatawarehouse.distribution.toUpperCase();
+        let hash_pattern = new RegExp('HASH\\s*\\(\\s*\\w*\\s*\\)\\s*');
```
BenBirt: `hashPattern`, and can you make this a constant defined somewhere nearish the top of this file?

protos/profiles.proto
```diff
@@ -30,3 +30,13 @@ message BigQuery {
...
 }
+
+message SQLDataWarehouse {
+  string server = 2;
+  int32 port = 3;
+  string username = 4;
+  string password = 5;
+  string database_name = 6;
+
+  reserved 1;
```
BenBirt: don't bother with this, you can just renumber all of your fields since this protobuf has never been used / changed before.

api/dbadapters/sqldatawarehouse.ts
```diff
@@ -0,0 +1,111 @@
...
+      server: sqlDataWarehouseCredentials.server,
+      port: sqlDataWarehouseCredentials.port,
+      user: sqlDataWarehouseCredentials.username,
+      password: sqlDataWarehouseCredentials.password,
+      database: sqlDataWarehouseCredentials.databaseName,
+      encrypt: true
+    };
+    const conn = new sql.ConnectionPool(config).connect();
+
+    conn.then(pool =>{
```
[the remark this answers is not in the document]
ichavez: it isn't, since I'm just throwing the error, the recommendation from the library though is to always attach an error listener: https://www.npmjs.com/package/mssql#connections-1

core/adapters/sqldatawarehouse.ts
ichavez: yes, it is

core/session.ts
ichavez: Some other dbadapters have these line to handle assertTasks: 
const target =
      assertion.target ||
      dataform.Target.create({
        schema: projectConfig.assertionSchema,
        name: assertion.name
      });
however, assertion.target always had the default "dataform" schema, not the desired "dataform_assertions"

api/dbadapters/sqldatawarehouse.ts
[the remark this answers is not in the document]
BenBirt: thanks for the link!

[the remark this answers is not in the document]
BenBirt: taking a slightly deeper look... there is a bit of a race here, i believe. if `execute` is called before this code is, no error handler will be attached before some real code is run.

i think we want to do something like:

this.pool = new Promise((resolve) => {
  const conn = new sql.ConnectionPool(config).connect();
  conn.then(pool => {
    pool.on('error', err => {
      throw new Error(err);
    });
    resolve(conn);
  });
});

core/adapters/index.ts
```diff
@@ -52,12 +57,20 @@ const requiredSnowflakeWarehouseProps: Array<keyof dataform.ISnowflake> = [
   "databaseName",
   "warehouse"
 ];
+const requiredSQLDataWarehouseProps: Array<keyof dataform.ISQLDataWarehouse> = [
+  "server",
```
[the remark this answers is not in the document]
ichavez: I would leave it as is to be consistent with Microsoft's documentation and the mssql library.
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/mssql/index.d.ts#L189

api/dbadapters/sqldatawarehouse.ts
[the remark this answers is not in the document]
ichavez: but, in the future if you decide to handle errors in the db adapters, it would just be clear where's the place to do it.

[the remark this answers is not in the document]
ichavez: got it (race condition), fixed.

api/dbadapters/sqldatawarehouse.ts
ichavez: this code is exactly the same for redshift and snowflake. i dont want sql to be the odd man out :). if you want to refactor to create a base for sql adapters?

core/table.ts
ichavez: i think i fixed this everywhere in the code, except in the config. The config would look like this:
```
config {
    type: "table",
    sqldatawarehouse: {
        distribution: "round_robin"
    },
    dependencies:["example"]
}
SELECT 1 AS test
```
I think I did it to be consistent with "bigquery": https://docs.dataform.co/guides/warehouses/bigquery/

But, let me know if you want me to change it.

core/session.ts
```diff
@@ -120,6 +128,8 @@ export class Session {
       return this.test(actionOptions.sqlxConfig.name).dataset(actionOptions.sqlxConfig.dataset);
     }
 
+    let schema = actionOptions.sqlxConfig.schema;
```
BenBirt: for some reason i can't respond to the existing comment thread on this, github not being helpful...

anyways: thanks a lot for finding this bug! would you mind merging in https://github.com/dataform-co/dataform/pull/324 to fix? it includes a unit test for the fix too.

BenBirt: Don't worry about this one any more - I wanted to get that fix in, so have merged it, if you pull from master (and revert your changes to this file) everything should be good now.

BenBirt:
Hey Isaac - we'd love to get this support in, it'd be a great addition. Unless you have any objections, I'm going to clone your branch and merge it into Dataform - feel free to get in touch on dataform-users.slack.com otherwise!

BenBirt:
Just to round this out - we merged in support for Azure SQL data warehouse in https://github.com/dataform-co/dataform/commit/38c149b14733217dc423dade968c02103448774f.

lewish:
Closing this now, as it's been merged in another PR!

PR #319, opened 2019-07-22
Change tslint.json props to allow use of console.log()

canetcesc:
This allows using `console.log()` statements eg. in `built.ts`, which I need for debugging purposes.

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -1,9 +1,9 @@
 {
-    "extends": ["tslint:recommended", "tslint-react", "tslint-config-prettier"],
-    "rules": {
-      "object-literal-sort-keys": false,
-      "max-classes-per-file": [false, 1],
-      "no-floating-promises": true
-    }
+  "extends": ["tslint:recommended", "tslint-react", "tslint-config-prettier"],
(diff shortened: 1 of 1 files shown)


BenBirt:
TBH I am a bit wary of turning this off -- I spent a bit of time before we launched the CLI removing a bunch of `console.log` calls (so that they didn't pollute CLI output) - I value this check as it's an indicator that this sort of thing should be removed (after debugging is complete).

What do you think?

canetcesc:
Sure, I was not aware of these issues, so let's reject this Pull Request for now.

By the way, it seems possible to configure what specific console verbs are alllowed, eg:

`"no-console": [ true, "log", "debug", "info", "time", "timeEnd", "trace" ]`

I just find myself having to turn this setting to `false` again and again on every new branch, which is daunting!

Perhaps if I knew how to configure the debugger I would be able to inspect breakpoints and avoid this  need. Maybe you can share with me your `.vscode/launch.json` file so I can see how you do it?

canetcesc:
Closed because it is problematic. See comments above.

Issue #307, opened 2019-07-16
Add documentation for `dataform help` command


BenBirt:
And perhaps more generally other useful commands/flags that aren't documented very well currently!

@canetcesc for visibility

BenBirt:
Fixed in https://github.com/dataform-co/dataform/pull/323

Issue #228, opened 2019-05-29
Add documentation to cover the --schema-suffix flag


lewish:
We should document the schema suffix flag 
- [ ] What it is used for
- [ ] How to use it

canetcesc:
Hi @lewish, I am taking issue #307 now (pull request #323 ).
Is this issue only related to CLI help? If so, I can piggyback the changes on that one.
Could you explain what to put down on the CLI  next to the `--schema-suffix` option of the `dataform run` command? - I don't know what is the meaning of this flag!

PR #323, opened 2019-07-23
Add yargs doc for `dataform help` command. Few fixes. [#307]

canetcesc:
Let me know if there are any other commands which require changing the CLI documentation.

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -141,6 +141,17 @@ const verboseOutputOption: INamedOption<yargs.Options> = {
 const builtYargs = createYargsCli({
   commands: [
     {
+      // This dummy command is a hack with the only goal of displaying "help" as a command in the CLI
+      // and we need it because of the limitations of yargs considering "help" as an option and not as a command.
+      format: "help [command]",
+      description: "Show help. If [command] is specified, the help is for the given command.",
+      positionalOptions: [],
+      options: [],
+      processFn: async argv => {
+        return false;
+      }
+    },
+    {
       format: "init <warehouse> [project-dir]",
       description: "Create a new dataform project.",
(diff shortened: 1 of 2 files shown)


BenBirt:
Thanks Cesc! Just to confirm, this doesn't break `dataform help` or `dataform help <command>`? (yargs doesn't see this as an override of the hack I added to make that work?)

If you're up for it, a small edit to the relevant docs page (https://github.com/dataform-co/dataform/blob/master/docs/pages/guides/command-line-interface.mdx, https://docs.dataform.co/guides/command-line-interface/) to add some details on `dataform help` would also be great, don't feel you need to though.

canetcesc:
1. The CLI seems to work just fine with the new change. For example we can test

`dataform help compile`:

Screenshot 2019-07-24 at 11 08 41

`dataform --help compile`:

Screenshot 2019-07-24 at 11 21 04

`dataform help`:
Screenshot 2019-07-24 at 11 09 54

`dataform --help`:

Screenshot 2019-07-24 at 11 22 45


**Fun fact:** If you try to run a `dataform help help` or `dataform --help help`  "meta-command" you will get the same output as if you just run `dataform help`. That is because the hacky `fixArgvForHelp()` function  will intercept it.  So, the description added in `index.ts` (Line 147) is never displayed to the user.

2. Sure, will add that to the docs.

3. Take a look at my comments on https://github.com/dataform-co/dataform/issues/228, we may want to document `--schema-sufix` in the same change.

lewish:
Hi Cesc, apologies for the delay here, Ben is on holiday :)

canetcesc:
Thanks, @lewish! Be aware that I do not have permissions to merge this, only you can.

PR #324, opened 2019-07-24
Fix assertion schema names in SQLX.

BenBirt:


core/session.ts
```diff
@@ -151,10 +151,12 @@ export class Session {
       }
     })();
     if (action.proto.target) {
-      action.proto.target = this.target(
-        actionOptions.sqlxConfig.name,
-        actionOptions.sqlxConfig.schema
-      );
+      const finalSchema =
+        actionOptions.sqlxConfig.schema ||
+        (actionOptions.sqlxConfig.type === "assertion"
```
faiyaz26: Can we have a enum for this types ?

BenBirt: We actually already enforce typing with this: https://github.com/dataform-co/dataform/blob/79268d531702599f984206b0546a639de03698b5/core/session.ts#L16

Issue #280, opened 2019-06-26
Allow users to have the same table name under different schema names


BenBirt:
We'll need to:
- key actions by their fully-qualified `schema.name` (or 'target')
- allow the user to specify `schema.name` in a call to `ref()`
- update ref() to handle the case where just passing `name` (and not `schema.name`) is ambiguous - i.e. where there are indeed >1 tables named `name` - by throwing an error

PR #337, opened 2019-08-02
Key actions by fully qualified name [Issue #280]

canetcesc:
The change is not yet ready for code review but I am creating a pull request for visibility as I'd like to discuss about it either tomorrow or on Monday.

Some things that I still need to understand are:

**[1]** Should wildcards return objects on other schemas?  Given two tables `a.tab_1`, `b.tab_2` and  an assertion `a.a` with a dependency on `tab_*`, there are three possible outcomes:
    
a) `a.tab_1` and `b.tab_2` are returned because the schema is ignored.

b)  Only `a.tab_1` is returned because it is in the same schema as `a.a`.

c) Neither `a.tab_1` or `b.tab_2` are returned because there are no schema names starting with `tab_`. A user can only use wildcards within a schema, for example using `a.tab_*` or `b.tab_*`, or if schema `schema1` and `schema2` are  valid schemas then `schema*` would be a valid expression too.

**[2]** Can we avoid adding a `default_schema` string inside the `CompileConfig` proto ?

**[3]** How to access `ProjectConfig.default_schema` from  `compilers.ts`. Unless this is achieved we need to add a new argument `defaultSchema: string` into `compileSqlx(results: ISqlxParseResults, path: string, defaultSchema: string)` and `compile(code: string, path: string, defaultSchema: string)`.

**[4]** How should a fully qualified name interact with the schema suffix? As of now the suffix-related `api` test cases (`api.specs.ts`) are failing.


canetcesc:
This PR has been substituted by https://github.com/dataform-co/dataform/pull/352.

PR #339, opened 2019-08-05
Adds an error when the dataform.json file is invalid

GJMcGowan:
Working towards solving #1272

### DIFF
--- a/api/commands/compile.ts
+++ b/api/commands/compile.ts
@@ -16,9 +16,7 @@ export async function compile(
     const dataformJson = fs.readFileSync(`${compileConfig.projectDir}/dataform.json`, "utf8");
     JSON.parse(dataformJson);
   } catch (e) {
-    const error = new Error("Compile Error: `dataform.json` is invalid");
-    error.stack = "Compile Error: `dataform.json` is invalid";
-    throw error;
+    throw new Error("Compile Error: `dataform.json` is invalid");
   }
 
   const returnedPath = await CompileChildProcess.forkProcess().compile(compileConfig);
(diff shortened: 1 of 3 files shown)


core/gen_index.ts
```diff
@@ -32,12 +32,23 @@ export function genIndex(base64EncodedConfig: string): string {
     })
     .join("\n");
 
+  const safeIncludeDataformJson = `
```
BenBirt: can we instead just do a JSON parse of dataform.json, in native typescript code?

api/commands/compile.ts
```diff
@@ -10,6 +10,17 @@ export async function compile(
 ): Promise<dataform.CompiledGraph> {
   // Resolve the path in case it hasn't been resolved already.
   path.resolve(compileConfig.projectDir);
+
+  try {
+    // check dataformJson is valid before we try to compile
+    const dataformJson = fs.readFileSync(`${compileConfig.projectDir}/dataform.json`, "utf8");
+    JSON.parse(dataformJson);
+  } catch (e) {
+    const error = new Error("Compile Error: `dataform.json` is invalid");
```
BenBirt: this stacktrace is probably a lot cleaner... can we just `throw new Error("Contents of 'dataform.json' file are invalid.");`?

version.bzl
```diff
@@ -1,3 +1,3 @@
 # NOTE: If you change the format of this line, you must change the bash command
 # in /scripts/publish to extract the version string correctly.
-DF_VERSION = "1.0.9"
+DF_VERSION = "1.0.10"
```
BenBirt: no need for this now.

GJMcGowan: Isn't there? If I want to require this in `dataform-co` do I not refer to the version number? I guess I don't really understand the release process for this.

PR #341, opened 2019-08-07
Add support for documenting SQLX actions.

BenBirt:


version.bzl
```diff
@@ -1,3 +1,3 @@
 # NOTE: If you change the format of this line, you must change the bash command
 # in /scripts/publish to extract the version string correctly.
-DF_VERSION = "1.0.10"
+DF_VERSION = "1.0.11"
```
lewish: 1.1.0 ? New functionality like this is probably a minor version.

lewish: Unless there are other things left

PR #351, opened 2019-08-08
Azure SQL Data Warehouse support.

BenBirt:
from https://github.com/dataform-co/dataform/pull/318

for https://github.com/dataform-co/dataform/issues/312


core/adapters/sqldatawarehouse.ts
```diff
@@ -0,0 +1,113 @@
...
+        as ${assertion.query}`)
+      )
+      .add(Task.assertion(`select sum(1) as row_count from ${this.resolveTarget(target)}`));
+  }
+
+  public dropIfExists(target: dataform.ITarget, type: string) {
+    if (type === "view") {
+      return `drop ${this.baseTableType(type)} if exists ${this.resolveTarget(target)} `;
+    }
+    return `if object_id ('${this.resolveTarget(
```
lewish: :O

core/adapters/sqldatawarehouse.ts
```diff
@@ -0,0 +1,113 @@
...
+      target
+    )}','U') is not null drop table ${this.resolveTarget(target)}`;
+  }
+
+  private createOrReplace(table: dataform.ITable) {
+    if (table.type === "view") {
+      return (
+        Tasks.create()
+          // Drop the view in case we are changing the number of column(s) (or their types).
+          .add(Task.statement(this.dropIfExists(table.target, this.baseTableType(table.type))))
```
lewish: It would be much nicer to user `alter view` if we can. I did this for redshift because there was no other way, but this increases downtime for consumers while we run.

I think this is as simple as changing `create` to `alter` when the view exists, i just did a test manually and it seems to work.

Issue #247, opened 2019-06-05
Replace ValidationErrors with CompilationErrors


BenBirt:
Validation errors are essentially errors which shouldn't be rendered alongside a stacktrace, so let's replace `ValidationError` with `CompilationError` (just don't set stacktraces in these cases).

This will require a bit of plumbing to get filenames through to the validation code.

canetcesc:
I've sort of started this in #352 . Makes sense for me to end it.

Issue #107, opened 2019-03-14
"api" package should not depend on "core"


BenBirt:
core is *only* responsible for running the JS in user's project files.

api is responsible for everything except that, i.e. connecting to the user's data warehouse, running the compiled graph, etc.

Projects themselves should only need to depend on core. Developers using dataform should only need to depend on the api (but not core).

Let's make a strict line where the api and core packages only communicate via the shared protobufs, and api has strictly no dependencies on core. Any shared code can be pulled out to a utils package.

canetcesc:
I'd suggest setting this issue as depending upon the completion of issue #247: if we can remove all the validation that occurs in `core/utils.ts` then there will be less of a dependency on core from api.

Ekrekr:
Closing as @dataform/api is no longer being published

PR #352, opened 2019-08-08
Key Actions by Fully Qualified Name [New] [Issue #280]

canetcesc:
This is a replacement of https://github.com/dataform-co/dataform/pull/337.


core/assertion.ts
```diff
@@ -78,13 +78,17 @@ export class AssertionContext {
     this.assertion = assertion;
   }
 
-  public ref(name: string) {
+  public ref(reference: string | IResolvable) {
+    const name =
+      typeof reference === "string" || typeof reference === "undefined"
```
BenBirt: i don't think we should add explicit handling of `undefined` here - if anything such a value should throw (but I just wouldn't bother adding code explicitly for that case)

core/assertion.ts
```diff
@@ -78,13 +78,17 @@ export class AssertionContext {
...
+        ? reference
+        : (reference as IResolvable).schema + "." + (reference as IResolvable).name;
```
BenBirt: you *shouldn't* need to type-assert `reference` as `IResolvable`: typescript should be clever enough to know that if `reference` is not a string (as in this case), it must be `IResolvable`

(typescript's type system is crazy good)

core/session.ts
```diff
@@ -191,6 +196,7 @@ export class Session {
           ? this.config.assertionSchema
           : this.config.defaultSchema);
       action.proto.target = this.target(actionOptions.sqlxConfig.name, finalSchema);
+      action.proto.name = action.proto.target.schema + "." + action.proto.target.name;
```
BenBirt: after this PR, `name` and `target` are equivalently useable for uniqueness. I think we should:
1. put a comment on the `name` field to indicate that it is deprecated, and to prefer using the `target` field
2. stop unsetting the `target` field in the case of `Operation` queries
3. (as a direct result of (2)) change `resolve` to throw if you try to resolve an `Operation` with `hasOutput = false`

core/session.ts
```diff
@@ -210,10 +216,19 @@ export class Session {
     });
   }
 
-  public resolve(name: string): string {
-    const table = this.tables[name];
+  public resolve(reference: string | IResolvable): string {
+    let [fQName, err] = [null, null];
+    if (typeof reference === "string") {
+      [fQName, err] = utils.matchFQName(reference, this.getAllFQNames());
```
BenBirt: i think we can simplify this a lot, especially if we also do my suggestion above (to always set the `target` field):

when trying to `resolve` a `reference`, iterate through all actions (tables, operations, assertions) looking for actions whose `target.name` matches `reference` (if it's a string) or whose `target` matches `reference` (if it's an `IResolvable`). collect all matching actions. if matching action count is >1, throw a compile error complaining that the `reference` is not specific enough.

we can define all of this as a helper method on `Session` and then callers such as `Table.dependencies` can reuse it.

core/session.ts
```diff
@@ -347,14 +363,7 @@ export class Session {
     allActions.forEach(action => {
       const uniqueDependencies: { [dependency: string]: boolean } = {};
```
BenBirt: this can be simplified now:

`action.dependencies = [...new Set(action.dependencies || [])];`

core/test.ts
```diff
@@ -24,7 +25,26 @@ export class Test {
     return this;
   }
 
-  public dataset(datasetToTest: string) {
+  public dataset(reference: string | IResolvable) {
+    var datasetToTest = null;
```
BenBirt: i don't think we need this code - let's just assign `reference` to `this.datasetToTest` and resolve the dataset inside `compile()`

core/table.ts
```diff
@@ -135,8 +141,11 @@ export class Table {
   public dependencies(value: string | string[]) {
     const newDependencies = typeof value === "string" ? [value] : value;
     newDependencies.forEach(d => {
-      const table = this.session.tables[d];
-
+      const [fQd, err] = utils.matchFQName(d, this.session.getAllFQNames());
```
BenBirt: this code is made complicated because dependency actions may not yet have been declared (when `dependencies()` is called). i think that we should collect all dependencies in `dependencies()` and then finally resolve them all in `compile()` (by which time all actions should have been declared).

core/utils.ts
```diff
@@ -116,20 +116,28 @@ export function validate(compiledGraph: dataform.ICompiledGraph): dataform.IGrap
...
   allActions.forEach(action => (actionsByName[action.name] = action));
 
-  // Check all dependencies actually exist.
+  // Check all dependencies actually exist and are not ambiguous.
   allActions.forEach(action => {
     const actionName = action.name;
     (action.dependencies || []).forEach(dependency => {
-      if (allActionNames.indexOf(dependency) < 0) {
-        const message = `Missing dependency detected: Node "${action.name}" depends on "${dependency}" which does not exist.`;
+      const [matchedDep, err] = matchFQName(dependency, allActionNames);
```
BenBirt: if we move dependency resolution into `compile()` functions as discussed above, we shouldn't need this code (i believe!).

docs/_methods.ts
```diff
@@ -3,14 +3,20 @@ import { Props as MethodProps } from "./components/method";
 export const methods: { [name: string]: MethodProps } = {
   ref: {
     name: "ref()",
-    signatures: ["ref(tableName)"],
+    signatures: ["ref(name)", 'ref({schema: "schema", name: "name"})'],
```
BenBirt: don't worry about changing this file, it's super out of date and not used in the documentation anyway. we should probably delete it, but i think we're still thinking about the right way of doing this documentation.

core/assertion.ts
canetcesc: Ref'ing `undefined` throws an error but it still adds the dependency to the list (see https://github.com/dataform-co/dataform/blob/58a5226eb5a8b57ae0e16bdd1852fb41a2ae6736/tests/core/core.spec.ts#L370).
Anyway, I've changed the code to make it look more elegant. Thanks.

core/assertion.ts
canetcesc: Done

core/session.ts
```diff
@@ -66,6 +66,11 @@ function mapColumnDescriptionToProto(
   );
 }
 
+export interface IResolvable {
```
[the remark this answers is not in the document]
canetcesc: Done

core/session.ts
canetcesc: Done. Notice that as a corollary of this, the following test will be changed:
https://github.com/dataform-co/dataform/blob/58a5226eb5a8b57ae0e16bdd1852fb41a2ae6736/tests/api/api.spec.ts#L824

core/session.ts
canetcesc: 1. This helper method which you mention is I _believe_ what I have called `utils.matchFQName`.  Unfortunately, it is also needed by the `validate` method to check for circular dependencies and ambiguous dependencies on validation time.  Since our long-term goal is to get rid of the validation process altogether and isolate core and api packages (see Issues https://github.com/dataform-co/dataform/issues/247, https://github.com/dataform-co/dataform/issues/107), I'd suggest we keep it as it is and once those two issues are addressed we can hopefully move `matchFQName`  from `utils` into `Session`. (I believe we _might_ be able to totally get rid of `utils` but that's another story).

2. But yes, I think `matchFQName` can now be largely simplified by using `{schema: string, name: string}` as a parameter, because now we can assume there will always be a target. I'll do that!

docs/_methods.ts
canetcesc: Done

core/test.ts
canetcesc: Done

core/utils.ts
canetcesc: That is right but then we need to change the validation errors into compilation errors and change the test cases accordingly. And it is not only this one but also circular dependencies, etc. Since this change is already quite large, I didn't want to mess too many things up. Let's do that on Issue https://github.com/dataform-co/dataform/issues/247 once and for all.

core/session.ts
canetcesc: ES6 is fantastic! Done.

core/table.ts
canetcesc: OK, will do. I see a pattern here where all action types should throw compilation errors only when `action.compile()` is run, and it makes sense to me.

A bit off-topic but for consistency reasons _perhaps_ the `dependencies` method should allow `Resolvable`'s or (arrays of) and become:

`public dependencies(ref: Resolvable | Resolvable[])`

Same with `preOps`, `postOps`, etc.

canetcesc:
@BenBirt: Can you CR this one? -- Everything is ready now except for the integration tests that are still failing. As per our slack discussion, I have changed:

1. All the occurences of `actionMap.XXX` to `actionMap.["df_integration_test.XXX"]`.
2. Removed the `compilationError()` which was created upon creation of an undefined / null reference. The error happens now as a `validationError`, i.e., exactly as it was before.

At this point I am sure it must be something super small missing and since I can't see the logs I think it will be far easier if you take a look at fix it + CR all the rest.

Talk to you tomorrow,

Cesc

core/session.ts
```diff
@@ -190,17 +193,13 @@ export class Session {
...
-      const finalSchema =
-        actionOptions.sqlxConfig.schema ||
-        (actionOptions.sqlxConfig.type === "assertion"
-          ? this.config.assertionSchema
-          : this.config.defaultSchema);
-      action.proto.target = this.target(actionOptions.sqlxConfig.name, finalSchema);
-    } else {
-      delete action.proto.target;
-    }
+    const finalSchema =
```
BenBirt: i have a suspicion that these lines that fix `target` and `name` for overridden `schema` settings are no longer needed after your https://github.com/dataform-co/dataform/pull/368. can you remove these lines and check the tests still pass?

core/session.ts
```diff
@@ -217,50 +216,66 @@ export class Session {
...
+      : this.target(typeof ref === "string" ? ref : ref.name);
     return this.adapter().resolveTarget(target);
   }
 
   public operate(name: string, queries?: OContextable<string | string[]>): Operation {
-    this.checkActionNameIsUnused(name);
     const operation = new Operation();
     operation.session = this;
-    operation.proto.name = name;
     operation.proto.target = this.target(name);
```
BenBirt: these 4 lines are common to all actions. suggest:

```
private setNameAndTarget(action: IActionProto, name: string) {
  action.target = this.target(name);
  this.checkTargetIsUnused(action.target);
  action.name = `${action.target.schema}.${action.target.name}`;
}
```

(note that `IActionProto` will probably have to change to include the `target` field)

core/session.ts
```diff
@@ -335,34 +346,32 @@ export class Session {
...
-        .forEach(dependency => (uniqueDependencies[dependency] = true));
-      // Match wildcard deps against all action names.
-      utils
-        .matchPatterns(dependencies.filter(d => d.includes("*")), allActionNames)
-        .forEach(dependency => (uniqueDependencies[dependency] = true));
-      action.dependencies = Object.keys(uniqueDependencies);
+      const fQDeps = action.dependencies.map(act => {
+        const allActs = this.findActions(act);
+        return allActs.length === 1
+          ? allActs[0].proto.target.schema + "." + allActs[0].proto.target.name
```
BenBirt: `allActs[0].name` would be cleaner as it means we aren't duplicating this string concatenation logic

core/session.ts
```diff
@@ -335,34 +346,32 @@ export class Session {
...
+          : act;
```
BenBirt: presumably we should throw a compileError if `allActs.length > 1`?

core/assertion.ts
```diff
@@ -66,7 +66,13 @@ export class Assertion {
   }
 
   public schema(schema: string) {
-    this.proto.target = this.session.target(schema, this.session.config.assertionSchema);
+    if (this.session.findActions({ schema: schema, name: this.proto.target.name }).length === 0) {
```
BenBirt: partly to prevent code duplication, and partly because we now have a bunch of different circumstances where action schemas/names can be updated fairly dynamically prior to compile time, i think we should move these checks out of each Action, plus calling `checkTargetIsUnused` inside `session.ts` into a single place: compile time. I think we should run `checkTargetIsUnused` inside `session.compileGraphChunk`, and only in there.

core/assertion.ts
```diff
@@ -86,13 +92,14 @@ export class AssertionContext {
     this.assertion = assertion;
   }
 
-  public ref(name: string) {
+  public ref(ref: Resolvable) {
+    const name = typeof ref === "object" ? ref.schema + "." + ref.name : ref;
```
BenBirt: nit: invert this and use `typeof ref === "string"? (in a few places)

core/operation.ts
```diff
@@ -125,20 +136,33 @@ export class OperationContext {
...
+    return this.resolve(ref);
   }
 
-  public resolve(name: string) {
-    return this.operation.session.resolve(name);
+  public resolve(ref: Resolvable) {
+    if (this.hasOutput) {
+      return this.operation.session.resolve(ref);
+    } else {
+      this.operation.session.compileError(
```
BenBirt: i think this is incorrect - operations should always be able to `resolve` anything. what's invalid is for somebody to `ref` or `resolve` an operation which has `hasOutput: false`. (which is checked in `session.ts)

core/table.ts
```diff
@@ -149,10 +155,10 @@ export class Table {
   public dependencies(value: string | string[]) {
     const newDependencies = typeof value === "string" ? [value] : value;
     newDependencies.forEach(d => {
-      const table = this.session.tables[d];
-
-      if (!!table && table.proto.type === "inline") {
-        table.proto.dependencies.forEach(childDep => this.addDependency(childDep));
+      const allResolved = this.session.findActions(d);
+      const resolved = allResolved.length > 0 ? allResolved[0] : undefined;
+      if (!!resolved && resolved instanceof Table && resolved.proto.type === "inline") {
```
BenBirt: this is not due to you, but i just realised this code exists and is a bit crazy, i am pretty sure it doesn't work in at least some circumstances! can you add a `// TODO: This code fails to function correctly if the inline table has not yet been attached to the session. This code probably needs to be moved to compile().`?

core/table.ts
```diff
@@ -214,6 +225,19 @@ export class Table {
     });
     this.contextablePostOps = [];
 
+    //Check dependencies are not ambiguous
```
BenBirt: we have this in session.ts too, right? i think this can be removed in favour of that check, which also happens at compile()-time.

core/test.ts
```diff
@@ -24,8 +24,12 @@ export class Test {
     return this;
   }
 
-  public dataset(datasetToTest: string) {
-    this.datasetToTest = datasetToTest;
+  public dataset(ref: Resolvable) {
+    const target =
```
BenBirt: let's not do this resolution here. instead, do it in compile(), when all actions must have been attached to the `session`. then look up the relevant `datasetToTest` in the session.

edit: in fact, you already do this! let's just remove this code; compile() does the right thing already. this can just be `this.datasetToTest = ref;`

core/utils.ts
```diff
@@ -1,5 +1,6 @@
 import { dataform } from "@dataform/protos";
 import { DistStyleTypes, ignoredProps, SortStyleTypes, TableTypes } from "./table";
+import { Resolvable } from "./session";
```
BenBirt: please prefer fully-qualified imports, i.e. `@dataform/core/session`

core/utils.ts
```diff
@@ -116,20 +117,34 @@ export function validate(compiledGraph: dataform.ICompiledGraph): dataform.IGrap
...
   allActions.forEach(action => (actionsByName[action.name] = action));
 
-  // Check all dependencies actually exist.
+  // Check all dependencies actually exist and are not ambiguous.
   allActions.forEach(action => {
     const actionName = action.name;
     (action.dependencies || []).forEach((dependency: string) => {
-      if (allActionNames.indexOf(dependency) < 0) {
-        const message = `Missing dependency detected: Node "${action.name}" depends on "${dependency}" which does not exist.`;
+      const [matchedDep, err] = matchFQName(dependency, allActionNames);
```
BenBirt: i don't think we need to do any matching here, right? my understanding of the code in `session.ts` is that it now fully-resolves dependencies. all we should need to do is something like:

```
if (!allActionNames.includes(dependency)) {
  // add a validation error
}
```

core/utils.ts
```diff
@@ -116,20 +117,34 @@ export function validate(compiledGraph: dataform.ICompiledGraph): dataform.IGrap
...
+      if (!!err && err.includes("Ambiguous")) {
+        const message = `Ambiguous dependency detected: ` + err;
```
BenBirt: isn't this impossible, isn't this checked in `session.ts`?

core/utils.ts
```diff
@@ -148,7 +163,14 @@ export function validate(compiledGraph: dataform.ICompiledGraph): dataform.IGrap
       validationErrors.push(dataform.ValidationError.create({ message, actionName }));
       return true;
     }
-    return (action.dependencies || []).some(d => {
+    const depsFQ: string[] = [];
+    (action.dependencies || []).forEach((d: string) => {
+      const [fQd, err] = matchFQName(d, allActionNames);
```
BenBirt: ditto, not sure this code is needed any longer, the dependencies are already fully-qualified and necessarily match action names.

tests/integration/bigquery.spec.ts
```diff
@@ -59,35 +59,51 @@ describe("@dataform/integration/bigquery", () => {
...
     let incrementalRows = await getTableRows(incrementalTable.target, adapter, dbadapter);
     expect(incrementalRows.length).equals(1);
 
     // Re-run some of the actions.
     executionGraph = await dfapi.build(
       compiledGraph,
-      { actions: ["example_incremental", "example_table", "example_view"] },
+      {
+        actions: [
+          "df_integration_test.example_incremental",
```
BenBirt: let's keep these unqualified, please - we need that to continue working. (of course if the name is ambiguous then we should throw an error somewhere). (same elsewhere)

core/session.ts
canetcesc: I’d rather not use any deprecated fields.

core/session.ts
canetcesc: Perhaps I should have mentioned this on Slack before implementing it but this line is here because I added the ability to refer to datasets in tests using `Resolvable`. Without this, datasets in tests cannot refer to datasets in arbitrary schemas. See test case `bigquery_language_v2/definitions/example_test_case_fq_ref.sqlx` in this same CR:

````
config {
    type: "test",
    dataset: {schema: "df_integration_test", name: "example_table"}
}
````

In fact I believe we will need to add the ability to use `Resolvable`'s in other places like `dependencies`, `preOps`, `postOps`, etc. - Maybe I should create an issue or something?

File: core/table.ts (hunk 18 above)
Comment (canetcesc) [reply core/table.ts]: Yes absolutely, in fact I already mentioned this to you as a potential problem in the previous CR:
https://github.com/dataform-co/dataform/pull/352#discussion_r312616391

File: core/assertion.ts (hunk 16 above)
Comment (canetcesc) [reply core/assertion.ts]: There is an actual reason why I compare it this way. The default case should include `undefined` so that you can do a `ref(undefined)`:

https://github.com/dataform-co/dataform/blob/c34588e0dd0057f1767b6d7517cae95a2656e379/tests/core/core.spec.ts#L382

But I see your point in doing this in a more type-safe fashion. I will break down the `Resolvable` part (the non-string one) into a different type, like this:

```
export type Resolvable = string  | {  schema: string;  name: string; };
```

into

````
export type FullyQualifiedName = {schema: string, name:string};
export type Resolvable = string | FullyQualifiedName;
````

File: core/test.ts (hunk 20 above)
Comment (canetcesc) [reply core/test.ts]: See my comments above about accepting `Resolvable` as an argument to `test.dataset()`. This was completely outside of the specs of the original issue so I am fine removing it.

File: core/session.ts (hunk 11 above)
Comment (BenBirt) [reply core/session.ts]: Can you clarify why this is required for tests to reference `Resolvable` datasets?

As far as I understand it:
- a dataset action registers itself with the `session`, possibly overriding its `schema` and/or `name`
- a test references a dataset, using a `Resolvable`
- at test `compile()`-time, the test can look up all actions in the `session`, searching for a dataset which matches the `Resolvable`

I guess I'm not clear on what these lines do that are required to make that work!

File: core/utils.ts (hunk 24 above)
Comment (canetcesc) [reply core/utils.ts]: It was required because the test case was not well designed. I have changed the definition of the test case and now this line can go.

File: core/utils.ts (hunk 22 above)
Comment (canetcesc) [reply core/utils.ts]: That's correct. Good catch!

File: core/session.ts (hunk 11 above)
Comment (canetcesc) [reply core/session.ts]: Sorry Ben, that was a mistake from my part. This has nothing to do with the `Resolvable` on `Test`, however as you correctly point out if we take out this line we get an error on this test case:

`api.spec.ts`, `L849`:
`expect(exampleUsingOverriddenSchema).to.not.be.undefined;`

I also think that this smells bad and the override schema should already been captured.

File: core/session.ts [hunk 26]
```diff
@@ -190,17 +193,13 @@ export class Session {
...
-    } else {
-      delete action.proto.target;
-    }
+    const finalSchema =
+      actionOptions.sqlxConfig.schema ||
+      (actionOptions.sqlxConfig.type === "assertion"
+        ? this.config.assertionSchema
+        : this.config.defaultSchema);
+    action.proto.target = this.target(actionOptions.sqlxConfig.name, finalSchema);
+    action.proto.name = action.proto.target.schema + "." + action.proto.target.name;
```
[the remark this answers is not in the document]
canetcesc: Done

core/session.ts
```diff
@@ -217,50 +216,66 @@ export class Session {
...
-    const operation =
-      !!this.operations[name] && this.operations[name].hasOutput && this.operations[name];
+  public resolve(ref: Resolvable): string {
+    const allResolved = this.findActions(ref);
+    if (allResolved.length > 1) {
+      const msg =
+        "Ambiguous Action name: " +
+        ref +
+        ". Did you mean one of: [" +
+        allResolved.join(", ") +
```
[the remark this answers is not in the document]
canetcesc: Fixed. Thx.

core/session.ts
canetcesc: This is probably captured elsewhere. The duplicity is checked in `checkTargetIsUnused` and in validation phase as a validation error. The tests pass, but I agree with you that this looks like the suitable place for throwing that error. I think there will be a lot more clarity as soon as I start moving the validation errors into compile.

core/assertion.ts
canetcesc: I'm a bit confused here. Earlier (https://github.com/dataform-co/dataform/pull/352#discussion_r314660766) you suggested to put `checkTargetIsUnused` inside `setNameAndTarget` and that is where it is located now.
I anyway agree with you that it would be better if that check happened just once and inside `compileGraphChunk` because that one is run for each and every action so the code would look cleaner and easier to maintain.

I actually tried to remove it from `setNameAndTarget` and put it only inside `compileGraphChunk` but it did not work (many of the duplicate action tests failed). Here is the code which I tested:

````
  public compileGraphChunk<T>(actions: Array<{ proto: IActionProto; compile(): T }>): T[] {
    const compiledChunks: T[] = [];

    actions.forEach(action => {
      try {
        if (!(action instanceof test.Test)) {
          this.checkTargetIsUnused(action.proto.target); // <---- THIS DID NOT WORK
        }
        const compiledChunk = action.compile();
        compiledChunks.push(compiledChunk);
      } catch (e) {
        this.compileError(e, action.proto.fileName);
      }
    });

    return compiledChunks;
  }
````

File: core/session.ts (hunk 12 above)
Comment (canetcesc) [reply core/session.ts]: Done

File: core/table.ts (hunk 19 above)
Comment (canetcesc) [reply core/table.ts]: No it wasn't there. I've just added it. Thx.

File: core/utils.ts [hunk 28]
```diff
@@ -263,3 +255,44 @@ export function flatten<T>(nestedArray: T[][]) {
     return previousValue.concat(currentValue);
   }, []);
 }
+
+export function matchFQName(
```
BenBirt: if i understand correctly, this code can be simplified a fair bit:

- this is only called by `matchPatterns`
- `matchPatterns` is called in two places; one for determining what dependencies should be included, and one for determining what actions can be included
- since dependencies are fully-qualified now, we can stop using `matchPatterns` in the first case - we simply need to find the intersection of `action.dependencies` and `allActionNames`, something like `action.dependencies.filter(dependency => allActionNames.includes(dependency))`
- in the case of actions, we only care about (a) if the requested action name includes a `.`, then we should be able to fully-match it to an action name in the graph, or (b) if the requested action name does not include a `.`, then we should only match against the last segment of each action name:

```
export function matchPatterns(patterns: string[], values: string[]) {
  const fQActs: string[] = [];
  patterns.forEach(pat => {
    if (pat.includes(".")) {
      if (values.includes(pat)) {
        fQActs.push(pat);
      }
    } else {
      const matchingActions = values.filter(value => pat === value.split(".").slice(-1)[0]);
      if (matchingActions.length === 0) {
        return;
      }
      if (matchingActions.length > 1) {
        throw new Error(...);
      }
      fQActs.push(matchingActions[0]);
    }
  });
  return fQActs;
}
```

tests/core/core.spec.ts
```diff
@@ -45,16 +45,42 @@ describe("@dataform/core", () => {
       expect(t.postOps).deep.equals(["post_op"]);
 
       const t2 = session
+        .publish("schema2.example", {
+          type: "table",
+          dependencies: ["schema1.example"],
```
BenBirt: sorry for not catching this earlier: i don't think we should let users do this; `dependencies` should be a `Resolvable[]`, and thus elements should either be a string not containing `.`, or full `Resolvable` objects. (to clarify, we should make the `dependencies` method accept `dependencies: Resolvable | Resolvable[]`)

core/session.ts
```diff
@@ -335,34 +337,49 @@ export class Session {
...
-      utils
-        .matchPatterns(dependencies.filter(d => d.includes("*")), allActionNames)
-        .forEach(dependency => (uniqueDependencies[dependency] = true));
-      action.dependencies = Object.keys(uniqueDependencies);
+      const fQDeps = action.dependencies.map(act => {
+        const allActs = this.findActions(act);
+        if (allActs.length === 1) {
+          return `${allActs[0].proto.target.schema}.${allActs[0].proto.target.name}`;
+        } else if (allActs.length >= 1) {
+          this.compileError(
```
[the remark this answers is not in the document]
canetcesc: Agree. Done.

core/test.ts
```diff
@@ -24,8 +24,14 @@ export class Test {
     return this;
   }
 
-  public dataset(datasetToTest: string) {
-    this.datasetToTest = datasetToTest;
+  public dataset(ref: Resolvable) {
+    const target2Resolvable = (t: dataform.ITarget) => {
```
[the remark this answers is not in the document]
canetcesc: We need to call `target()` to add the suffix. If not, then the following test will not pass: `bigquery using v2 language compiles with suffix "suffix"` for the file `biguquery_language_v2/definitions/example_test_case_fq_ref`.

Notice that the ability to use FQ refs on tests is a functionality that nobody asked for. If it is controvesial, I am happy removing it but what will happen is that you will not be able to reference a FQ `dataset` on a Test like this:

```
config {
    type: "test",
    dataset: {schema: "df_integration_test", name: "example_table"}
}
```

This only occurs in the case where there is a **SUFFIX**.  We need to call the `target()` function to add the suffix.

I have changed the code and moved the `target()` resolution inside the compile phase so that the code is cleaner.

This is arguably a very strange case because two things must happen for this code to make sense: (1) User must refer a Fully-Qualified dataset on a Test; (2) User must use a suffix. But still, it can happen!

core/test.ts
```diff
@@ -47,8 +53,9 @@ export class Test {
         this.proto.fileName
       );
     } else {
-      const dataset = this.session.tables[this.datasetToTest];
-      if (!dataset) {
+      const allResolved = this.session.findActions(this.datasetToTest);
+      const dataset = allResolved.length > 0 ? allResolved[0] : undefined;
```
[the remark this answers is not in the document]
canetcesc: Done.

core/utils.ts
canetcesc: Done!

tests/core/core.spec.ts
canetcesc: Done. Thx for the catch.

canetcesc:
@BenBirt - Sorry to pester you with this one. Please see my comment responses and in case something is not clear we can chat about it tomorrow!

Bons point: I have moved the circular dependency check from validation time to compile time. This is part of #247 but it made sense for me to put it in.

core/session.ts
```diff
@@ -66,7 +66,37 @@ function mapColumnDescriptionToProto(
...
 
-export type Resolvable = string | { schema: string; name: string };
+export type FullyQualifiedName = { schema: string; name: string };
+export type Resolvable = string | FullyQualifiedName;
+
+export function isResolvable(res: any) {
+  return typeof res === "string" || (!!res.schema && !!res.name);
+}
+
+export function resolvable2string(res: Resolvable) {
```
BenBirt: nit: `stringifyResolvable`

core/session.ts
```diff
@@ -66,7 +66,37 @@ function mapColumnDescriptionToProto(
...
+  return typeof res === "string" ? res : `${res.schema}.${res.name}`;
+}
+
+export function resolvable2FQame(res: Resolvable) {
```
BenBirt: doesn't appear to be used anywhere, remove?

core/session.ts
```diff
@@ -66,7 +66,37 @@ function mapColumnDescriptionToProto(
...
+  return typeof res === "string"
+    ? { schema: res.split(".").slice(-1)[0], name: res.split(".").slice(-1)[1] }
+    : res;
+}
+
+export function ambiguousActionNameMsg(
```
BenBirt: nit: suggest moving this to `core/utils.ts`

core/session.ts
```diff
@@ -411,7 +458,7 @@ export class Session {
     }
   }
 
-  private checkTestNameIsUnused(name: string) {
+  public checkTestNameIsUnused(name: string) {
```
BenBirt: this should stay `private`

core/test.ts
```diff
@@ -53,11 +52,26 @@ export class Test {
         this.proto.fileName
       );
     } else {
-      const allResolved = this.session.findActions(this.datasetToTest);
+      const target2Resolvable = (t: dataform.ITarget) => {
```
BenBirt: nit: `asResolvable`?

tests/core/core.spec.ts
```diff
@@ -47,7 +47,7 @@ describe("@dataform/core", () => {
       const t2 = session
         .publish("schema2.example", {
           type: "table",
-          dependencies: ["schema1.example"],
+          dependencies: [{ schema: "schema1", name: "example" }],
```
BenBirt: this is great. one last thing... I'm betting this doesn't work if the user requests schema suffixes? (in exactly the same way as referencing `dataset` in tests?)

core/test.ts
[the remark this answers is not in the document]
BenBirt: Ahhhhh, thank you for explaining, I definitely didn't consider the case for schema suffixes!

tests/core/core.spec.ts
canetcesc: That's right it was not working. I have made some modifications and changed the test case `it("config",...`. Now it runs with and without a suffix. Now it should work

core/test.ts
canetcesc: Done

core/session.ts
canetcesc: Done

core/session.ts
canetcesc: Done

core/session.ts
canetcesc: Done

core/session.ts
canetcesc: Done

core/assertion.ts
```diff
@@ -68,7 +69,9 @@ export class Assertion {
 
   public schema(schema: string) {
     this.proto.target.schema = schema;
```
BenBirt: i'm a bit concerned about having this block of code repeated in each Action, it seems ripe for something to break. i propose instead:

```
public schema(schema: string) {
  this.session.setNameAndTarget(this.proto, this.proto.target.name, schema);
  return this;
}
```

(i realise that i have also changed it to `return this;`, it's a small bug that we don't have that; i don't mind if you leave it out - we can add it in a later PR)

core/session.ts
```diff
@@ -458,11 +427,14 @@ export class Session {
     }
   }
 
-  public checkTestNameIsUnused(name: string) {
+  private checkTestNameIsUnused(name: string) {
```
BenBirt: nit: when i look at these changes in my editor, VSCode gets very unhappy because private members are coming before public - i'm surprised that your VSCode isn't formatting that on save - out of interest, have you ever restarted it?

core/utils.ts
```diff
@@ -1,6 +1,14 @@
 import { dataform } from "@dataform/protos";
-import { DistStyleTypes, ignoredProps, SortStyleTypes, TableTypes } from "./table";
-import { ambiguousActionNameMsg } from "@dataform/core/session";
+import { Assertion } from "@dataform/core/assertion";
```
BenBirt: ditto, super surprised VSCode isn't alphabetically ordering these imports on save.

tests/integration/redshift.spec.ts
```diff
@@ -14,7 +14,6 @@ describe("@dataform/integration/redshift", () => {
     });
 
     expect(compiledGraph.graphErrors.compilationErrors).to.eql([]);
-    expect(compiledGraph.graphErrors.validationErrors).to.eql([]);
```
BenBirt: let's keep this line for now! it's important to assert that, while we still ever put stuff in that field, the integration tests never have validation errors.

canetcesc: Oh sorry my bad. I've been in parallel working on #247 and this change actually belongs there.

core/session.ts
canetcesc: Yes I'm surprised too. Something bad happened recently while I was troubleshooting the linter issue with `includes(...)`. I've had to restart VSCode many times. Sometimes IntelliCode works, sometimes it doesn't...:S

canetcesc: We anyway have to change it back to `public` so it can be called from within the actions!

core/assertion.ts
canetcesc: 1. The master branch / status quo does not perform that check either. In fact, if you do the suggested change (using `setNameAndTarget`) it crashes with the following error:

      -    {
      -      "fileName": "definitions/override_schema_example.sqlx"
      -      "message": "Duplicate action name detected. Names within a schema must be unique across tables, assertions, and operations: \"override_schema.override_schema_example\""
      -      "stack": "Error: Duplicate action name detected. Names within a schema must be unique across tables, assertions, and operations: \"override_schema.override_schema_example\"\n    at Session.checkTargetIsUnused (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/124/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/df/examples/bigquery_language_v2/node_modules/@dataform/core/session.js:303:35)\n    at Session.setNameAndTarget (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/124/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/df/examples/bigquery_language_v2/node_modules/@dataform/core/session.js:211:18)\n    at Table.schema (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/124/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/df/examples/bigquery_language_v2/node_modules/@dataform/core/table.js:161:26)\n    at Table.config (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/124/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/df/examples/bigquery_language_v2/node_modules/@dataform/core/table.js:84:22)\n    at Session.sqlxAction (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/124/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/df/examples/bigquery_language_v2/node_modules/@dataform/core/session.js:123:18)\n    at /private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/124/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/df/examples/bigquery_language_v2/definitions/override_schema_example.sqlx:21:24\n    at Object.EXTENSIONS.(anonymous function) [as .sqlx] (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/124/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/npm/node_modules/vm2/lib/sandbox.js:83:5)\n    at _require (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/124/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/npm/node_modules/vm2/lib/sandbox.js:320:24)\n    at Object.<anonymous> (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/124/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/df/examples/bigquery_language_v2/index.js:225:13)\n    at NodeVM.run (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/124/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/npm/node_modules/vm2/lib/main.js:448:23)"
      -    }

Same for assertions:

```
           {
      -      "fileName": "definitions/example_assertion.sqlx"
      -      "message": "Duplicate action name detected. Names within a schema must be unique across tables, assertions, and operations: \"hi_there.example_assertion\""
      -      "stack": "Error: Duplicate action name detected. Names within a schema must be unique across tables, assertions, and operations: \"hi_there.example_assertion\"\n    at Session.checkTargetIsUnused (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/144/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/df/examples/bigquery_language_v2/node_modules/@dataform/core/session.js:303:35)\n    at Session.setNameAndTarget (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/144/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/df/examples/bigquery_language_v2/node_modules/@dataform/core/session.js:211:18)\n    at Assertion.schema (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/144/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/df/examples/bigquery_language_v2/node_modules/@dataform/core/assertion.js:61:26)\n    at Assertion.config (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/144/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/df/examples/bigquery_language_v2/node_modules/@dataform/core/assertion.js:29:22)\n    at Session.sqlxAction (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/144/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/df/examples/bigquery_language_v2/node_modules/@dataform/core/session.js:123:18)\n    at /private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/144/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/df/examples/bigquery_language_v2/definitions/example_assertion.sqlx:22:24\n    at Object.EXTENSIONS.(anonymous function) [as .sqlx] (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/144/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/npm/node_modules/vm2/lib/sandbox.js:83:5)\n    at _require (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/144/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/npm/node_modules/vm2/lib/sandbox.js:320:24)\n    at Object.<anonymous> (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/144/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/df/examples/bigquery_language_v2/index.js:33:13)\n    at NodeVM.run (/private/var/tmp/_bazel_cesc/e27ac56fa175c4de03c842b466cd7c26/sandbox/darwin-sandbox/144/execroot/df/bazel-out/darwin-py2-fastbuild/bin/tests/api/api.runfiles/npm/node_modules/vm2/lib/main.js:448:23)"
      -    }
      -    {
```

Since this issue is unrelated to this change, I propose we leave it here for now and file it as an issue.

2. I've changed the function a little bit so that now there is no mention to the suffix in the function. It is all taken care of by `this.session.target(...)`.

core/utils.ts
```diff
@@ -248,6 +246,11 @@ export function targetAsResolvable(t: dataform.ITarget) {
   return { schema: t.schema, name: t.name };
 }
 
+export function prependSuffixToSchema(d: Resolvable, suffix: string) {
```
BenBirt: nit: `append` i think

canetcesc:
The problem was calling `checkTargetIsUnused()` after changing the target. It should happen before:

```
-  private setNameAndTarget(action: IActionProto, name: string, overrideSchema?: string) {
-    action.target = overrideSchema ? this.target(name, overrideSchema) : this.target(name);
-    this.checkTargetIsUnused(action.target);
+  public setNameAndTarget(action: IActionProto, name: string, overrideSchema?: string) {
+    const newTarget = overrideSchema ? this.target(name, overrideSchema) : this.target(name);
+    this.checkTargetIsUnused(newTarget);
+    action.target = newTarget;
```

Issue #60, opened 2019-01-11
Implement declare() API


lewish:
It should be possible to declare external tables that your project depends on, but are not generated by the dataform project. For example, if I automatically push my web logs to BigQuery to the table `weblogs.logs` then in my dataform project I can create a file:

`definitions/external_tables.js`:
```
declare("weblogs.logs");
```

And reference them in other queries, e.g. `definitions/example.sql`:
```
select * from ${ref("weblogs.logs")}
```

This would allow us to show external tables as part of the DAG, and would be useful for debugging, as well as simplifying queries (in BigQuery).

- [ ] Add a new proto type `Declaration`
- [ ] Add `declarations` to `CompiledGraph` proto
- [ ] Make it possible to `ref()` a declaration
- [ ] Implement a `Declaration` class similar to `Materialization` that just stores a `Target` which is derived from the provided name
- [ ] Make sure declarations complete immediately as a no-op during execution of the graph

BenBirt:
suggested API: https://docs.google.com/document/d/1Ch2O7bydqd3fUZJGtek8rbVroELpMlE58_IhS-Snzno/

PR #353, opened 2019-08-08
Add a declare() API.

BenBirt:
fixes https://github.com/dataform-co/dataform/issues/60


core/session.ts
```diff
@@ -293,6 +317,18 @@ export class Session {
     return newTest;
   }
 
+  public declare(fullyQualifiedDataset: { schema: string; name: string }): Declaration {
```
canetcesc: You'll notice that in my yet-to-be-approved change this structure `{schema: string, name: string}` has become an interface in Session.ts which I paste here in case you want  to use it too:

export interface IResolvable {
  schema: string;
  name: string;
}

core/session.ts
```diff
@@ -293,6 +317,18 @@ export class Session {
...
+    const { name } = fullyQualifiedDataset;
+    this.checkActionNameIsUnused(name);
+    const declaration = new Declaration();
+    declaration.proto.name = name;
```
canetcesc: nit - I'd switch lines 324 and 325 because going forward name will mimic target.

core/session.ts
```diff
@@ -367,8 +404,13 @@ export class Session {
 
   private checkActionNameIsUnused(name: string) {
     // Check for duplicate names
-    if (this.tables[name] || this.operations[name] || this.assertions[name]) {
-      const message = `Duplicate action name detected. Names must be unique across tables, assertions, and operations: "${name}"`;
+    if (
+      this.tables[name] ||
+      this.operations[name] ||
+      this.assertions[name] ||
+      this.declarations[name]
```
canetcesc: Good to know! If it goes in this list then it is my understanding that declarations will also contain a schema in front of the name (just as tables, ops and assertions). If this change is approved before mine then I'll need to adjust my code accordingly. Do let me know what do you think makes more sense to merge earlier, either this one or mine.

BenBirt:
After making this work such that actions cannot depend on `declaration` actions, I now realise that actually it makes sense to allow that. I'll need to change the `run.ts` code to assume that all `declaration` dependencies immediately complete - I'll do that tomorrow.

core/session.ts
BenBirt: Yup, that's correct, declarations *must* have a schema too.

core/session.ts
BenBirt: Thanks, that's a good idea. I'm going to hold off for now - I left a comment suggesting that we can actually make `Resolvable` be either this interface or a string, in which case it wouldn't be good to use here - but if we don't end up doing that, cleaning this up will be pretty simple in any case!

core/session.ts
BenBirt: Done!

BenBirt:
PR now updated to allow actions to depend directly on declarations, PTAL!

protos/core.proto
```diff
@@ -157,6 +157,17 @@ message Assertion {
   reserved 4, 5, 6;
 }
 
+message Declaration {
+  string name = 1;
```
canetcesc: It doesn't make sense to me to create a variable `name` here knowing that it will become deprecated (use of Target is preferred).

BenBirt: I think you're right, and I think that's a good reason to block this PR on yours. (That's no issue for me, also no rush!)

core/declaration.ts
```diff
@@ -0,0 +1,43 @@
+import { IColumnsDescriptor, mapToColumnProtoArray, Session } from "@dataform/core/session";
+import { dataform } from "@dataform/protos";
+
+export interface DConfig {
+  description?: string;
+  columns?: IColumnsDescriptor;
```
canetcesc: Do we want to include `schema` here?  - I see this one related to https://github.com/dataform-co/dataform/issues/357.

My idea in #337 was to add it in _operations_, _tables_, _assertions_. 

Also on L21, add:

```
    if (config.schema) {
      this.schema(config.schema);
    }
```
and then add this:

```
public schema(schema: string) {
   this.proto.target.schema = schema;
}
```

BenBirt: I have added `schema` as a field in `DConfig`, because that's absolutely needed here.

I haven't done the other suggestions because:
- I don't think we should expose a `schema()` method, users should be required to fully-qualify schema & name when they call `declare()` in the JS API
- the `schema` field is actually already set at a higher-level than this, inside `Session.declare()` itself

core/session.ts
```diff
@@ -277,6 +294,19 @@ export class Session {
     return assertion;
   }
 
+  public declare(dataset: FullyQualifiedName): Declaration {
```
canetcesc: Why not `Resolvable`?

protos/core.proto
canetcesc: Will `name` finally stay?

core/session.ts
BenBirt: For declarations, I don't think we should ever let users just pass a table's name, because if they do, we don't really know what schema to use. We could use the schema(s) defined in `dataform.json`, but then the question is: what's the correct behaviour when the user asks for schema suffixes? I think for clarity / user understandability sake, forcing them to always specify the schema is the right thing to do.

BenBirt:
For some reason GitHub won't let me reply directly to your comment about `name`, but: I do think we eventually want to get rid of `name`, but (a) I just don't think that we can do it now, or probably before a v2, and (b) I think Declarations should be as similar as possible to the other types (so that they all either have `name` or none of them do).

PR #359, opened 2019-08-12
Document inline tables

canetcesc:


docs/pages/guides/inline-tables.mdx
```diff
@@ -0,0 +1,88 @@
...
+```sql
+config { type: "inline"}
+select id from ${ref("source_table")}
+```
+You can now reference the inline table from a normal table in the `WHERE` clause in another file such as:
+
+`definitions/table_using_WHERE.sqlx`:
+
+```sql
+select * from ${ref("outer_table")} where user_id in ${ref("inline_table")}
```
lewish: should be `ref("inline_table_ids")`

canetcesc: Done

PR #361, opened 2019-08-12
Turn on noImplicitAny.

BenBirt:


core/utils.ts
```diff
@@ -201,51 +194,49 @@ export function validate(compiledGraph: dataform.ICompiledGraph): dataform.IGrap
...
+          validationErrors.push(dataform.ValidationError.create({ message, actionName }));
+        }
+      };
+      const validatePropertiesDefined = (
+        opts: dataform.IRedshiftOptions,
+        props: Array<keyof dataform.IRedshiftOptions>
+      ) => props.forEach(prop => validatePropertyDefined(opts, prop));
+      const validatePropertyValueInValues = (
+        opts: dataform.IRedshiftOptions,
+        prop: keyof dataform.IRedshiftOptions & ("distStyle" | "sortStyle"),
```
lewish: why do we need the distStyle | sortStyle here?

BenBirt: We need to tighten the type otherwise this might be `sortKeys`, whose value is a `string[]` not `string`.

Issue #346, opened 2019-08-07
Add compile time validation warnings for dataform.json


BenBirt:
Had entered dataform instead of bigquery for the warehouse, didn't get any error message pointing in the right direction

---
Moved from dataform-co/dataform-co#134
/cc @G2H

canetcesc:
@BenBirt - Basides defaultSchema and warehouse, do we need to check for warehouse-specific required properties? See list:

```
const requiredBigQueryWarehouseProps: Array<keyof dataform.IBigQuery> = [
  "projectId",
  "credentials"
];
const requiredJdbcWarehouseProps: Array<keyof dataform.IJDBC> = [
  "host",
  "port",
  "username",
  "password",
  "databaseName"
];
const requiredSnowflakeWarehouseProps: Array<keyof dataform.ISnowflake> = [
  "accountId",
  "username",
  "password",
  "role",
  "databaseName",
  "warehouse"
];
const requiredSQLDataWarehouseProps: Array<keyof dataform.ISQLDataWarehouse> = [
  "server",
  "port",
  "username",
  "password",
  "database"
];
```

BenBirt:
Those are properties that must go in the `credentials` JSON file, not `dataform.json`. We should only need to check fields as exist in the `ProjectConfig` protobuf.

PR #369, opened 2019-08-15
Add compile-time dataform.json validation warnings (Issue #346)

canetcesc:


examples/dataform_json_checks/invalid_defaultschema/BUILD
```diff
@@ -0,0 +1,19 @@
...
+    srcs = glob([
+        "**/*.*",
+    ]),
+)
+
+node_modules(
+    name = "node_modules",
+    deps = [
+        "//core:package_tar",
+        "//crossdb:package_tar",
```
BenBirt: none of your code actually depends on this, remove it from each `BUILD` file?

examples/dataform_json_checks/invalid_defaultschema/BUILD
```diff
@@ -0,0 +1,19 @@
...
+        "//protos:package_tar",
```
BenBirt: ditto, i don't believe this is actually used anywhere either

tests/api/api.spec.ts
```diff
@@ -1145,6 +1145,59 @@ describe("@dataform/api", () => {
         expect(e.message).to.equal("Compilation timed out");
       }
     });
+
+    it("dataform.json is valid", async () => {
+      const graphInvalidWarehouse = await compile({
+        projectDir: path.resolve("df/examples/dataform_json_checks/invalid_warehouse")
+      });
+      const graphInvalidWarehouseErrors = utils.validate(graphInvalidWarehouse);
```
BenBirt: why bother running `validate`?

tests/api/api.spec.ts
```diff
@@ -1145,6 +1145,59 @@ describe("@dataform/api", () => {
...
+      expect(graphInvalidWarehouseErrors)
```
BenBirt: i realise you're following the pattern here, but i think we can start cleaning this up, at least for new tests. it would be much cleaner if we could just do:

```
expect(await compile(...)).to.eql({ ... });
```

note that if you do this, you might get cryptic errors from chai, complaining that 'foo' is not equal to 'foo' - this usually happens because you're comparing an instance of `proto.Foo` against a `proto.IFoo`, in which case just wrap your expected value in `proto.Foo.create(...)`.

let me know if you run into issues, or if it's a pain feel free to ignore this comment!

api/commands/compile.ts
```diff
@@ -10,23 +10,74 @@ export async function compile(
...
+      compErrors.push(compileError);
+    }
+  });
+
+  const gErrors: dataform.GraphErrors =
+    compErrors !== null
+      ? dataform.GraphErrors.create({ compilationErrors: compErrors })
+      : dataform.GraphErrors.create({ compilationErrors: [], validationErrors: [] });
 
   const returnedPath = await CompileChildProcess.forkProcess().compile(compileConfig);
```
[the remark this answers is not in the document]
canetcesc: I guess you mean throwing a normal error (not a `compileError`). That is the same strategy which we had here:

https://github.com/dataform-co/dataform/blob/c34588e0dd0057f1767b6d7517cae95a2656e379/api/commands/compile.ts#L19

I've thus changed all the  invalid dataform.json `compilationErrors` into normal Errors.

examples/dataform_json_checks/BUILD
```diff
@@ -0,0 +1,19 @@
+package(default_visibility = ["//tests:__subpackages__"])
```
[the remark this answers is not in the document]
canetcesc: You are right about this particular `BUILD` file. It can be deleted. *BUT* all other `BUILD` files have to remain because each `dataform.json` test requires a new independent folder.

examples/dataform_json_checks/invalid_defaultschema/BUILD
canetcesc: Can't delete it. See comments above. Same for all similar ones.

api/commands/compile.ts
```diff
@@ -10,23 +10,74 @@ export async function compile(
...
+    if (!(prop in dataformJsonParsed)) {
+      const compileError = dataform.CompilationError.create();
+      compileError.message =
+        "`dataform.json` does not have mandatory property defined: " + prop + ".";
+      compErrors.push(compileError);
+    }
+  });
+
+  const validDWHs = ["bigquery", "postgres", "redshift", "sqldatawarehouse", "snowflake"];
+  if (!!dataformJsonParsed.warehouse && validDWHs.indexOf(dataformJsonParsed.warehouse) === -1) {
```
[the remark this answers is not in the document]
canetcesc: Done and it works but the linter complains with an error.

api/commands/compile.ts
```diff
@@ -10,23 +10,74 @@ export async function compile(
...
   } catch (e) {
     throw new Error("Compile Error: `dataform.json` is invalid");
   }
+  const mandatoryProps = ["warehouse", "defaultSchema"];
+  mandatoryProps.forEach(prop => {
+    if (!(prop in dataformJsonParsed)) {
+      const compileError = dataform.CompilationError.create();
+      compileError.message =
+        "`dataform.json` does not have mandatory property defined: " + prop + ".";
+      compErrors.push(compileError);
```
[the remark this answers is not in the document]
canetcesc: I've changed all errors to `throw()` errors.

api/commands/compile.ts
```diff
@@ -10,23 +10,74 @@ export async function compile(
...
+      ". Should be one of: " +
+      validDWHs.join(", ");
+    compErrors.push(compileError);
+  }
+
+  const checkSimpleValidityProps = [
+    "assertion_schema",
+    "schema_suffix",
+    "gcloud_project_id",
+    "defaultSchema"
```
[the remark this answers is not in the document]
canetcesc: I'll change it to alphanumeric / underscore / hyphens. In fact, we enforce it already for the suffix:
https://github.com/dataform-co/dataform/blob/374e5b5cae3ef981d0268cab27b3984e19fde1cc/cli/index.ts#L106

examples/dataform_json_checks/invalid_defaultschema/BUILD
canetcesc: If I delete that I would get this error:

```
ERROR: /Users/cesc/dataform/tests/api/BUILD:29:1: no such target '//examples/dataform_json_checks/invalid
_warehouse:node_modules': target 'node_modules' not declared in package 'examples/dataform_json_checks/in
valid_warehouse' defined by /Users/cesc/dataform/examples/dataform_json_checks/invalid_warehouse/BUILD and referenced by '//tests/api:api_bin'
```

By the way, have you seen the changes in `tests/api/BUILD` file ?

BenBirt: I think we had a bit of confusion here! I'm not saying we can get rid of these `BUILD` files (except `examples/dataform_json_checks/BUILD` which you have already removed), but instead that you can get rid of this dependency on `//crossdb:package_tar` (and `//protos:package_tar`, I believe).

api/commands/compile.ts
```diff
@@ -14,9 +22,12 @@ export async function compile(
   try {
     // check dataformJson is valid before we try to compile
     const dataformJson = fs.readFileSync(`${compileConfig.projectDir}/dataform.json`, "utf8");
-    JSON.parse(dataformJson);
+    const checkValidity = checkDataformJsonValidity(JSON.parse(dataformJson));
```
BenBirt: we can just let `checkDataformJsonValidity` throw directly; no need to return the error string and then throw that.

tests/api/api.spec.ts
```diff
@@ -1145,6 +1145,51 @@ describe("@dataform/api", () => {
         expect(e.message).to.equal("Compilation timed out");
       }
     });
+
+    it("dataform.json is valid", async () => {
+      try {
+        utils.validate(
```
BenBirt: i don't think we need to do this `utils.validate()`, we can just call `compile()` and let it throw.

tests/api/api.spec.ts
```diff
@@ -1145,6 +1145,51 @@ describe("@dataform/api", () => {
...
+          /Invalid value on property defaultSchema: rock&roll. Should only contain alphanumeric characters, underscores and\/or hyphens./
+        );
+      }
+      try {
+        utils.validate(
+          await compile({
+            projectDir: path.resolve("df/examples/dataform_json_checks/valid_dataform_json")
+          })
+        );
+        fail("Forcing a failure.");
```
BenBirt: just don't bother wrapping this case in a `try/fail/catch` at all; if `compile` throws (without a wrapping `try/catch`) then the test case will automatically fail.

api/commands/compile.ts
```diff
@@ -73,3 +84,30 @@ class CompileChildProcess {
     }
   }
 }
+
+function checkDataformJsonValidity(dataformJsonParsed: { [prop: string]: string }): string {
+  const invalidWarehouseProp = function(): string {
```
BenBirt: nit: the linter should be complaining here; better style is to use `() => { ... }` than `function(): { ... }` because the fat-arrow operator correctly binds `this`.

canetcesc:
> Thanks Cesc!
> 
> Re the comment about using `includes` and your response saying the linter has an error - I checked out this branch and couldn't see my linter complaining, but please ping me if there is still an issue and we can figure it out.

The only thing that works for me to avoid the problem with `includes` is adding `es7` lib complierOptions in `tsconfig.json`  as described in https://stackoverflow.com/questions/51811239/ts2339-property-includes-does-not-exist-on-type-string.

api/commands/compile.ts
canetcesc: Done

api/commands/compile.ts
canetcesc: Done (though linter was not complaining).

tests/api/api.spec.ts
canetcesc: Done

tests/api/api.spec.ts
canetcesc: Done

PR #378, opened 2019-08-19
For interactive queries, don't use BigQuery jobs.

BenBirt:
fixes https://github.com/dataform-co/dataform/issues/374

### DIFF
--- a/api/dbadapters/index.ts
+++ b/api/dbadapters/index.ts
@@ -10,8 +10,8 @@ export type OnCancel = (handleCancel: () => void) => void;
 export interface IDbAdapter {
   execute(
     statement: string,
-    onCancel?: OnCancel,
     options?: {
+      onCancel?: OnCancel;
       interactive?: boolean;
     }
   ): Promise<any[]>;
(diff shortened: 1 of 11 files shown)


api/dbadapters/index.ts
```diff
@@ -8,7 +8,13 @@ import { dataform } from "@dataform/protos";
 export type OnCancel = (handleCancel: () => void) => void;
 
 export interface IDbAdapter {
-  execute(statement: string, onCancel?: OnCancel): Promise<any[]>;
+  execute(
+    statement: string,
+    onCancel?: OnCancel,
```
lewish: let's put this in options?

tests/integration/utils.ts
```diff
@@ -1,6 +1,8 @@
+import { Credentials } from "@dataform/api/commands/credentials";
 import * as dbadapters from "@dataform/api/dbadapters";
 import * as adapters from "@dataform/core/adapters";
 import { dataform } from "@dataform/protos";
+import { query } from "df/api";
```
lewish: @dataform/api otherwise our npm packages will break!

PR #383, opened 2019-08-21
Add CLI support for non-US/EU BigQuery data regions.

BenBirt:
for https://github.com/dataform-co/dataform-web-tracking/issues/158

### DIFF
--- a/cli/credentials.ts
+++ b/cli/credentials.ts
@@ -23,7 +23,7 @@ export function getBigQueryCredentials(): dataform.IBigQuery {
   ]);
   let location = locationIndex === 0 ? "US" : "EU";
   if (locationIndex === 2) {
-    location = question("Enter the location's region name (e.g. 'us-west2'):");
+    location = question("Enter the location's region name (e.g. 'asia-south1'):");
   }
   return {
     projectId: cloudCredentials.project_id,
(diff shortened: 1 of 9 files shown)


cli/credentials.ts
```diff
@@ -18,12 +18,17 @@ export function getBigQueryCredentials(): dataform.IBigQuery {
   const cloudCredentials = require(cloudCredentialsPath);
   const locationIndex = selectionQuestion("Enter the location of your datasets:", [
     "US (default)",
-    "EU"
+    "EU",
+    "other"
   ]);
+  let location = locationIndex === 0 ? "US" : "EU";
+  if (locationIndex === 2) {
+    location = question("Enter the location's region name (e.g. 'us-west2'):");
```
canetcesc: If I understand correctly, this question is only raised when user chooses `other`, therefore the example given is not very intuitive because it cannot happen that the user now chooses a US or EU region name, perhaps you can change it to:
```
location = question("Enter the location's region name (e.g. 'asia-south1'):");
```

BenBirt: so FWIW the user may still want `us-west2` instead of just `US`, but i do think your example is better! fixed.

canetcesc:
Just noticed that the checks did not pass!

PR #385, opened 2019-08-21
Fix some broken BUILD rules.

BenBirt:


BenBirt:
Specifically those rules - I was wrong to ask you to remove dependencies on `"//protos:package_tar"`, because of course `core` depends upon those.

canetcesc:
In what way did you experience this to be broken? Tests passed with or without `//protos:package_tar`

PR #390, opened 2019-08-22
Only create snowflake schemas if they don't exist

lewish:


api/dbadapters/snowflake.ts
```diff
@@ -97,6 +97,15 @@ export class SnowflakeDbAdapter implements IDbAdapter {
     );
   }
 
+  public async schemas(): Promise<string[]> {
+    const rows = await this.execute(
+      `select SCHEMA_NAME
```
BenBirt: supernit: pull this query onto a single line?

PR #393, opened 2019-08-22
Add a mechanism to override project config settings via compile API

lewish:


api/commands/compile.ts
```diff
@@ -27,6 +27,11 @@ export async function compile(
     throw new Error(`Compile Error: 'dataform.json' is invalid. ${e}`);
   }
 
+  // Schema overrides field can be set in two places, projectConfigOverride is the source of truth.
+  if (compileConfig.projectConfigOverride && compileConfig.projectConfigOverride.schemaSuffix) {
+    compileConfig.schemaSuffixOverride = compileConfig.projectConfigOverride.schemaSuffix;
```
BenBirt: doing this sort of indicates that the plan is to keep `CompileConfig.schema_suffix_override` around long-term, is that the plan?

otherwise i'd expect the inverse; copying `compileConfig.schemaSuffixOverride` into `compileConfig.projectConfigOverride.schemaSuffix` (unless it already has a value)

protos/core.proto
```diff
@@ -38,6 +38,9 @@ message CompileConfig {
 
   // Compile time configuration options.
   string schema_suffix_override = 2;
```
BenBirt: `// Deprecated`?

api/commands/compile.ts
```diff
@@ -27,6 +27,18 @@ export async function compile(
     throw new Error(`Compile Error: 'dataform.json' is invalid. ${e}`);
   }
 
+  // Create an empty projectConfigOverride if not set.
+  compileConfig = { projectConfigOverride: {}, ...compileConfig };
+
+  // Schema overrides field can be set in two places, projectConfigOverride.schemaSuffix takes precedent.
+  if (compileConfig.schemaSuffixOverride) {
+    // Maybe create an empty project config override.
```
BenBirt: nit: comment seems wrong

core/gen_index.ts
```diff
@@ -32,15 +32,22 @@ export function genIndex(base64EncodedConfig: string): string {
...
+    dataform.ProjectConfig.create(config.compileConfig.projectConfigOverride).toJSON()
+  );
+
   return `
     require("@dataform/core");
     const protos = require("@dataform/protos");
     const { util } = require("protobufjs");
     ${includeRequires}
-    const projectConfig = require("./dataform.json");
+    let projectConfig = { ...require("./dataform.json") };
```
BenBirt:
```suggestion
    let projectConfig = require("./dataform.json");
```

PR #396, opened 2019-08-23
Change compilationError for ValidationError [#247]

canetcesc:


canetcesc:
What should we do with the leftovers of `utils.validate()`

BenBirt:
Let's get rid of it!

PR #402, opened 2019-08-29
Adding docs for tags

dwl285:


docs/pages/guides/tags.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+```js
+publish("users").query("SELECT ...").tags(["daily"]);
+```
+
+Tags can be added to datasets, assertions, operations and declarations. You can add more than one tag to each action.
+
+## Using tags
+
+Once you have assigned tags to actions, you can use the tags to define a run or schedule. For example, in the schedule creation
```
BenBirt: you mention the scheduling UI here - maybe link to the docs page about scheduling?

docs/pages/guides/tags.mdx
```diff
@@ -0,0 +1,41 @@
...
+screen, simply choose the tags you'd like to include in that schedule.
+
+Tags can also be used with the CLI using the --tags command
```
BenBirt:
```suggestion
Tags can also be used with the CLI using the `--tags` argument.
```

docs/pages/guides/tags.mdx
dwl285: Have I done that right?

PR #409, opened 2019-08-30
Always uppercase all Snowflake identifiers.

BenBirt:


core/adapters/base.ts
```diff
@@ -1,35 +1,39 @@
 import { dataform } from "@dataform/protos";
 
 export abstract class Adapter {
-  public where(query: string, where: string) {
-    return `select * from (${query}) as subquery
-        where ${where}`;
+  public abstract resolveTarget(target: dataform.ITarget): string;
+
+  public normalizeIdentifier(identifier: string) {
+    return identifier;
```
GJMcGowan: What's the point of this function?

tests/api/api.spec.ts
```diff
@@ -1097,7 +1097,7 @@ describe("@dataform/api", () => {
 
       // Make sure it compiles.
       expect(tableNames).includes("example");
-      const example = graph.tables.filter((t: dataform.ITable) => t.name == "example")[0];
+      const example = graph.tables.filter((t: dataform.ITable) => t.name === "example")[0];
```
GJMcGowan: I wonder if there's a rule we can turn on somewhere to require `===` instead of `==`

core/adapters/base.ts
BenBirt: It's to give a default implementation of this function (defined on the interface) for all other Adapters.

tests/api/api.spec.ts
BenBirt: https://palantir.github.io/tslint/rules/triple-equals/ - will enable in a later PR!

Issue #411, opened 2019-09-03
Incorrect compile error when depending on a fully-qualified dependency


BenBirt:
```
config {
  type: "view",
  schema: "dependency",
  name: "foo"
}

select 1 as test
```

```
config {
  schema: "dependent",
  name: "bar"
}

select * from ${ref({schema: "dependency", name: "foo"})}
```

results in:
```Error: Missing dependency detected: Node "dependent.bar" depends on "dependency.foo" which does not exist.```

(the error also has no attached filename)

PR #413, opened 2019-09-04
Fix fully-qualified dependencies.

BenBirt:
- Move schema suffixing to only happen at compile()-time, after all contextable queries have run (because otherwise the code has to constantly be aware of actions that effectively have two names, one with and one without the suffix)
- Move dependency checking / qualification to compile()-time (the only safe time to do it since we allow schema overrides)

fixes https://github.com/dataform-co/dataform/issues/411

### DIFF
--- a/core/session.ts
+++ b/core/session.ts
@@ -212,14 +212,6 @@ export class Session {
     return action;
   }
 
-  public target(name: string, schema: string): dataform.ITarget {
-    const adapter = this.adapter();
-    return dataform.Target.create({
-      name: adapter.normalizeIdentifier(name),
-      schema: adapter.normalizeIdentifier(schema)
-    });
-  }
-
   public resolve(ref: Resolvable): string {
     const allResolved = this.findActions(ref);
     if (allResolved.length > 1) {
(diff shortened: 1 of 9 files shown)


core/session.ts
```diff
@@ -445,33 +393,134 @@ export class Session {
...
+            ),
+            action.fileName
+          );
+        }
+        return dependencyName;
+      });
+    });
+  }
+
+  private appendSchemaSuffix(actions: IActionProto[]) {
```
GJMcGowan: For my sanity, what does this function do? It doesn't return anything and every variable it changes is inside it's scope, no?

BenBirt: It mutates the state of the protos which get passed into it; those protos are still in scope in the caller function.

I'm not a fan of how stateful and non-idempotent this makes `Session.compile()` (it's fairly horrible), but this doesn't make anything worse - compilation was unfortunately previously not idempotent.

FWIW making it stateless is on my wishlist for the open-source repo: https://docs.google.com/document/d/10ewZ0xhUuTVs6SHEycqTfDOyj-5UYPVfYzlWph1TpA0/edit#

PR #417, opened 2019-09-10
Change schedules.json validation to include CompiledGraph action names without potential schema prefixes.

BenBirt:


BenBirt:
test added, thanks!

PR #420, opened 2019-09-17
Push query compilation deeper into @dataform/core, so that it doesn't have to use a possibly mismatching version of gen_index.ts built into @dataform/api.

BenBirt:
- push custom return value code down into gen_index.ts so that it's versioned inside @dataform/core
- maintain backwards compatibility with (some) old versions of @dataform/core

related to (but cannot fix) https://github.com/dataform-co/dataform-co/issues/1645


BenBirt:
for posterity on the PR but as discussed in person: i do think we should do that, but the API for query compilation should remain <here is a query string> and nothing else; this enforces that code in @dataform/core generates the line of code you describe, which is important because that generation needs to be versioned alongside the "current" version of the API exposed by @dataform/core.

PR #427, opened 2019-09-24
Add @dataform/sqlx package, currently used by @dataform/core but also to be used by @dataform/api.

BenBirt:
for https://github.com/dataform-co/dataform-co/issues/142


sqlx/lexer.ts
```diff
@@ -182,50 +223,61 @@ class SqlxParseState {
   }
 }
 
+function getValueMappings() {
+  const valueMappings: {
```
lewish: why not just a top level exported const? You'll get a nicer type too

sqlx/sqlx.package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "@dataform/sqlx",
+  "description": "Dataform SQLX language tools.",
+  "main": "index.js",
+  "types": "index.d.ts",
+  "dependencies": {
+    "moo": "^0.5.0"
```
lewish: Is moo still a direct dependency of @dataform/core? If not, maybe remove it from core.package.json (otherwise these will inevitably get out of sync!). If it is, then maybe it shouldn't be (if you only need the types, you can still remove it).

sqlx/lexer.ts
BenBirt: because sadly, java/typescript don't let you instantiate a top-level object whose keys are fields of other objects.

sqlx/sqlx.package.json
BenBirt: good call, done!

PR #428, opened 2019-09-24
Add formatter to @dataform/api.

BenBirt:
for dataform-co/dataform-co#142


api/commands/format.ts
```diff
@@ -0,0 +1,182 @@
...
+  return postProcessFormattedSqlx(
+    getIndividualSqlxStatements(node.contents)
+      .map(individualStatement => {
+        const placeholders: {
+          [placeholderId: string]: ISyntaxTreeNode;
+        } = {};
+        const unformattedPlaceholderSql = individualStatement
+          .map(child => {
+            if (typeof child !== "string") {
+              const placeholderId = crypto.randomBytes(16).toString("hex");
```
lewish: Can we seed this easily so it's consistent?

api/commands/format.ts
```diff
@@ -0,0 +1,182 @@
...
+      ? wholeLineContainingPlaceholderId.slice(0, leftOfPlaceholder.length + placeholderId.length)
+      : placeholderId;
+  return sqlx.replace(textToReplace, formattedChild);
+}
+
+function formatChildSyntaxTreeNode(node: ISyntaxTreeNode, jsIndent: number): string {
+  if (node.startTokenType.endsWith("_startJsPlaceholder")) {
+    return formatJavaScriptPlaceholder(node, jsIndent);
+  }
+  if (node.startTokenType.endsWith("_startJs") || node.startTokenType.endsWith("_startConfig")) {
```
lewish: ditto, feels like these tokens would benefit from some more structure (e.g. contentType: "js" | "sql")

api/commands/format.ts
```diff
@@ -0,0 +1,182 @@
...
+        );
+      })
+      .join("\n\n---\n\n")
+  );
+}
+
+function getIndividualSqlxStatements(nodeContents: Array<string | ISyntaxTreeNode>) {
+  const sqlxStatements: Array<Array<string | ISyntaxTreeNode>> = [[]];
+  nodeContents.forEach(child => {
+    if (typeof child !== "string" && child.startTokenType.endsWith("_statementSeparator")) {
```
lewish: Can we reference the _statementSeperator from where it's defined

api/commands/format.ts
```diff
@@ -0,0 +1,182 @@
...
+}
+
+function formatChildSyntaxTreeNode(node: ISyntaxTreeNode, jsIndent: number): string {
+  if (node.startTokenType.endsWith("_startJsPlaceholder")) {
+    return formatJavaScriptPlaceholder(node, jsIndent);
+  }
+  if (node.startTokenType.endsWith("_startJs") || node.startTokenType.endsWith("_startConfig")) {
+    return `\n\n${formatJavaScript(concatenateSyntaxTreeContents(node))}\n\n`;
+  }
+  // This node must be an "inner" SQL block, e.g. "pre_operations { ... }",
```
lewish: Should we be explicit here? We will inevitably add new types that might not wanted to be treated like this. Might be better to just leave such blocks as they were

api/commands/format.ts
BenBirt: yup, good call, done something very close to that.

api/commands/format.ts
BenBirt: fixed as per your other comment

api/commands/format.ts
BenBirt: done!

api/commands/format.ts
BenBirt: I couldn't find a way to seed random numbers in node short of depending on another package :(

PR #436, opened 2019-09-27
More formatter fixes to make it run on dataform-data.

BenBirt:
- run post-processing only after the whole file is formatted
- fix double-quote regexp lifting regexp
- push comments to their own lines
- change formatter to always place config/JS blocks first, then SQL statements, then "inner" SQL blocks

for https://github.com/dataform-co/dataform-co/issues/142

### DIFF
--- a/api/commands/format.ts
+++ b/api/commands/format.ts
@@ -44,7 +44,7 @@ export async function formatFile(
   return formattedText;
 }
 
-function formatSqlx(node: ISyntaxTreeNode) {
+function formatSqlx(node: ISyntaxTreeNode, indent: string = "") {
   return getIndividualSqlxStatements(node.contents)
     .map(individualStatement => {
       const placeholders: {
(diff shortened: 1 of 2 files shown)


api/commands/format.ts
```diff
@@ -117,16 +185,32 @@ function formatJavaScript(text: string) {
...
+  let formatted = sqlFormatter.format(text) as string;
+  // Unfortunately sql-formatter does not always produce final formatted output (even on plain SQL) in a single pass.
+  for (let attempts = 0; attempts < MAX_SQL_FORMAT_ATTEMPTS; attempts++) {
+    const newFormatted = sqlFormatter.format(formatted) as string;
+    if (newFormatted === formatted) {
+      return newFormatted;
+    }
+    formatted = newFormatted;
+  }
+  throw new Error(
```
lewish: I'd argue that it would be more annoying for the formatter to throw here than for it to flip-flop, wdyt?

BenBirt: If we let it flip-flop, I need to remove the error checking at the top level that the whole formatter doesn't flip-flop-flap-etc; I don't yet feel comfortable doing that because so far it has only shown up buggy formatting (where it actually breaks the code).

PR #451, opened 2019-10-14
Add a docs page on environments.

BenBirt:
for https://github.com/dataform-co/dataform-co/issues/1623

### DIFF
--- a/docs/pages/platform_guides/environments.mdx
+++ b/docs/pages/platform_guides/environments.mdx
@@ -9,9 +9,9 @@ export default ({ children }) => <Documentation {...meta}>{children}</Documentat
 
 ## Introduction
 
-Environments can be used to configure different variations of your schedules. For each environment
-you define, Dataform will run all of the schedules defined in that environment, using that environment's
-version of the project's code.
+Environments can be used to configure different variations of your schedules. **For each environment
+you define, using that environment's version of the project code, Dataform will run all of the
+project's schedules.**
 
 A common use-case for environments is to run a staged release process; you might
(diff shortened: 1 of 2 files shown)


docs/pages/platform_guides/environments.mdx
```diff
@@ -0,0 +1,115 @@
...
+      }
+    }
+  ]
+}
+```
+
+This configuration creates a `staging` environment which always runs the latest version of your project's code
+(as exists on the `master` branch). When you have confirmed that the code works as expected, you can push that to production
+by merging the `master` branch into the `production` branch.
+
```
G2H: I think here would be a good place to explain how this is going to impact schedules and that schedules will run for every environments.

BenBirt: I've changed the copy in the Intro section and the Multiple environments section to hopefully make this clearer, lmk!

docs/pages/platform_guides/environments.mdx
```diff
@@ -0,0 +1,117 @@
...
+  title: "Configure environments",
+  __filename
+};
+
+export default ({ children }) => <Documentation {...meta}>{children}</Documentation>;
+
+## Introduction
+
+Environments can be used to configure different variations of your schedules. For each environment
+you define, Dataform will run all of the schedules defined in that environment, using that environment's
```
G2H: If I understand things correctly, users do not define schedules in environment, right? I would have imagined that "Dataform will run all of the schedules defined in your project" was more explicit in that case. Am I missing anything?

I would emphasize on this sentence as well since the behaviour is quite important. With bolded text or else

PR #452, opened 2019-10-14
Refactor the Runner:

BenBirt:
- general cleanup
- never leave a Promise hanging
- run actions as soon as they become runnable
- introduce 3 layers of execution: graph/action/task
- notify listeners of an update *after* an action is added to the result, not before

for https://github.com/dataform-co/dataform-co/issues/1968

### DIFF
--- a/api/commands/run.ts
+++ b/api/commands/run.ts
@@ -101,30 +101,8 @@ export class Runner {
   }
 
   private async executeGraph() {
-    const executeAllActionsReadyForExecution = async () => {
-      const allSuccessfulActions = this.result.actions
-        .filter(isSuccessfulAction)
-        .map(fn => fn.name);
-      const isReadyForExecution = (action: dataform.IExecutionAction) => {
-        for (const dependency of action.dependencies) {
-          if (!allSuccessfulActions.includes(dependency)) {
-            return false;
-          }
-        }
-        return true;
-      };
-      const readyForExecutionActions = this.pendingActions.filter(isReadyForExecution);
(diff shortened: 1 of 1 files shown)


api/commands/run.ts
```diff
@@ -96,107 +100,84 @@ export class Runner {
...
-        this.result.actions.push({
-          name: action.name,
-          status: dataform.ActionExecutionStatus.SKIPPED,
-          deprecatedSkipped: true
-        });
-      } else {
-        this.pendingActions.push(action);
-      }
+  private async executeGraph() {
+    const executeAllActionsReadyForExecution = async () => {
```
lewish: Doesn't look like this has any reason to be inside executeGraph - make it class level function?

api/commands/run.ts
```diff
@@ -96,107 +100,84 @@ export class Runner {
...
+    for (const task of action.tasks) {
+      if (allSuccessful) {
+        const executedTask = await this.executeTask(task);
+        executedTasks.push(executedTask);
+        allSuccessful = allSuccessful && executedTask.ok;
+      }
     }
+
+    const endTime = process.hrtime(startTime);
+    const executionTime = endTime[0] * 1000 + Math.round(endTime[1] / 1000000);
```
lewish: This is super weird (I know it was already there). To get timestamp millis, just do new Date().valueOf().
(doubles are a safe way to represent milli timestamps until the year 2255 AIUI, although the above code doesn't even really circumvent that).

PR #453, opened 2019-10-15
Bring back SKIPPED action logic, and handling run cancellation in the Runner.

BenBirt:
for dataform-co/dataform-co#1968

### DIFF
--- a/api/commands/run.ts
+++ b/api/commands/run.ts
@@ -117,7 +117,9 @@ export class Runner {
       const allPendingActions = this.pendingActions;
       this.pendingActions = [];
       await Promise.all(
-        allPendingActions.map(async pendingAction => this.skipAction(pendingAction))
+        allPendingActions.map(async pendingAction =>
+          this.onActionExecutionComplete(toSkippedAction(pendingAction))
+        )
       );
       return;
     }
@@ -139,7 +141,11 @@ export class Runner {
     );
 
     await Promise.all([
-      Promise.all(skippableActions.map(async skippableAction => this.skipAction(skippableAction))),
+      Promise.all(
+        skippableActions.map(async skippableAction =>
+          this.onActionExecutionComplete(toSkippedAction(skippableAction))
+        )
+      ),
       Promise.all(
         executableActions.map(async executableAction =>
           this.onActionExecutionComplete(await this.executeAction(executableAction))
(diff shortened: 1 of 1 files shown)


api/commands/run.ts
```diff
@@ -112,24 +112,54 @@ export class Runner {
...
-        this.result.actions.push(await this.executeAction(action));
-        await this.triggerChange();
-        await this.executeAllActionsReadyForExecution();
-      })
+    // If the run has been cancelled, skip all pending actions.
+    if (this.cancelled) {
+      const allPendingActions = this.pendingActions;
+      this.pendingActions = [];
+      await Promise.all(
+        allPendingActions.map(async pendingAction => this.skipAction(pendingAction))
```
lewish: These aren't promises, no need to await or promise.all() here. `skipAction` implies it mutates state, but it does not - `skippedAction` or `getSkippedAction` ? Or could just be inline tbh...

api/commands/run.ts
```diff
@@ -112,24 +112,54 @@ export class Runner {
...
+          this.onActionExecutionComplete(await this.executeAction(executableAction))
+        )
+      )
+    ]);
+  }
+
+  private async onActionExecutionComplete(executedAction: dataform.IExecutedAction) {
+    this.result.actions.push(executedAction);
+    await this.triggerChange();
+    await this.executeAllActionsReadyForExecution();
```
lewish: Hmmm. AIUI, you build up this massive dependency tree of promises that's the inverse of the dependecy graph, that's going to recursively come all the way back up to the top when it's done?
Feels like when this goes wrong it's going to be very hard to debug / reason about.
Personally I thought the for loop was more readable. (Not going to block on this though).

api/commands/run.ts
BenBirt: fixed, i had accidentally not called `onActionExecutionComplete` on these (which does return a Promise)

api/commands/run.ts
BenBirt: unless I misunderstood the previous code, this code is very similar to the loop, which was itself recursive (or at least, triggered calls to itself).

i could be misunderstanding your comment, but I don't think it works like that. this code does not build an inverse of the dependency graph, it's very like the old code in that upon each iteration it (a) finds and executes all executable actions, (b) skips all skippable actions, (c) triggers another iteration of all of the above, until the list of pending actions is exhausted.

PR #455, opened 2019-10-16
Convert Runner internal logic to use new {Run/Action/Task}Result protobufs.

BenBirt:
for https://github.com/dataform-co/dataform-co/issues/1968

### DIFF
--- a/api/commands/run.ts
+++ b/api/commands/run.ts
@@ -10,31 +10,6 @@ const isSuccessfulAction = (actionResult: dataform.IActionResult) =>
   actionResult.status === dataform.ActionResult.ExecutionStatus.SUCCESSFUL ||
   actionResult.status === dataform.ActionResult.ExecutionStatus.DISABLED;
 
-const actionExecutionStatusMap = new Map<
-  dataform.ActionResult.ExecutionStatus,
-  dataform.ActionExecutionStatus
->();
-actionExecutionStatusMap.set(
-  dataform.ActionResult.ExecutionStatus.SUCCESSFUL,
-  dataform.ActionExecutionStatus.SUCCESSFUL
-);
-actionExecutionStatusMap.set(
-  dataform.ActionResult.ExecutionStatus.FAILED,
-  dataform.ActionExecutionStatus.FAILED
-);
-actionExecutionStatusMap.set(
-  dataform.ActionResult.ExecutionStatus.SKIPPED,
-  dataform.ActionExecutionStatus.SKIPPED
-);
-actionExecutionStatusMap.set(
-  dataform.ActionResult.ExecutionStatus.DISABLED,
-  dataform.ActionExecutionStatus.DISABLED
-);
-actionExecutionStatusMap.set(
-  dataform.ActionResult.ExecutionStatus.CANCELLED,
-  dataform.ActionExecutionStatus.FAILED
-);
-
 export function run(graph: dataform.IExecutionGraph, credentials: Credentials): Runner {
   const runner = Runner.create(
     dbadapters.create(credentials, graph.projectConfig.warehouse),
@@ -111,7 +86,7 @@ export class Runner {
         .filter(action => action.status !== dataform.ActionResult.ExecutionStatus.RUNNING)
         .map(actionResult => ({
           name: actionResult.name,
-          status: actionExecutionStatusMap.get(actionResult.status),
+          status: toActionExecutionStatus(actionResult.status),
           tasks: actionResult.tasks.map((taskResult, i) => {
             const executedTask: dataform.IExecutedTask = {
               ok: taskResult.status === dataform.TaskResult.ExecutionStatus.SUCCESSFUL,
@@ -184,9 +159,9 @@ export class Runner {
       allPendingActions.forEach(pendingAction =>
         this.runResult.actions.push({
           name: pendingAction.name,
-          status: dataform.ActionResult.ExecutionStatus.CANCELLED,
+          status: dataform.ActionResult.ExecutionStatus.SKIPPED,
           tasks: pendingAction.tasks.map(() => ({
-            status: dataform.TaskResult.ExecutionStatus.CANCELLED
+            status: dataform.TaskResult.ExecutionStatus.SKIPPED
           }))
         })
       );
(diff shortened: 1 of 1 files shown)


api/commands/run.ts
```diff
@@ -6,9 +6,34 @@ import * as Long from "long";
...
-  action.status == dataform.ActionExecutionStatus.DISABLED;
+const isSuccessfulAction = (actionResult: dataform.IActionResult) =>
+  actionResult.status === dataform.ActionResult.ExecutionStatus.SUCCESSFUL ||
+  actionResult.status === dataform.ActionResult.ExecutionStatus.DISABLED;
+
+const actionExecutionStatusMap = new Map<
+  dataform.ActionResult.ExecutionStatus,
+  dataform.ActionExecutionStatus
+>();
+actionExecutionStatusMap.set(
```
lewish: This feels like a bad idea. Can we wrap in method or do something like  `new Map<...>([[key, value], [key, value]...])` ?

protos/core.proto
```diff
@@ -263,6 +263,56 @@ message ExecutedTask {
   string error = 3;
 }
 
+message Timing {
+  int64 start_time_millis = 1;
```
lewish: You might think I'm crazy, but I'd actually use `double` here. Long's are a PITA, doubles are safe for millis until like the year 200,000+ or something crazy. It also makes them much more readable in JSON format and anyone who wants to consume CLI output

api/commands/run.ts
```diff
@@ -61,134 +83,223 @@ export class Runner {
...
-    return this.executionTask;
+  public async resultPromise(): Promise<dataform.IExecutedGraph> {
+    await this.executionTask;
+    return this.resultAsExecutedGraph();
   }
 
   private triggerChange() {
-    return Promise.all(this.changeListeners.map(listener => listener(this.result)));
+    const executedGraph = this.resultAsExecutedGraph();
+    return Promise.all(this.changeListeners.map(listener => listener(executedGraph)));
```
lewish: feels like you are going to need another PR in this case to do the close-source switch. Would it not be better to expose resultAsExecutedGraph to user and change this API to it's final form here and now? Or you'll need some other way to add listeners for the actual RunResult object

api/commands/run.ts
```diff
@@ -61,134 +83,223 @@ export class Runner {
...
       const allPendingActions = this.pendingActions;
       this.pendingActions = [];
-      await Promise.all(
-        allPendingActions.map(async pendingAction =>
-          this.onActionExecutionComplete(toSkippedAction(pendingAction))
-        )
+      allPendingActions.forEach(pendingAction =>
+        this.runResult.actions.push({
+          name: pendingAction.name,
+          status: dataform.ActionResult.ExecutionStatus.CANCELLED,
```
lewish: IMO these are skipped, not cancelled. Or rather, I interpreted "cancelled" to mean "interrupted" actions/tasks. I think they is value in knowing which ones were killed, as opposed to which ones never started because the run was cancelled

api/commands/run.ts
```diff
@@ -61,134 +83,223 @@ export class Runner {
...
     );
     const executableActions = this.pendingActions.filter(allDependenciesHaveExecutedSuccessfully);
     this.pendingActions = this.pendingActions.filter(
       action => !allDependenciesHaveExecutedSuccessfully(action)
     );
 
     // Determine what actions we should skip (necessarily excluding executable actions).
-    const allDependenciesHaveExecuted = allDependenciesHaveBeenExecuted(this.result.actions);
+    const allDependenciesHaveExecuted = allDependenciesHaveBeenExecuted(
+      this.runResult.actions.filter(
```
lewish: Seems like this should happen inside `allDependenciesHaveBeenExecuted`

api/commands/run.ts
```diff
@@ -61,134 +83,223 @@ export class Runner {
...
+            status: dataform.TaskResult.ExecutionStatus.CANCELLED
+          }))
+        })
       );
+      await this.triggerChange();
       return;
     }
 
     // Determine what actions we can execute.
     const allDependenciesHaveExecutedSuccessfully = allDependenciesHaveBeenExecuted(
```
lewish: Could clean this up a bit? Why not two class methods `allDependenciesHaveExecuted and allDependenciesHaveExecutedSuccesfully` that do as they describe and drop the function returning function.

api/commands/run.ts
```diff
@@ -61,134 +83,223 @@ export class Runner {
...
+    await this.triggerChange();
 
-    const executedTasks: dataform.IExecutedTask[] = [];
-    let allSuccessful = true;
     for (const task of action.tasks) {
-      if (allSuccessful) {
-        const executedTask = await this.executeTask(task);
-        executedTasks.push(executedTask);
-        allSuccessful = allSuccessful && executedTask.ok;
+      if (this.cancelled) {
```
lewish: See above comment, if you agree this should be skipped too.

api/commands/run.ts
```diff
@@ -61,134 +83,223 @@ export class Runner {
...
+        actionResult.tasks.push({
+          status: dataform.TaskResult.ExecutionStatus.CANCELLED
+        });
+        await this.triggerChange();
```
lewish: this (and below) will trigger more often than it needs to I think? (ideally, we'd reduce the number of triggerChanges calls). As you are probably to skip many other actions and tasks.
This might be tricky to do so I'm not that bothered, but ideally triggerChange would fire only when:
- The moment a task finishes executing successfully
- After a task fails/cancels AND we mark everything else that can be as skipped

lewish: Actually, because you trigger at the end of the loop and during execute task, if you remove this and below, I think that would be absolutely fine.

api/commands/run.ts
```diff
@@ -61,134 +83,223 @@ export class Runner {
...
-        : dataform.ActionExecutionStatus.FAILED,
-      tasks: executedTasks,
-      executionTime: Long.fromNumber(executionTime),
-      deprecatedOk: allSuccessful
-    };
+    if (actionResult.status === dataform.ActionResult.ExecutionStatus.RUNNING) {
+      actionResult.status = dataform.ActionResult.ExecutionStatus.SUCCESSFUL;
+    }
+    actionResult.timing = timer.end();
+    await this.triggerChange();
```
lewish: This also looks like it will trigger twice for the final task if it executes successfully, once with just the task marked as successful, and once with the task and the surround action marked as successful.

protos/core.proto
BenBirt: Yeah, I fairly strongly disagree here.
- part of the reason we use protos here is because we want to be language agnostic; I don't think using a type to represent a timestamp which we wouldn't use in any other language but to make things nicer in JS is a good idea
- presentation-layer concerns should be dealt with at the presentation-layer, not at the API layer
- I have once before dealt with a system that used `double`s to represent timestamps, it was a total pain; this might be fine in JS but bring those fields into any other language (and remember we do; in Java) and weird stuff starts to happen with floating point precision, and that causes some really difficult to track down bugs

lewish: I'm sold :)I guess if we want nicer JSON, we can just write Long's as numbers when we we know it is safe to do so!

PR #466, opened 2019-10-22
Bring back scheduleOptions

faiyaz26:


protos/schedules.proto
```diff
@@ -2,18 +2,23 @@ syntax="proto3";
...
 
-import "protos/core.proto";
-
 message NotificationSettings {
   repeated string emails = 1;
   bool on_success = 2;
   bool on_failure = 3; 
 }
 
+message ScheduleOptions {
```
BenBirt: please add a comment here and on RunConfig, that these should be kept in sync (in terms of field names)

PR #468, opened 2019-10-23
Add notification channel to schedules

faiyaz26:


protos/schedules.proto
```diff
@@ -2,12 +2,36 @@ syntax="proto3";
 
 package dataform.schedules;
 
+
+message EmailNotificationChannel {
+  repeated string emails = 1;
```
BenBirt: i think we should call this field `to`

protos/schedules.proto
```diff
@@ -2,12 +2,36 @@ syntax="proto3";
...
+message NotificationChannel {
+ string name = 1;
+ oneof channel {
+    EmailNotificationChannel email_notification = 1;
+    SlackNotificationChannel slack_notification = 2;
+  } 	
+}
+
+message ScheduleNotification {
+ repeated environments = 1;
```
BenBirt: presumably this should be a `string`?

protos/schedules.proto
```diff
@@ -2,12 +2,36 @@ syntax="proto3";
...
+ repeated string onEvent = 2;
```
BenBirt: protobuf fields are snake_case

BenBirt: we should put a comment on this, pointing at the enum of possible values

protos/schedules.proto
```diff
@@ -20,11 +44,15 @@ message Schedule {
   bool disabled = 2;
   ScheduleOptions options = 3;
   string cron = 4; // required field
+  // to support legacy
```
BenBirt: `// Deprecated`

protos/schedules.proto
```diff
@@ -2,12 +2,36 @@ syntax="proto3";
...
+
+message NotificationChannel {
+ string name = 1;
+ oneof channel {
+    EmailNotificationChannel email_notification = 1;
+    SlackNotificationChannel slack_notification = 2;
+  } 	
+}
+
+message ScheduleNotification {
```
BenBirt: @lewish now's the right time to make the final decision about whether we have a `schedules` field in here or not.

lewish: Personally, I'm happy with this...

protos/schedules.proto
faiyaz26: Added an Enum, and added a comment here

protos/schedules.proto
BenBirt: ok!

protos/schedules.proto
```diff
@@ -2,12 +2,42 @@ syntax="proto3";
 
 package dataform.schedules;
 
+enum NotificationEventType {
+  ON_ANY = 0;
```
BenBirt: i don't think we need this value (it corresponds to a non-existent `events` list IMO)

protos/schedules.proto
```diff
@@ -2,12 +2,42 @@ syntax="proto3";
...
+  ON_SUCCESS = 1;
```
BenBirt: IMO we should drop the "ON_" prefixes

protos/schedules.proto
faiyaz26: I get error, that Enum should have a index of 0, what should I do about that?

lewish: Do we need this enum at all?
If so, I'd just call the first value - UKNOWN (that's what we do everywhere else).

protos/schedules.proto
lewish: +1

PR #487, opened 2019-11-20
Action retries

### LINKED ISSUE #484: Implement automatic action retries
- automatically retry failed actions, where those actions are idempotent (non operations)
- add a 'idempotentActionRetries' setting to ProjectConfig, defaulted to 3 (?), which determines the number of retries

### LINKED ISSUE #485: Change action failure error messages
Include a warehouse specific prefix in errors thrown by dbadapter.execute.

e.g. "BigQuery error: ..."

GJMcGowan:
Fixes #484
Fixes #485


api/utils/retry_promise.ts
```diff
@@ -0,0 +1,14 @@
+type RetryReturnType = (func: () => Promise<any>, retries: number, error?: Error) => Promise<any>;
+
+export const retryPromise: RetryReturnType = async (func, retries, error) => {
```
BenBirt:
```suggestion
export async function retryPromise (func: () => Promise<any>, retries: number, error?: Error) => {
```

api/utils/retry_promise.ts
```diff
@@ -0,0 +1,14 @@
...
+  if (retries < 0) {
```
BenBirt: i think this would be cleaner if we put it inside the `catch` block, and check for when `retries === 0`. that way, you don't need to pass the `Error` down the recursion stack.

api/utils/retry_promise.ts
```diff
@@ -0,0 +1,14 @@
...
+    throw error;
+  }
+
+  try {
+    const funcResult = await func();
+    return funcResult;
```
BenBirt:
```suggestion
    return await func();
```

api/utils/retry_promise.ts
BenBirt: can we just call this function, and file, `retry`?

api/commands/run.ts
```diff
@@ -247,10 +248,15 @@ export class Runner {
...
-        maxResults: 1
-      });
+      // Retry this function a given number of times, configurable by user
+      const rows = await retryPromise(
+        () =>
+          this.adapter.execute(task.statement, {
+            onCancel: handleCancel => this.eEmitter.on(CANCEL_EVENT, handleCancel),
+            maxResults: 1
+          }),
+        this.graph.projectConfig.idempotentActionRetries || 0
```
BenBirt: we should *always* have zero retries for any non-idempotent action, i.e. any task of type "operations"

(longer term we will want to let the user mark their actions as idempotent or not, but for now we can assume that operations aren't, everything else is.)

api/commands/run.ts
```diff
@@ -264,7 +270,9 @@ export class Runner {
       taskResult.status = this.cancelled
         ? dataform.TaskResult.ExecutionStatus.CANCELLED
         : dataform.TaskResult.ExecutionStatus.FAILED;
-      taskResult.errorMessage = e.message;
+      const warehouse = this.graph.projectConfig.warehouse;
+      const capitalisedWarehouse = `${warehouse.charAt(0).toUpperCase()}${warehouse.slice(1)}`;
```
BenBirt: let's not bother, this will end up being super weird for e.g. "Bigquery" and "Sqldatawarehouse". i'm fine with just lowercasing it, but if you prefer then feel free to add a user visible name mapping along the lines of the stuff we have near `WarehouseType`

protos/core.proto
```diff
@@ -12,6 +12,7 @@ message ProjectConfig {
   string assertion_schema = 5;
 
   string schema_suffix = 7;
+  int32 idempotent_action_retries = 8;
```
BenBirt: nit: let's put it above `schema_suffix` which folks aren't really supposed to set in `dataform.json`.

api/commands/run.ts
BenBirt: what about a default value of `3`?

api/utils/retry_promise.ts
GJMcGowan: Had to do a manual type, but I implemented the spirit of your suggestion:

```
export async function retryPromise(
  func: () => Promise<any>,
  retries: number,
  error?: Error
): Promise<any> {
```

api/commands/run.ts
GJMcGowan: Scary, done

GJMcGowan: Seems like it could make runs a lot slower to fail if a ton of stuff is failing though

GJMcGowan: Changed line to this:

```
task.type === "operation" ? 0 : this.graph.projectConfig.idempotentActionRetries || 3
```

So operations will never retry.

PR #492, opened 2019-11-27
Make building much faster

lewish:
- Introduce a prune command to apply all filters against a compiled graph
- Reduce algorithmic complexity of a lot of the filtering code significantly
- Only compute warehouse state for actions that will actually be run

### DIFF
--- a/api/commands/build.ts
+++ b/api/commands/build.ts
@@ -4,18 +4,22 @@ import * as dbadapters from "@dataform/api/dbadapters";
 import { adapters } from "@dataform/core";
 import * as utils from "@dataform/core/utils";
 import { dataform } from "@dataform/protos";
+import { prune } from "df/api/commands/prune";
 
-export function build(
+export async function build(
   compiledGraph: dataform.ICompiledGraph,
   runConfig: dataform.IRunConfig,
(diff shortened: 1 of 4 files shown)


api/commands/build.ts
```diff
@@ -4,18 +4,22 @@ import * as dbadapters from "@dataform/api/dbadapters";
...
 import { dataform } from "@dataform/protos";
+import { prune } from "df/api/commands/prune";
 
-export function build(
+export async function build(
   compiledGraph: dataform.ICompiledGraph,
   runConfig: dataform.IRunConfig,
-  credentials: Credentials
+  credentials: Credentials,
+  overrideState?: dataform.IWarehouseState
```
BenBirt: what's the thinking here?

lewish: Just for testing, I could change the tests to just test the prune() method directly but that leaves this code untested except for e2e or example tests. Maybe that's fine if you'd prefer it?

BenBirt: removing it seems best IMO since this is a public API and we'd rather people don't add dependencies on this parameter.

lewish: Done!

PR #498, opened 2019-11-28
Changes warehouse state calculation to throw an error instead of swallowing it

### LINKED ISSUE #495: Errors not handled during state computation can lead to unexpected behaviour
When we attempt to compute state in the warehouse, and get information about a table / view at run startup, we silently catch errors and treat it as meaning the table does not exist.

This is not OK, because the table fetch request may fail for other reasons, such as flakiness. This can lead to failed runs, as the wrong built SQL will be generated.

GJMcGowan:
Fixes #495

### DIFF
--- a/api/commands/state.ts
+++ b/api/commands/state.ts
@@ -6,11 +6,13 @@ export async function state(
   dbadapter: IDbAdapter
 ): Promise<dataform.IWarehouseState> {
   const allTables = await Promise.all(
-    compiledGraph.tables
-      .map(async t => dbadapter.table(t.target))
-      // Skip tables that don't exist.
-      .filter(async table => !!(await table).type)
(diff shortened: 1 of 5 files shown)


api/commands/state.ts
```diff
@@ -1,23 +1,18 @@
...
 import { dataform } from "@dataform/protos";
 
-export function state(
+export async function state(
   compiledGraph: dataform.ICompiledGraph,
   dbadapter: IDbAdapter
 ): Promise<dataform.IWarehouseState> {
-  const tables: dataform.ITableMetadata[] = [];
+  const allTables = await Promise.all(
+    compiledGraph.tables
```
BenBirt:
```suggestion
    compiledGraph.tables
      .map(async t => dbadapter.table(t.target))
      // Skip tables that don't exist.
      .filter(async table => !!(await table).type)
```

PR #499, opened 2019-11-28
Implement late binding views in redshift by default

### LINKED ISSUE #227: Automatically create all views in Redshift as late binding views
Late binding views are probably a good idea as we cascade drop tables in Redshift. This should generally mean that downstream views are still query-able even after only a portion of the dataform run has executed.

https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_VIEW.html

- [ ] Automatically specify `WITH NO SCHEMA BINDING` for views
- [ ] Add a redshift flag called `bind` that when set to true, doesn't specify the above
- [ ] Update the documentation to cover this

lewish:
This avoids possible transaction conflicts, also reduces downtime due to the fact that we have to do a drop cascade.

Fixes #227

### DIFF
--- a/core/adapters/redshift.ts
+++ b/core/adapters/redshift.ts
@@ -67,9 +67,8 @@ export class RedshiftAdapter extends Adapter implements IAdapter {
     const createQuery = `create or replace view ${this.resolveTarget(target)} as ${query}`;
     if (bind) {
       return createQuery;
-    } else {
-      return `${createQuery} with no schema binding`;
     }
+    return `${createQuery} with no schema binding`;
   }
 
   public createOrReplace(table: dataform.ITable) {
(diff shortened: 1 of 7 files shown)


core/adapters/redshift.ts
```diff
@@ -63,9 +63,13 @@ export class RedshiftAdapter extends Adapter implements IAdapter {
...
   }
 
-  public createOrReplaceView(target: dataform.ITarget, query: string) {
-    return `
-      create or replace view ${this.resolveTarget(target)} as ${query}`;
+  public createOrReplaceView(target: dataform.ITarget, query: string, bind = false) {
+    const createQuery = `create or replace view ${this.resolveTarget(target)} as ${query}`;
+    if (bind) {
+      return createQuery;
+    } else {
```
BenBirt: nit: no need for else, just return

BenBirt:
you'll also need to increment `version.bzl`

PR #503, opened 2019-12-02
Add versioning to make bind default in redshift before 1.4.1

lewish:
- bind changes correctness of some SQL, so we will only make it the default on versions from where it is supported.

### DIFF
--- a/core/adapters/redshift.ts
+++ b/core/adapters/redshift.ts
@@ -1,14 +1,12 @@
 import { dataform } from "@dataform/protos";
+import * as semver from "semver";
 import { Task, Tasks } from "../tasks";
 import { Adapter } from "./base";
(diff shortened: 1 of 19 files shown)


protos/core.proto
```diff
@@ -234,6 +234,8 @@ message CompiledGraph {
 
   GraphErrors graph_errors = 7;
 
+  string version = 10;
```
BenBirt: `dataform_core_version`

core/adapters/redshift.ts
```diff
@@ -1,14 +1,12 @@
 import { dataform } from "@dataform/protos";
+import * as semver from "semver";
 import { Task, Tasks } from "../tasks";
```
BenBirt: can we fix these imports?

PR #512, opened 2019-12-11
Docs site refresh

lewish:
image

Make it easier to write new docs
- Remove MDX, use vanilla markdown and front-matter
- Start using content from `content/`
- Add support for loading content from local FS or Github branches

Restyling of docs site
- Move search bar to the top of the page
- Add page overview and edit page links to right of content
- Increase the max width of the site

Other stuff
- Upgrade to NextJS 9 for dynamic paths
- Rewrite nextjs bazel rules using the new npm_package_bin libraries
- Make the a new docs site docker image which runs next instead of a static site


docs/components/header.tsx
```diff
@@ -11,73 +10,34 @@ export interface IPage {
...
-  pages?: IPage[];
-  currentPath?: string;
   maxWidth?: string;
 }
 
-interface IState {
-  showFixedNav?: boolean;
-  showMobileMenu?: boolean;
-}
+interface IState {}
```
GJMcGowan: Can we just delete from from the class?

docs/components/navigation.css
```diff
@@ -1,28 +1,37 @@
-.navigation {
+.navigation ul {
+  list-style-type: none;
+  padding-inline-start: 0px;
 }
 
-.navigation h4 {
-  margin-left: 10px;
+.navigation ul:not(.depth0) {
+  padding-inline-start: 8px;
```
GJMcGowan: This is pretty experimental (I'd never heard of it!) and specifically Edge wouldn't understand it (https://developer.mozilla.org/en-US/docs/Web/CSS/padding-inline-start#Browser_compatibility).

Can we change it to something more normal?

docs/components/navigation.css
```diff
@@ -1,28 +1,37 @@
...
-.sidebar > .bp3-menu {
-  background-color: unset;
+.navigation li {
+  padding: 0px 8px;
+  border-radius: 8px;
 }
 
-.menu {
-  color: #2f2d44;
+.navigation li.depth0.hasChildren {
```
GJMcGowan: Many of these styles are quite specific and brittle. Can they each be a class and put directly on the element?

docs/components/page_links.css
```diff
@@ -0,0 +1,9 @@
+ul.pageLinks {
+  list-style-type: none;
+  padding-inline-end: 10px;
```
GJMcGowan: As above, Edge won't like it.

docs/components/page_links.tsx
```diff
@@ -0,0 +1,25 @@
...
+interface IProps {
+  links: IHeaderLink[];
+}
+
+export class PageLinks extends React.Component<IProps> {
+  public render() {
+    return (
+      <ul className={styles.pageLinks}>
+        {this.props.links.map(link => (
+          <li>
```
GJMcGowan: Needs a key

docs/components/page_links.tsx
```diff
@@ -0,0 +1,25 @@
...
+export interface IHeaderLink {
+  id: string;
+  text: string;
+}
+
+interface IProps {
+  links: IHeaderLink[];
+}
+
+export class PageLinks extends React.Component<IProps> {
```
GJMcGowan: In general for small bits like this I would use a functional component instead of a class, less code and easier to think about.

docs/components/search/index.css
```diff
@@ -30,3 +30,6 @@
     left: 0;
   }
 }
+
+.instantSearch {
```
GJMcGowan: Can be deleted?

docs/components/search/index.tsx
```diff
@@ -74,14 +78,13 @@ class SearchWidget extends React.Component<IWidgetProps, IWidgetState> {
...
-          <MenuItem
-            icon="search"
-            text="Search..."
-            onClick={() => this.setState({ isOpen: true })}
-          />
-        </Menu>
+      <React.Fragment>
+        <Button
+          icon="search"
+          style={{ color: "black" }}
```
GJMcGowan: Move to css please

docs/components/search/index.tsx
```diff
@@ -74,14 +78,13 @@ class SearchWidget extends React.Component<IWidgetProps, IWidgetState> {
...
     return (
-      <div>
-        <Menu>
-          <MenuItem
-            icon="search"
-            text="Search..."
-            onClick={() => this.setState({ isOpen: true })}
-          />
-        </Menu>
+      <React.Fragment>
```
GJMcGowan: Can just be `<>`

docs/global.css
```diff
@@ -2,25 +2,9 @@
...
-  -webkit-box-shadow: unset !important;
-  border-radius: 4px;
-  background-image: unset !important;
-}
-
-:global body .bp3-button * {
-  color: #fff;
-  font-size: 12px;
+:global body {
+  overflow-y: scroll !important;
```
GJMcGowan: Does this need `!important`? Placing this style on the `body` is not ideal either - would be nice to have it in some lower bit of html.

docs/components/navigation.css
```diff
@@ -1,28 +1,37 @@
...
-  letter-spacing: 0.04em;
+.navigation li.active * {
+  color: #fff;
+}
+
+.navigation li.depth0.hasChildren:not(.active) a {
+  color: unset;
+}
+
+.navigation a:hover {
```
GJMcGowan: You have something similar in your main styles css, is this also necessary?
```
:global body a:hover {
  text-decoration: none;
}
```

docs/components/page_links.tsx
lewish: My general position on this has been to always do the same thing everywhere. However, I will try it, for you.

docs/global.css
lewish: I can't see an easy way to avoid this - it's purpose is for the search omnibar, which causes a big render jump when it opens because the scrollbar disappears. In practice, the entire site is scrolling on every page anyway.

PR #513, opened 2019-12-11
Added exit codes to CLI, where non-zero indicates failure

### LINKED ISSUE #394: CLI should return a non-zero exit code if any compile/build/test/run step fails

Ekrekr:
Fixes #394 

I also added an additional `process.exit()` for watcher SIGINT to conform with others.

Also unsure whether `process.exit()` could be substituted for `return` - CLI tests might be useful for.

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -153,7 +153,7 @@ const builtYargs = createYargsCli({
       positionalOptions: [],
       options: [],
       processFn: async argv => {
-        return false;
+        return 0;
       }
     },
     {
@@ -201,20 +201,20 @@ const builtYargs = createYargsCli({
       ],
       processFn: async argv => {
         print("Writing project files...\n");
-        printInitResult(
-          await init(
-            argv["project-dir"],
-            {
-              warehouse: argv.warehouse,
-              gcloudProjectId: argv["gcloud-project-id"]
-            },
-            {
-              skipInstall: argv["skip-install"],
-              includeSchedules: argv["include-schedules"],
-              includeEnvironments: argv["include-environments"]
-            }
-          )
+        const initResult = await init(
+          argv["project-dir"],
+          {
+            warehouse: argv.warehouse,
+            gcloudProjectId: argv["gcloud-project-id"]
+          },
+          {
+            skipInstall: argv["skip-install"],
+            includeSchedules: argv["include-schedules"],
+            includeEnvironments: argv["include-environments"]
+          }
         );
+        printInitResult(initResult);
+        return 0;
       }
     },
     {
@@ -278,6 +278,7 @@ const builtYargs = createYargsCli({
         const filePath = path.resolve(argv["project-dir"], credentials.CREDENTIALS_FILENAME);
         fs.writeFileSync(filePath, prettyJsonStringify(finalCredentials));
         printInitCredsResult(filePath);
+        return 0;
       }
     },
     {
@@ -311,9 +312,11 @@ const builtYargs = createYargsCli({
           if (compiledGraphHasErrors(compiledGraph)) {
             print("");
             printCompiledGraphErrors(compiledGraph.graphErrors);
+            return true;
           }
+          return false;
         };
-        await compileAndPrint();
+        const graphHasErrors = await compileAndPrint();
 
         if (argv.watch) {
           let timeoutID: NodeJS.Timer = null;
@@ -366,11 +369,13 @@ const builtYargs = createYargsCli({
             watching = false;
             process.exit(1);
           });
+          if (!watching) {
+            return graphHasErrors ? 1 : 0;
+          }
           while (watching) {
             await new Promise((resolve, reject) => setTimeout(() => resolve(), 100));
           }
         }
-        process.exit(0);
       }
     },
     {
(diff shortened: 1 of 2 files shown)


cli/index.ts
```diff
@@ -359,11 +364,13 @@ const builtYargs = createYargsCli({
           process.on("SIGINT", () => {
             watcher.close();
             watching = false;
+            process.exit(1);
           });
           while (watching) {
             await new Promise((resolve, reject) => setTimeout(() => resolve(), 100));
           }
         }
+        process.exit(0);
```
BenBirt: i'm not sure this belongs here. this code should never really execute if `watching === true`. i would expect instead a `return 0` if `watching === false`, before we enter the `if (watching)` block.

in fact, that `return` should really return a 0 or 1 as according to whether or not the `CompiledGraph` contains errors: https://source.cloud.google.com/tada-analytics/github_dataform-co_dataform/+/master:protos/core.proto;l=235?q=compiledgraph

cli/index.ts
```diff
@@ -399,6 +406,7 @@ const builtYargs = createYargsCli({
           compiledGraph.tests
         );
         testResults.forEach(testResult => printTestResult(testResult));
+        return 0;
```
BenBirt: this should only `return 0` if all tests were successful: https://source.cloud.google.com/tada-analytics/github_dataform-co_dataform/+/master:protos/core.proto;l=362?q=testresult

cli/index.ts
```diff
@@ -511,7 +519,9 @@ const builtYargs = createYargsCli({
         };
 
         runner.onChange(printExecutedGraph);
-        printExecutedGraph(await runner.resultPromise());
+        const runResult = await runner.resultPromise()
+        printExecutedGraph(runResult);
+        return ((runResult.status === dataform.RunResult.ExecutionStatus.FAILED) ? 1 : 0);
```
BenBirt: are you using Prettier with vscode? i'd expect it to remove all these brackets upon save (if you have format-on-save turned on)

BenBirt:
You'll also need to change `yargswrapper` to handle the new returned exit code, and exit with that code.

cli/index.ts
Ekrekr: Now records the result of `compiledGraphHasErrors` and returns according to the result of that if watching is not set

cli/index.ts
Ekrekr: 👍

cli/index.ts
Ekrekr: 👍

Ekrekr:
Replaced `any` type of processFn with `Promise<number>`, added return statements with correct type accordingly.

cli/index.ts
```diff
@@ -201,20 +201,20 @@ const builtYargs = createYargsCli({
...
+            gcloudProjectId: argv["gcloud-project-id"]
+          },
+          {
+            skipInstall: argv["skip-install"],
+            includeSchedules: argv["include-schedules"],
+            includeEnvironments: argv["include-environments"]
+          }
         );
+        printInitResult(initResult);
+        return 0;
```
Ekrekr: I was expecting `init` to have a failure condition that can be checked for; nothing wrong with leading this expanded for when that is implemented later.

Ekrekr: *meant to be a single line comment, oops

cli/index.ts
```diff
@@ -366,11 +369,13 @@ const builtYargs = createYargsCli({
             watching = false;
             process.exit(1);
           });
+          if (!watching) {
```
BenBirt: i don't *believe* this line can ever be called? (or if it is, it gets called when the watcher is about to start running, but this will have the effect of killing the process at that time)

i'd expect this to be above the `if (argv.watch) {` line, and only return if `!argv.watch`

cli/index.ts
```diff
@@ -405,8 +410,14 @@ const builtYargs = createYargsCli({
...
-        testResults.forEach(testResult => printTestResult(testResult));
-        return 0;
+        let allTestsSuccessful = true;
+        testResults.forEach(testResult => {
+          printTestResult(testResult);
+          if (testResult.successful === false) {
+            allTestsSuccessful = false;
+          }
+        });
+        return allTestsSuccessful ? 0 : 1;
```
BenBirt: nit: i think this might be a little cleaner as `return testResults.some(testResult => !testResult.successful) ? 1 : 0`

cli/index.ts
```diff
@@ -519,9 +530,9 @@ const builtYargs = createYargsCli({
         };
 
         runner.onChange(printExecutedGraph);
-        const runResult = await runner.resultPromise()
+        const runResult = await runner.resultPromise();
         printExecutedGraph(runResult);
-        return ((runResult.status === dataform.RunResult.ExecutionStatus.FAILED) ? 1 : 0);
+        return runResult.status === dataform.RunResult.ExecutionStatus.FAILED ? 1 : 0;
```
BenBirt: suggest inverting this to only return `0` when the status is `SUCCESSFUL`; that way all other statuses are guaranteed to return `1` (which is probably more future-compatible with new statuses, or at least more safely compatible with them)

cli/yargswrapper.ts
```diff
@@ -26,8 +26,8 @@ export function createYargsCli(cli: ICli) {
       command.description,
       (yargsChainer: yargs.Argv) => createOptionsChain(yargsChainer, command),
       async (argv: { [argumentName: string]: any }) => {
-        await command.processFn(argv);
-        process.exit();
+        const result = await command.processFn(argv);
```
BenBirt: nit: `exitCode`

cli/index.ts
Ekrekr: nit 2.0: `testResults.every(testResult => testResult.successful) ? 0 : 1`? One less operator to think through.

cli/index.ts
Ekrekr: 👍

cli/index.ts
Ekrekr: Ah, good spot :)

cli/yargswrapper.ts
Ekrekr: 👍

PR #517, opened 2019-12-11
Convert --verbose to --json, useful for piping output

### LINKED ISSUE #377: Create a --json flag for compilation to output the compiled graph as pure json
currently `dataform compile --verbose` will output raw JSON, however it also prints out `Compiling project...` at the top which makes it invalid json so it can't be used for automation, bash scripts, etc.

Proposal:
- Change --verbose to --json (because that's what it's actually outputting)
- Make sure that --json only ever prints out the pure json, and no other logging messages

Ekrekr:
Fixes #377


cli/index.ts
```diff
@@ -133,11 +133,11 @@ const warehouseOption: INamedOption<yargs.PositionalOptions> = {
...
 };
 
-const verboseOutputOption: INamedOption<yargs.Options> = {
-  name: "verbose",
+const jsonOutputOption: INamedOption<yargs.Options> = {
+  name: "json",
   option: {
     describe:
-      "If true, the full contents of command output will be output (containing fully compiled SQL, etc).",
+      "Exclusively output a JSON describing the non-executable graph (other output is suppressed, excluding errors).",
```
lewish: "Outputs a JSON representation of the compiled project" - I think is slightly more clear.

Ekrekr: Ah nice, I guess they'll see what happens anyway when they run it

PR #520, opened 2019-12-11
Throw validation error for partitionBy if BigQuery view

### LINKED ISSUE #421: If a user tries to add a partitionBy statement for a view in bigquery, throw a compilation error
partitionBy is only supported for tables

---
Moved from dataform-co/dataform-co#1749
/cc @dwl285

Ekrekr:
Fixes #421 

* Added a validation error for if a user tries to specify a `partitionBy` while the action is for `bigquery` of type `view` (BigQuery only allows `partitionBy` for type `table`).

* Added a test, similar to the existing redhshift tests, that tests that the validation error is indeed ✋💨🥎 as expected.

* Changed the `TEST_CONFIG` environment variable in tests as there are currently 3 and likely to be more in the future. See comment below.

### DIFF
--- a/tests/core/core.spec.ts
+++ b/tests/core/core.spec.ts
@@ -7,19 +7,27 @@ import { expect } from "chai";
 import { asPlainObject } from "df/tests/utils";
 import * as path from "path";
 
-const TEST_CONFIG: dataform.IProjectConfig = {
-  warehouse: "redshift",
-  defaultSchema: "schema"
-};
+class TEST_CONFIG {
+  public static REDSHIFT: dataform.IProjectConfig = {
+    warehouse: "redshift",
+    defaultSchema: "schema"
+  };
 
-const TEST_CONFIG_WITH_SUFFIX: dataform.IProjectConfig = {
-  ...TEST_CONFIG,
-  schemaSuffix: "suffix"
-};
+  public static REDSHIFT_WITH_SUFFIX: dataform.IProjectConfig = {
+    warehouse: "redshift",
+    defaultSchema: "schema",
+    schemaSuffix: "suffix"
+  };
+
+  public static BIGQUERY: dataform.IProjectConfig = {
+    warehouse: "bigquery",
+    defaultSchema: "schema"
+  };
+}
 
 describe("@dataform/core", () => {
   describe("publish", () => {
-    [TEST_CONFIG, TEST_CONFIG_WITH_SUFFIX].forEach(testConfig => {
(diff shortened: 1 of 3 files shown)


tests/core/core.spec.ts
```diff
@@ -7,19 +7,27 @@ import { expect } from "chai";
...
+  public static REDSHIFT: dataform.IProjectConfig = {
+    warehouse: "redshift",
+    defaultSchema: "schema"
+  };
+
+  public static REDSHIFT_WITH_SUFFIX: dataform.IProjectConfig = {
+    warehouse: "redshift",
+    defaultSchema: "schema",
+    schemaSuffix: "suffix"
+  };
```
Ekrekr: This is a less elegant method of duplicating the prior `REDSHIFT` that `  ...TEST_CONFIG,
  schemaSuffix: "suffix"`; this is because the variables have to be static in order to be used in the way they currently are.

Is there a workaround for this?

tests/core/core.spec.ts
```diff
@@ -7,19 +7,27 @@ import { expect } from "chai";
...
-const TEST_CONFIG: dataform.IProjectConfig = {
-  warehouse: "redshift",
-  defaultSchema: "schema"
-};
-
-const TEST_CONFIG_WITH_SUFFIX: dataform.IProjectConfig = {
-  ...TEST_CONFIG,
-  schemaSuffix: "suffix"
-};
+class TEST_CONFIG {
```
lewish: I'd call this `TestConfigs` (we have classes as CamelCase, we should add a tslint rule for this!).

tests/core/core.spec.ts
lewish: Does it not work if you do: 
```
{
  ...TEST_CONFIG.REDSHIFT
}
```
?

tests/core/core.spec.ts
```diff
@@ -311,8 +319,35 @@ describe("@dataform/core", () => {
       ).to.have.deep.members(expectedResults);
     });
 
+    it("validation_bigquery_fail", () => {
```
lewish: I think you can give this a slightly more specific name

Ekrekr: Ah, I replicated from the `"validation_redshift_fail"` from above, as that's used as a parent wrapper for all the redshift specific validation failure tests. Likewise this is the parent wrapper for all bigquery specific validation failure tests (although now there is only one).

tests/core/core.spec.ts
Ekrekr: Thanks, much prettier now

tests/core/core.spec.ts
Ekrekr: Yes that's it! cheers.

PR #521, opened 2019-12-12
Force new line if premature end of file after js block

### LINKED ISSUE #401: js blocks in SQLX files aren't picked up if there isn't at least one line after the block closes
Steps to reproduce:

Copy these 5 lines into a SQLX file:
select ${TEST}

js {
  const TEST = "test";
}
Expected result: works
Actual result: compile error "missing ) after argument list"

---
Moved from dataform-co/dataform-co#1521
/cc @dwl285

Ekrekr:
Fixes #401 

This issue was caused by the `jsBlockLexer` which expects one or more of any character after the block, which is described as `/[\s\S]+?/`. Attempting to allow end of the file term (`/[\s\S$]+?/`) did not work, nor did replacing `+` with `*` as that would allow empty strings, which is illegal in moo.js.

The current solution appends a new line if the end of the file has nothing after a "}". Sorry @BenBirt.


BenBirt:
Are you sure about the cause here? `jsBlockLexer` doesn't expect a character after the block, `/[\s\S]+?/` captures all text inside the block which doesn't match one of the previous regexps.

Ekrekr:
> Are you sure about the cause here? `jsBlockLexer` doesn't expect a character after the block, `/[\s\S]+?/` captures all text inside the block which doesn't match one of the previous regexps.

Ah that might be adding `$` to the regex wasn't working. I'll have another play around with it tomorrow.

lewish:
I'll hold off review on this

lewish:
(FWIW if was my suggestion to add newlines :laughing: )

lewish:
Any luck with this?

Ekrekr:
I can't find an easy solution to this in the lexer. It may be something to do with moo.js; POSIX standards say that a line is only a line if it [ends in a terminating character](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206), so it may be to do with this.

Ekrekr:
Issue was being cause by an edge case where curly braces are only truncated from the end of the JS block if the next state is not JS. As there was no next state, just the end of the file, the curly brace was not being truncated.

PR #523, opened 2019-12-13
Open source contribution extension

Ekrekr:
Instructions should work out the box (at least for now), preventing confusion about the `env.yaml` file if they try to build the entire solution.

I also added some info on events and such, which I think is nice.

### DIFF
--- a/contributors.md
+++ b/contributors.md
@@ -9,7 +9,7 @@ For help making changes, join our `#development` channel in [Slack](https://slac
 
 ## Getting Started
 
-First clone this repository, and navigate within.
+First :fork_and_knife: [fork this repository](https://github.com/dataform-co/dataform/fork), clone it to your desktop, and navigate within.
 
 ### Run the CLI
 
@@ -19,7 +19,7 @@ Print out the default help information:
 ./scripts/run help
 ```
 
-Create a new `redshift` project in a temp directory:
+Create a new `redshift` toy project:
 
 ```bash
 mkdir /tmp/test_project
@@ -30,7 +30,7 @@ More commands can be found by substituting `dataform` with `./scripts/run` in th
 
 ### Test
 
-To test the project, run the following command:
+To test the Dataform project, run the following command:
 
 ```bash
 bazel test -- tests/... -tests/integration/...
@@ -38,7 +38,7 @@ bazel test -- tests/... -tests/integration/...
 
 This runs all tests excluding the integration tests. If you need to run integration tests, please [get in touch](mailto:opensource@dataform.co) with the team.
 
-_Note: A `java failed` error suggests that java needs to be installed._
+_Note: A `java failed` error suggests that Java needs to be installed._
 
 ### Building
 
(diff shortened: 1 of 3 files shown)


contributors.md
```diff
@@ -1,49 +1,58 @@
 # Contributing
 
 Dataform is a TypeScript project, and is fairly easy to build and run locally.
-For help making changes, join our `#development` channel in [Slack](https://slack.dataform.co).
+For help making changes, join our `#development` channel in [Slack](https://slack.dataform.co), or [email us](mailto:opensource@dataform.co) directly.
```
BenBirt: is this actually an email address we have? or will this just all go to guillaume?

contributors.md
```diff
@@ -1,49 +1,58 @@
...
+Create a new `redshift` toy project:
 
 ```bash
-bazel test ...
+mkdir /tmp/test_project
+./scripts/run init redshift /tmp/test_project
 ```
 
-_Note: Some integration tests may fail due to missing credentials. If you need to run these tests, reach out to the team on [Slack](https://slack.dataform.co)._
+More commands can be found by substituting `dataform` with `./scripts/run` in the [CLI documentation](https://docs.dataform.co/guides/command-line-interface).
```
BenBirt:
```suggestion
More commands can be found by running `./scripts/run help`.
```

(alternatively just get rid of this line since you already mention this above)

contributors.md
```diff
@@ -1,49 +1,58 @@
...
+
+Running the CLI will build the required components. To build other components, for example the api, use:
 
 ```bash
-./scripts/run init redshift /tmp/test_project
+bazel build api
 ```
 
-See the [CLI documentation](https://docs.dataform.co/guides/command-line-interface/) for more commands.
+_Note: Building the entire project with `Bazel build ...` will not work unless you have credentials provided by the team. A workaround to this is to create an empty `tools/stackdriver-github-bridge/env.yaml` file._
```
BenBirt: this recommendation seems a little heavy-handed... can we just suggest they build everything except that which is inside `//tools/...`?

contributors.md
```diff
@@ -1,49 +1,58 @@
...
 
-## Run the documentation site
+_Note: If you are running Bazel on a **Mac**, `bazel build ...` may fail with a `Too many open files in system` error. This is [due to a limitation](https://github.com/angular/angular-bazel-example/issues/178) on the default maximum open file descriptors. You can increase the limit by running `sudo sysctl -w kern.maxfiles=<LARGE_NUMBER>` (we use `65536`)._
```
BenBirt: IMO this should still come early in the doc, since `/scripts/run` already does a very large build.

contributors.md
```diff
@@ -52,3 +61,11 @@ bazel run docs
...
 
 You can view the documentation site at [localhost:3001](http://localhost:3001).
+
+## Deciding what to contribute
+
+A good place to start is by solving [Issues](https://github.com/dataform-co/dataform/issues), with the aim of understanding the code base well enough to create new features.
+
+We also regularly run hackathons, but you'll have to [ask us directly](mailto:opensource@dataform.co) if you would like to participate.
+
+Come meet us at [one of our events](https://www.google.com/search?q=dataform+events)!
```
BenBirt: can you check with Josie if there is some alternative link / mailing list / something that she might like here?

contributors.md
Ekrekr: Have rejigged the order now so it should be a bit more clear.

contributors.md
Ekrekr: Have moved upwards and changed message a tad

contributors.md
Ekrekr: Good idea, `bazel build -- ... -tools/...` works

contributors.md
Ekrekr: team@dataform.co should work

Issue #501, opened 2019-11-28
Support changing database/google project ID per action


lewish:
Currently all datasets referenced, declared and created in a dataform project are scoped within a single database.

Snowflake and BigQuery support writing between databases, which is useful.

Design proposal here: https://docs.google.com/document/d/17PIgu2o87V5d670d0uEta0iEeEwIr_sJtLPyquleD7U/edit#

PR #529, opened 2019-12-18
Allow users to override 'database' on a per-action basis.

BenBirt:
To keep things sane, I've had to change how we track and resolve dependencies:
- similarly to how we have deprecated the `name` field for actions (preferring `target`), I have now deprecated the string form of `dependencies` in favour of a `dependency_targets` field
- during action compilation, `dependency_targets` contains all requested dependencies (which may not be fully-qualified `Target` protobufs)
- at the end of compilation, we find all requested dependencies, resolving any special cases (like `inline` tables), performing any necessary de-deplication

fixes https://github.com/dataform-co/dataform/issues/501


core/session.ts
```diff
@@ -424,38 +448,48 @@ export class Session {
...
+            fullyQualifiedDependencies[protoDep.name] = protoDep.target;
+          }
+        } else {
+          // Too many targets matched the dependency.
+          this.compileError(new Error(utils.ambiguousActionNameMsg(dependency, possibleDeps)));
         }
-        return dependencyName;
-      });
+      }
+      action.dependencies = Object.keys(fullyQualifiedDependencies);
```
lewish: Why store this at all now? It's not warehouse specific, and simple to recompute

lewish: Ah I see, it's just during the deprecation period

tests/integration/snowflake_project/definitions/sample_data_2.sqlx
```diff
@@ -1,4 +1,4 @@
-config { type: "view" }
+config { type: "view", database: "TADA2" }
```
lewish: Does the snowflake integration test actually check if we created this successfully?

PR #532, opened 2019-12-19
Fix double preops/postops scripts on incremental table

faiyaz26:
Fixes: https://github.com/dataform-co/dataform-co/issues/2571


core/table.ts
```diff
@@ -208,7 +208,6 @@ export class Table {
   public compile() {
```
BenBirt: nit: can you revert this file to `master` since there are no changes in it?

PR #534, opened 2019-12-23
Added backtick replacing regex to fix compilation by interpreting bac…

Ekrekr:
…kticks literally when text is selected

Fixes https://github.com/dataform-co/dataform-co/issues/2004

### DIFF
--- a/api/commands/query.ts
+++ b/api/commands/query.ts
@@ -58,6 +58,10 @@ export async function compile(
   // Resolve the path in case it hasn't been resolved already.
   const projectDir = path.resolve(compileConfig.projectDir);
 
+  // This captures any single backticks that aren't escaped with a preceding \.
+  const RAW_BACKTICKS_REGEX = /(?!\\)`/g;
+  query = query.replace(RAW_BACKTICKS_REGEX, "\\`");
+
   return await CompileChildProcess.forkProcess().compile({
     ...compileConfig,
     projectDir,
--- a/api/commands/query.ts
+++ b/api/commands/query.ts
@@ -58,7 +58,6 @@ export async function compile(
   // Resolve the path in case it hasn't been resolved already.
   const projectDir = path.resolve(compileConfig.projectDir);
 
-  console.log("RAW QUERY:", query);
   // This captures any single backticks that aren't escaped with a preceding \.
   const RAW_BACKTICKS_REGEX = /(?!\\)`/g;
   query = query.replace(/\\/g, "\\\\").replace(RAW_BACKTICKS_REGEX, "\\`");
--- a/api/commands/query.ts
+++ b/api/commands/query.ts
@@ -58,9 +58,10 @@ export async function compile(
   // Resolve the path in case it hasn't been resolved already.
   const projectDir = path.resolve(compileConfig.projectDir);
 
+  console.log("RAW QUERY:", query);
   // This captures any single backticks that aren't escaped with a preceding \.
   const RAW_BACKTICKS_REGEX = /(?!\\)`/g;
-  query = query.replace(RAW_BACKTICKS_REGEX, "\\`");
+  query = query.replace(/\\/g, "\\\\").replace(RAW_BACKTICKS_REGEX, "\\`");
 
   return await CompileChildProcess.forkProcess().compile({
     ...compileConfig,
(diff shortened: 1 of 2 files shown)


tests/api/api.spec.ts
```diff
@@ -556,6 +556,18 @@ describe("@dataform/api", () => {
           );
         });
     });
+    it("bigquery example with input backticks", () => {
+      return query
+        .compile("select 1 from `tada-analytics.df_integration_test.example_view`", {
```
BenBirt: please add a test with text that contains a backslash, i think there is still a bug in the code for mishandling that case. (this is the exact same problem as you fixed in the lexer).

Ekrekr: Good point!

tests/api/api.spec.ts
```diff
@@ -556,6 +556,28 @@ describe("@dataform/api", () => {
...
+    it("bigquery example with a backslash in regex", () => {
+      return query
+        .compile("select regexp_extract('01a_data_engine', '^(\\d{2}\\w)')", {
+          projectDir: "df/examples/common_v1",
+          projectConfigOverride: { warehouse: "bigquery", defaultDatabase: "tada-analytics" }
+        })
+        .then(compiledQuery => {
+          expect(compiledQuery).equals("select regexp_extract('01a_data_engine', '^(\\d{2}\\w)')");
+        });
+    });
```
Ekrekr: I'm correct in thinking that the injected code we want in the query should have the same number of backslashes as we input, right? In the lexer they are doubled because it's yet to go through a JS abstraction, whereas here it is going through that abstraction, if you get what I mean...

api/commands/query.ts
```diff
@@ -58,6 +58,11 @@ export async function compile(
   // Resolve the path in case it hasn't been resolved already.
   const projectDir = path.resolve(compileConfig.projectDir);
 
+  console.log("RAW QUERY:", query);
+  // This captures any single backticks that aren't escaped with a preceding \.
+  const RAW_BACKTICKS_REGEX = /(?!\\)`/g;
+  query = query.replace(/\\/g, "\\\\").replace(RAW_BACKTICKS_REGEX, "\\`");
```
Ekrekr: Note backslashes are getting doubled here

tests/api/api.spec.ts
BenBirt: yes, that's right!

api/commands/query.ts
```diff
@@ -58,6 +58,10 @@ export async function compile(
   // Resolve the path in case it hasn't been resolved already.
   const projectDir = path.resolve(compileConfig.projectDir);
 
+  // This captures any single backticks that aren't escaped with a preceding \.
+  const RAW_BACKTICKS_REGEX = /(?!\\)`/g;
```
BenBirt: i'm not clear on why we only want to do this for backticks that aren't already escaped; IMO that's confusing. i think we should do it for *all* backticks; if you're escaping one then you want to send those bytes (i.e. a backslash and a backtick) to your warehouse; if we don't re-escape the backtick then those bytes will *not* be sent to the warehouse.

tests/api/api.spec.ts
```diff
@@ -556,6 +556,28 @@ describe("@dataform/api", () => {
           );
         });
     });
+    it("bigquery example with input backticks", () => {
+      return query
```
BenBirt: nit: can we refactor these tests to use await/async please?

api/commands/query.ts
Ekrekr: Ah yes, that makes sense

tests/api/api.spec.ts
Ekrekr: Sure, I think I've done this right

Ekrekr: But I have been wrong before

tests/api/api.spec.ts
```diff
@@ -544,8 +544,8 @@ describe("@dataform/api", () => {
   });
 
   describe("query", () => {
-    it("bigquery_example", () => {
-      return query
+    it("bigquery_example", async () => {
+      return await query
```
BenBirt: this isn't quite right, you want:

```
const compiledQuery = await query.compile(...);
expect(compiledQuery).equals(...);
```

PR #540, opened 2019-12-30
Dataform core API cleanup and reference documentation

lewish:
image

This contains a big cleanup of various interfaces and types within `@dataform/core`, and a starting point for the reference documentation.
- Documentation is generated using typedoc via a bazel rule
- All exposed interfaces have been cleaned up and given some documentation using jsdoc/typedoc annotations
- All protobuf interfaces have been removed from the exposed API. The protos don't always reflect the types we want to show to the user,  and documenting them consistently is tricky
- The reference docs themselves still need a bit of work before we publish them (UX wise)
- Some legacy context methods have been removed entirely from exposed interfaces (although will continue to work, if used)

You can run the docs site with `bazel run docs` and view the generated docs at `http://localhost:3001/reference`


core/common.ts
```diff
@@ -0,0 +1,116 @@
...
+ */
+export interface ITarget {
+  database?: string;
+
+  schema?: string;
+
+  name?: string;
+}
+
+/**
```
BenBirt: this text isn't rendering

core/assertion.ts
```diff
@@ -1,17 +1,20 @@
...
+export interface IAssertionContext extends ICommonContext {}
 
-export interface AConfig {
-  dependencies?: Resolvable | Resolvable[];
-  tags?: string[];
-  description?: string;
-  database?: string;
-  schema?: string;
-}
+export interface IAssertionConfig extends ICommonOutputConfig {}
```
BenBirt: can we override some of the documentation here? it's confusing that it refers to 'datasets' for assertion related code.

core/common.ts
```diff
@@ -0,0 +1,116 @@
...
+   */
+  resolve: (ref: Resolvable | string[], ...rest: string[]) => string;
+}
+
+/**
+ * @hidden
+ */
+export interface ICommonConfig {
+  /**
+   * One or more explicit dependencies for the table action. This is typically not needed, as it is inferred from using
```
BenBirt: this text is a little confusing/clumsy. suggest something more like ```One or more explicit dependencies for this action. Dependency actions will run before dependent actions. Typically this would remain unset, because most dependencies are declared as a by-product of using the `ref` function.```

core/common.ts
```diff
@@ -0,0 +1,116 @@
...
+   */
+  database?: string;
+
+  /**
+   * The schema (or dataset, in BigQuery) for this dataset (or table / view, in BigQuery).
+   */
+  schema?: string;
+
+  /**
+   * A description of the dataset that will be used to populate the data catalog.
```
BenBirt: I don't think we should mention "data catalog" here, partly because (a) that's one place where it might be used, but we may indeed put it other places and (b) this is open-source code and IMO we should have a light touch on referring to closed-source features.

core/common.ts
```diff
@@ -0,0 +1,116 @@
...
+ * @hidden
+ */
+export interface ICommonOutputConfig extends ICommonConfig {
+  /**
+   * The database (or Google Cloud project ID) for this dataset.
+   */
+  database?: string;
+
+  /**
+   * The schema (or dataset, in BigQuery) for this dataset (or table / view, in BigQuery).
```
BenBirt: this is somewhat confusing because of the BQ stuff! a suggestion: in the reference documentation, can we just settle on using 'database', 'schema', and 'dataset' everywhere, but wherever they are mentioned, we link to a special glossary page that explains the BQ specialness?

core/common.ts
```diff
@@ -0,0 +1,116 @@
...
+ */
+export interface ICommonConfig {
+  /**
+   * One or more explicit dependencies for the table action. This is typically not needed, as it is inferred from using
+   * the `ref` function within SQLX files.
+   */
+  dependencies?: Resolvable | Resolvable[];
+
+  /**
+   * A list of tags that should be applied to this action. This is useful for managing large projects.
```
BenBirt: stylistically, IMO we don't need to suggest people to only use features in "large" or "small" projects. suggest `A list of user-defined tags with which the action should be labeled.`

core/common.ts
```diff
@@ -0,0 +1,116 @@
+/**
+ * @hidden
+ */
+export interface ICommonContext {
+  /**
+   * Returns a valid SQL string that can be used to reference the dataset produced by this action.
```
BenBirt: ```, equivalent to `resolve(name())`.``` (i think this example makes it a little clearer what this does)

core/common.ts
```diff
@@ -0,0 +1,116 @@
...
+   */
+  self: () => string;
+
+  /**
+   * Returns the name of this dataset.
+   */
+  name: () => string;
+
+  /**
+   * References another action, returning valid SQL to be used in a `from` expression and adds it as a dependency to this action.
```
BenBirt: invert the order of clauses here?

`References another action, adding it as a dependency to this action, returning valid SQL to be used ...`

core/common.ts
```diff
@@ -0,0 +1,116 @@
...
+  description?: string;
+
+  /**
+   * A descriptor for columns within the dataset.
+   */
+  columns?: IColumnsDescriptor;
+}
+
+/**
+ * Describes columns in a dataset, used for populating the data catalog.
```
BenBirt: ditto re mentioning data catalog in docs

core/common.ts
```diff
@@ -0,0 +1,116 @@
...
+ */
+export interface IColumnsDescriptor {
+  [name: string]: string | IRecordDescriptor;
+}
+
+/**
+ * Describes a struct, object or record in a dataset that has nested columns.
+ */
+export interface IRecordDescriptor {
+  description?: string;
```
BenBirt: it's irritating that typedoc is reordering these (and other fields elsewhere), it makes the generated docs harder to read top-to-bottom. is that configurable? can we switch this sorting off?

core/declaration.ts
```diff
@@ -1,19 +1,18 @@
...
 import { dataform } from "@dataform/protos";
+import { IColumnsDescriptor, ICommonOutputConfig } from "@dataform/core/common";
 
-export interface DConfig {
-  description?: string;
-  columns?: IColumnsDescriptor;
-  database?: string;
-  schema?: string;
-}
+export interface IDeclarationConfig extends ICommonOutputConfig {}
```
BenBirt: this `extends` means that `IDeclarationConfig` ends up with a `dependencies` field which it doesn't have (and which is confusing IMO).

core/operation.ts
```diff
@@ -1,34 +1,45 @@
...
+  IColumnsDescriptor,
+  ICommonContext,
+  ICommonOutputConfig,
+  Resolvable
+} from "@dataform/core/common";
+import { Contextable } from "@dataform/core/contextable";
+
+export interface IOperationConfig extends ICommonOutputConfig {
+  /**
+   * Tells Dataform whether this operations action creates an output table
```
BenBirt: ```Declares that this `operations` action creates a dataset which should be referenceable using the `ref` function.```

core/operation.ts
```diff
@@ -1,34 +1,45 @@
...
+   * and is allowed to be referenced using the <code>ref</code> function.
+   *
+   * When set to true, the SQL operations should create a table or view of the given name.
```
BenBirt: ```If set to true, this action should create a dataset with its configured name, using the `self()` context function.```

core/operation.ts
```diff
@@ -1,34 +1,45 @@
...
+   * This can be accomplished using the <code>self()</code> context function.
+   *
+   * For example, an operation with output typically contains SQL such as:
```
BenBirt: ```For example:```

core/operation.ts
```diff
@@ -1,34 +1,45 @@
...
+   * ```sql
+   * create or replace table ${self()} as select ...
+   * ```
+   */
   hasOutput?: boolean;
-  database?: string;
-  schema?: string;
 }
 
+export interface IOperationContext extends ICommonContext {}
```
BenBirt: ditto with the confusing `dataset` terminology caused by extending `ICommonContext`, can we override some of the docs, or perhaps make the docs in `ICommonContext` a bit more generic?

core/table.ts
```diff
@@ -1,63 +1,165 @@
...
-export enum SortStyleTypes {
-  COMPOUND = "compound",
-  INTERLEAVED = "interleaved"
+
+/**
+ * Options for creating tables within BigQuery projects.
+ */
+export interface IBigQueryOptions {
+  /**
+   * The key for partitioning the table by, typically a timestamp or date.
```
BenBirt: ```The key with which to partition the table. Typically the name of a timestamp or date column.```

core/table.ts
```diff
@@ -1,63 +1,165 @@
...
+   *
+   * For more information, read the BigQuery documentation [here](https://cloud.google.com/bigquery/docs/partitioned-tables#next_steps).
```
BenBirt: seems weird to link to `#next_steps`

core/table.ts
```diff
@@ -1,63 +1,165 @@
...
+ * Redshift specific warehouse options.
+ */
+export interface IRedshiftOptions {
+  /**
+   * Sets the DISTKEY property when creating tables.
+   *
+   * For more information, read the AWS documentation [here](https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_TABLE_examples.html#r_CREATE_TABLE_NEW-diststyle-distkey-and-sortkey-options).
+   */
+  distKey?: string;
+  distStyle?: string;
```
BenBirt: are these intentionally undocumented?

core/table.ts
```diff
@@ -1,63 +1,165 @@
...
+  sortStyle?: string;
+  bind?: boolean;
 }
-export enum DistStyleTypes {
-  EVEN = "even",
-  KEY = "key",
-  ALL = "all"
+
+/**
+ * Options for creating tables within Azure SQL Datawarehouse projects.
```
BenBirt: nit: `Azure SQL Data Warehouse`

core/table.ts
```diff
@@ -1,63 +1,165 @@
...
-
-type ValueOf<T> = T[keyof T];
-export type TContextable<T> = T | ((ctx: TableContext) => T);
-export type TableType = ValueOf<TableTypes>;
-
-export interface TConfig {
+/**
+ * General options that can be provided to a table.
+ */
+export interface ITableConfig extends ICommonOutputConfig {
```
BenBirt: I feel like the name of this is confusing! Can we call it `IDatasetConfig`?

core/table.ts
```diff
@@ -1,63 +1,165 @@
...
-  ]
-};
-
-type ValueOf<T> = T[keyof T];
-export type TContextable<T> = T | ((ctx: TableContext) => T);
-export type TableType = ValueOf<TableTypes>;
-
-export interface TConfig {
+/**
+ * General options that can be provided to a table.
```
BenBirt: "General" here seems misleading, it implies that there are more specific settings you can configure some other way - when there aren't.

BenBirt: nit: `dataset`

core/table.ts
```diff
@@ -1,63 +1,165 @@
...
+   */
   protected?: boolean;
-  redshift?: dataform.IRedshiftOptions;
-  bigquery?: dataform.IBigQueryOptions;
-  sqldatawarehouse?: dataform.ISQLDataWarehouseOptions;
-  database?: string;
-  schema?: string;
+
+  /**
+   * Redshift specific warehouse options.
```
BenBirt: These should probably be hyphenated, `Redshift-specific`, (for the below options too)

core/table.ts
```diff
@@ -1,63 +1,165 @@
...
+   */
+  redshift?: IRedshiftOptions;
+
+  /**
+   * BigQuery specific warehouse options.
+   */
+  bigquery?: IBigQueryOptions;
+
+  /**
+   * Azure SQL data warehouse specific warehouse options.
```
BenBirt: nit: `Data Warehouse`

core/table.ts
```diff
@@ -1,63 +1,165 @@
...
-export type TContextable<T> = T | ((ctx: TableContext) => T);
-export type TableType = ValueOf<TableTypes>;
-
-export interface TConfig {
+/**
+ * General options that can be provided to a table.
+ */
+export interface ITableConfig extends ICommonOutputConfig {
+  /**
+   * The type of the table. For more information, check out the documentation [guides](guides).
```
BenBirt: "documentation guides" is a little clumsy, what are you trying to say? "dataset documentation" perhaps?

BenBirt: `dataset` plz

core/table.ts
```diff
@@ -1,63 +1,165 @@
...
+   */
   type?: TableType;
-  dependencies?: Resolvable | Resolvable[];
-  tags?: string[];
-  description?: string;
-  columns?: IColumnsDescriptor;
+
+  /**
+   * When set to true, the action and SQL queries will not be executed, however they will still be compiled
```
BenBirt: ```If set to true, this action will not be executed. However, the action may still be depended upon. Useful for ...```

core/table.ts
```diff
@@ -1,63 +1,165 @@
...
+   * and the action can still be dependend upon. Useful for temporarily turning off broken actions.
+   */
   disabled?: boolean;
+
+  /**
+   * Only allowed when the table type is <code>"incremental"</code>.
```
BenBirt: nit: rm the quotes around `incremental`?

core/table.ts
```diff
@@ -1,63 +1,165 @@
...
+   * When set to true, the full-refresh option will have no effect. This is useful for tables that
```
BenBirt: ```If set to true, running this action will ignore the full-refresh option. This is useful for tables which are built from transient data, to ensure that historical data is never lost.```

core/table.ts
```diff
@@ -1,63 +1,165 @@
...
+  bigquery?: IBigQueryOptions;
+
+  /**
+   * Azure SQL data warehouse specific warehouse options.
+   */
+  sqldatawarehouse?: ISQLDataWarehouseOptions;
 }
 
+/**
+ * Context methods are available when evaluating contextable SQL code, such as
```
BenBirt: can you add a similar comment to the other `Context` types?

core/table.ts
```diff
@@ -1,63 +1,165 @@
...
+import {
+  IColumnsDescriptor,
+  ICommonContext,
+  ICommonOutputConfig,
+  Resolvable
+} from "@dataform/core/common";
+import { Contextable } from "@dataform/core/contextable";
+
+/**
+ * @hidden
```
BenBirt: it doesn't seem like this is being obeyed? i don't mind it being documented, but let's keep "inline" out please

core/test.ts
```diff
@@ -1,24 +1,33 @@
...
 import { dataform } from "@dataform/protos";
+import { Resolvable, ICommonConfig, ICommonContext } from "@dataform/core/common";
+import { Contextable } from "@dataform/core/contextable";
 
-export type TContextable<T> = T | ((ctx: TestContext) => T);
-
-export interface TConfig {
+export interface ITestConfig {
+  /**
+   * The dataset that this unit test is to be run against.
```
BenBirt: ```The dataset that this unit test tests.```

core/test.ts
```diff
@@ -1,24 +1,33 @@
...
+   */
   dataset?: Resolvable;
 }
 
+export interface ITestContext extends ICommonContext {}
```
BenBirt: IMO this should be @hidden, nobody should really know this exists.

core/contextable.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Contextable arguments can either pass a plain value for their
+ * generic type `T` or can pass a function that will be called
+ * with the context object for this type of operation.
+ */
+export type Contextable<Context, T> = T | ((ctx: Context) => T);
```
BenBirt: why not just put this in `common.ts`?

core/session.ts
```diff
@@ -198,28 +187,28 @@ export class Session {
...
+            name: sqlxConfig.name
+          }).config(sqlxConfig);
         case "test":
-          return this.test(actionOptions.sqlxConfig.name);
+          return this.test(sqlxConfig.name).config(sqlxConfig);
         default:
-          throw new Error(`Unrecognized action type: ${actionOptions.sqlxConfig.type}`);
+          throw new Error(`Unrecognized action type: ${(sqlxConfig as SqlxConfig).type}`);
       }
-    })().config(actionOptions.sqlxConfig);
```
BenBirt: it was intentional to have this line here for DRY purposes, wdyt?

core/table.ts
```diff
@@ -1,63 +1,165 @@
...
-
-export interface TConfig {
+/**
+ * General options that can be provided to a table.
+ */
+export interface ITableConfig extends ICommonOutputConfig {
+  /**
+   * The type of the table. For more information, check out the documentation [guides](guides).
+   */
   type?: TableType;
```
Ekrekr: I think get rid of the second sentence here. `TableType` already links to the relevant section.

core/common.ts
lewish: SGTM!

core/common.ts
lewish: Unclear, having a look...

core/declaration.ts
lewish: Woops, doesn't support tags either. Just refactored the interfaces a bit

core/operation.ts
lewish: Made the common descriptions a bit more generic!

core/operation.ts
```diff
@@ -1,30 +1,40 @@
...
 } from "@dataform/core/common";
-import { Contextable } from "@dataform/core/contextable";
+import { Contextable } from "@dataform/core/common";
+import { mapToColumnProtoArray, Session } from "@dataform/core/session";
+import * as utils from "@dataform/core/utils";
+import { dataform } from "@dataform/protos";
 
-export interface IOperationConfig extends ICommonOutputConfig {
+/**
+ * Configuration options for operations.
```
BenBirt: nit: ```... for `operations` actions.```

PR #541, opened 2019-12-31
Make it possible to use Gazelle, and clean up proto BUILD rules.

BenBirt:


package.json
```diff
@@ -92,5 +92,8 @@
     "vm2": "^3.8.1",
     "webpack": "4.20.2",
     "yargs": "^12.0.1"
+  },
+  "devDependencies": {
+    "@bazel/hide-bazel-files": "latest"
```
lewish: Can we not use this plz, move it up to dependencies (also, lock a specific version)

PR #542, opened 2020-01-06
Add a protomongo package which exports a Go MongoDB codec for encoding/decoding protobuf objects (with protobuf field numbers as keys).

BenBirt:
Also add a `go.mod` file, which Gazelle can use to automatically update the contents of `WORKSPACE`:
- update `go.mod` by running `go build <dir>`, e.g. `go build ./protomongo`
- update `WORKSPACE` to match by running `bazel run //:gazelle -- update-repos -from_file=go.mod`


protomongo/protomongo.go
```diff
@@ -0,0 +1,134 @@
...
+	"go.mongodb.org/mongo-driver/bson/bsoncodec"
+	"go.mongodb.org/mongo-driver/bson/bsonrw"
+)
+
+// ProtobufCodec is a MongoDB codec. It encodes protobuf objects using the protobuf field numbers as document
+// keys. This means that stored protobufs can survive normal protobuf definition changes, e.g. renaming a field.
+type ProtobufCodec struct {
+}
+
+func (e *ProtobufCodec) EncodeValue(ectx bsoncodec.EncodeContext, vw bsonrw.ValueWriter, val reflect.Value) error {
```
lewish: `pc`?

protomongo/protomongo.go
```diff
@@ -0,0 +1,134 @@
...
+	for f, fvr, err := dr.ReadElement(); err != bsonrw.ErrEOD; f, fvr, err = dr.ReadElement() {
+		if err != nil {
+			return err
+		}
+
+		prop, isProp := indexedProps[f]
+		oneof, isOneof := indexedOneofProps[f]
+
+		// Skip any field that we don't recognize.
+		if !isProp && !isOneof {
```
lewish: This definitely means the tag doesn't exist, and there aren't any other possible types?

protomongo/protomongo_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+var (
+	tests = []struct {
+		name string
+		pb   proto.Message
+		new  func() proto.Message
+	}{
+		{
+			name: "simple message",
+			pb: &dataform.ProjectConfig{
```
lewish: Seems weird to use / depend on dataform protos. Why not make a testing proto message with every type in it?

protomongo/protomongo_test.go
```diff
@@ -0,0 +1,124 @@
...
+		if err = bson.UnmarshalWithRegistry(reg, b, &out); err != nil {
+			t.Errorf("bson.UnmarshalWithRegistry error = %v", err)
+		}
+		if !proto.Equal(testCase.pb, out) {
+			t.Errorf("failed: in=%#v, out=%#v", testCase.pb, out)
+		}
+	}
+}
+
+func TestMarshalUnmarshalWithPointers(t *testing.T) {
```
lewish: What's the point of this?

protomongo/protomongo.go
BenBirt: Yup! We must have either seen the tag in `props.Prop` or in `props.OneofTypes`, all of which are put into the relevant maps and then checked (`isProp` and `isOneof` being the result of those checks).

protomongo/protomongo_test.go
BenBirt: The MongoDB codec registry is a little tricksy, it seems to use registered codecs for non-pointer types for pointer-types too (if that's all it can resolve, at least). This caused me some issues when initially writing the code; I've made the library handle either case, and having both tests enforces that this behaviour continues to work.

PR #543, opened 2020-01-06
Web API documentation

lewish:
Auto generated API documentation from the web API swagger spec.

image


package.json
```diff
@@ -28,11 +28,13 @@
     "@types/rimraf": "^2.0.2",
     "@types/semver": "^6.2.0",
     "@types/stack-trace": "^0.0.29",
+    "@types/swagger-schema-official": "^2.0.20",
     "@types/yargs": "^11.1.1",
     "@zeit/next-css": "1.0.1",
     "@zeit/next-mdx": "^1.2.0",
     "@zeit/next-typescript": "^1.1.1",
     "algoliasearch": "^3.32.1",
+    "axios": "^0.19.0",
```
GJMcGowan: I'm surprised that this addition didn't change the `yarn.lock`

docs/components/swagger.css
```diff
@@ -0,0 +1,62 @@
...
+
+.propertyName {
+  max-width: 300px;
+  width: 50%;
+  display: inline-block;
+}
+
+.propertyDescription {
+  padding: 5px;
+  padding-left: 300px;
```
GJMcGowan: This is a very weird thing to do. Instead, consider just merging the `code` and `propertyDescription` elements into one and then having the page flow normally.

docs/global.css
```diff
@@ -98,13 +98,24 @@
   color: #2f2d44;
 }
 
+:global body h5 {
```
GJMcGowan: This whole file is quite repetitive.

Can you have a shared block for all the `h*`s which contains:
```
  font-family: "Poppins";
  font-weight: 500;
  letter-spacing: -0.02em;
  color: #2f2d44;
```

and then apply custom styling for each of them

docs/global.css
```diff
@@ -98,13 +98,24 @@
...
+  color: colorsTextMain;
```
GJMcGowan: Also, this color is overridden at the bottom of this block, so it's pointless.

docs/components/swagger.tsx
```diff
@@ -0,0 +1,204 @@
...
+      .reduce((acc, curr) => [...acc, ...curr], []);
+
+    return (
+      <div className={styles.container}>
+        <div className={styles.sidebar} />
+        <div className={styles.mainContent}>
+          <div className={styles.titleBlock}>
+            <h1>Dataform Web API</h1>
+            {allOperations.map((operation, i) => (
+              <Operation key={i} apiHost={apiHost} operation={operation} />
```
GJMcGowan: [Don't use indexes as keys](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318)

docs/components/swagger.tsx
```diff
@@ -0,0 +1,204 @@
...
+            ))}
+            {Object.keys(spec.definitions).map((name, i) => (
+              <Definition key={i} name={name} schema={spec.definitions[name]} />
```
GJMcGowan: Also here

GJMcGowan: (And a lot of other places)

package.json
lewish: It's already in there, just adding it as an explicit dependency now.

docs/components/swagger.tsx
lewish: As a general rule yes - however this content is completely static, it never changes as the data model will never change unless the user refreshes, at which point it doesn't matter.

lewish: OK OK I changed them

docs/components/swagger.css
lewish: Moved to flex

lewish:
> What's up with the left hand sidebar being empty?

This is on it's own at the moment as we don't want to publish it (Well, it's open-source, but just not promoted). I'm shortly about to make a big change to how the sidebar works that will somewhat fix this.

docs/components/swagger.css
```diff
@@ -39,23 +39,24 @@
...
 
 .properties {
-  
 }
 
 .propertyName {
   max-width: 300px;
-  width: 50%;
-  display: inline-block;
+  min-width: 300px;
```
GJMcGowan: Perhaps just `width` if the min and max are the same.

docs/global.css
```diff
@@ -55,58 +55,49 @@
...
+  letter-spacing: -0.03em;
+  color: #2f2d44;
+}
+
+:global body h1 {
   font-weight: 500;
   letter-spacing: -0.03em;
   line-height: 58px;
   font-size: 31px;
-  margin: 0px;
```
GJMcGowan: I think you forgot to put the `margin: 0` on the default styles.

docs/global.css
```diff
@@ -55,58 +55,49 @@
...
+:global h4,
+:global h5 {
   font-family: "Poppins";
+  letter-spacing: -0.03em;
+  color: #2f2d44;
+}
+
+:global body h1 {
   font-weight: 500;
   letter-spacing: -0.03em;
```
GJMcGowan: `letter-spacing` can probably just be the one in the default styles as well.

GJMcGowan: Same with `font-weight`

docs/global.css
lewish: Didn't actually ever need this

GJMcGowan: `h*` elements have a default margin which this overrides. Your docs may look weird with more margin. Your choice though 😄

PR #544, opened 2020-01-06
Stop regex r format space

Ekrekr:
Fixes https://github.com/dataform-co/dataform-web-tracking/issues/204

Text lift patterns in the formatter weren't doing as they should have been, replaced them.

Added API test for formatting a file to check that it works as it should do.

### DIFF
--- a/api/commands/format.ts
+++ b/api/commands/format.ts
@@ -13,7 +13,7 @@ const JS_BEAUTIFY_OPTIONS: JsBeautifyOptions = {
 
 const MAX_SQL_FORMAT_ATTEMPTS = 5;
 
-const TEXT_LIFT_PATTERNS = [/r'(\\['\\]|[^\n'\\])*'/, /r"(\\["\\]|[^\n"\\])*"/];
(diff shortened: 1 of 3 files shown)


api/commands/format.ts
```diff
@@ -13,7 +13,7 @@ const JS_BEAUTIFY_OPTIONS: JsBeautifyOptions = {
 
 const MAX_SQL_FORMAT_ATTEMPTS = 5;
 
-const TEXT_LIFT_PATTERNS = [/r'(\\['\\]|[^\n'\\])*'/, /r"(\\["\\]|[^\n"\\])*"/];
+const TEXT_LIFT_PATTERNS = [/r'.*'/, /r".*"/];
```
BenBirt: i am a bit worried that these are very greedy. e.g. what happens with `select r'foo', 'bar'`? i'm betting it lifts all of `r'foo', 'bar'` into a single string.

Ekrekr: Ah good spot. I updated regex to not be greedy, which then broke the parsing of escaped quotes, so I added a negative lookahead to fix that.

PR #550, opened 2020-01-09
Add metadata to Bigquery

faiyaz26:


api/dbadapters/index.ts
```diff
@@ -7,6 +7,11 @@ import { dataform } from "@dataform/protos";
 
 export type OnCancel = (handleCancel: () => void) => void;
 
+export interface IQueryResult {
+  rows: any[];
+  metadata?: any;
```
BenBirt: this definitely needs a type. for now i suspect we want to do something like

```
metadata: {
  bigquery?: {
    // useful fields here
  }
}
```

that is, unless we get abstract enough fields that we don't need to namespace them by `bigquery`

api/commands/run.ts
```diff
@@ -258,14 +258,15 @@ export class Runner {
     await this.triggerChange();
     try {
       // Retry this function a given number of times, configurable by user
-      const rows = await retry(
+      const queryResult = await retry(
```
```suggestion
      const { rows } = await retry(
```

api/dbadapters/index.ts
```diff
@@ -7,6 +7,11 @@ import { dataform } from "@dataform/protos";
 
 export type OnCancel = (handleCancel: () => void) => void;
 
+export interface IQueryResult {
```
BenBirt: `IExecutionResult`?

tests/integration/bigquery.spec.ts
```diff
@@ -134,4 +143,19 @@ describe("@dataform/integration/bigquery", () => {
       });
     }
   });
+
+  describe("bq metadata check", async () => {
```
BenBirt:
```suggestion
  describe("metadata", async () => {
```

tests/integration/bigquery.spec.ts
```diff
@@ -134,4 +143,19 @@ describe("@dataform/integration/bigquery", () => {
...
+    it("should get jobReference and statistics as metadata", async () => {
```
BenBirt:
```suggestion
    it("includes jobReference and statistics", async () => {
```

tests/api/api.spec.ts
```diff
@@ -838,7 +847,7 @@ describe("@dataform/api", () => {
             delete taskResult.timing;
           });
         });
-
+        console.log("result", result);
```
BenBirt: rm

tests/integration/bigquery.spec.ts
```diff
@@ -123,7 +130,9 @@ describe("@dataform/integration/bigquery", () => {
 
     for (const interactive of [true, false]) {
       it(`with interactive=${interactive}`, async () => {
-        expect(await dbadapter.execute(query, { interactive, maxResults: 2 })).eql([
+        const { rows } = await dbadapter.execute(query, { interactive, maxResults: 2 });
+        console.log("rows", rows);
```
BenBirt: rm

tests/integration/bigquery.spec.ts
```diff
@@ -134,4 +143,15 @@ describe("@dataform/integration/bigquery", () => {
       });
     }
   });
+
+  describe("metadata", async () => {
+    it("includes jobReference and statistics", async () => {
+      const query = `select 1 as test`;
+      const { metadata } = await dbadapter.execute(query, { interactive: false, maxResults: 2 });
+      const { bigquery: bqMetadata } = metadata;
+      expect(bqMetadata).to.have.property("jobId");
```
BenBirt: can we assert something a little stronger here? e.g. jobId matches some regexp, has >0 characters? and something about the bytes counts?

protos/core.proto
```diff
@@ -320,6 +320,16 @@ message ActionResult {
   Timing timing = 4;
 }
 
+
+message ExecutionMetadata {
+  message BigqueryMetadata {
+    string job_id = 1;
+    string total_bytes_processed = 2;
```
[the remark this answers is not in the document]
faiyaz26: they provide all the value as string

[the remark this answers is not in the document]
faiyaz26: I think we should keep it as string because BQ provides it as a string, its upto the client of dataform to convert this field to integer or whatever.

tests/integration/bigquery.spec.ts
faiyaz26: Add regex match for uuid

protos/core.proto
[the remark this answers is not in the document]
BenBirt: IMO we should expose a clean interface here; if that means doing the conversion ourselves, let's do it.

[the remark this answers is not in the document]
faiyaz26: converted it to number

tests/integration/bigquery.spec.ts
```diff
@@ -134,4 +143,20 @@ describe("@dataform/integration/bigquery", () => {
       });
     }
   });
+
+  describe("metadata", async () => {
+    it("includes jobReference and statistics", async () => {
+      const query = `select 1 as test`;
+      const { metadata } = await dbadapter.execute(query, { interactive: false, maxResults: 2 });
```
BenBirt: rm the options here?

PR #554, opened 2020-01-13
Testing library to remove dependency on mocha

lewish:
Mocha causes us a few issues, as it needs to be run through the CLI, we have complex bazel rules to deal with this. It has annoyingly globally scoped types and does a lot more than we need it to.

This PR introduces a simple testing library that can be used without a CLI, where each test is simply considered a binary with a exit set according to whether it passes or not. It also adds rules `ts_test` and `ts_test_suite` for making writing bazel rules easier.

The library itself is otherwise very similar to `mocha`, without the globals, CLI, and slightly more strict usage patterns.

Features:
- Parallel test suites
- Timeouts
- Sub suites
- Hooks: beforeEach, afterEach, setUp, tearDown
- Deep diffing on log output
- Log format that matches bazel

In the future, we should I think be able to use `bazel` for coverage metrics without any additional effort with this set up after this work is finished: https://github.com/bazelbuild/rules_nodejs/pull/1135

### DIFF
--- a/testing/BUILD
+++ b/testing/BUILD
@@ -17,7 +17,7 @@ ts_library(
 load("//testing:index.bzl", "ts_test_suite")
 
 ts_test_suite(
-    name = "test_suite",
+    name = "tests",
     srcs = glob(["*.spec.ts"]),
     deps = [
         ":testing",
--- a/testing/BUILD
+++ b/testing/BUILD
@@ -1,0 +1,26 @@
+package(default_visibility = ["//visibility:public"])
+
+load("@npm_bazel_typescript//:index.bzl", "ts_library")
+
+ts_library(
+    name = "testing",
+    srcs = glob(
+        ["*.ts"],
+        exclude = ["*.spec.ts"],
+    ),
+    deps = [
+        "@npm//@types/node",
+        "@npm//chalk",
+    ],
+)
+
+load("//testing:index.bzl", "ts_test_suite")
+
+ts_test_suite(
+    srcs = glob(["*.spec.ts"]),
+    deps = [
+        ":testing",
+        "@npm//@types/chai",
+        "@npm//chai",
+    ],
+)
--- a/testing/BUILD
+++ b/testing/BUILD
@@ -17,6 +17,7 @@ ts_library(
 load("//testing:index.bzl", "ts_test_suite")
 
 ts_test_suite(
+    name = "test_suite",
     srcs = glob(["*.spec.ts"]),
     deps = [
         ":testing",
--- a/testing/hook.ts
+++ b/testing/hook.ts
@@ -21,7 +21,7 @@ export class Hook {
     optionsOrFn: Omit<IHookOptions, "name"> | IHookFunction,
     fn?: IHookFunction
   ) {
-    let options: IHookOptions = { name: null };
+    let options: IHookOptions = typeof nameOrOptions === "string" ? { name: nameOrOptions } : nameOrOptions;
     if (typeof nameOrOptions === "string") {
       options.name = nameOrOptions;
     } else {
--- a/testing/hook.ts
+++ b/testing/hook.ts
@@ -14,7 +14,7 @@ export type IHookHandler = (
 ) => void;
 
 export class Hook {
-  public static readonly DEFAULT_TIMEOUT = 30000;
+  public static readonly DEFAULT_TIMEOUT_MILLIS = 30000;
 
   public static create(
     nameOrOptions: IHookOptions | string,
--- a/testing/hook.ts
+++ b/testing/hook.ts
@@ -1,0 +1,69 @@
+import { IRunContext, IRunResult } from "df/testing";
+
+export type IHookFunction = () => any;
+
+export interface IHookOptions {
+  name: string;
+  timeout?: number;
+}
+
+export type IHookHandler = (
+  nameOrOptions: IHookOptions | string,
+  optionsOrFn: Omit<IHookOptions, "name"> | IHookFunction,
+  fn?: IHookFunction
+) => void;
+
+export class Hook {
+  public static readonly DEFAULT_TIMEOUT = 30000;
+
+  public static create(
+    nameOrOptions: IHookOptions | string,
+    optionsOrFn: Omit<IHookOptions, "name"> | IHookFunction,
+    fn?: IHookFunction
+  ) {
+    let options: IHookOptions = { name: null };
+    if (typeof nameOrOptions === "string") {
+      options.name = nameOrOptions;
+    } else {
+      options = { ...nameOrOptions };
+    }
+    if (typeof optionsOrFn === "function") {
+      fn = optionsOrFn;
+    } else {
+      options = { ...options, ...optionsOrFn };
+    }
+    return new Hook(options, fn);
+  }
+
+  constructor(public readonly options: IHookOptions, private readonly fn: IHookFunction) {}
+
+  public async run(ctx: IRunContext) {
+    let timer: NodeJS.Timer;
+    const timeout = this.options.timeout || Hook.DEFAULT_TIMEOUT;
+    const result: Partial<IRunResult> = {
+      path: [...ctx.path, `${this.options.name} (hook)`]
+    };
+    try {
+      await Promise.race([
+        this.fn(),
+        new Promise((_, reject) => {
+          timer = setTimeout(() => {
+            result.outcome = "timeout";
+            reject(new Error(`Timed out (${timeout}ms).`));
+          }, timeout);
+        })
+      ]);
+      result.outcome = "passed";
+    } catch (e) {
+      if (result.outcome !== "timeout") {
+        result.outcome = "failed";
+      }
+      result.err = e;
+      ctx.results.push(result as IRunResult);
+      // If hooks fail, we throw anyway.
+      throw e;
+    }
+
+    clearTimeout(timer);
+  }
+}
--- a/testing/hook.ts
+++ b/testing/hook.ts
@@ -7,11 +7,17 @@ export interface IHookOptions {
   timeout?: number;
 }
 
-export type IHookHandler = (
+export function hook(name: string | IHookOptions, fn: IHookFunction): Hook;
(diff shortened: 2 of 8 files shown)


testing/runner.ts
```diff
@@ -0,0 +1,114 @@
...
+      Runner.resultPromise = Runner.run();
+    }
+  }
+
+  // tslint:disable: no-console
+  public static async run() {
+    // We tell the runner to start running at the end of current block of
+    // synchronously executed code. This will typically be after all the
+    // suite definitions are evaluated.
+    await promisify(process.nextTick)();
```
BenBirt: i have no idea why exactly 1 tick is correct

testing/runner.ts
```diff
@@ -0,0 +1,114 @@
...
+      ctx.results.push({
+        path: [],
+        err: new Error("No tests found in top level test suite."),
+        outcome: "failed"
+      });
+    }
+    const indent = (value: string, levels = 4) =>
+      value
+        .split("\n")
+        .map(line => `${new Array(levels).fill(" ").join("")}${line}`)
```
BenBirt:
```suggestion
        .map(line => `${" ".repeat(4).join("")}${line}`)
```

testing/suite.ts
```diff
@@ -0,0 +1,153 @@
...
+  name: string;
+  parallel?: boolean;
+}
+
+export interface ISuiteContext {
+  suite: typeof suite;
+  test: typeof test;
+  beforeEach: IHookHandler;
+  afterEach: IHookHandler;
+  setUp: IHookHandler;
```
BenBirt: i think we should be consistent with the naming of these hook handlers. I propose: `before`, `after`, `beforeAll`, `afterAll`

testing/suite.ts
```diff
@@ -0,0 +1,153 @@
...
+  tearDown: IHookHandler;
+}
+
+export function suite(
+  nameOrOptions: ISuiteOptions | string,
```
BenBirt: similar comment as above re: this is hard to read

testing/suite.ts
```diff
@@ -0,0 +1,153 @@
...
+  optionsOrFn: Omit<ISuiteOptions, "name"> | ((ctx?: ISuiteContext) => void),
+  fn?: (ctx?: ISuiteContext) => void
+): void {
+  Suite.globalStack.slice(-1)[0].addSuite(Suite.create(nameOrOptions, optionsOrFn, fn));
+  Runner.queueRunAndExit();
+}
+
+export class Suite {
+  public static readonly globalStack: Suite[] = [new Suite({ name: undefined }, () => null)];
```
BenBirt: IMO we shouldn't have a root suite, if possible

testing/suite.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  public static create(
+    nameOrOptions: ISuiteOptions | string,
+    optionsOrFn: Omit<ISuiteOptions, "name"> | ((ctx?: ISuiteContext) => void),
+    fn?: (ctx?: ISuiteContext) => void
+  ) {
+    let options: ISuiteOptions = { name: null };
```
BenBirt: similar comment as above re: use ternary operator

testing/test.ts
```diff
@@ -0,0 +1,69 @@
...
+  nameOrOptions: ITestOptions | string,
+  optionsOrFn: Omit<ITestOptions, "name"> | (() => void),
+  fn?: () => void
+): void {
+  Suite.globalStack.slice(-1)[0].addTest(Test.create(nameOrOptions, optionsOrFn, fn));
+  Runner.queueRunAndExit();
+}
+
+export class Test {
+  public static readonly DEFAULT_TIMEOUT = 30000;
```
BenBirt:
```suggestion
  public static readonly DEFAULT_TIMEOUT_MILLIS = 30000;
```

testing/test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  public static create(
+    nameOrOptions: ITestOptions | string,
+    optionsOrFn: Omit<ITestOptions, "name"> | (() => void),
+    fn?: () => void
+  ) {
+    let options: ITestOptions = { name: null };
```
BenBirt: similar comment ternary operator

testing/test.ts
```diff
@@ -0,0 +1,69 @@
...
+    } catch (e) {
+      if (result.outcome !== "timeout") {
+        result.outcome = "failed";
+      }
+      result.err = e;
+    }
+
+    clearTimeout(timer);
+
+    ctx.results.push(result as IRunResult);
```
BenBirt: similar comment let's not break type-safety

testing/suite.ts
lewish: `before, after, beforeEach, afterEach` as it's the same as mocha. `beforeAll` is confusing IMO, as it can be kind of interpreted either way

testing/runner.ts
lewish: This is equivelant to setTimeout(..., 0).
It basically means - finish processing whats currently on the node event queue and then run the tests.

testing/index.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+    ctx.tearDown("change counter", () => {
+      this.counter = 2;
+    });
+  }
+}
+
+const _ = (async () => {
+  let exampleFixture: ExampleFixture;
+  suite("suite", () => {
+    test("passes", async () => true);
```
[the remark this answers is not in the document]
lewish: I've done it, but I'm a bit on the fence about it. It's a bit confusing / annoying to actually write, maybe we'll get used to it quickly...

testing/hook.ts
```diff
@@ -0,0 +1,69 @@
...
+      if (result.outcome !== "timeout") {
+        result.outcome = "failed";
+      }
+      result.err = e;
+      ctx.results.push(result as IRunResult);
+      // If hooks fail, we throw anyway.
+      throw e;
+    }
+
+    clearTimeout(timer);
```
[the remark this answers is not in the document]
lewish: Wooops!!

testing/index.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+    });
+
+    suite("with set up and tear down", ctx => {
+      exampleFixture = new ExampleFixture(ctx);
+      test({ name: "set up is called" }, () => {
+        expect(exampleFixture.counter).equals(1);
+      });
+    });
+
+    suite("can execute in parallel", { parallel: true }, async () => {
```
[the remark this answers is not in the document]
lewish: What do you mean by spin look?
Both tests start on the same-ish cycle, and counter is 0, then they both wait 10ms, and both reset counter, so they should both run their expectation before either of them resets the counter.

lewish:
Types for test/suite etc are clearer to read for consumers now:
image

Bazel timeout - not clear what we can do here? Bazel hard times out tests after something like 5 mins.

R.e. output, it's bazel-esque, unfortunately I can't get the colors working properly, but it used the same colors for PASSED, FAILED etc:

```bash
suite > fails on promise rejection                                       FAILED

    Error: fail-async
        at testing/index.spec.ts:28:30
        at Generator.next (<anonymous>)
        at /home/lewis/.cache/bazel/_bazel_lewis/34cd37228b66f7a6f769c3c562621852/sandbox/linux-sandbox/19/execroot/df/bazel-out/k8-py2-fastbuild/bin/testing/index.spec.sh.runfiles/df/testing/index.spec.js:6:71
        at new Promise (<anonymous>)
        at __awaiter (/home/lewis/.cache/bazel/_bazel_lewis/34cd37228b66f7a6f769c3c562621852/sandbox/linux-sandbox/19/execroot/df/bazel-out/k8-py2-fastbuild/bin/testing/index.spec.sh.runfiles/df/testing/index.spec.js:2:12)
        at Test.fn (testing/index.spec.ts:27:53)
        at Test.<anonymous> (testing/test.ts:51:14)
        at Generator.next (<anonymous>)
        at /home/lewis/.cache/bazel/_bazel_lewis/34cd37228b66f7a6f769c3c562621852/sandbox/linux-sandbox/19/execroot/df/bazel-out/k8-py2-fastbuild/bin/testing/index.spec.sh.runfiles/df/testing/test.js:6:71
        at new Promise (<anonymous>)

suite > times out                                                       TIMEOUT

    Error: Timed out (10ms).
        at Timeout.<anonymous> (testing/test.ts:55:20)
        at listOnTimeout (internal/timers.js:531:17)
        at processTimers (internal/timers.js:475:7)


Tests failed.
================================================================================
INFO: Elapsed time: 1.064s, Critical Path: 0.80s
INFO: 4 processes: 2 linux-sandbox, 2 worker.
INFO: Build completed successfully, 5 total actions
//testing:index.spec                                                     PASSED in 0.2s
```

testing/index.spec.ts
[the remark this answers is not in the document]
BenBirt: a spinlock is a while loop that just waits for a condition to be true, sleeping every so often.

personally i would like to not rely on nodejs promise queueing, feels very un-obvious to the reader IMO.

testing/runner.ts
BenBirt: sorry, yes, i know what it means, but why are we doing it on the next tick? especially since the comment says "typically", it implies it might be the case but not always (?)

BenBirt:
timeouts - i only took a quick scan, but IMO your runner should be obeying the various parameters in https://docs.bazel.build/versions/master/test-encyclopedia.html, including timeouts.

testing/BUILD
```diff
@@ -17,6 +17,7 @@ ts_library(
 load("//testing:index.bzl", "ts_test_suite")
 
 ts_test_suite(
+    name = "test_suite",
```
BenBirt: `tests`?

lewish:
> timeouts - i only took a quick scan, but IMO your runner should be obeying the various parameters in https://docs.bazel.build/versions/master/test-encyclopedia.html, including timeouts.

As discussed, reading that page I don't think the test code is _supposed_ to do anything with these flags. Bazel will kill processes and tests if they time out itself.

lewish:
As discussed, re-allowing calling of top level suite/test methods inside tests, as this was IMO more annoying than it was worth after trying to migrate tests.

PR #558, opened 2020-01-14
Ask me when incremental

Ekrekr:
Fixes https://github.com/dataform-co/dataform-co/issues/2967

1. [Design doc for conditional globals](https://docs.google.com/document/d/1DWw9DKm6X0m6OLAtZrLbzmcMk-5yliQ_NXFp0VLKYt4)

1. Implementation of `when()`, renamed `isIncremental()` to `incremental()`, removed `ifIncremental()`. Added test to example compilation files.

1. Added documentation on new features, with examples. Removed hidden status so that they appear in the API reference.

Note that I changed the private variable `incremental` to `isIncremental` to prevent a clash with the new function `incremental()`.


protos/core.proto
```diff
@@ -152,6 +152,8 @@ message Table {
   // Pre/post operations.
   repeated string pre_ops = 13;
   repeated string post_ops = 14;
+  repeated string inc_pre_ops = 28;
```
BenBirt:
```suggestion
  repeated string incremental_pre_ops = 28;
```

protos/core.proto
```diff
@@ -152,6 +152,8 @@ message Table {
...
+  repeated string inc_post_ops = 29;
```
BenBirt:
```suggestion
  repeated string incremental_post_ops = 29;
```

core/test.ts
```diff
@@ -168,13 +168,12 @@ class RefReplacingContext implements ITableContext {
     return "";
   }
 
-  public isIncremental() {
-    return false;
+  public when(cond: boolean, trueCase: string, falseCase?: string) {
+    return "";
```
BenBirt: this should have the same behaviour as `table.ts`

core/table.ts
```diff
@@ -426,12 +433,12 @@ export class TableContext implements ITableContext {
     return "";
   }
 
-  public isIncremental() {
-    return !!this.incremental;
+  public when(cond: boolean, trueCase: string, falseCase?: string) {
+    return cond ? trueCase : falseCase ? falseCase : "";
```
BenBirt:
```suggestion
    return cond ? trueCase : falseCase || "";
```

core/table.ts
```diff
@@ -355,29 +356,35 @@ export class Table {
     const incrementalContext = new TableContext(this, true);
 
     this.proto.query = context.apply(this.contextableQuery);
+
+    const createOps = (
```
BenBirt:
```suggestion
    const contextifyOps = (
```

BenBirt: nit: can this instead be a private method?

core/table.ts
```diff
@@ -355,29 +356,35 @@ export class Table {
...
+      contextableOps: Array<Contextable<ITableContext, string | string[]>>,
+      currentContext: TableContext
+    ) => {
+      let protoOps = Array<string>();
```
BenBirt:
```suggestion
      let protoOps: []string = [];
```

core/table.ts
```diff
@@ -355,29 +356,35 @@ export class Table {
...
+      contextableOps.forEach(contextableOp => {
+        const appliedOps = currentContext.apply(contextableOp);
+        protoOps = (protoOps || []).concat(
```
BenBirt:
```suggestion
        protoOps.concat(
```

core/table.ts
```diff
@@ -160,14 +160,15 @@ export interface ITableConfig extends ITargetableConfig, IDocumentableConfig, ID
  */
 export interface ITableContext extends ICommonContext {
   /**
-   * @hidden
+   * Shorthand if condition that uses an empty string in the case of the condition not being satisfied.
+   * `falseCase` is not required.
```
BenBirt:
```suggestion
   * `falseCase` is optional, and defaults to an empty string.
```

core/table.ts
```diff
@@ -160,14 +160,15 @@ export interface ITableConfig extends ITargetableConfig, IDocumentableConfig, ID
...
    */
-  isIncremental: () => boolean;
+  when: (cond: boolean, trueCase: string, falseCase?: string) => string;
```
BenBirt:
```suggestion
  when: (cond: boolean, trueCase: string, falseCase: string = "") => string;
```

core/table.ts
```diff
@@ -160,14 +160,15 @@ export interface ITableConfig extends ITargetableConfig, IDocumentableConfig, ID
  */
 export interface ITableContext extends ICommonContext {
   /**
-   * @hidden
+   * Shorthand if condition that uses an empty string in the case of the condition not being satisfied.
```
BenBirt:
```suggestion
   * Shorthand for an `if` condition. Equivalent to `cond ? trueCase : falseCase`.
```

api/commands/build.ts
```diff
@@ -82,9 +82,13 @@ export class Builder {
     const tasks = t.disabled
       ? emptyTasks
       : emptyTasks.concat(
-          (t.preOps || []).map(pre => ({ statement: pre })),
+          t.incrementalQuery
```
BenBirt: - we shouldn't check whether the table is incremental or not using the existence of this field; old versions of `@dataform/core` don't know how to set it. we should instead use `t.type === "incremental"`
- i'm not sure this is quite right... what about old versions of `@dataform/core`? we need to be backwards compatible with those; in which case `incPreOps` and `incPostOps` won't be set, and we should continue to use `preOps` and `postOps`

content/docs/how-to-guides/incremental-datasets.md
```diff
@@ -179,3 +179,20 @@ SELECT ...
...
+
+For functionality other than a `WHERE` clause, `when()` combined with `incremental()` can be used. For example if there was a separate field `weblogs.user_alternative_actions` containing an alternative set of user events, and it was desirable for the script to use this only when incremental, then a neat way to manage this would be to use:
+
+```js
+config { type: "incremental" }
+
+SELECT timestamp, action
+FROM ${ when(incremental(), weblogs.user_alternative_actions, weblogs.user_actions) }
+
+incremental_where {
```
BenBirt: IMO we should get rid of all mention of `incremental_where` from the docs.

content/docs/how-to-guides/incremental-datasets.md
```diff
@@ -179,3 +179,20 @@ SELECT ...
 Since incremental datasets are usually timestamp based, it's a best practice to set up dataset partioning on the timestamp field to speed up downstream queries.
 
 For more information, check out the [BigQuery](warehouses/bigquery) and [Redshift](warehouses/redshift) guides.
+
+## Conditional code if incremental
+
+For functionality other than a `WHERE` clause, `when()` combined with `incremental()` can be used. For example if there was a separate field `weblogs.user_alternative_actions` containing an alternative set of user events, and it was desirable for the script to use this only when incremental, then a neat way to manage this would be to use:
```
BenBirt:
```suggestion
For functionality other than a `WHERE` clause, `when()` combined with `incremental()` can be used. For example, assuming there is a field `weblogs.user_alternative_actions` containing an alternative set of user events, and it is desirable for the script to use this only when running incrementally, you could write:
```

content/docs/how-to-guides/sqlx.md
```diff
@@ -52,32 +52,20 @@ In-line Javascript can be used anywhere SQL is written in order to dynamically m
...
 ### Built-in functions
 
-Built in functions have special functionality should be executed within [in-line Javascript](#in-line-javascript).
+Built in functions have special functionality and can be executed either within [in-line Javascript](#in-line-javascript) or [javascript blocks](#javascript-blocks).
 
-#### `ref()`
+For all built in functions, see [`ITableContext` in the API reference](/reference#ITableContext). Some useful examples can be found here:
 
-`ref()` enables you to easily reference another dataset in your project without having to provide the full SQL dataset name. `ref()` also adds the referenced dataset to the set of dependencies for the query.
+- [Reference another dataset in your project with `ref()`](datasets/#referencing-other-datasets).
```
BenBirt: IMO the earlier sub-headings called out these examples a bit more nicely, fwiw these functions, especially `ref()` are going to be 99% of all context method invocations, i think it would be a good idea to continue calling them out loudly

content/docs/how-to-guides/sqlx.md
```diff
@@ -52,32 +52,20 @@ In-line Javascript can be used anywhere SQL is written in order to dynamically m
...
+- [Execute code only if the script is for an incremental dataset using `incremental()`](incremental-datasets/#conditional-code-if-incremental).
 
 ## Additional Features
 
-- [Pre-operations and post-operations](#pre-operations-and-post-operations): Pre and post operations are only valid for some table types.
-
-### Pre-operations and post-operations
-
-Pre-operation and post-operation blocks are defined in SQLX by writing `pre_operations { }` and `post_operations { }` respectively.
+- **Post-operations**: defined in SQLX by writing `post_operations { }`, SQL written inside will be executed after the central SQL. This can be useful for granting permissions, as can be seen in the [publishing datasets guide](/how-to-guides/datasets/#example-granting-dataset-access-with-post_operations). **Actions may only include pre_operations if they create a dataset**, for example with `type: "table"` or `type: "view"` or `type: "incremental"` in their config.
```
BenBirt: this is an overview, IMO we should leave the details of where it's ok to use `pre_operations` etc in the specific guide (`/datasets`)

content/docs/how-to-guides/sqlx.md
```diff
@@ -52,32 +52,20 @@ In-line Javascript can be used anywhere SQL is written in order to dynamically m
...
 
-SQL written in pre-operation blocks is executed before the central chunk of SQL, while post-operation blocks are executed afterwards.
+- **Pre-operations**: the same as post-operations, but defined with `pre_operations { }`, and takes place before the central SQL.
```
BenBirt: seems logical to mention pre-operations before post-operations

content/docs/how-to-guides/sqlx.md
```diff
@@ -52,32 +52,20 @@ In-line Javascript can be used anywhere SQL is written in order to dynamically m
...
 
-They are useful for purposes such as setting permissions. An example can be found in the [publishing datasets guide](datasets).
+- **Incremental where**: defined in SQLX by writing `incremental_where { }`. Only valid for incremental tables (those with `type: "incremental"` in the config). This wraps the main block of SQL in a `WHERE` clause, with the argument of whatever's inside the block. For examples see the [incremental datasets how to guide](http://localhost:3001/how-to-guides/incremental-datasets).
```
BenBirt: ditto re mentioning `incremental_where`, IMO let's get rid of it

protos/core.proto
Ekrekr: Good point

core/test.ts
Ekrekr: Ah ok

core/table.ts
Ekrekr: Sure, does indeed look nicer.

core/table.ts
Ekrekr: 👍

core/table.ts
Ekrekr: Oof, did not catch that!

api/commands/build.ts
Ekrekr: Ah nice catch. I've added a check for `t.incrementalPreOps` to the ternary, I believe this will work but I'm not sure. Does the `@dataform/api` ever get recompiled whilst depending an older version of `@dataform/core`? The build file just says `//core` so I assume not.

content/docs/how-to-guides/incremental-datasets.md
Ekrekr: Sounds good to me

content/docs/how-to-guides/sqlx.md
Ekrekr: Hm I disagree, I think the information here is more suited in this guide than in `/datasets`, as when publishing a dataset, a dataset will always be created; users searching about how to use post_operations in SQLX when trying to understand why they they get an error while defining it with `inline` would end up here first.

isn't in `/datasets`

Ekrekr:
Some potentially important replies are hidden, just a heads up

content/docs/how-to-guides/incremental-datasets.md
```diff
@@ -82,26 +82,22 @@ config { type: "incremental" }
...
-incremental_where {
-  timestamp > (SELECT MAX(timestamp) FROM ${self()})
-}
+${ when(incremental(), `WHERE timestamp > (SELECT MAX(timestamp) FROM ${self()})` }
 ```
 
 First the script sets the type of the dataset to `incremental`.
 
-It then specifies a `WHERE` clause using an `incremental_where` block:
+It then specifies a `WHERE` clause using a `when(incremental(), ...)` in-line:
```
BenBirt: this is odd english; "in-line" is not a noun, there can't be "a" "in-line". 

```
It then specifies a `WHERE` clause using the `when()` and `incremental()` functions:
```

content/docs/how-to-guides/sqlx.md
```diff
@@ -56,16 +56,28 @@ Built in functions have special functionality and can be executed either within
...
+
+Some examples can be found [here](datasets/#referencing-other-datasets).
+
+#### `resolve()`
+
+`resolve()` works similarly to `ref()`, but doesn't add the dataset to the dependency list for the query.
+
+#### `self()`
+
+`self()` returns the name of the current dataset. If the default schema or dataset name is overridden in the `config{}` block, `self()` will return the full and correct dataset name.
```
BenBirt: `If the database, schema, or dataset name is overridden ...`

content/docs/how-to-guides/sqlx.md
```diff
@@ -56,16 +56,28 @@ Built in functions have special functionality and can be executed either within
...
+
+An example of `self()` being used to set up incremental tables is [here](incremental-datasets/#a-simple-example).
```
BenBirt: ```[Here](...) is an example of an incremental table using the `self()` function.```

content/docs/how-to-guides/sqlx.md
```diff
@@ -56,16 +56,28 @@ Built in functions have special functionality and can be executed either within
...
 
 ## Additional Features
 
-- **Post-operations**: defined in SQLX by writing `post_operations { }`, SQL written inside will be executed after the central SQL. This can be useful for granting permissions, as can be seen in the [publishing datasets guide](/how-to-guides/datasets/#example-granting-dataset-access-with-post_operations). **Actions may only include pre_operations if they create a dataset**, for example with `type: "table"` or `type: "view"` or `type: "incremental"` in their config.
-
-- **Pre-operations**: the same as post-operations, but defined with `pre_operations { }`, and takes place before the central SQL.
+- **Pre-operations**: defined in SQLX by writing `pre_operations { }`, SQL written inside will be executed before the central SQL. This can be useful for granting permissions, as can be seen in the [publishing datasets guide](/how-to-guides/datasets/#example-granting-dataset-access-with-post_operations). **Actions may only include pre_operations if they create a dataset**, for example with `type: "table"` or `type: "view"` or `type: "incremental"` in their config.
```
BenBirt: IMO we shouldn't introduce a new term like "central SQL". how about "main SQL"?

content/docs/how-to-guides/sqlx.md
```diff
@@ -56,16 +56,28 @@ Built in functions have special functionality and can be executed either within
...
 
-- **Incremental where**: defined in SQLX by writing `incremental_where { }`. Only valid for incremental tables (those with `type: "incremental"` in the config). This wraps the main block of SQL in a `WHERE` clause, with the argument of whatever's inside the block. For examples see the [incremental datasets how to guide](http://localhost:3001/how-to-guides/incremental-datasets).
+- **Post-operations**: the same as pre-operations, but defined with `post_operations { }`, and takes place after the central SQL.
```
BenBirt: ```... runs after the main SQL.```

api/commands/build.ts
```diff
@@ -82,12 +82,12 @@ export class Builder {
     const tasks = t.disabled
       ? emptyTasks
       : emptyTasks.concat(
-          t.incrementalQuery
-            ? (t.incPreOps || []).map(pre => ({ statement: pre }))
+          t.type === "incremental" && t.incrementalPreOps
```
BenBirt: i think this is still incorrect. we should only use `incrementalPreOps` in the same circumstances where we'll run the incremental query.

see https://source.cloud.google.com/tada-analytics/github_dataform-co_dataform/+/master:core/adapters/bigquery.ts;l=29

this logic determines whether to run the plain `query`, or to run the `incrementalQuery` (and it also has fallback logic to the old-incremental style in that case). i would suggest pushing preops/postops down into `publishTasks`, so that the logic can be shared.

api/commands/build.ts
BenBirt: `@dataform/api` (effectively) doesn't depend on `@dataform/core`; `core` is responsible for creating `CompiledGraph` instances, and `api` acts upon those; `CompiledGraph` is (intended to be) the only interface between the two. thus, `api` must be backwards-compatible with old versions of `core`.

core/table.ts
```diff
@@ -355,38 +356,45 @@ export class Table {
...
-      const appliedPosts = context.apply(contextablePostOps);
-      this.proto.postOps = (this.proto.postOps || []).concat(
-        typeof appliedPosts === "string" ? [appliedPosts] : appliedPosts
-      );
-    });
-    this.contextablePostOps = [];
     return this.proto;
   }
+
+  private contextifyOps = (
```
BenBirt:
```suggestion
  private contextifyOps(...) {
```

protos/core.proto
```diff
@@ -152,6 +152,8 @@ message Table {
   // Pre/post operations.
```
BenBirt: github won't let me comment on the relevant line, but please change `string where = 8;` to `string where = 8 [deprecated = true];`

content/docs/how-to-guides/sqlx.md
Ekrekr: Main SQL is a new term too, but sure

protos/core.proto
Ekrekr: Ah very cool

api/commands/build.ts
Ekrekr: Have migrated them to individual adapters

api/commands/build.ts
```diff
@@ -81,15 +81,7 @@ export class Builder {
 
     const tasks = t.disabled
       ? emptyTasks
```
BenBirt: cleanup: let's just return `? [] as dataform.IExecutionTask[]` and remove `emptyTasks`

core/adapters/bigquery.ts
```diff
@@ -19,6 +19,10 @@ export class BigQueryAdapter extends Adapter implements IAdapter {
     tableMetadata: dataform.ITableMetadata
   ): Tasks {
     const tasks = Tasks.create();
+
+    (table.incrementalPreOps || []).forEach(pre => tasks.add(Task.statement(pre)));
```
BenBirt: this code isn't correct, we're adding `incrementalPreOps` even if the table doesn't already exist. it should be something more like:

```
if (table.type !== "incremental" || dataformCoreVersion <= "current_version") {
  (table.preOps || []).forEach(pre => tasks.add(Task.statement(pre)));
}
```

then, inside the `if (table.type === "incremental")` branch below, take special action.
- if `dataformCoreVersion > "current_version"`, and inside the "full refresh" branch, we should add `preOps`.
- if `dataformCoreVersion > "current_version"`, and inside the "non-full refresh" branch, we should add `incrementalPreOps`

(a similar comment applies to post-ops, and for the other adapters too)

core/adapters/bigquery.ts
```diff
@@ -61,16 +69,6 @@ export class BigQueryAdapter extends Adapter implements IAdapter {
...
-  public createEmptyIfNotExists(table: dataform.ITable) {
-    return `create ${this.baseTableType(table.type)} if not exists ${this.resolveTarget(
-      table.target
-    )} ${
-      table.bigquery && table.bigquery.partitionBy
-        ? `partition by ${table.bigquery.partitionBy}`
-        : ""
-    } as ${this.where(table.query, "false")}`;
-  }
-
```
Ekrekr: Not used anywhere; removed

tests/integration/bigquery.spec.ts
```diff
@@ -87,6 +88,23 @@ describe("@dataform/integration/bigquery", () => {
...
+    expect(actionMap["dataform-integration-tests.df_integration_test.example_view"].status).equals(
+      dataform.ActionResult.ExecutionStatus.SUCCESSFUL
+    );
+    expect(actionMap["dataform-integration-tests.df_integration_test.sample_data_2"].status).equals(
+      dataform.ActionResult.ExecutionStatus.SUCCESSFUL
+    );
+    expect(actionMap["dataform-integration-tests.df_integration_test.sample_data"].status).equals(
+      dataform.ActionResult.ExecutionStatus.SUCCESSFUL
+    );
+
```
Ekrekr: As we discussed, this should have been being tested but weren't, so have added for each adapter

tests/integration/bigquery.spec.ts
```diff
@@ -144,6 +163,70 @@ describe("@dataform/integration/bigquery", () => {
...
+      const buildsFromIncrement = bqadapter
+        .publishTasks(table, { fullRefresh: false }, { fields: [] })
+        .build();
+
+      buildsFromIncrement.forEach((build, i) => {
+        expect(build.statement).to.eql(expectedIncrementStatements[i]);
+      });
+    });
+  });
+
```
Ekrekr: I think this is as short as I can get it, and doesn't read too bad IMO. Refresh and increment statements do differ between different adapters.

core/adapters/bigquery.ts
Ekrekr: Large updates, this has been applied

core/table.ts
Ekrekr: can't instantiate in an interface

core/adapters/base.ts
```diff
@@ -22,18 +23,50 @@ export abstract class Adapter {
...
 
   protected where(query: string, where: string) {
-    return `select * from (${query}) as subquery
-        where ${where || "true"}`;
+    return `
+  select * from (${query}) as subquery
+    where ${where || "true"}`;
+  }
+
+  protected refreshIncrementCondition(
```
BenBirt: i would suggest inverting this to `!runConfig.fullRefresh && tableMetadata && tableMetadata.type !== "view"`, and call it `shouldWriteIncrementally`

core/adapters/base.ts
```diff
@@ -22,18 +23,50 @@ export abstract class Adapter {
...
+    runConfig: dataform.IRunConfig,
+    tableMetadata: dataform.ITableMetadata
```
BenBirt:
```suggestion
    tableMetadata?: dataform.ITableMetadata
```

core/adapters/base.ts
```diff
@@ -22,18 +23,50 @@ export abstract class Adapter {
...
+  ) {
+    return runConfig.fullRefresh || !tableMetadata || tableMetadata.type === "view";
+  }
+
+  protected addPreOps(
```
BenBirt: IMO it's weird that this causes state changes. i think it should return the list of preops which should be added, as a list of `Task`s, and should be called `preOps`

core/adapters/base.ts
```diff
@@ -22,18 +23,50 @@ export abstract class Adapter {
...
+    table: dataform.ITable,
+    dataformCoreVersion: string,
+    tasks: Tasks,
+    refreshIncrementCondition: boolean
```
BenBirt: I don't think this should be passed in; instead, `addPreOps` should accept `tableMetadata?: dataform.ITableMetadata` and call `refreshIncrementCondition` itself.

(same with `addPostOps`)

core/adapters/bigquery.ts
```diff
@@ -19,29 +19,38 @@ export class BigQueryAdapter extends Adapter implements IAdapter {
...
+      if (refreshIncrementCondition) {
         tasks.add(Task.statement(this.createOrReplace(table)));
       } else {
         tasks.add(
           Task.statement(
             this.insertInto(
               table.target,
               tableMetadata.fields.map(f => f.name),
-              this.where(table.incrementalQuery || table.query, table.where)
+              this.dataformCoreVersion > "1.4.8"
```
BenBirt: this shouldn't change, it breaks users

tests/integration/bigquery.spec.ts
```diff
@@ -145,6 +150,70 @@ suite("@dataform/integration/bigquery", ({ after }) => {
...
+      buildsFromRefresh.forEach((build, i) => {
+        expect(build.statement).to.eql(expectedRefreshStatements[i]);
+      });
+
+      const buildsFromIncrement = bqadapter
+        .publishTasks(table, { fullRefresh: false }, { fields: [] })
+        .build();
+
+      buildsFromIncrement.forEach((build, i) => {
+        expect(build.statement).to.eql(expectedIncrementStatements[i]);
```
BenBirt: this is testing *a lot* more than just pre/post ops. can we thin down this test please?

core/adapters/base.ts
Ekrekr: Sure

core/adapters/base.ts
Ekrekr: Also added `||` check to incrementalPreOps as proto allows a null value

core/adapters/base.ts
Ekrekr: I had it separate because I was expecting adapters to have different `refreshIncrementCondition` in the future, but no need to separate it preemptively I guess

core/adapters/bigquery.ts
Ekrekr: Added change where if the `table.where` protobuf is set then it will follow that, otherwise it will follow the new `${ when(incremental(), ... ) }` pattern. I think users will actually be able to use both, but I won't write a test for this unless you say as `when_incremental` is being deprecated anyways.

tests/integration/bigquery.spec.ts
Ekrekr: Have narrowed the scope massively

core/adapters/base.ts
```diff
@@ -22,18 +23,56 @@ export abstract class Adapter {
...
+  }
+
+  protected shouldWriteIncrementally(
+    runConfig: dataform.IRunConfig,
+    tableMetadata?: dataform.ITableMetadata
+  ) {
+    return !runConfig.fullRefresh && tableMetadata && tableMetadata.type !== "view";
+  }
+
+  protected addPreOps(
```
BenBirt: naming is still a bit odd, `preOps` and `postOps` please?

core/adapters/base.ts
```diff
@@ -22,18 +23,56 @@ export abstract class Adapter {
...
+    table: dataform.ITable,
+    dataformCoreVersion: string,
```
BenBirt: you shouldn't need to pass this in. all subclasses of this receive this value in their constructor; so add a constructor to this class and store it in a member variable.

core/adapters/base.ts
```diff
@@ -22,18 +23,56 @@ export abstract class Adapter {
...
+    runConfig: dataform.IRunConfig,
+    tableMetadata: dataform.ITableMetadata
+  ): Task[] {
+    return (dataformCoreVersion > "1.4.8" &&
```
BenBirt: IMO we could make this a tad more readable:

```
let preOps = table.preOps;
if (<conditions>) {
  preOps = table.incrementalPreOps;
}
return (preOps || []).map(pre => Task.statement(pre));
```

core/adapters/bigquery.ts
```diff
@@ -19,14 +19,19 @@ export class BigQueryAdapter extends Adapter implements IAdapter {
     tableMetadata: dataform.ITableMetadata
   ): Tasks {
     const tasks = Tasks.create();
-    // Drop views/tables first if they exist.
-    if (tableMetadata && tableMetadata.type != this.baseTableType(table.type)) {
+    const shouldWriteIncrementally = this.shouldWriteIncrementally(runConfig, tableMetadata);
```
BenBirt: we only use this in one place, let's inline it?

core/adapters/bigquery.ts
```diff
@@ -19,14 +19,19 @@ export class BigQueryAdapter extends Adapter implements IAdapter {
...
+
+    const preOps = this.addPreOps(table, this.dataformCoreVersion, runConfig, tableMetadata);
```
BenBirt:
```suggestion
    this.addPreOps(table, this.dataformCoreVersion, runConfig, tableMetadata).forEach(statement => tasks.add(statement));
```

(same for postops)

tests/integration/bigquery.spec.ts
```diff
@@ -145,6 +148,43 @@ suite("@dataform/integration/bigquery", ({ after }) => {
     }
   });
 
+  suite("publish tasks", async () => {
+    test("incremental pre and post ops, core version <= 1.4.8", async () => {
+      const projectConfig: dataform.IProjectConfig = {
```
BenBirt: inline this?

core/adapters/base.ts
Ekrekr: 👍

core/adapters/base.ts
Ekrekr: 👍

core/adapters/base.ts
Ekrekr: 👍

core/adapters/bigquery.ts
Ekrekr: 👍

core/adapters/bigquery.ts
Ekrekr: 👍

tests/integration/bigquery.spec.ts
Ekrekr: 👍

PR #567, opened 2020-01-20
Add hooks to Runner to allow the user to set a configurable per-run timeout.

BenBirt:


api/commands/run.ts
```diff
@@ -74,7 +82,11 @@ export class Runner {
   }
 
   public async resultPromise(): Promise<dataform.IRunResult> {
-    return this.executionTask;
+    try {
+      return await this.executionTask;
+    } finally {
+      clearTimeout(this.timeout);
```
lewish: Should you check this.timeout exists first?

PR #570, opened 2020-01-21
Add packages section of docs site

lewish:
Including the new Segment package!

Also cleans up a bunch of the github CMS stuff, adds caching, simplifies interfaces and supports an oauth token set via the process environment (when we deploy to k8s).

image


lewish:
> Packages is a nice addition and code looks much neater. Just a few things that should be fixed:
> 
> * Suggest edits button is broken again, I think just removing `.md` from `href={`[https://github.com/dataform-co/dataform/blob/master/content/docs/${this.props.current.path}.md`}`](https://github.com/dataform-co/dataform/blob/master/content/docs/$%7Bthis.props.current.path%7D.md%60%7D%60) will fix it.
> * Parent page for `packages` is blank, same as with `how-to-guides`. Is now a good time to add in the table of contents generator?

Fixed edits button. Agree on the content, although would argue out of scope for this PR, but I will follow up with something!

PR #574, opened 2020-01-23
Compiled query highlighting

GJMcGowan:
Open source part of #3430

Extracting line number of errors in compiled query.


api/utils/error_parsing.ts
```diff
@@ -0,0 +1,100 @@
...
+interface IAzureEvalError {
+  originalError?: {
+    info?: {
+      message?: string;
+    };
+  };
+  errorLocation?: IErrorLocation;
+}
+
+export const AzureEvalErrorParser = (error: IAzureEvalError) => {
```
BenBirt:
```suggestion
export function parseAzureEvaluationError(error: IAzureEvaluationError) {
```

api/utils/error_parsing.ts
```diff
@@ -0,0 +1,100 @@
+interface IErrorLocation {
+  line?: number;
+  column?: number;
+}
+
+interface IAzureEvalError {
```
BenBirt:
```suggestion
interface IAzureEvaluationError {
```

(similar below)

api/utils/error_parsing.ts
```diff
@@ -0,0 +1,100 @@
...
+  originalError?: {
+    info?: {
+      message?: string;
+    };
+  };
+  errorLocation?: IErrorLocation;
```
BenBirt: this seems a bit weird to me. we mutate the original `Error`, instead of just returning a new `EvaluationError` instance.

api/utils/error_parsing.ts
```diff
@@ -0,0 +1,100 @@
...
+export const RedshiftEvalErrorParser = (statement: string, error: IRedshiftEvalError) => {
+  // expected error format:
+  // // e.position = "123" - position is the number of characters into the query that the error was found at
+  // // including \n characters
+
+  if (!error.position) {
+    return error;
+  }
+  // split statement into lines
+  const statementSplitByLine = statement.split("\n");
```
BenBirt: i think this logic is a little confusing, and i worry that it's calculating the wrong number (by removing these characters). how about something like:

```
const statementBeforeError = statement.substring(0, error.position);
const lineNumber = statementBeforeError.split("\n").length;
const lineIncludingError = statementBeforeError.split("\n").slice("-1")[0];
const colNumber = lineIncludingError.length;
```

api/utils/error_parsing.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export const BigqueryEvalErrorParser = (error: IBigqueryEvalError) => {
+  // expected error format:
+  // // e.message = Syntax error: Unexpected identifier "asda" at [2:1]
+  if (!error.message) {
+    return error;
+  }
+
+  // extract all characters between '[' and ']' (inclusive)
+  const matches = error.message.match(/\[(.*?)\]/g);
```
BenBirt: it's a minor thing, but we can probably use a similar regexp as the redshift case to pull out the `lineNumber` and `colNumber` in one go here.

protos/core.proto
```diff
@@ -116,6 +116,25 @@ message CompilationError {
...
 }
 
+
+message EvaluationError {
+  message ErrorLocation {
+    int32 line = 1;
+    int32 column = 2;
+  }
+  string message = 1;
+  ErrorLocation error_location = 2;
```
BenBirt: leave a comment on this line stating that if no location can be determined, this field will be unset?

api/utils/error_parsing.ts
```diff
@@ -0,0 +1,100 @@
+interface IErrorLocation {
```
BenBirt: for all of these parsers: should their bodies be wrapped in try/catch clauses just in case some string splitting/matching/etc breaks for some reason? (so that we always return the original error)

protos/core.proto
```diff
@@ -116,6 +116,25 @@ message CompilationError {
   string stack = 3;
 }
 
+
+message EvaluationError {
```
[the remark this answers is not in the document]
GJMcGowan: I think I will need to use this elsewhere, seems cleaner to be able to call it without going through EvaluationResponse

api/dbadapters/snowflake.ts
```diff
@@ -89,8 +89,15 @@ export class SnowflakeDbAdapter implements IDbAdapter {
     };
   }
 
-  public evaluate(statement: string): Promise<void> {
-    throw new Error("Unimplemented");
+  public evaluate(statement: string): Promise<dataform.EvaluationResponse> {
+    return new Promise(resolve =>
```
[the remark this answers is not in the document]
GJMcGowan: In this case it's because the `IDbAdapter` type has to return `EvaluationResponse` (for all db adapters) - I've added an `any`

api/utils/error_parsing.ts
GJMcGowan: Updated all to create a new instance.

api/utils/error_parsing.ts
GJMcGowan: sure

Issue #593, opened 2020-02-05
Non full-refresh incremental post operations are replaced with pre-operations


lewish:
Running an incremental table update with pre operations results in the pre-operations being run as both pre and post operations, and the post operations themselves not being run.

PR #594, opened 2020-02-05
Fix incremental pre/post ops.

BenBirt:
fixes https://github.com/dataform-co/dataform/issues/593

### DIFF
--- a/tests/api/api.spec.ts
+++ b/tests/api/api.spec.ts
@@ -3,12 +3,12 @@ import { IDbAdapter } from "@dataform/api/dbadapters";
 import { BigQueryDbAdapter } from "@dataform/api/dbadapters/bigquery";
 import * as utils from "@dataform/core/utils";
 import { dataform } from "@dataform/protos";
-import { assert, config, expect } from "chai";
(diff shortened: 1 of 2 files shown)


tests/api/api.spec.ts
```diff
@@ -137,6 +137,72 @@ suite("@dataform/api", () => {
...
+              postOps: ["postOp"],
+              incrementalPostOps: ["incremental postOp"]
+            }
+          ],
+          dataformCoreVersion: "1.4.9"
+        });
+
+        test(`${warehouse} when running non incrementally`, () => {
+          const action = new Builder(graph, {}, TEST_STATE).build().actions[0];
+          expect(action.tasks[0]).eql(
```
lewish: Would be a slightly tighter test if you just checked array equality on all the tasks, including the query (e.g. this would not catch our previous issue of duplicated pre / post ops)

BenBirt: discussed on slack but for posterity: it's a fair point, I'm trying to avoid having per-warehouse SQL in here though which would be hard to avoid in that case.

perhaps we should just give up and have per-warehouse DB adapter tests.

PR #595, opened 2020-02-05
Make test filtering using Bazel's --test_filter flag work.

BenBirt:
I have no idea how to write tests for this :P

### DIFF
--- a/testing/runner.ts
+++ b/testing/runner.ts
@@ -9,6 +9,7 @@ export interface IRunResult {
 }
 
 export interface IRunContext {
+  testNameMatcher: RegExp;
   path: string[];
   results: IRunResult[];
   beforeEaches: Hook[];
@@ -37,9 +38,10 @@ export class Runner {
     try {
       // We tell the runner to start running at the end of current block of
       // synchronously executed code. This will typically be after all the
-      // suite definitions are evaluated. This is equivelant to setTimeout(..., 0).
+      // suite definitions are evaluated. This is equivalent to setTimeout(..., 0).
       await promisify(process.nextTick)();
       const ctx: IRunContext = {
+        testNameMatcher: new RegExp(process.env.TESTBRIDGE_TEST_ONLY) || /.*/,
         path: [],
         results: [],
         beforeEaches: [],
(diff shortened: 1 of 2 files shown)


testing/runner.ts
```diff
@@ -37,9 +38,10 @@ export class Runner {
     try {
       // We tell the runner to start running at the end of current block of
       // synchronously executed code. This will typically be after all the
-      // suite definitions are evaluated. This is equivelant to setTimeout(..., 0).
+      // suite definitions are evaluated. This is equivalent to setTimeout(..., 0).
       await promisify(process.nextTick)();
       const ctx: IRunContext = {
+        testNameMatcher: new RegExp(process.env.TESTBRIDGE_TEST_ONLY) || /.*/,
```
lewish: Small thing, I don't care too much - but if this makes a suite empty, we'll still run before and afters for the suite. You could possibly solve this by adding the filtering the actual `test()` method, so that these tests never even get registered, and change it so that suites with no tests just don't run anything (and also not run before afters).

testing/test.ts
```diff
@@ -46,9 +46,12 @@ export class Test {
   constructor(public readonly options: ITestOptions, private readonly fn: () => any) {}
 
   public async run(ctx: IRunContext) {
-    let lastResult: IRunResult;
     const path = [...ctx.path, this.options.name];
+    if (!path.some(pathPart => pathPart.match(ctx.testNameMatcher))) {
```
lewish: One other thing, it might be useful to filter on a test in a specific suite for when we programmatically create suites.

In that case, I'd be tempted to just join the path parts with " > " like we do in the logs then match that

testing/runner.ts
BenBirt: I couldn't find any nice way of doing this TBH, so leaving it for now.

If I do it somewhere around `test()`, I need to futz around in `globalStack` to figure out the current test path. If I do it instead at runtime, I need to recursively calculate all the tests that will run, before they run - which means constructing new `IRunContext` objects all the way down the Suite tree, so it becomes a bit of a mess.

PR #596, opened 2020-02-06
Add cluster by

Ekrekr:
Fixes https://github.com/dataform-co/dataform-co/issues/3889

API changes, tests changes, and docs updates.

### DIFF
--- a/content/docs/how-to-guides/warehouses/bigquery.md
+++ b/content/docs/how-to-guides/warehouses/bigquery.md
@@ -40,5 +40,5 @@ config {
     clusterBy: ["name", "revenue"]
   }
 }
-SELECT CURRENT_TIMESTAMP() AS ts
+SELECT CURRENT_TIMESTAMP(), name, revenue AS ts
 ```
--- a/content/docs/how-to-guides/warehouses/bigquery.md
+++ b/content/docs/how-to-guides/warehouses/bigquery.md
@@ -30,7 +30,7 @@ PARTITION BY DATE(ts)
 AS (SELECT CURRENT_TIMESTAMP() AS ts)
 ```
 
-If desired, partitions can be clustered by using the `clusterBy` field, for example:
+If desired, partitions can be clustered by using the `clusterBy` option, for example:
 
 ```js
 config {
--- a/content/docs/how-to-guides/warehouses/bigquery.md
+++ b/content/docs/how-to-guides/warehouses/bigquery.md
@@ -29,3 +29,16 @@ CREATE OR REPLACE TABLE dataform.example
 PARTITION BY DATE(ts)
 AS (SELECT CURRENT_TIMESTAMP() AS ts)
 ```
+
+If desired, partitions can be clustered by using the `clusterBy` field, for example:
+
+```js
+config {
+  type: "table",
+  bigquery: {
+    partitionBy: "DATE(ts)",
+    clusterBy: ["name", "revenue"]
+  }
+}
+SELECT CURRENT_TIMESTAMP() AS ts
+```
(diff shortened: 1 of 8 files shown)


core/table.ts
```diff
@@ -112,6 +112,13 @@ export interface IBigQueryOptions {
    * For more information, read the [BigQuery partitioned tables docs](https://cloud.google.com/bigquery/docs/partitioned-tables).
    */
   partitionBy?: string;
+
+  /**
+   * The keys by which to cluster partitions by.
```
Ekrekr: @dwl285 Is this enough? Note the description for partitionBy is `The key with which to partition the table. Typically the name of a timestamp or date column.`

tests/api/api.spec.ts
```diff
@@ -532,7 +532,74 @@ suite("@dataform/api", () => {
           tasks: [
             {
               type: "statement",
-              statement: "create or replace table `deeb.schema.name`  as select 1 as test"
+              statement: "create or replace table `deeb.schema.name` as select 1 as test"
```
Ekrekr: Whitespace change in bigquery adapter removes double space after `\`deeb.schema.name\``

core/utils.ts
```diff
@@ -190,7 +190,14 @@ export function validate(compiledGraph: dataform.ICompiledGraph): dataform.IGrap
     if (!!action.bigquery) {
       if (action.bigquery.partitionBy && action.type === "view") {
         const error = dataform.ValidationError.create({
-          message: `partitionBy is not valid for BigQuery views; it is only valid for tables`,
+          message: `partitionBy/clusterBy are not valid for BigQuery views; they are only valid for tables`,
```
Ekrekr: Potentially controversial change. My reasoning is that:

* Simplifies error logic, don't have to check for the case with just `partitionBy`, just `clusterBy`, and then the case with the combination of both.

* It's still immediately obvious to the user in any of these scenarios that what they did is not valid for views.

core/table.ts
dwl285: LGTM

content/docs/how-to-guides/warehouses/bigquery.md
```diff
@@ -29,3 +29,16 @@ CREATE OR REPLACE TABLE dataform.example
...
+
+```js
+config {
+  type: "table",
+  bigquery: {
+    partitionBy: "DATE(ts)",
+    clusterBy: ["name", "revenue"]
+  }
+}
+SELECT CURRENT_TIMESTAMP() AS ts
```
lewish: Can you add a field such as "name" and use it in the clusterBy, so that this whole example works if you copy paste it?

core/utils.ts
lewish: SGTM!

tests/api/api.spec.ts
```diff
@@ -532,7 +532,74 @@ suite("@dataform/api", () => {
...
+          ]
+        }
+      ];
+      const executionGraph = new Builder(testGraph, {}, dataform.WarehouseState.create({})).build();
+      expect(asPlainObject(executionGraph.actions)).deep.equals(
+        asPlainObject(expectedExecutionActions)
+      );
+    });
+
+    test("bigquery_clusterby", () => {
```
lewish: We don't actually have an end to end test here checking that when we set cluster by on a table, it comes out in the final graph. This only checks from compilation onwards, and your example project only has a failure mode.

content/docs/how-to-guides/warehouses/bigquery.md
Ekrekr: Oops, nice catch

tests/api/api.spec.ts
Ekrekr: Ah, have added test to core tests for both `partitionBy` and `clusterBy`

PR #597, opened 2020-02-06
Quick patch to flag lines that are different to as expected in tests.

Ekrekr:


testing/runner.ts
```diff
@@ -96,7 +97,16 @@ export class Runner {
               console.error(`\n    Actual:\n`);
               console.error(indent(JSON.stringify(result.err.actual, null, 4), 8));
             }
-            console.error("\n");
+            console.error(`\n    Overall diff:`);
```
lewish: In some cases, either expected or actual is missing, can you wrap this in an if statement?

testing/runner.ts
```diff
@@ -96,7 +97,16 @@ export class Runner {
...
+            Diff.diffLines(
+              indent(JSON.stringify(result.err.expected)),
```
lewish: I think you probably want to indent these after?

testing/runner.ts
Ekrekr: Ah good catch

testing/runner.ts
Ekrekr: Meh, either works

PR #598, opened 2020-02-07
Add identical info to tests, make comparisons print objects with keys in order

Ekrekr:


testing/runner.ts
```diff
@@ -86,36 +86,42 @@ export class Runner {
...
-              let toLog = "";
-              diffs.forEach(diff => {
-                // This diff won't show well for users with either default green or red text.
-                const colorPrefix = diff.added ? red : diff.removed ? green : reset;
-                toLog += `${colorPrefix}${diff.value}`;
-              });
-              console.error(toLog);
+              if (diffs.length === 1 && !diffs[0].added && !diffs[0].removed) {
+                console.error(
+                  `\n    ${green}JSONs appear identical! Have all protobufs been casted?`
```
lewish: I'd like this to not be too specific to us, can you put a more generic message, e.g. "Are you comparing objects with functions?"

Ekrekr: Sure

Ekrekr: Sure

PR #600, opened 2020-02-10
Fix docs links

lewish:
Re-use old URLs because we aren't getting re-indexes, strip trailing slashes so relative links work properly / consistently.


Ekrekr:
This PR will break some links in the front end due to the new paths, such as here

image

lewish:
Ack on the links, I'll prep a PR to change them

lewish:
Yup, I won't push this until our release tomorrow (and will prep the closed source changes now!)

PR #601, opened 2020-02-10
Dataform server boilerplate

lewish:
This introduces the bazel set up for us to start a server. It doesn't actually do anything interesting, but get's everything in place for future development.

- Adds rules for protobufjs ts service code generation, copied from closed-source
- Adds a grpc-web rpc_impl for protobufjs, also from closed source
- Adds bazel webpack rules, dependencies and a minimal starter impl for the UI
- Adds a boilerplate grpc-service impl for the backend
- Serves an HTTP server via express, grpc server, and grpc-web-proxy from a single node binary

Can be run with:
```
bazel run server:bin --project-dir <path-to-some-dataform-project>
```


protos/server.proto
```diff
@@ -0,0 +1,15 @@
+syntax="proto3";
+
+package dataform.server;
+
+import "protos/core.proto";
+
+message Empty {}
+
+message MetadataResponse {
+    string project_dir = 2;
```
faiyaz26: why not tag id = 1 ?

app/service.ts
```diff
@@ -0,0 +1,16 @@
...
+  public static get(): Service {
+    if (!Service.instance) {
+      Service.instance = new Service();
+    }
+    return Service.instance;
+  }
+  private static instance: Service;
+
+  constructor() {
+    super(rpcImpl("http://localhost:9111", "dataform.server.Service"), false, false);
```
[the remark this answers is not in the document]
lewish: Unfortunately without some template/variable solution, this is hard to do :/

protos/server.proto
lewish: Fixed

PR #602, opened 2020-02-11
Handled 404s and add robots.txt to docs site

lewish:
Makes sure that when a path isn't found, it returns a 404 instead of a 500 error.
Also add an allow all robots.txt.

### DIFF
--- a/docs/public/robots.txt
+++ b/docs/public/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow:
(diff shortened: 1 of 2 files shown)


docs/public/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
+Disallow:
```
Ekrekr:
```suggestion
Allow: /

Sitemap: https://docs.dataform.co/sitemap.xml
```

Sitemap not required but does help; easiest option is to use an [externally generated sitemap](https://www.xml-sitemaps.com/details-docs.dataform.co-8ee834eda.html), but could add a package to do it instead.

`Disallow: /404.html` can be added when there's a 404 page.

PR #609, opened 2020-02-12
Create schemas in correct databases

### LINKED ISSUE #591: New schemas always created in the default database / project-id
Dataform creates new schemas/datasets automatically, however it always does so in the default database, instead of using the overriden database from the target table.

https://github.com/dataform-co/dataform/blob/master/api/commands/run.ts#L105

Current workaround is to create these schemas manually.

lewish:
Currently we always create schemas in the default database, additionally, we were also not always fetching metadata from the correct database either.

This applies for both BigQuery and snowflake, the other databases don't support override database in, so it is ignored.

- Make sure table metadata methods use the correct database when fully qualified names are available for BQ and Snowflake
- Update integration tests to delete schemas at the beginning so schema creation is tested properly

Fixes #591


api/dbadapters/bigquery.ts
```diff
@@ -178,12 +175,39 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
 
+  public async dropSchema(database: string, schema: string): Promise<void> {
+    const client = this.getClient(database);
+    try {
+      await client.dataset(schema).getMetadata();
+    } catch (e) {
+      // If metadata call fails, it probably doesn't exist, so don't do anything.
+      return;
+    }
+    await this.getClient(database)
```
faiyaz26:
```suggestion
    await client
```

api/dbadapters/bigquery.ts
```diff
@@ -178,12 +175,39 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
   public async close() {}
 
+  private getClient(projectId = this.bigQueryCredentials.projectId) {
+    if (!this.clients.has(projectId)) {
+      this.clients.set(
+        projectId,
+        new BigQuery({
+          projectId,
+          credentials: JSON.parse(this.bigQueryCredentials.credentials),
+          scopes: ["https://www.googleapis.com/auth/drive"]
```
faiyaz26: pull this to a constant maybe?

lewish:
> Nice changes 🧤
> 
> I don't think there's a test in the API spec that `defaultDatabase` is correctly overridden by a `configOverride`, may be worth adding one.

Done!

PR #614, opened 2020-02-19
Add metadata related code to Proto file and dbadapter

faiyaz26:


protos/core.proto
```diff
@@ -424,4 +429,12 @@ message TableMetadata {
   Target target = 1;
   string type = 2; // "table" or "view"
   repeated Field fields = 3;
+  int64 lastUpdatedMillis = 4;
+}
+
+message PersistedTableMetadata {
+  TableMetadata table = 1;
```
lewish: I don't think you need any of this in the metadata table? It's information about columns and types which I don't think we need to keep track of.

protos/core.proto
```diff
@@ -424,4 +429,12 @@ message TableMetadata {
...
+  string create_query = 2;
```
lewish: I've just realized a problem.
Storing just the query is not enough, you need to also check the pre_operations and post_operations, as well as other metadata such as the table type, any warehouse specific settings (partitionBy, clusterBy).

These are also going to vary between each table type (view, incremental, table, as well as declarations, assertions and operations).

Therefore, I'd strongly suggest storing a hash instead. Storing all of these things is going to be a pain in the ass, whereas computing a hash from a JSON object is much easier...

If you still want to store the exact query, then you could possibly store the list of queries in the "tasks" object of an ExecutionAction

protos/core.proto
```diff
@@ -424,4 +429,12 @@ message TableMetadata {
...
+  int64 lastUpdatedMillis = 3;
```
lewish: shouldn't be camelCase

protos/core.proto
```diff
@@ -424,4 +429,12 @@ message TableMetadata {
...
+  repeated PersistedTableMetadata dependencies = 4;
```
lewish: Recursive protos are annoying, also this is slightly incorrect (one of these repeated objects shouldn't have any children itself).
I think you can avoid this here by breaking this into 2 messages

protos/core.proto
```diff
@@ -424,4 +429,12 @@ message TableMetadata {
   Target target = 1;
   string type = 2; // "table" or "view"
   repeated Field fields = 3;
+  int64 lastUpdatedMillis = 4;
+}
+
+message PersistedTableMetadata {
```
lewish: We need to store the action name in here as a string right?

protos/core.proto
```diff
@@ -424,4 +429,12 @@ message TableMetadata {
   Target target = 1;
   string type = 2; // "table" or "view"
   repeated Field fields = 3;
+  int64 lastUpdatedMillis = 4;
```
lewish: not camelCase

api/dbadapters/bigquery.ts
```diff
@@ -190,6 +194,24 @@ export class BigQueryDbAdapter implements IDbAdapter {
 
   public async close() {}
 
+  public async persistedStateMetadata(
+    projectConfig: dataform.IProjectConfig
+  ): Promise<dataform.IPersistedTableMetadata[]> {
+    try {
+      const { rows } = await this.runQuery(
+        `SELECT * FROM \`${projectConfig.defaultDatabase}.${projectConfig.defaultSchema}.${STATE_PERSIST_TABLE_NAME}\``,
```
lewish: Can we re-use the resolveTarget method from the adpater that already does this?

protos/core.proto
lewish: While you don't need it in this PR, you will definitely need it to write updates correctly in the future!

protos/core.proto
faiyaz26: I changed the name to create_query_hash, which will store the hash of the publish task object.

protos/core.proto
faiyaz26: I added the target object, which is actually resolved to action name.

protos/core.proto
faiyaz26: moved to another message

protos/core.proto
faiyaz26: broke into two message

api/dbadapters/bigquery.ts
faiyaz26: done

api/dbadapters/bigquery.ts
```diff
@@ -7,6 +8,8 @@ import { QueryResultsOptions } from "@google-cloud/bigquery/build/src/job";
 import * as Long from "long";
 import * as PromisePool from "promise-pool-executor";
 
+import { STATE_PERSIST_TABLE_NAME } from "@dataform/api/dbadapters/index";
```
lewish: you can import this in the same line above

api/dbadapters/bigquery.ts
```diff
@@ -190,6 +195,28 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+  public async persistedStateMetadata(
+    compiledGraph: dataform.ICompiledGraph
+  ): Promise<dataform.IPersistedTableMetadata[]> {
+    const adapter = adapters.create(
+      compiledGraph.projectConfig,
+      compiledGraph.dataformCoreVersion || "1.0.0"
+    );
+    try {
+      const { rows } = await this.runQuery(
+        `SELECT * FROM \`${adapter.resolveTarget({ name: STATE_PERSIST_TABLE_NAME })}\``,
```
lewish: I thought the intention is this table is fixed regardless of environments etc. The problem here is, this will create a different table for each environment (schemaSuffix) and in the future will also end up respecting tablePrefix option we add.

The design doc implies that the schema is fixed regardless of the environment, in which case you should have a fixed schema name and table name and pass both in to resolveTarget.

My suggestion would be to create a table:
schema: dataform_meta
table: cache_state

Can we add this to the design doc?

protos/core.proto
```diff
@@ -424,4 +429,17 @@ message TableMetadata {
...
+}
+
+message PersistedTableDependency{
+  Target target = 1;
+  int64 last_updated_millis = 2;
+}
+
+message PersistedTableMetadata {
+  Target target = 1;
+  string create_query_hash = 2;
```
lewish: nit: I think this is a bit misleading, as it's more than just the create query, and for other files in the future (e.g.operatations) there isn't necessarily a create query at all.

I'd suggest something like: definition_hash or just hash, that's generic?

protos/core.proto
```diff
@@ -424,4 +429,17 @@ message TableMetadata {
   Target target = 1;
   string type = 2; // "table" or "view"
   repeated Field fields = 3;
+  int64 last_updated_millis = 4;
+}
+
+message PersistedTableDependency{
```
lewish: nit whitespace before {

api/dbadapters/bigquery.ts
faiyaz26: Added: https://docs.google.com/document/d/1cksnd-lFnQeWLIbhEaYAENl3ylKWcmsm-JMeUXIgabU/edit?ts=5e3d3f19#heading=h.2g1udzmih3xp

api/dbadapters/bigquery.ts
```diff
@@ -1,11 +1,13 @@
 import { Credentials } from "@dataform/api/commands/credentials";
+import { adapters } from "@dataform/core";
```
BenBirt: i think this is a bit sketchy. we want to be removing dependencies on `core` from `api`, not adding them. it looks like we only use this to determine the metadata table's name, do we really need to do that instead of hardcoding the table name?

api/dbadapters/bigquery.ts
```diff
@@ -190,6 +194,28 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+    compiledGraph: dataform.ICompiledGraph
+  ): Promise<dataform.IPersistedTableMetadata[]> {
+    const adapter = adapters.create(
+      compiledGraph.projectConfig,
+      compiledGraph.dataformCoreVersion || "1.0.0"
+    );
+    try {
+      const { rows } = await this.runQuery(
+        `SELECT * FROM ${adapter.resolveTarget(STATE_PERSIST_TABLE_TARGET)}`,
+        5000 // not expecting to have more dataset than this
```
BenBirt: this is definitely a bit concerning, as we have seen some users with insane numbers of tables, especially with a few users attached to a large project which has IDS turned on.

i wonder if this suggests we should be rethinking the metadata table's schema, to have separate columns for database/schema/table, so that we can do smaller lookups.

in any case, there should be a check here that the returned number of rows is not equal to 5000 - and if it is, throw an Error or something as at that point we can't reliably return results

api/dbadapters/bigquery.ts
```diff
@@ -190,6 +194,28 @@ export class BigQueryDbAdapter implements IDbAdapter {
 
   public async close() {}
 
+  public async persistedStateMetadata(
+    compiledGraph: dataform.ICompiledGraph
```
BenBirt: it definitely feels a bit weird to have an API db adapter depend on the CompiledGraph proto. if we stop depending on core, we can get rid of this dependency

api/dbadapters/bigquery.ts
```diff
@@ -190,6 +194,28 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+        `SELECT * FROM ${adapter.resolveTarget(STATE_PERSIST_TABLE_TARGET)}`,
+        5000 // not expecting to have more dataset than this
+      );
+      const peristedMetadata = rows.map(row => {
+        const encodedProto = Buffer.from(row.proto, "base64");
+        return dataform.PersistedTableMetadata.decode(encodedProto);
+      });
+      return peristedMetadata || [];
+    } catch (err) {
+      return [];
```
BenBirt: it feels pretty unsafe to be doing this at this level! perhaps in the runner this is OK, but inside the DB adapter this seems a bit dangerous.

api/dbadapters/bigquery.ts
```diff
@@ -190,6 +194,28 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+    try {
+      const { rows } = await this.runQuery(
+        `SELECT * FROM ${adapter.resolveTarget(STATE_PERSIST_TABLE_TARGET)}`,
+        5000 // not expecting to have more dataset than this
+      );
+      const peristedMetadata = rows.map(row => {
+        const encodedProto = Buffer.from(row.proto, "base64");
+        return dataform.PersistedTableMetadata.decode(encodedProto);
+      });
+      return peristedMetadata || [];
```
BenBirt: nit: persistedMetadata cannot be falsy, just return it

api/dbadapters/bigquery.ts
```diff
@@ -190,6 +194,28 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+    const adapter = adapters.create(
+      compiledGraph.projectConfig,
+      compiledGraph.dataformCoreVersion || "1.0.0"
+    );
+    try {
+      const { rows } = await this.runQuery(
+        `SELECT * FROM ${adapter.resolveTarget(STATE_PERSIST_TABLE_TARGET)}`,
+        5000 // not expecting to have more dataset than this
+      );
+      const peristedMetadata = rows.map(row => {
```
BenBirt:
```suggestion
      const persistedMetadata = rows.map(row => {
```

api/dbadapters/index.ts
```diff
@@ -7,6 +7,11 @@ import { dataform } from "@dataform/protos";
 
 export type OnCancel = (handleCancel: () => void) => void;
 
+export const STATE_PERSIST_TABLE_TARGET: dataform.ITarget = {
```
BenBirt:
```suggestion
export const CACHED_STATE_TABLE_TARGET: dataform.ITarget = {
```

cli/index.ts
```diff
@@ -207,7 +207,8 @@ const builtYargs = createYargsCli({
           argv["project-dir"],
           {
             warehouse: argv.warehouse,
-            defaultDatabase: argv["default-database"]
+            defaultDatabase: argv["default-database"],
+            useRunCache: false
```
BenBirt: nit: seems unnecessary, this is the default value

protos/core.proto
```diff
@@ -424,4 +427,16 @@ message TableMetadata {
   Target target = 1;
   string type = 2; // "table" or "view"
   repeated Field fields = 3;
+  int64 last_updated_millis = 4;
+}
+
+message PersistedTableMetadata {
+  Target target = 1;
+  message CacheKey{
```
BenBirt: i think i change my mind back on this, sorry!

these two fields aren't really the cache key. the cache key really is just `definition_hash`. `last_updated_millis` is only used for computing whether a table's *dependencies* have changed; `last_updated_millis` for a table `A` is never used in the context of determining whether the cache for `A` is broken or not.

i think this was right the first time, just having these fields part of the outer message.

please put comments on the fields though, it's tricky to know precisely what they mean!

api/dbadapters/bigquery.ts
faiyaz26: Added comment

api/dbadapters/bigquery.ts
```diff
@@ -190,6 +195,20 @@ export class BigQueryDbAdapter implements IDbAdapter {
 
   public async close() {}
 
+  public async persistedStateMetadata(
+    database: string
```
BenBirt: we shouldn't need this. queries always run by default in the scope of the database to which we're connected, which is the default database.

protos/core.proto
```diff
@@ -424,4 +427,13 @@ message TableMetadata {
   Target target = 1;
   string type = 2; // "table" or "view"
   repeated Field fields = 3;
+  int64 last_updated_millis = 4;
+}
+
+message PersistedTableMetadata {
+  Target target = 1;
+  string definition_hash = 2; // sha1 hash of ExectionAction object
+  int64 last_updated_millis = 3; // when the table/view last updated, extracted from warehouse
```
BenBirt: this comment seems a bit confusing to me. 'extracted from warehouse' makes it sound like it's the warehouse's last modified timestamp, which it is, but it sounds like this value might change even if no runs happen (which it can't, right?).

i would suggest `Timestamp of the last modification to this dataset (where that modification happened as a result of a Dataform run).`

PR #622, opened 2020-02-26
Store cache metadata to warehouse

faiyaz26:


api/commands/state.ts
```diff
@@ -8,11 +8,19 @@ export async function state(
...
   });
 
-  return { tables: tablesWithValues };
+  let cachedStates: dataform.IPersistedTableMetadata[] = null;
+
+  if (compiledGraph.projectConfig.useRunCache) {
+    try {
+      cachedStates = await dbadapter.persistedStateMetadata();
+    } catch (err) {
+      cachedStates = [];
```
BenBirt: leave a comment, clarifying that the worst case is that we were unable to load cached state so this is treated equivalently to the cache being empty?

api/dbadapters/bigquery.ts
```diff
@@ -199,13 +200,68 @@ export class BigQueryDbAdapter implements IDbAdapter {
       `SELECT * FROM ${CACHED_STATE_TABLE_NAME}`,
       5000 // TODO: Add pagination for 5000+ rows
     );
-    const persistedMetadata = rows.map(row => {
-      const encodedProto = Buffer.from(row.proto, "base64");
+    const persistedMetadata = rows.map((row: IMetadataRow) => {
```
BenBirt: i feel like we should put this code, plus the `IMetadataRow` interface, inside some common utility location.

api/dbadapters/bigquery.ts
```diff
@@ -199,13 +200,68 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+      const encodedProto = Buffer.from(row.metadata_proto, "base64");
       return dataform.PersistedTableMetadata.decode(encodedProto);
     });
     return persistedMetadata;
   }
 
+  public async persistStateMetadata(executionGraph: dataform.IExecutionGraph): Promise<void> {
+    const { actions } = executionGraph;
```
BenBirt: i would argue for just passing in an array of `IExecutionAction` objects instead of the whole execution graph

api/dbadapters/bigquery.ts
```diff
@@ -199,13 +200,68 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+    await Promise.all(
+      actions.map(async action => {
+        tableMetadataMap.set(action.target, await this.getMetadata(action.target));
+      })
+    );
+    const updateQueries = actions.map(action => {
+      const definitionHash = hashTableDefinition(action);
+      const dependencies = action.dependencyTargets;
+      const metadata =
+        action.type === "operation"
```
BenBirt: IMO this is confusing from a readability standpoint. i get that you're adding special code to effectively cachebust for each `operation`, but the code isn't very clear about that. IMO we shouldn't even be writing data about `operation` actions to the cache.

api/dbadapters/bigquery.ts
```diff
@@ -199,13 +200,68 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+    `;
+    await this.runQuery(metadataTableCreateQuery);
+    const tableMetadataMap = new Map<dataform.ITarget, IBigQueryTableMetadata>();
+    await Promise.all(
+      actions.map(async action => {
+        tableMetadataMap.set(action.target, await this.getMetadata(action.target));
+      })
+    );
+    const updateQueries = actions.map(action => {
+      const definitionHash = hashTableDefinition(action);
```
BenBirt: all of this code to do with computing `PersistedTableMetadata` as a result of table metadata feels like it should probably be in a common utility location.

api/dbadapters/bigquery.ts
```diff
@@ -199,13 +200,68 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+        persistTable.toJSON()
+      )}', '${encodedProtoBuffer.toString("base64")}')
+          WHEN MATCHED THEN
+            UPDATE SET metadata_json = '${JSON.stringify(persistTable.toJSON())}',
+                metadata_proto = '${encodedProtoBuffer.toString("base64")}'
+          ;`;
+
+      return updateQuery;
+    });
+    const batchQuery = updateQueries.join("\n");
```
BenBirt: instead of one `MERGE` statement per table, why not have one `MERGE` statement and have one row in the merged target per table?

tests/integration/bigquery.spec.ts
```diff
@@ -200,4 +201,19 @@ suite("@dataform/integration/bigquery", ({ after }) => {
       expect(bqMetadata.totalBytesProcessed).to.eql(Long.fromNumber(0));
     });
   });
+
+  suite("run cache", async () => {
```
BenBirt: IMO this should really just be part of the main integration test above

api/commands/state.ts
faiyaz26: Added

api/dbadapters/bigquery.ts
```diff
@@ -223,43 +226,41 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
-        action.type === "operation"
-          ? { lastModifiedTime: `${new Date().valueOf()}` }
-          : tableMetadataMap.get(action.target);
-      const persistTable = dataform.PersistedTableMetadata.create({
-        target: action.target,
-        lastUpdatedMillis: Long.fromString(metadata.lastModifiedTime),
-        definitionHash,
-        dependencies
-      });
+    const queries = actions
```
BenBirt: IMO all of the code that produces `queries` belongs inside `run_cache.ts`

protos/core.proto
```diff
@@ -281,7 +281,7 @@ message ExecutionTask {
 
 message ExecutionAction {
   string name = 1;
-  repeated string dependencies = 3;
+  repeated string dependencies = 3 [deprecated = true];
```
BenBirt: nit: move this to the bottom of the proto like in `Table`

api/dbadapters/bigquery.ts
faiyaz26: This code will be different for different warehouse. I have moved the selectQuery part to run_cache

PR #623, opened 2020-02-26
Added mux lock and unlock

Ekrekr:


protomongo/protomongo.go
```diff
@@ -13,13 +14,14 @@ import (
 )
 
 type protobufCodec struct {
+	mux          sync.Mutex
```
BenBirt: nit: `mux` to me implies something to do with multiplexing. just call it `m`?

BenBirt: surely this should be a `*sync.Mutex` (pointer)

protomongo/protomongo.go
```diff
@@ -139,8 +141,12 @@ type protoHelper struct {
 func (pc *protobufCodec) protoHelper(pb descriptor.Message, t reflect.Type) *protoHelper {
 	// Try to load a pre-existing protoHelper from cache, if it exists.
 	messageName := proto.MessageName(pb)
-	if ph, ok := pc.protoHelpers[messageName]; ok {
-		return ph
+
+	pc.mux.Lock()
```
BenBirt: don't bother locking/unlocking on read; the vast majority of the time the read will return successfully so this will just cost time.

protomongo/protomongo.go
```diff
@@ -167,7 +173,11 @@ func (pc *protobufCodec) protoHelper(pb descriptor.Message, t reflect.Type) *pro
 		oneofPropsByTag[strconv.Itoa(oneof.Prop.Tag)] = oneof
 	}
 	ph := &protoHelper{normalPropsByTag, oneofPropsByTag, oneofFieldWrapperProps}
+
+	pc.mux.Lock()
 	pc.protoHelpers[messageName] = ph
+	pc.mux.Unlock()
```
BenBirt: this isn't quite correct - if code between `Lock()` and `Unlock()` returns early (or panics), the mutex will never unlock. we should really `defer pc.mux.Unlock()` straight after locking it.

Ekrekr: Ah very cool

PR #635, opened 2020-03-03
Markdown CMS fixes

lewish:
- Make it so that a index.md file isn't required
- Give the markdown cms library a module_name so it can be used properly
- Update references in docs
- Make the attributes generic so they can be customized (different attributes e.g. for blog posts)
- Fix tsconfig settings
- Make sure next/webpack can resolve custom module names

### DIFF
--- a/docs/content_tree.ts
+++ b/docs/content_tree.ts
@@ -1,6 +1,6 @@
 import { GitHubCms } from "@dataform-tools/markdown-cms/github";
 import { LocalCms } from "@dataform-tools/markdown-cms/local";
-import { IBaseAttributes, Tree } from "@dataform-tools/markdown-cms/tree";
+import { Tree } from "@dataform-tools/markdown-cms/tree";
 import NodeCache from "node-cache";
 
 export interface IExtraAttributes {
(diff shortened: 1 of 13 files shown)


docs/content_tree.ts
```diff
@@ -1,19 +1,23 @@
-import { GitHubCms } from "df/tools/markdown-cms/github";
-import { LocalCms } from "df/tools/markdown-cms/local";
-import { Tree } from "df/tools/markdown-cms/tree";
```
Ekrekr: IBaseAttributes no longer needed

tools/markdown-cms/local.ts
```diff
@@ -1,21 +1,23 @@
...
 
   public async get(path = "") {
     const isDir = await this.isDirectory(path);
     const actualFilePath = isDir
       ? join(this.rootPath, path, "index.md")
       : join(this.rootPath, path);
-    const rawContent = await promisify(fs.readFile)(actualFilePath, "utf8");
+    const rawContent = (await promisify(fs.exists)(actualFilePath))
+      ? await promisify(fs.readFile)(actualFilePath, "utf8")
+      : "";
```
Ekrekr:
```suggestion
    const rawContent = fs.existsSync(actualFilePath)
      ? fs.readFileSync(actualFilePath, "utf8")
      : "";
```

All file `promisify` stuff can me removed using `fs-extra`; looks neater but not really required

Ekrekr: Nice stuff!

lewish: Ack - don't want to add this dep yet as we don't have it in open-source, although I agree it's nicer!

PR #636, opened 2020-03-03
Fix useRunCache flag check and change the timer position for true run timing

faiyaz26:


api/commands/build.ts
```diff
@@ -64,6 +64,7 @@ export class Builder {
       runConfig: {
         ...this.runConfig,
         useRunCache:
+          this.runConfig.hasOwnProperty("useRunCache") === false ||
```
lewish: !this.runConfig.hasOwnProperty("useRunCache") ||

PR #637, opened 2020-03-04
Skip disabled action during storing metadata

faiyaz26:
Fixes: https://github.com/dataform-co/dataform-co/issues/4412


api/commands/run.ts
```diff
@@ -110,7 +110,11 @@ export class Runner {
     this.runResult.timing = timer.end();
 
     if (this.graph.runConfig && this.graph.runConfig.useRunCache) {
-      await this.adapter.persistStateMetadata(this.graph.actions);
+      // Currently, we don't support caching for operation and its dependents)
```
```suggestion
      // Currently, we don't support caching for operations (and any dependents)
```

PR #641, opened 2020-03-05
Store metadata only for successful actions

faiyaz26:


tests/integration/bigquery.spec.ts
```diff
@@ -129,8 +129,10 @@ suite("@dataform/integration/bigquery", ({ after }) => {
     const exampleViewExecutionAction = executionGraph.actions.find(
       action => action.name === "dataform-integration-tests.df_integration_test.example_view"
     );
-
     expect(exampleView.definitionHash).to.eql(hashExecutionAction(exampleViewExecutionAction));
+
+    const exampleAssertionFail = data.find(table => table.target.name === "example_assertion_fail");
```
BenBirt: IMO we should really check *all* failed actions (there is more than one)

faiyaz26: Added an expect statement to match number of metadata row to number of successfulActions

PR #651, opened 2020-03-10
Stop bq false line number

Ekrekr:
Fixes https://github.com/dataform-co/dataform-co/issues/4178

Error only seems to happen in BQ projects.

### DIFF
--- a/api/utils/error_parsing.ts
+++ b/api/utils/error_parsing.ts
@@ -38,6 +38,7 @@ export function parseRedshiftEvalError(statement: string, error: IRedshiftEvalua
   // expected error format:
   // e.position = "123" - position is the number of characters into the query that the error was found at
   // including \n characters
+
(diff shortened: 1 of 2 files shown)


api/utils/error_parsing.ts
```diff
@@ -75,7 +75,9 @@ export function parseBigqueryEvalError(error: IBigqueryEvaluationError) {
     // extract everything after the very last [ in the string
     const bracketsString = error.message.split("[").slice(-1)[0];
     const [_, lineNumber, columnNumber] = bracketsString.match(/([0-9]*)[^0-9]*([0-9]*).*/);
-    evalError.errorLocation = { line: Number(lineNumber), column: Number(columnNumber) };
+    if (lineNumber && columnNumber) {
```
BenBirt: this needs fixing for errors at `(0, 0)`. probably need to check the match status from the above line.

PR #656, opened 2020-03-17
Implement skip publish algorithm

faiyaz26:


api/commands/run.ts
```diff
@@ -1,9 +1,11 @@
 import { Credentials } from "@dataform/api/commands/credentials";
 import * as dbadapters from "@dataform/api/dbadapters";
 import { retry } from "@dataform/api/utils/retry";
+import { hashExecutionAction } from "@dataform/api/utils/run_cache";
 import { dataform } from "@dataform/protos";
 import * as EventEmitter from "events";
 import * as Long from "long";
+import * as _ from "lodash";
```
BenBirt: please name this import

api/commands/run.ts
```diff
@@ -118,7 +120,19 @@ export class Runner {
 
       // Currently, we don't support caching for operations (and any dependents)
       await this.adapter.persistStateMetadata(
-        successfulActions.filter(action => action.type !== "operation")
+        successfulActions.filter(action => action && action.type !== "operation")
+      );
+      await this.adapter.deleteStateMetadata(
```
BenBirt: feels like it might be prudent to do this before we persist anything new, as it reduces the chance of the persisting failing causing the deletion to not run.

BenBirt: what's the risk of this failing? if the metadata fails to be deleted, will the next run be able to run correctly?

api/commands/run.ts
```diff
@@ -118,7 +120,19 @@ export class Runner {
...
+        this.runResult.actions
+          .filter(
+            action =>
+              action.status !== dataform.ActionResult.ExecutionStatus.SUCCESSFUL &&
+              action.status !== dataform.ActionResult.ExecutionStatus.CACHE_SKIPPED
+          )
+          .map(action =>
+            this.graph.actions.find(executionAction => action.name === executionAction.name)
+          )
+          .filter(action => !!action)
```
BenBirt: this line concerns me a bit. AFAIUI it should be impossible for `RunResult` to not contain an action entry for each action in `ExecutionGraph`

api/commands/run.ts
```diff
@@ -329,6 +353,78 @@ export class Runner {
     await this.triggerChange();
     return taskResult.status;
   }
+
+  private isActionCacheable(action: dataform.IExecutionAction): boolean {
```
BenBirt: i would suggest something like `actionHasCacheHit` for clarity

api/commands/run.ts
```diff
@@ -329,6 +353,78 @@ export class Runner {
...
+
+    for (const dependencyTarget of action.dependencyTargets) {
+      const dependencyAction = this.graph.actions.find(
+        action => action.target === dependencyTarget
+      );
+      if (!dependencyAction) {
+        continue;
+      }
+
+      if (dependencyAction.type === "operation") {
```
BenBirt: it seems weird to repeat this logic. already below you check that all dependencies of this action ran, and completed with `CACHE_SKIPPED`, thus this check must have already been run for the dependency, right?

api/commands/run.ts
```diff
@@ -329,6 +353,78 @@ export class Runner {
...
+      );
+
+      if (
+        runResultAction &&
+        runResultAction.status !== dataform.ActionResult.ExecutionStatus.CACHE_SKIPPED
+      ) {
+        return false;
+      }
+
+      const cachedState = this.graph.warehouseState.cachedStates.find(state =>
```
BenBirt: as above, haven't these checks already been run for the dependency, because before the dependency ran it had these checks run on itself (the code below, outside of this loop), and thus if it were cached it must have completed with `CACHE_SKIPPED`?

cli/console.ts
```diff
@@ -327,6 +327,37 @@ export function printExecutedAction(
       }
       return;
     }
+    case dataform.ActionResult.ExecutionStatus.CACHE_SKIPPED: {
+      switch (executionAction.type) {
+        case "table": {
+          writeStdOut(
+            `${warningOutput("Cached dataset creation: ")} ${datasetString(
```
BenBirt:
```suggestion
            `${warningOutput("Skipped dataset creation (cache hit): ")} ${datasetString(
```

similar below

tests/integration/bigquery_project/definitions/example_operation.sqlx
```diff
@@ -0,0 +1 @@
+select 1 as test
```
BenBirt: nit: newline at eof

api/commands/run.ts
faiyaz26: thats the official name :D anyway, changing it to Lodash

api/commands/run.ts
faiyaz26: moved it before persisting new information

api/commands/run.ts
```diff
@@ -110,15 +112,24 @@ export class Runner {
...
+          ].includes(action.status)
+        )
         .map(action =>
           this.graph.actions.find(executionAction => action.name === executionAction.name)
         );
 
       // Currently, we don't support caching for operations (and any dependents)
       await this.adapter.persistStateMetadata(
-        successfulActions.filter(action => action.type !== "operation")
+        successfulActions.filter(action => action && action.type !== "operation")
```
BenBirt: how can `action` possibly be falsey? AFAICS `action` comes from a lookup in the `ExecutionGraph` for the action with the same name in `RunResult`. i can't see how an action would appear in `RunResult` if it weren't in `ExecutionGraph` - indeed, if this happens, we should definitely blow up as this implies something has gone Really Wrong

api/commands/run.ts
```diff
@@ -329,6 +350,57 @@ export class Runner {
...
+      if (!dependencyAction) {
+        continue;
+      }
+
+      const runResultAction = this.runResult.actions.find(
+        action => action.name === dependencyAction.name
+      );
+
+      if (
+        runResultAction &&
```
BenBirt: as above: if `runResultAction` is falsey, IMO we should be blowing up, or at least returning false to this being a cache hit - case this should be impossible. i'd just remove this line altogether

api/commands/run.ts
```diff
@@ -1,9 +1,11 @@
 import { Credentials } from "@dataform/api/commands/credentials";
 import * as dbadapters from "@dataform/api/dbadapters";
 import { retry } from "@dataform/api/utils/retry";
+import { hashExecutionAction } from "@dataform/api/utils/run_cache";
 import { dataform } from "@dataform/protos";
 import * as EventEmitter from "events";
 import * as Long from "long";
+import * as Lodash from "lodash";
```
```suggestion
import * as lodash from "lodash";
```

api/commands/run.ts
faiyaz26: seems not needed.

faiyaz26:
@BenBirt fixed the test

PR #657, opened 2020-03-18
Push documentation to bigquery

GJMcGowan:
fixes #4772

* Adds actionDescriptor from compiled graph to executed graph
* Adds 'setMetadata' function to adapters (only bigquery uses)
* Calls 'setMetadata' when a run is going on


api/commands/run.ts
```diff
@@ -280,6 +280,10 @@ export class Runner {
       }
     }
 
+    if (action.actionDescriptor) {
```
BenBirt: IMO we should only do this for `SUCCESSFUL` actions.

api/dbadapters/bigquery.ts
```diff
@@ -256,6 +258,61 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
   }
 
+  public async setMetadata(action: dataform.IExecutionAction): Promise<void> {
+    const { target, actionDescriptor } = action;
+
+    await this.pool
+      .addSingleTask({
+        generator: async () => {
+          try {
+            const metaData = await this.getMetadata(target);
```
BenBirt:
```suggestion
            const metadata = await this.getMetadata(target);
```

api/dbadapters/bigquery.ts
```diff
@@ -256,6 +258,61 @@ export class BigQueryDbAdapter implements IDbAdapter {
     await this.runQuery(updateQuery);
   }
 
+  public async setMetadata(action: dataform.IExecutionAction): Promise<void> {
+    const { target, actionDescriptor } = action;
+
+    await this.pool
```
BenBirt: we shouldn't run this code for any action that isn't of type `table`, right?

api/dbadapters/bigquery.ts
```diff
@@ -256,6 +258,61 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+      .addSingleTask({
+        generator: async () => {
+          try {
+            const metaData = await this.getMetadata(target);
+            const schemaWithDescription = addDescriptionToMetadata(actionDescriptor?.columns, metaData.schema.fields)
```
BenBirt: IMO just return immediately at the start of this function if there is no `actionDescriptor` - it's faster

api/dbadapters/bigquery.ts
```diff
@@ -256,6 +258,61 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+              .table(target.name)
+              .setMetadata({
+                description: actionDescriptor.description,
+                schema: schemaWithDescription
+              });
+            return table;
+          } catch (e) {
+            if (e && e.errors && e.errors[0] && e.errors[0].reason === "notFound") {
+              // if the table can't be found, just return null
+              return null;
```
BenBirt: this seems weird. if this code is being called and this case happens, something has gone Really Wrong - I don't think we should swallow this error

api/dbadapters/bigquery.ts
```diff
@@ -416,3 +449,33 @@ function convertField(field: IBigQueryFieldMetadata): dataform.IField {
   }
   return result;
 }
+
+function addDescriptionToMetadata (
+  columnDescriptions: dataform.IColumnDescriptor[],
+  metadataArray: IBigQueryFieldMetadata[]
+): IBigQueryFieldMetadata[]  {
+  const findNextField = (fields: IBigQueryFieldMetadata[], path: string[]) => fields?.find((field) => field.name === path[0])
```
BenBirt: TBH i'm having some trouble reading this code. i would expect:
1. loop through `metadataArray`
2. find the column descriptor that matches this column, if one exists
3. append its description
4. recurse for any nested fields

I *think* you're doing the inverse of that loop?

tests/integration/bigquery.spec.ts
```diff
@@ -137,6 +137,47 @@ suite("@dataform/integration/bigquery", ({ after }) => {
     expect(exampleAssertionFail).to.be.eql(undefined);
 
     expect(persistedMetaData.length).to.be.eql(9);
+
+    // metadata
+    const target = {
```
BenBirt: we should be able to get this out of the `ExecutionGraph` instead of re-creating it

tests/integration/bigquery_project/definitions/example_incremental.sqlx
```diff
@@ -1,14 +1,28 @@
...
+    columns: {
+        user_timestamp: "the timestamp",
+        user_id: "the id",
+        nested_data: {
+          description: "some nested data with duplicate fields",
+          columns: {
+            user_timestamp: "nested timestamp",
+            user_id: "nested id"
+          }
+        }
```
BenBirt: can you add `not_a_column_name` to check that nothing breaks?

tests/integration/bigquery_project/definitions/example_incremental.sqlx
```diff
@@ -1,14 +1,28 @@
 WITH example_data AS (
```
BenBirt: i realize you haven't changed this, but IMO this is confusing - move this `WITH` block down below the `config` block?

api/dbadapters/bigquery.ts
GJMcGowan: Far enough, this was a copy paste from `getMetadata`

api/dbadapters/bigquery.ts
GJMcGowan: So I added an integration test for `view` type and it seems to work. Should I just run the code for actions of type `view || table`?

BenBirt: yes please

api/dbadapters/bigquery.ts
```diff
@@ -433,3 +464,30 @@ function convertField(field: IBigQueryFieldMetadata): dataform.IField {
...
 }
+
+function addDescriptionToMetadata(
+  columnDescriptions: dataform.IColumnDescriptor[],
+  metadataArray: IBigQueryFieldMetadata[]
+): IBigQueryFieldMetadata[] {
+  const findDescription = (path: string[]) =>
+    columnDescriptions.find(column => column.path.join("") === path.join(""));
+
+  const mapDescriptionToMetadata = (metadata: IBigQueryFieldMetadata, previousPath: string[]) => {
```
BenBirt:
```suggestion
  const mapDescriptionToMetadata = (metadata: IBigQueryFieldMetadata, path: string[]) => {
```

GJMcGowan:
confirmed in the bigquery ui

Screenshot 2020-03-19 at 16 02 16
Screenshot 2020-03-19 at 16 02 20

PR #658, opened 2020-03-19
Add minimal analytics tracking for CLI behind explicit opt-in

lewish:
- Establishes a new configuration settings file in `$USER/.dataform/settings.json`
- Asks user if they would like to opt in to anonymous usage and error tracking when using some commands, will only ask once
- Stores a unique user ID using UUID in the configuration settings along with opt-in choice
- Tracks all command usage by name only, and only if user has opted in


cli/analytics.ts
```diff
@@ -0,0 +1,98 @@
...
+import { v4 as uuidv4 } from "uuid";
+
+const analytics = new Analytics("eR24ln3MniE3TKZXkvAkOGkiSN02xXqw");
+
+export async function maybeConfigureAnalytics() {
+  const settings = await getConfigSettings();
+  const dockerStatus = await getDockerStatus();
+  // We should only ask if users want to track analytics if they aren't in docker, e.g. for CI.
+  if (!dockerStatus && settings.allowAnonymousAnalytics === undefined) {
+    const optInResponse = question(
```
BenBirt: can you expose a new method in `console.ts` that uses https://www.npmjs.com/package/readline-sync#keyinyn?

```
function ynQuestion(questionText: string, default: boolean = false): boolean {
  const response = readlineSync.keyInYN(questionText);
  if (response === '') {
    return default;
  }
  return response;
}
```

cli/analytics.ts
```diff
@@ -0,0 +1,98 @@
...
+      `
+To help improve the quality of our products, we collect anonymized usage data and anonymized stacktraces when crashes are encountered.
+This can be changed at any point by modifying the file: ${getConfigSettingsPath()}
+
+Do you want to opt in to anonymous usage and error tracking (y/N)?`
+    );
+    await upsertConfigSettings({
+      allowAnonymousAnalytics: optInResponse.toLowerCase() === "y",
+      userId: uuidv4()
```
BenBirt:
```suggestion
      anonymousUserId: uuidv4()
```

cli/config.ts
```diff
@@ -0,0 +1,46 @@
+import { exists, mkdir, readFile, writeFile } from "fs";
+import { homedir } from "os";
+import { join } from "path";
+import { promisify } from "util";
+
+export interface IConfigSettings {
```
BenBirt: can you put this in a protobuf definition? or at least put a loud comment on the interface saying that existing members of this interface *must* not be changed or we lose access to those user settings?

cli/config.ts
```diff
@@ -0,0 +1,46 @@
...
+  return join(homedir(), ".dataform");
+}
+
+export function getConfigSettingsPath() {
+  return join(getConfigDirPath(), "settings.json");
+}
+
+export async function getConfigSettings(): Promise<IConfigSettings> {
+  try {
+    if (await promisify(exists)(getConfigSettingsPath())) {
```
BenBirt: use `fs-extra`, it's a bajillion times better

cli/config.ts
```diff
@@ -0,0 +1,46 @@
...
+      return JSON.parse(
+        await promisify(readFile)(getConfigSettingsPath(), "utf8")
+      ) as IConfigSettings;
+    }
+  } catch (e) {
+    // If we can't get the config, just return the default.
+  }
+  return {};
```
BenBirt: i realize that it's already effectively the default, but can you add an explicit `allowAnonymousAnalytics: false` here (more as documentation than anything else)?

cli/config.ts
```diff
@@ -0,0 +1,46 @@
...
+export async function upsertConfigSettings(config: Partial<IConfigSettings>) {
+  try {
+    const existingConfig = await getConfigSettings();
+    const newConfig = { ...existingConfig, ...config };
+    if (!(await promisify(exists)(getConfigDirPath()))) {
+      await promisify(mkdir)(getConfigDirPath());
+    }
+    await promisify(writeFile)(getConfigSettingsPath(), `${JSON.stringify(newConfig, null, 4)}\n`);
+  } catch (e) {
+    // If we can't set the configuration, fail quietly.
```
BenBirt: not going to block on this, but this PR contains several places that silently swallow errors - that makes sense for end-users, but could make us debugging very confusing. it might be nice to have a `--debug` flag that means that certain errors are still logged

cli/index.ts
```diff
@@ -328,6 +333,10 @@ const builtYargs = createYargsCli({
         jsonOutputOption
       ],
       processFn: async argv => {
+        if (!argv.json && !argv.watch) {
```
BenBirt: why protect this behind `--json`? that seems quite confusing

cli/index.ts
```diff
@@ -216,6 +220,7 @@ const builtYargs = createYargsCli({
         }
       ],
       processFn: async argv => {
+        await maybeConfigureAnalytics();
```
BenBirt: you haven't added this to all commands, why? why not just put it in `yargswrapper.ts`?

cli/analytics.ts
```diff
@@ -0,0 +1,98 @@
...
+    });
+  }
+}
+
+export async function trackCommand(command: string) {
+  const config = await getConfigSettings();
+  if (!config.allowAnonymousAnalytics) {
+    return;
+  }
+  // Don't block execution on the result of the analytics call.
```
BenBirt: maybe i'm missing something, but isn't this comment wrong?

cli/yargswrapper.ts
```diff
@@ -26,7 +27,10 @@ export function createYargsCli(cli: ICli) {
       command.description,
       (yargsChainer: yargs.Argv) => createOptionsChain(yargsChainer, command),
       async (argv: { [argumentName: string]: any }) => {
+        const analyticsTrack = trackCommand(command.format.split(" ")[0]);
         const exitCode = await command.processFn(argv);
+        // These can take a while, so wait up to 2 seconds for them to finish.
+        await analyticsTrack;
```
BenBirt: there doesn't actually appear to be any timeout here, as the comment suggests

go.mod
```diff
@@ -5,5 +5,8 @@ go 1.13
 require (
 	github.com/go-stack/stack v1.8.0 // indirect
 	github.com/golang/protobuf v1.3.2
+	github.com/google/go-cmp v0.4.0 // indirect
```
BenBirt: is this stuff coming from something VSCode told you to install (this happened to George)? can you please revert these files?

BenBirt:
It would also be nice to have a beforeAll/afterAll in the CLI type in yargswrapper, so they we can cleanly extract that code.

cli/analytics.ts
lewish: Sorry yeah, I flipflopped on this!

cli/config.ts
lewish: I'm really trying to limit the number of packages we use here given that these are packages we actually publish.

cli/config.ts
lewish: This would change the behaviour, there are 3 states here - false, true, undefined. I explicitly check for undefined so I can't set it here (so I know if to ask or not).

cli/index.ts
lewish: I wanted to be careful here to not ask for terminal input on commands like `dataform compile --json`. So it's only on specific commands and dependent on some of those flags.

cli/index.ts
lewish: If you are using --json, you probably are doing something like `dataform compile --json > dump.json`. It would be really annoying if that asked you a question and messed up the stdout!

cli/analytics.ts
```diff
@@ -0,0 +1,98 @@
...
+      },
+      () => {
+        resolve();
+        // Silently fail on tracking errors.
+      }
+    );
+  });
+}
+
+async function getDockerStatus() {
```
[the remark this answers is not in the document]
lewish: Found a much nicer solution - we can just detect if the terminal is a TTY, and then we won't ask the question during CI, Docker, or when someone does something like `dataform compile | more` or `dataform compile --json > blah.json`

cli/config.ts
lewish: Added a loud comment. IMO we shouldn't store things as protos unless we need to send them over network? Easy to migrate it if we need to.

cli/config.ts
lewish: Ack, obviously swallowing was deliberate (analytics should be best effort). Of course it's easy enough to test locally and unswallow.

cli/index.ts
lewish: OK ignore this, see other comment about detecting interactive terminals. I've moved this to all commands in the wrapper.

cli/index.ts
lewish: See other comment, resolved.

cli/config.ts
BenBirt: i don't think i agree with that being the dividing line, I think protos have value in just being a nice clean way to define DTOs, but this case is fine.

cli/config.ts
BenBirt: - fs-extra has a total of 3 dependencies
- the user installs the dataform CLI once and then updates rarely, it's not a big deal (like it might be for @dataform/core)

cli/config.ts
BenBirt: OK, but doesn't that mean that users who are in some buggy state where the file can never be read (I don't know, weird environment or something), they keep being asked over and over?

PR #662, opened 2020-03-23
Incremental merge

Ekrekr:
Fixes https://github.com/dataform-co/dataform-co/issues/4357

[Design doc](https://docs.google.com/document/d/10iAXrnwFDyyJwMcs4LhLNDxCp878QECH7iiNuEJtW_A/edit#heading=h.f7jt8fcqsk6a)

This looks long, but a lot of it is relatively similar code between adapters and their tests.

* Added `uniqueKey` proto

* Added new merge statements for `insertInto`

* Updated/added appropriate integration tests

* Added docs updates

* Bumped version

Azure was being a pain so I haven't implemented, as following Lewis' instructions. Merge adapter is same as original.


content/docs/guides/incremental-datasets.md
```diff
@@ -111,6 +111,26 @@ This dataset may not exist in the warehouse yet. That's OK, because the `WHERE`
   Dataform Web.
 </div>
 
+## A merge modification
```
lewish: I think a concrete example of where you would use this would be useful. For example, if your rows for transactions get updated historically. Walk through an example where you have a config like:

```
config {
  type: "incremental",
  uniqueKey: "transaction_id",
}

select * from raw.transactions
${when(incremental(), "where last_modified > select max(last_modified) from ${self()}")}
```

core/adapters/bigquery.ts
```diff
@@ -89,4 +90,24 @@ export class BigQueryAdapter extends Adapter implements IAdapter {
   public dropIfExists(target: dataform.ITarget, type: string) {
     return `drop ${this.baseTableType(type)} if exists ${this.resolveTarget(target)}`;
   }
+
+  public insertInto(
```
lewish: R.e. my comment in the design doc, can we stick with insert when no uniqueKey is specified?

core/adapters/bigquery.ts
```diff
@@ -89,4 +90,24 @@ export class BigQueryAdapter extends Adapter implements IAdapter {
...
+    columns: string[],
+    query: string,
+    uniqueKey: string[]
+  ) {
+    return `
+merge ${this.resolveTarget(target)} T
+using (${query}) S
+on ${
+      uniqueKey && uniqueKey.length > 0
+        ? uniqueKey.map(uk => `T.${uk} = S.${uk}`).join(` and `)
```
lewish: uk > uniqueKey
and call uniqueKey uniqueKeys so you don't get the conflict :)

core/adapters/bigquery.ts
```diff
@@ -89,4 +90,24 @@ export class BigQueryAdapter extends Adapter implements IAdapter {
...
+        : `false`
+    }
+when matched then
+  update set ${columns.map(c => `${c} = S.${c}`).join(",")}
```
lewish: ditto: c > column

core/adapters/snowflake.ts
```diff
@@ -80,4 +81,24 @@ export class SnowflakeAdapter extends Adapter implements IAdapter {
       table.target
     )} as ${table.query}`;
   }
+
+  public insertInto(
```
lewish: afaict this is identical to bigquery. Can we re-use a function? I realize this code is all a mess, but a good opportunity to clean it up.

core/table.ts
```diff
@@ -192,6 +192,13 @@ export interface ITableConfig extends ITargetableConfig, IDocumentableConfig, ID
    * If configured, relevant assertions will automatically be created and run as a dependency of this dataset.
    */
   assertions?: ITableAssertions;
+
+  /**
+   * Unique keys for merge criteria for incremental tables.
+   *
+   * If configured, records with matching unique key(s) will be updated, rather than new rows being inserted.
+   */
+  uniqueKey?: string[];
```
lewish: should this be uniqueKeys?

core/adapters/sqldatawarehouse.ts
```diff
@@ -122,4 +122,12 @@ export class SQLDataWarehouseAdapter extends Adapter implements IAdapter {
      ) 
      as ${table.query}`;
   }
+
+  public insertInto(target: dataform.ITarget, columns: string[], query: string) {
```
lewish: Can you add a validation rule for sql data warehouse that throws an error during compilation if uniqueKeys is specified?

content/docs/guides/incremental-datasets.md
```diff
@@ -111,6 +111,26 @@ This dataset may not exist in the warehouse yet. That's OK, because the `WHERE`
...
+
+<div className="bp3-callout bp3-icon-info-sign" markdown="1">
+  Incremental merging requires <code>@dataform/core</code> version <code>1.4.21</code>.<br>
```
lewish: As discussed in DD, we shouldn't just change this behavior. We should only apply the new logic if the user is on a version >= 1.4.21.

That means gating all the logic for merge statements in the adapter SQL generation code based on the version I'm afraid :(

tests/integration/bigquery_project/definitions/example_incremental_merge.sqlx
```diff
@@ -0,0 +1,15 @@
...
+  (SELECT 3 AS ts,	21	AS id_1, 31 AS id_2, 'new' AS val) UNION ALL
+  (SELECT 4 AS ts,	22	AS id_1, 32 AS id_2, 'new' AS val)
+)
+
+config { type: "incremental", uniqueKey: ["id_1", "id_2"] }
+
+SELECT ts, id_1, id_2, val
+FROM example_data
+
+${ when(incremental(), `WHERE ts > 2`, `WHERE ts <= 2`) }
```
lewish: nice!

version.bzl
```diff
@@ -1,3 +1,3 @@
 # NOTE: If you change the format of this line, you must change the bash command
 # in /scripts/publish to extract the version string correctly.
-DF_VERSION = "1.4.18"
+DF_VERSION = "1.4.21"
```
lewish: This is definitely a minor version upgrade (We've not been following semver rules very well). 1.5.0

core/adapters/snowflake.ts
Ekrekr: Ah, I did that originally but they do have slightly different syntax, e.g. `merge into` rather than just `merge`. With BQ now having the partition filter I think it makes sense not to have a generic function

core/adapters/sqldatawarehouse.ts
Ekrekr: Have added to `core/utils`

content/docs/guides/incremental-datasets.md
Ekrekr: I'm confused, isn't this what I've done here already?

content/docs/guides/incremental-datasets.md
Ekrekr: I added a new section with an example, does read nicer

core/table.ts
```diff
@@ -397,7 +418,7 @@ export class Table {
   public assertions({ uniqueKey, nonNull, rowConditions }: ITableAssertions) {
     if (!!uniqueKey) {
       const indexCols = typeof uniqueKey === "string" ? [uniqueKey] : uniqueKey;
-      this.session.assert(`${this.proto.target.name}_assertions_uniqueKey`, ctx =>
+      this.session.assert(`${this.proto.target.name}_assertions_uniqueKeys`, ctx =>
```
lewish: Hmm, I wouldn't change this, someone might depend on this behaviour.

core/table.ts
```diff
@@ -192,6 +199,13 @@ export interface ITableConfig extends ITargetableConfig, IDocumentableConfig, ID
    * If configured, relevant assertions will automatically be created and run as a dependency of this dataset.
    */
   assertions?: ITableAssertions;
+
+  /**
+   * Unique keys for merge criteria for incremental tables.
+   *
+   * If configured, records with matching unique key(s) will be updated, rather than new rows being inserted.
+   */
+  uniqueKeys?: string[];
```
lewish: I'm sorry to do this to you, but I'm just realizing we already use `uniqueKey` for assertions. This kind of sucks if they are different. Maybe we should leave this as it was (sorry!!).

core/utils.ts
```diff
@@ -135,6 +140,15 @@ export function validate(compiledGraph: dataform.ICompiledGraph): dataform.IGrap
         const message = `Invalid value for sqldatawarehouse distribution: "${distribution}"`;
         validationErrors.push(dataform.ValidationError.create({ message, actionName }));
       }
+
+      if (action.uniqueKeys) {
```
lewish: This is inside the wrong IF statement, you don't care about if the distribution is set or not (line 132).

content/docs/guides/incremental-datasets.md
lewish: Sorry weird place to put comment. - The DD suggested we only use the new behaviour (in code) if the user updates.

However, now that we are using the old insert behaviour by default, this release should be a no-op for all existing projects, and we only use merge once they set a uniqueKey value.

So, ignore my comment, we can leave as is :)

core/table.ts
Ekrekr: That's ok, one global find and replace

core/utils.ts
Ekrekr: Oops, thanks

PR #663, opened 2020-03-24
Some tweaks to ts_gprc proto gen to handle closed-source services

lewish:
- Adds a way to provide a list of root paths for resolving to imports
- Removes the default root import behaviour. This was causing all kind of weirdness and doesn't seem like we depend on it currently (or should).


lewish:
> It might be worth renaming `root_paths` and/or `root`, it's not immediately obvious from the API what the difference means.

Have added better descriptions, changed root to root-name, I think that helps...

PR #668, opened 2020-03-25
Adds the 'tablePrefix' to the projectConfig

GJMcGowan:
open source part of #4769

I managed to test it all the way through with the frontend and it seemed to work

### DIFF
--- a/core/session.ts
+++ b/core/session.ts
@@ -367,6 +367,12 @@ export class Session {
       );
     }
 
+    if (!!this.config.tablePrefix) {
+      this.prependTablePrefix(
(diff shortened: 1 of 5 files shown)


core/session.ts
```diff
@@ -463,6 +469,32 @@ export class Session {
     });
   }
 
+  private getTablePrefixWithUnderscore() {
+    return !!this.config.tablePrefix ? `${this.config.tablePrefix}_` : "";
+  }
+
+  private prependTablePrefix(actions: IActionProto[]) {
```
lewish: This code is basically the same as the appendSuffix code, can we put this all into one method plz?

PR #672, opened 2020-03-27
Fix run cache when downstream is supposed to run if upstream changed

faiyaz26:
Fixes: https://github.com/dataform-co/dataform-co/issues/4902

### DIFF
--- a/tests/integration/bigquery.spec.ts
+++ b/tests/integration/bigquery.spec.ts
@@ -158,9 +158,7 @@ suite("@dataform/integration/bigquery", ({ after }) => {
       name: 'dataform-integration-tests.df_integration_test.depends_on_example_view',
       type: 'view',
       target: testTarget,
-      query: '\n' +
-        '\n' +
-        'select * from `dataform-integration-tests.df_integration_test.example_view`'
+      query: 'select * from `dataform-integration-tests.df_integration_test.example_view`'
     })
     compiledGraph.targets.push(testTarget)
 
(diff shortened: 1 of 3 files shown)


tests/integration/bigquery.spec.ts
```diff
@@ -143,6 +143,25 @@ suite("@dataform/integration/bigquery", ({ after }) => {
 
     // run cache assertions
 
+    const testTarget = {
+      schema: 'df_integration_test',
+      name: 'depends_on_example_view',
+      database: 'dataform-integration-tests'
+    }
+    compiledGraph.tables.push({
```
BenBirt: can you just add this as a file instead of inserting it into the compilation graph? otherwise this is bound to get out of sync with the compilation logic.

tests/integration/bigquery.spec.ts
```diff
@@ -67,7 +67,25 @@ suite("@dataform/integration/bigquery", ({ after }) => {
     ]);
 
     // Run the project.
-    let executionGraph = await dfapi.build(compiledGraph, {}, credentials);
+
+    const actionsToRun = [
+      ...compiledGraph.tables
+        .filter(
+          table => table.name !== "dataform-integration-tests.df_integration_test.depends_on_example_view"
```
BenBirt: why not just run it? for that matter, we already have actions that depend on `example_view`, why not get rid of it entirely and use one of those?

PR #673, opened 2020-03-30
Fix tslint warnings. Add tslint-config-security.

BenBirt:


api/commands/run.ts
```diff
@@ -16,16 +16,7 @@ const isSuccessfulAction = (actionResult: dataform.IActionResult) =>
 
 export function run(graph: dataform.IExecutionGraph, credentials: Credentials): Runner {
   const dbadapter = dbadapters.create(credentials, graph.projectConfig.warehouse);
-  const runner = Runner.create(dbadapter, graph);
-  const executeAndCloseDbAdapter = async () => {
```
GJMcGowan: Why remove this bit? Seems important.

api/commands/run.ts
```diff
@@ -76,21 +67,24 @@ export class Runner {
         this.cancel();
       }, this.graph.runConfig.timeoutMillis);
     }
-    return this.resultPromise();
```
GJMcGowan: wat, this is super weird

api/commands/run.ts
BenBirt: I have just moved this code into execute()/result(), so that we don't have unhandled promises.

api/commands/run.ts
BenBirt: you mean removing this? or what?

my changes here constitute some API cleanup, driven really by tslint warnings.

GJMcGowan: I mean returning `this` - I've never seen that done before.

PR #676, opened 2020-03-31
Fix documentation push

GJMcGowan:
Fixes #4903

Restores functionality removed in https://github.com/dataform-co/dataform/commit/8bc539e3ec1a5fd735ef26bb8afdcaab9d518147

The issue is that:
* We call `setMetadata` which internally calls `getMetadata`
* So they are both added to the promise pool (which only has room for 16 promises)
* The promise pool fills up with `setMetada` calls, no room to call `getMetadata` therefore the `setMetadata` function never returns

Solution is to not put `getMetadata` in the promise pool. It is only called in volume from within `setMetadata`, which is already rate limited. 

Verified against Livup's warehouse with their code.


api/commands/run.ts
```diff
@@ -300,6 +300,12 @@ export class Runner {
       actionResult.status = dataform.ActionResult.ExecutionStatus.SUCCESSFUL;
     }
 
+    if (
+      action.actionDescriptor &&
+      actionResult.status === dataform.ActionResult.ExecutionStatus.SUCCESSFUL
+    ) {
```
BenBirt: do you think we should put this behind a temporary `RunConfig` flag (defaulting to running this code), so that we can turn it off if necessary?

api/dbadapters/bigquery.ts
```diff
@@ -296,27 +297,20 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
-              return null;
-            }
-            // otherwise throw the error as normal
-            throw e;
-          }
-        }
-      })
-      .promise();
-    return metadataResult && metadataResult[0];
+    try {
```
BenBirt: since we now have this method allowed for public callers (which is fine), i do think it should still live inside the promise pool. however, i would suggest extracting a private method (e.g. `getMetadataOutsidePromisePool` or something), which both `setMetadata` and `getMetadata` call, and keep the promise pool logic for `getMetadata`.

PR #684, opened 2020-04-06
Specify application parameter for Snowflake

dwl285:
see docs here: https://docs.snowflake.com/en/user-guide/jdbc-configure.html#connection-parameters

These docs are for JDBC but I've been told by a solutions engineer at Snowflake that the same thing applies to the node.js driver (even though this specific bit isn't documented).

### DIFF
--- a/api/dbadapters/snowflake.ts
+++ b/api/dbadapters/snowflake.ts
@@ -212,7 +212,8 @@ async function connect(snowflakeCredentials: dataform.ISnowflake) {
           password: snowflakeCredentials.password,
           database: snowflakeCredentials.databaseName,
           warehouse: snowflakeCredentials.warehouse,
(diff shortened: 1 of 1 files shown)


dwl285:
This is the context. This person is not a software engineer though, so I think it's fair to say we should be sceptical of his information.
image

I've checked the file you linked to, and his example (`jsTreatIntegerAsBigInt`) is mentioned. So yeah, I have no reason to believe you're wrong. I'll chase my contact at Snowflake with your points here and see if he can help us get to the bottom of this.

BenBirt:
Closing this in favour of https://github.com/dataform-co/dataform/pull/696 and https://github.com/snowflakedb/snowflake-connector-nodejs/issues/100.

PR #687, opened 2020-04-07
Don't set with no schema binding in PG

### LINKED ISSUE #686: postgres bug: `with no schema binding` causes SQL error on postgres view creation
as far as I can tell, postgres doesn't support the "with no schema binding" option
https://www.postgresql.org/docs/9.2/sql-createview.html
but redshift does:
https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_VIEW.html

image

---
Moved from dataform-co/dataform-co#4933
/cc @dwl285

lewish:
Fixes #686

### DIFF
--- a/tests/api/api.spec.ts
+++ b/tests/api/api.spec.ts
@@ -481,7 +481,7 @@ suite("@dataform/api", () => {
       });
       const testState = dataform.WarehouseState.create({});
       const expectedSQL = [
-        'create or replace view "schema"."redshift_view" as query',
+        'create or replace view "schema"."postgres_view" as query',
       ];
 
       const builder = new Builder(testGraph, {}, testState);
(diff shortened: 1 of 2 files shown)


tests/api/api.spec.ts
```diff
@@ -463,6 +463,42 @@ suite("@dataform/api", () => {
...
+        .to.have.lengthOf(1);
+
+      executionGraph.actions.forEach((action, index) => {
+        expect(action.tasks.length).greaterThan(0);
+
+        const statements = action.tasks.map(item => item.statement);
+        expect(statements).includes(expectedSQL[index]);
+      });
+    });
+
```
Ekrekr, line 501: Why the low core version?

lewish: Just using the same as most of the file, although this has no effect here anyway.
We need to refactor these tests quite fundamentally I think...

PR #694, opened 2020-04-09
Navigator sqlx protos

Ekrekr:
Small PR for protos. Does everything look right?

### DIFF
--- a/core/session.ts
+++ b/core/session.ts
@@ -63,8 +63,12 @@ function mapColumnDescriptionToProto(
   const columnDescriptor: dataform.IColumnDescriptor[] = description.description
     ? [
         dataform.ColumnDescriptor.create({
+          path: currentPath,
           description: description.description,
-          path: currentPath
+          displayName: description.displayName,
+          dimensionType: mapDimensionType(description.dimension),
+          aggregation: mapAggregation(description.aggregation),
+          expression: description.expression
         })
       ]
     : [];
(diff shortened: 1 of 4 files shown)


core/session.ts
```diff
@@ -81,6 +85,29 @@ function mapColumnDescriptionToProto(
   );
 }
 
+function mapAggregation(aggregation: string) {
+  switch (aggregation) {
```
BenBirt, line 89: it's so sad we have to do this manually, but protobufjs :(

core/session.ts
```diff
@@ -81,6 +85,29 @@ function mapColumnDescriptionToProto(
...
+    case "sum":
+      return dataform.ColumnDescriptor.Aggregation.SUM;
+    case "distinct":
+      return dataform.ColumnDescriptor.Aggregation.DISTINCT;
+    case "derived":
+      return dataform.ColumnDescriptor.Aggregation.DERIVED;
+    default:
+      return null;
```
BenBirt, line 97: IMO this should throw

core/session.ts
```diff
@@ -81,6 +85,29 @@ function mapColumnDescriptionToProto(
...
+  }
+}
+function mapDimensionType(dimensionType: string) {
+  switch (dimensionType) {
+    case "category":
+      return dataform.ColumnDescriptor.DimensionType.CATEGORY;
+    case "timestamp":
+      return dataform.ColumnDescriptor.DimensionType.TIMESTAMP;
+    default:
+      return null;
```
BenBirt, line 107: ditto

core/session.ts
Ekrekr: Nice idea

core/session.ts
Ekrekr: ditto

PR #695, opened 2020-04-14
Remove old Executed* protobufs.

BenBirt:


BenBirt:
These have been deprecated for a long while now actually. The code compiles (as of https://github.com/dataform-co/dataform-co/pull/5172), so nothing should break - if it does, then we wrote some very bad code!

PR #697, opened 2020-04-14
Fix the validator to take account of tablePrefix when validating schedules

GJMcGowan:


api/commands/validate.ts
```diff
@@ -77,10 +80,12 @@ export function validateSchedules(
...
           .map(operation => operation.target.name)
       );
+
       schedule.options.actions.forEach(action => {
-        if (!allActionNames.includes(action)) {
+        const prefixedActionName = tablePrefix ? `${tablePrefix}_${action}` : action;
+        if (!allActionNames.includes(prefixedActionName)) {
           errors.push(
-            `Action "${action}" included on schedule ${schedule.name} doesn't exist in the project.`
+            `Action "${prefixedActionName}" included on schedule ${schedule.name} doesn't exist in the project.`
```
BenBirt, line 88:
```suggestion
            `Action "${prefixedActionName}" included in schedule ${schedule.name} doesn't exist in the project.`
```

version.bzl
```diff
@@ -1,3 +1,3 @@
 # NOTE: If you change the format of this line, you must change the bash command
 # in /scripts/publish to extract the version string correctly.
-DF_VERSION = "1.5.1"
+DF_VERSION = "1.5.2"
```
BenBirt, line 3: i don't think we need this? your changes aren't in `core`, so nobody needs to update their package.json.

PR #702, opened 2020-04-17
Big cleanup of lexing/parsing code.

BenBirt:
- all callers now use the new `SyntaxTreeNode` codepath, i.e. they all process a (basic) AST
- string literals are a new addition to the AST
- the formatter code is simpler (I promise it is, really) as a result of putting string literals into the AST
- string replacements that need to happen because we drop raw chunks of SQLX code into the javascript template string are now cleanly moved into the compiler code
- tests for string replacements are now deleted and replaced by tests that check compiled output against the contents of a file


tests/sqlx/lexer.spec.ts
```diff
@@ -1,92 +1,15 @@
-import { parseSqlx, SyntaxTreeNode, SyntaxTreeNodeType } from "@dataform/sqlx/lexer";
+import { SyntaxTreeNode, SyntaxTreeNodeType } from "@dataform/sqlx/lexer";
 import { suite, test } from "@dataform/testing";
 import { expect } from "chai";
 
 suite("@dataform/sqlx", () => {
-  suite("outer SQL lexing", () => {
-    test("backslashes are duplicated, so that they act literally when included in a JavaScript template string", () => {
```
lewish, line 7: Where did all these tests go??

BenBirt: I have now brought them back in the form of compiler tests which check the result of compiling these against the full expected compiler output.

tests/core/backslashes-act-literally.js.test
```diff
@@ -0,0 +1,145 @@
+
```
lewish, line 1: OK I realize I'm u-turning here, but this seems not super useful...

Would be much easier to test this if you just compiled it and looked at the resulting action.
With this test, refactoring JS generation now is going to require rewriting these tests entirely, and it feels like we shouldn't care how it compiles, just that the result i the same. WDYT?

BenBirt: I disagree, I think there is definite value here.

For the same reason that it was valuable to have the original tests in `lexer.spec.ts` - because it's hard to make sure that we properly escape all the relevant escape characters, handle special text in strings properly, etc etc - these test cases are very valuable, because they run before any transpiled code is actually run.

Otherwise, when trying to track down - for example - how we are mishandling `\` characters, it's very difficult to actually run any tests, because compilation (to a `CompiledGraph`) just fails, with invalid transpiled JavaScript and nonsense output.

PR #705, opened 2020-04-21
Make config-defined-assertions have the same tags as their parent dataset.

BenBirt:


core/table.ts
```diff
@@ -374,6 +377,12 @@ export class Table {
     newTags.forEach(t => {
       this.proto.tags.push(t);
     });
+    if (!!this.uniqueKeyAssertion) {
+      this.uniqueKeyAssertion.tags(value);
+    }
+    if (!!this.mergedRowConditionsAssertion) {
+      this.mergedRowConditionsAssertion.tags(value);
+    }
```
Ekrekr, line 385: Don't these values get completely overridden @430, @442, and not used before then?

BenBirt: This is here to solve ordering concerns, if using the JS API. Consider the difference between calling `.assertions(...).tags(...)` vs `.tags(...).assertions(...)`. The code in `assertions()` solves the second case; at that point tags are already defined on the dataset and thus can be piped directly through to the assertion(s). However, this code solves the first case, where tags are defined *after* assertions, and thus must be piped through to them after they're created.

Ekrekr: Ah got it, cheers

PR #706, opened 2020-04-21
Add snowflake query validation API

lewish:


api/dbadapters/snowflake.ts
```diff
@@ -87,8 +87,38 @@ export class SnowflakeDbAdapter implements IDbAdapter {
     };
   }
 
-  public evaluate(statement: string): Promise<dataform.QueryEvaluationResponse> {
-    throw new Error("Unimplemented");
+  public async evaluate(statement: string): Promise<dataform.QueryEvaluationResponse> {
+    const getErrorLocation = (value: string) => {
```
BenBirt, line 91: we have so far been keeping all of this code inside `error_parsing.ts`

PR #707, opened 2020-04-21
Fixes issue with tablePrefix and ref/resolve

GJMcGowan:
From:
Screenshot 2020-04-21 at 11 45 29

To:
Screenshot 2020-04-21 at 11 45 23


BenBirt:
I feel like there should be a test for this!

GJMcGowan:
@BenBirt added a test

PR #708, opened 2020-04-22
Make expression default to column name

Ekrekr:
Tests were wrong!


core/session.ts
```diff
@@ -68,7 +68,8 @@ function mapColumnDescriptionToProto(
           displayName: description.displayName,
           dimensionType: mapDimensionType(description.dimension),
           aggregation: mapAggregation(description.aggregator),
-          expression: description.expression
+          // Expression defaults to name used for column.
+          expression: description.expression || currentPath[currentPath.length - 1]
```
BenBirt, line 72: i'm not sure we should do this here TBH. i suggest doing this inside the navigator code itself

PR #714, opened 2020-04-23
Add validation for extra properties in column descriptors

lewish:
This is all strongly typed, I think we can convert some other code to this pattern now.


core/column_descriptors.ts
```diff
@@ -0,0 +1,93 @@
...
+          path: currentPath
+        })
+      ];
+    }
+    const extraFields = Object.keys(description).filter(
+      key => !IRecordDescriptorFields.includes(key)
+    );
+    if (extraFields.length > 0) {
+      throw new Error(
+        `Unknown property "${extraFields[0]}" in column descriptor. Allowed properties are: ${IRecordDescriptorFields}`
```
BenBirt, line 38: Dan and I just agreed that we wouldn't fix this... It feels quite un-JavaScripty to do this, are you sure? I guess I'm happy if you are, though if we are going to do it I think we should be consistent with all `config` fields, as it'll be confusing if users start to expect us to do that and we don't

lewish: I'm happy to take this all the way... Having just written some configs, getting no feedback about if I'm doing it right I found very frustrating, particularly without good documentation for this alpha. 
We definitely get a fair few support requests where people have typos. We could downgrade the errors to not fully throw, but put them in the graph instead.

lewish: Have taken it all the way

scripts/lint
```diff
@@ -1,6 +1,3 @@
 #!/bin/bash
 
-set -e
-
-bazel build //:tslint
-./bazel-bin/tslint "ts/**/*.{ts,tsx}"
+bazel build @npm//tslint/bin:tslint && bazel-bin/external/npm/tslint/bin/tslint.sh --project .
```
BenBirt, line 3: nit: newline at EOF

tests/core/core.spec.ts
```diff
@@ -571,6 +572,25 @@ suite("@dataform/core", () => {
...
+            type: "table",
+            columns: {
+              dimension_column: {
+                unknownProperty: "unknown"
+              } as any
+            }
+          })
+          .query(_ => "select 1 as test");
+      } catch (e) {
+        return;
```
BenBirt, line 589: please add an `expect` asserting something about the expected Error

PR #715, opened 2020-04-24
Add alpha support for SSH tunnel connections to Redshift/Postgres

lewish:
Allows additional ssh tunnel configuration that can be used by Redshift/PG connections.

When provided, connects to the destination SSH server using the given private key, username, and tunnels a local port (randomly assigned via getPort) through the SSH connection to emerge on the other side.

Implementation was mainly followed from here - https://github.com/mscdex/ssh2/issues/67

- Refactors dbadapter usage a lot. The tunnel adds startup time to adapters, and in our code (and tests specifically) we recreate adapters all the time.
- Changes the constructor setup for dbadapters to allow async methods, which we need in this case.

#716


api/ssh_tunnel_proxy.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    const proxy = net.createServer(sock => {
+      sshClient.forwardOut(
+        sock.remoteAddress,
+        sock.remotePort,
+        destination.host,
+        destination.port,
+        (err, stream) => {
+          if (err) {
+            return sock.destroy();
```
BenBirt, line 28: does this silently swallow the error? it seems like we'd want that bubbled up through the db adapter, somehow, if at all possible, or at least logged somewhere.

api/ssh_tunnel_proxy.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    return new SSHTunnelProxy(sshClient, proxy, localPort);
+  }
+
+  // We need to keep track of connections so we can destroy them, as the proxy server won't close
+  // when there are keep-alive connections going on.
+  private connections = new Set<net.Socket>();
+
+  private constructor(
+    private sshClient: ssh2.Client,
```
BenBirt, line 55: these should all be `readonly`

api/ssh_tunnel_proxy.ts
```diff
@@ -0,0 +1,71 @@
...
+    });
+
+    await new Promise(resolve => sshClient.on("ready", () => resolve()));
+
+    return new SSHTunnelProxy(sshClient, proxy, localPort);
+  }
+
+  // We need to keep track of connections so we can destroy them, as the proxy server won't close
+  // when there are keep-alive connections going on.
+  private connections = new Set<net.Socket>();
```
BenBirt, line 52: `readonly`

protos/profiles.proto
```diff
@@ -41,3 +43,12 @@ message SQLDataWarehouse {
...
 }
+
+
+message SSHTunnel {
+  string private_key = 1;
+  string username = 2;
+  string password = 3;
+  string host = 4;
+  int32 port = 5;
+}
```
BenBirt, line 54: nit: newline at EOF

api/dbadapters/redshift.ts
```diff
@@ -11,21 +12,38 @@ interface ICursor {
...
+        host: jdbcCredentials.host,
+        port: jdbcCredentials.port
+      });
+      const queryExecutor = new PgPoolExecutor({
+        ...baseClientConfig,
+        host: "127.0.0.1",
+        port: sshTunnel.localPort
+      });
+      return new RedshiftDbAdapter(queryExecutor, sshTunnel);
+    } else {
```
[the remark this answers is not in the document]
lewish: Without the else this causes a no-shadowed-variable warning... :/

api/ssh_tunnel_proxy.ts
lewish: Looks like we can feed the error into the destroy call, so caller should get this propagated back.

PR #720, opened 2020-04-27
Update version to 1.6.3 for dfweb

lewish:
We are using this version number as a flag for a dfweb change - https://github.com/dataform-co/dataform-web-tracking/releases/tag/1.6.3


lewish:
Blocking on the next DF Web release.

PR #721, opened 2020-04-28
Add new protobufs for schedules-in-environments.

BenBirt:

### DIFF
--- a/protos/core.proto
+++ b/protos/core.proto
@@ -26,22 +26,62 @@ message ProjectConfig {
   string gcloud_project_id = 6;
 }
 
-message Environment {
-  string name = 1;
+message Environments {
+  message Environment {
+    string name = 1;
+    
+    message GitReference {
+      oneof git_reference {
+        string branch = 1;
+        string commit_sha = 2;
+      }
+    }
   
-  message GitReference {
-    oneof git_reference {
-      string branch = 1;
-      string commit_sha = 2;
+    GitReference git_reference = 2;
+    ProjectConfig config_override = 3;
+
+    message Schedule {
+      // Required fields.
+      string name = 1;
+      string cron = 2;
+      repeated string tags = 3;
+  
+      // Optional.
+      bool include_dependencies = 4;
+      bool full_refresh = 5;
+      bool disabled = 6;
+  
+      message NotificationRequirement {
+        repeated string channels = 1;
+  
+        enum CompletionStatus {
+          UNKNOWN = 0;
+          SUCCESS = 1;
+          FAILURE = 2;
+        }
+        repeated CompletionStatus statuses = 2;
+      }
+      repeated NotificationRequirement notify = 7;
+    }
+    repeated Schedule schedules = 4;  
+  }
+  repeated Environment environments = 1;
+
+  message NotificationChannel {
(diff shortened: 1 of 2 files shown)


protos/core.proto
```diff
@@ -26,22 +26,62 @@ message ProjectConfig {
...
-message Environments {
-  repeated Environment environments = 1;
+    message EmailNotificationChannel {
+      repeated string to = 1;
+    }
+    message SlackNotificationChannel {
+      string webhook_url = 1;
+    }
+    oneof channel {
+      EmailNotificationChannel email = 2;
```
GJMcGowan, line 80: Out of interest, why 2/3 here and not 1/2? Seems scoped

BenBirt: Fields inside a `oneof` are actually embedded in the `oneof`'s parent Message (a `oneof` is really just enforced at serialization/deserialization time). Here the parent Message's tag `1` is already used.

protos/core.proto
```diff
@@ -26,22 +26,62 @@ message ProjectConfig {
   string gcloud_project_id = 6;
 }
 
-message Environment {
-  string name = 1;
+message Environments {
+  message Environment {
```
lewish, line 30: I'm not a fan of this, it's going to make all our imports/Types horribly long for no obvious benefit?
Would it not be better to put this in a new namespace?

protos/core.proto
```diff
@@ -26,22 +26,62 @@ message ProjectConfig {
...
+      string name = 1;
+      string cron = 2;
+      repeated string tags = 3;
   
-  message GitReference {
-    oneof git_reference {
-      string branch = 1;
-      string commit_sha = 2;
+      // Optional.
+      bool include_dependencies = 4;
```
lewish, line 50: Why not put this stuff in a new type?
It's basically a stripped down RunConfig, I understand why we don't use that directly here but will be easier to keep in sync if we have seperate messages

protos/core.proto
```diff
@@ -26,22 +26,62 @@ message ProjectConfig {
...
+      bool full_refresh = 5;
+      bool disabled = 6;
+  
+      message NotificationRequirement {
```
lewish, line 54: ditto, writing `dataform.Environments.Environment.Schedule.NotificationRequirement` is going to be annoyingly verbose.
can we make Schedule a top level thing?
I think a new namespace would help a lot here

protos/core.proto
BenBirt: We can put it in a different namespace, that's fine with me, but I do think it was a mistake not to nest `Environment` inside `Environments` in the first place. 

For me at least, nesting messages inside the scope where they're used feels much cleaner, and that makes it more readable/understandable.

I actually wish that more of these protobufs were nested, I often find it hard to navigate our protobuf files and figure out what messages are used where.

BenBirt: Also I don't really think we should structure protobufs to suit any specific target language. If our TS types become verbose, there are workarounds for that, e.g. type aliases.

protos/core.proto
BenBirt: https://github.com/dataform-co/dataform/pull/722

PR #723, opened 2020-04-30
Rollup modules

lewish:
This completely changes how we build NPM packages.
- All published NPM packages live under `packages` in a structure that mirrors their real names
- We use rollup bundles to create fully contained packages, and remove hidden packages (no sqlx, protos, api any more, just core, cli)
- All code is now imported using the workspace name (df) instead of package names

Tests:
- We now generate package JSONs entirely automatically, locking versions to those in the root package.json and throwing if a required package isn't explicitly listed (woo!)
- Added an e2e test for the CLI where we actually install the CLI, run dataform init, create a file, run dataform compile, and check the response

Things I had to change:
- Added new tsconfig flags, esModuleInterop and allowSynthenticblah that seems to be required as rollup is built around es6 modules, this changes the way we import some stuff
- Changed the way we determine the compile subprocess script location so we can actually test the non bazel path inside a bazel context

Known issues:
- Rollup is not generating typescript definitions. This should be possible, but it's not trivial, and probably not currently required for cli or core packages


packages/@dataform/assertion_utils/BUILD
```diff
@@ -0,0 +1,35 @@
...
+    name = "json",
+    package_name = "@dataform/assertion_utils",
+    description = "Assertion utilities for Dataform.",
+    external_deps = [],
+    layers = [
+        "//:package.json",
+        "//packages/@dataform:package.layer.json",
+    ],
+    main = "index.js",
+    types = "index.d.js",
```
BenBirt, line 25: `index.d.ts` not `index.d.js`, or am i missing something? wait, i thought we couldn't generate types?

packages/gen_package_json.ts
```diff
@@ -0,0 +1,53 @@
+import * as fs from "fs";
+import * as yargs from "yargs";
+
```
BenBirt, line 3: supernit: these double empty lines are a bit odd, is prettier not cleaning this file up?

packages/rollup.config.js
```diff
@@ -0,0 +1,55 @@
...
+];
+
+const importsToBundle = ["df", /df\/.*$/, /^bazel\-.*$/];
+
+const checkImports = imports => {
+  const allowedImports = [...imports, ...knownBuiltIns].map(pattern => {
+    if (pattern instanceof RegExp) {
+      return pattern;
+    }
+    const normalized = pattern.replace(/[\\^$*+?.()|[\]{}]/g, "\\$&");
```
BenBirt, line 24: i have no idea what this is doing, can you leave a comment?

tests/cli/cli.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+    // Install packages manually to get around bazel sandbox issues.
+    const installChildProcess = execFile(npmPath, [
+      "install",
+      "--prefix",
+      projectDir,
+      "--cache",
+      npmCacheDir,
+      corePackageTarPath
+    ]);
+    installChildProcess.stdout.pipe(process.stdout);
```
BenBirt, line 39: it's probably worth being consistent and/or wrapping in a (local) utility function how we handle these child processes. (it makes sense to pipe output to this process, but we probably want to do it above too). that function can (a) set the correct pipes up, (b) wait for the child process to complete (with a zero exit code) and (c) return any output it produced

tests/cli/cli.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+    fs.writeFileSync(
+      path.join(projectDir, "definitions", "example.sqlx"),
+      `
+config { type: "table" }
+select 1 as test
+`
+    );
+
+    // Compile the project using the CLI.
+    const compileChildProcess = execFile(nodePath, [
```
BenBirt, line 56: here you execute node running the CLI, but for `init` above you fork into the CLI. not sure whether these need to be different, can they be consistent?

tests/cli/cli.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+      cliEntryPointPath,
+      "compile",
+      projectDir,
+      "--json"
+    ]);
+    compileChildProcess.stdout.pipe(process.stdout);
```
BenBirt, line 62: nit: i would suggest naming these according to the thing that is being run, not necessarily the command. e.g. `cliChildProcess` and `npmChildProcess` (above). IMO it makes it clearer what is being run, and avoids confusion (i thought for a few seconds that this meant the *actual* compile child process, but it doesn't of course).

api/commands/compile.ts
```diff
@@ -55,8 +55,17 @@ export async function compile(
...
   public static forkProcess() {
-    // Run the bin_loader script if inside bazel, otherwise don't.
-    const forkScript = process.env.BAZEL_TARGET ? "../vm/compile_loader" : "../vm/compile";
+    // Runs the worker_bundle script we generate for the package (see packages/@dataform/cli/BUILD)
+    // if it exists, otherwise run the bazel compile loader target.
+    const findForkScript = () => {
+      try {
+        const workerBundlePath = require.resolve("./worker_bundle");
+        return workerBundlePath;
+      } catch (e) {
```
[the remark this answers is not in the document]
lewish: This is what I tried originally, unfortunately I need to apply the node resolution logic from require.resolve, and the only signal it gives is to throw or not to throw. Otherwise I have to start replicating that method, which makes me feel more uncomfortable than this.

packages/@dataform/assertion_utils/BUILD
lewish: Woops! Thanks.
This doesn't use rollup as it's a self contained package, so we can keep them here.

tests/cli/cli.spec.ts
lewish: Kind of not relevant now, have refactored

packages/@dataform/assertion_utils/BUILD
lewish: (Once I can do .d.ts files, I'll make this all consistent)

packages/rollup.config.js
```diff
@@ -21,6 +21,7 @@ const checkImports = imports => {
     if (pattern instanceof RegExp) {
       return pattern;
     }
+    // If it's a string, turn it into a regex.
```
BenBirt, line 24: :P i knew what the intention was, i just have no idea what the replacer regex does. is it just escaping any regex characters?

tests/cli/cli.spec.ts
```diff
@@ -16,32 +16,21 @@ suite(__filename, () => {
...
-      "init",
-      "redshift",
-      projectDir,
-      "--skip-install"
-    ]);
-
-    await new Promise(resolve => {
-      initChildProcess.on("close", resolve);
-    });
+    await getProcessResult(
```
BenBirt, line 19: nit: why not push the executed file and list of args down into `getProcessResult`

lewish: This is my new thing - why do it that way? At some point I'll probably want to change from execFile to exec, or fork, or something else. ChildProcess is the interface here, it seems less likely I'll have to refactor this in future!

packages/index.bzl
```diff
@@ -0,0 +1,53 @@
+load("@npm_bazel_rollup//:index.bzl", "rollup_bundle")
+load("@build_bazel_rules_nodejs//:index.bzl", "pkg_npm")
+
+def pkg_json(name, package_name, description, version, external_deps = [], layers = [], main = "", types = ""):
+    native.genrule(
+        name = name,
+        srcs = layers,
+        tools = ["//packages:gen-package-json-bin"],
+        outs = ["package.json"],
+        cmd = """$(location //packages:gen-package-json-bin) \\
```
alexeagle, line 10: check out https://github.com/bazelbuild/rules_nodejs/blob/master/packages/typescript/checked_in_ts_project.bzl#L23-L39 - maybe this is simpler for you

PR #726, opened 2020-05-01
Add a 'hermetic' config option which must be set if run caching is turned on.

BenBirt:


core/assertion.ts
```diff
@@ -62,10 +63,18 @@ export class Assertion {
   private contextableQuery: AContextable<string>;
 
   public config(config: IAssertionConfig) {
-    checkExcessProperties(config, IAssertionConfigProperties, "assertion config");
+    checkExcessProperties(
+      (e: Error) => this.session.compileError(e),
```
lewish, line 67: Don't love this, is it also necessary? Each definition require() call basically does this wrapping AIUI, why not just throw?
Or, `try { ... } catch (e) { compileError(e) }`?

lewish: Would be good to understand intention here!

core/session.ts
```diff
@@ -506,6 +517,21 @@ export class Session {
...
+
+  private checkRunCachingCorrectness(actionsWithOutput: IActionProto[]) {
+    actionsWithOutput.forEach(action => {
+      if (action.dependencies?.length > 0) {
+        return;
+      }
+      if ([dataform.ActionHermeticity.HERMETIC, dataform.ActionHermeticity.NON_HERMETIC].includes(action.hermeticity)) {
+        return;
+      }
+      this.compileError(
```
lewish, line 529: Can we add a todo to link to docs in this error?

protos/core.proto
```diff
@@ -303,6 +309,12 @@ message Assertion {
   reserved 4, 5, 6;
 }
 
+enum ActionHermeticity {
+  UNKNOWN = 0;
```
lewish, line 313: Just so I understand - this is because with proto3 there is no way to tell the difference between set and default value?

lewish: I don't think you actually need to do all this for the purpose of validation. you could just validate this when setting the config, using hasOwnProperty (validation happens before we proto-ify these values)

core/assertion.ts
BenBirt: The problem with just throwing is that it short circuits the function, preventing any further code from running. This makes error messages really confusing: the compiler complains that `hermetic` is missing, when it really isn't.

I could move this check to the bottom of the function, or try/catch (though that seems messy when it's invoked in several places across the codebase).

However, IMO because `core` is essentially a transpiler, it should strongly prefer to output as many compile errors as it can; given that, I don't think we should be throwing uncaught Errors basically anywhere (unless the following code really cannot continue to run).

core/session.ts
BenBirt: I don't think it's a good idea to link to docs in compile errors... I think it'll be very brittle, we'll change URLs / move content around and forget to update links. (This is made even worse by the compiler being inside `core`, so there would be absolutely no way of updating links without pushing out a new version of `core`.)

protos/core.proto
BenBirt: Yep

core/assertion.ts
lewish: thank you, I understand now!

PR #728, opened 2020-05-01
Some fixes to integrate into closed source repo

lewish:
- Remove strange proto imports so they are the same pattern as closed source
- Upgrade rules_nodejs to latest and a couple of fixes
- Add a fake module_name'd ts_library and depend on it to make sure workspace imports work when importing this repo elsewhere

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -58,3 +58,13 @@ load("@bazel_gazelle//:def.bzl", "gazelle")
 # gazelle:proto package
 # gazelle:proto_group go_package
 gazelle(name = "gazelle")
+
+load("@npm_bazel_typescript//:index.bzl", "ts_library")
+
+# TODO: This is only here in order to workaround a bug in the way bazel resolves
+# workspace imports when in nested repositories, and can be removed once that is fixed.
+ts_library(
+    name = "modules-fix",
+    srcs = [],
+    module_name = "df",
+)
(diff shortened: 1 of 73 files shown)


BUILD
```diff
@@ -58,3 +58,13 @@ load("@bazel_gazelle//:def.bzl", "gazelle")
 # gazelle:proto package
 # gazelle:proto_group go_package
 gazelle(name = "gazelle")
+
+load("@npm_bazel_typescript//:index.bzl", "ts_library")
+
+# TODO: This is only here in order to workaround a bug in the way bazel resolves
```
BenBirt, line 64: can you link to an issue number, if it exists? (or some way for somebody to check whether the bug is fixed)

tools/css_typings.bzl
```diff
@@ -1,11 +1,13 @@
+load("@build_bazel_rules_nodejs//:providers.bzl", "DeclarationInfo", "JSEcmaScriptModuleInfo", "JSNamedModuleInfo")
```
BenBirt, line 1: doesn't seem like you're using these last two?

BUILD
lewish: I'm currently talking to someone on slack about how to fix this, don't have a bug yet but should soon, need to do some more testing a repros first.

PR #734, opened 2020-05-06
Check for excess properties on assertions property in table config block.

BenBirt:


BenBirt:
It should be on everything, but we have missed a couple of places (like here).

PR #735, opened 2020-05-06
More version syncing

lewish:
- Downgrade to react-router 3 (4/5 and 3 APIs are completely different)
- Have generated grpc services return the non interface protos types (as that's what they actually return) to make testing equality easier


lewish:
dataform-co/dataform-co#5707

PR #743, opened 2020-05-13
Cleanup global styles/theme config

lewish:
We had a weird mix of local and non local rules here, cleaning it up so we can do something that works for docs and the app.

- Make the colors a global scss sheet
- Scrap Root, it didn't really do anything, and instead expose an inject global styles method for all global styles
- Bundle all the global customized blueprint/theme scss including colors into one single CSS and inject it into the docs site
- Move root.scss into custom.global.css (as it is pretty much all global)


docs/pages/_document.tsx
```diff
@@ -18,15 +18,11 @@ export default class DefaultDocument extends Document {
...
-          />
-          <link
-            rel="stylesheet"
-            href="https://unpkg.com/@blueprintjs/select@3.8.1/lib/css/blueprint-select.css"
+            href="/blueprint.css"
           />
           <link rel="shortcut icon" href={"/static/images/favicon.ico"} type="image/png" />
         </Head>
-        <body>
+        <body className="light">
```
Ekrekr, line 25: We should add a dark mode button at some point

lewish: Hells yes

PR #744, opened 2020-05-13
Add card masonry style

Ekrekr:
Dynamically populates two columns.

*Gap between columns has been shrunk to be the same as between rows


components/card.scss
```diff
@@ -81,17 +94,20 @@
...
 }
 
-.fullWithCardContentContainer {
+.fullWidthCardContentContainer {
   padding: 0px;
 }
 
 .cardGrid {
   display: grid;
-  grid-row-gap: 20px;
```
GJMcGowan, line 90: Deleting these lines smushes the cards together in various places.

Screenshot 2020-05-13 at 16 52 54

components/card.scss
```diff
@@ -14,6 +14,19 @@
...
+  display: flex;
+}
+
+.masonryCardContainer {
+  display: inline-block;
+  margin: 10px;
+  width: 100%;
+  width: -moz-available;
+  width: -webkit-fill-available;
+  width: fill-available;
```
GJMcGowan, line 27: `fill-available` isn't widely supported enough to be used, the fact that even chrome required the prefixed version to work indicates that. 

I played around, and I think you can re-write like so:

```
.cardMasonry {
    columns: 2 200px;
    column-gap: 10px;
    margin: 10px;
}

.masonryCardContainer {
  display: inline-block;
  margin: 5px 0;
  width: 100%;
}
```

components/card.tsx
```diff
@@ -107,3 +110,16 @@ export const CardList = ({
     </div>
   );
 };
+
+export const CardMasonry = ({
+  minWidth = 200,
```
GJMcGowan, line 115: `minWidth` doesn't do anything I think

components/card.scss
GJMcGowan: https://caniuse.com/#search=fill-available

`Chrome does not yet unprefix stretch (aka fill/fill-available), because the CSSWG is not ready for that yet. See Chromium bug.`

Also I betcha Safari won't like it

GJMcGowan:
This didn't build locally for me so I merged master and then it did, so I pushed master

components/card.scss
Ekrekr: Ah I meant to revert that bit, adding in the second card type is a nicer alternative

components/card.scss
Ekrekr: Ah nice, thanks for help

Ekrekr: Ah nice, thanks for help

components/card.tsx
```diff
@@ -16,6 +17,7 @@ export const Card = ({
   className,
   minimal,
   fullWidth,
+  masonryCard,
```
lewish, line 20: It would be really nice to avoid this. I think you can do what you want via CSS selectors? It's going to be annoying to have to remember to add this prop to all cards in side CardMasonry

Ekrekr: Yeah but then I'd have to duplicate the current cardContent items. It doesn't work out badly I think, because whenever we display cards in a masonry they'll be dynamically inserted, so only one attribute set is needed. Can always update this later if that changes

PR #746, opened 2020-05-14
Add fix and test for undefined warehouse issue

### LINKED ISSUE #745: Dataform CLI doesn't pass the warehouse from the project config, tries to pull the value from argv which goes through as undefined.
https://github.com/dataform-co/dataform/blob/3192913889e6033bcf476c9c547f1b65e443359d/cli/index.ts#L498

lewish:
Fixes  #745


cli/index.ts
```diff
@@ -650,7 +650,7 @@ export function runCli() {
         const message = err && err.message ? err.message.split("\n")[0] : msg;
```
BenBirt, line 650:
```suggestion
        const message = err?.message ? err.message.split("\n")[0] : msg;
```

PR #748, opened 2020-05-14
Avoid some null pointers

lewish:


components/charts/common.tsx
```diff
@@ -50,7 +52,7 @@ export function generateAxis({
 }
 
 export function formatStringValue(value: React.ReactText, maxLength = 14): string {
-  const stringValue = value.toString();
+  const stringValue = String(value);
```
Ekrekr, line 55: Ah neat, because `"null"` actually is the value shown

PR #749, opened 2020-05-14
Add canonical targets

lewish:
A canonical target is a base64 encoded Target proto that would be the resolved target of the action if using only the contents of dataform.json and excluding all overrides set through environments, flags or otherwise.

Also open sources common/protos from closed source.


common/protos/proto_utils.ts
```diff
@@ -0,0 +1,29 @@
...
+    util.base64.decode(value, buf, 0);
+    return buf;
+  }
+
+  public static encode64(value: Uint8Array): string {
+    return util.base64.encode(value, 0, value.length);
+  }
+
+  public static decode<Proto>(protoType: new () => Proto, encodedValue: string): Proto {
+    if (!encodedValue || encodedValue === "" || encodedValue === "_") {
```
BenBirt, line 15: this stuff seems bug prone, do we need it?

common/protos/proto_utils.ts
```diff
@@ -0,0 +1,29 @@
...
+      return (protoType as any).create();
+    }
+    return (protoType as any).decode(ProtoUtils.decode64(encodedValue)) as Proto;
+  }
+
+  public static encode<Proto>(protoType: new () => Proto, value: Proto): string {
+    if (!value) {
```
BenBirt, line 22: ditto

core/session.ts
```diff
@@ -11,6 +11,11 @@ import { dataform } from "df/protos/ts";
 import { util } from "protobufjs";
 import { default as TarjanGraphConstructor, Graph as TarjanGraph } from "tarjan-graph";
 
+const DEFAULT_CONFIG = {
```
BenBirt, line 14: what's this needed for? would rather not have it if at all possible, just to prevent drift from the default `dataform.json`

protos/core.proto
```diff
@@ -223,6 +223,7 @@ message ColumnDescriptor {
 
 message Table {
   Target target = 4;
+  string canonical_target = 32;
```
BenBirt, line 226: what do you think about calling this `id` everywhere? that's effectively what it is

lewish: One thing from our conversation I don't think was made clear - we do need to decode this value, it won't always be one way. It makes me a bit uneasy having code that decodes what is apparently just an ID as a structured object, at least canonical_target implies what it is, or maybe it should actually be encoded_canonical_target

common/protos/proto_utils.ts
lewish: Removed, no longer needed

core/session.ts
lewish: Just preserving existing behaviour, refactoring code from below. I think this is mainly used in tests

lewish:
Now also open sources common/strings for doing sane-ish set equality semantics on target protos.

PR #750, opened 2020-05-14
Add a combined git reference field, deprecate the old one.

BenBirt:

### DIFF
--- a/protos/core.proto
+++ b/protos/core.proto
@@ -37,7 +37,12 @@ message Environments {
       }
     }
   
-    GitReference git_reference = 2;
+    oneof git_settings {
+      string git = 5;
+      // Deprecated. Please use 'git' instead.
+      GitReference git_reference = 2;
+    }
+
     ProjectConfig config_override = 3;
 
     message Schedule {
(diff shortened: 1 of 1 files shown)


protos/core.proto
```diff
@@ -37,7 +37,12 @@ message Environments {
       }
     }
   
-    GitReference git_reference = 2;
+    oneof git_settings {
+      string git = 5;
```
lewish, line 41: this doesn't feel right to me, it should really probably be `git_reference`, `reference`, `ref`, `git_ref` or something? Can we rename some stuff?

BenBirt: Amusingly I actually started with `git_ref` but changed my mind. Going with that...

protos/core.proto
```diff
@@ -37,7 +37,12 @@ message Environments {
       }
     }
   
-    GitReference git_reference = 2;
+    oneof git_settings {
+      string git_ref = 5;
+      // Deprecated. Please use 'ref' instead.
```
lewish, line 42: "git_ref"

PR #751, opened 2020-05-14
Update github readme.

josiehall:


josiehall:
@lewish Yep @G2H is still working on the DAG- will add it in when he's finished

PR #758, opened 2020-05-19
Add a BUILD rule for decrypting gcloud KMS secrets.

BenBirt:


lewish:
> Can we keep the test_credentials gitignore? As people will have files lying around from before this change

Forget about it

PR #764, opened 2020-05-22
Docs Q2: Code

lewish:
This is a partial PR, it does not fully implement full docs. Sending you this without the content changes which we are putting in another PR that I'll send out. once they are both approved, I'll merge and send a single PR.

As a result, you can't run really this right now (you can, but it will have old content and broken links), but the final PR you will be able to run when I send it!


docs/layouts/documentation.tsx
```diff
@@ -33,6 +35,17 @@ export default class Documentation extends React.Component<IProps> {
...
   public render() {
     const currentHeaderLinks = this.props.headerLinks || this.getHeaderLinks();
+    const tree = Tree.createFromIndex<IExtraAttributes>(this.props.index);
+    const current = tree.getChild(this.props.current.path);
+    const pathParts = this.props.current.path.split("/");
+    const parentPath = pathParts.slice(0, Math.max(pathParts.length - 1, 0)).join("/");
+    const parent = tree.getChild(parentPath);
+    const darkModeSaved =
+      typeof localStorage !== "undefined" &&
+      Boolean(localStorage.getItem("dataform-dark-mode") === "true");
```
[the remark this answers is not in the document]
GJMcGowan: maybe. At the very least I think the Boolean is irrevelant

docs/layouts/documentation.css
```diff
@@ -5,22 +5,33 @@
 }
 
 .sidebar {
-  min-width: 320px;
+  min-width: 360px;
+  max-width: 360px;
```
GJMcGowan, line 9: This reminds me of the code in @ghh's [other pr](https://github.com/dataform-co/dataform-co/pull/6085#discussion_r428730314) - looks like it can be `width: 360px` only. 

Might turn out I'm wrong though!

docs/layouts/documentation.css
```diff
@@ -31,14 +42,14 @@
...
+  border-left: var(--border);
   padding-left: 30px;
   padding-top: 30px;
-  background-color: #fff;
+  padding-right: 5px;
   flex-grow: 1;
-  border-left: 1px solid #e9e9e9;
   min-width: 500px;
-  padding-bottom: 800px;
+  padding-bottom: 300px;
```
GJMcGowan, line 51: You're defining all aspects of padding, so just collapse to:
```
  padding: 30px 5px 300px 30px;
```

docs/layouts/documentation.css
```diff
@@ -47,15 +58,38 @@
...
+}
+
+.subheader {
+  color: var(--textSecondary);
+  font-family: "Open Sans", sans-serif;
+  font-size: 18px;
+  line-height: 28px;
+}
+
+figcaption {
```
GJMcGowan, line 81: This style is targeting an html element instead of a class - can it be removed?

docs/layouts/base.css
```diff
@@ -20,6 +20,44 @@
...
+
+:not(pre) > code {
+  font-family: "Roboto Mono", monospace;
+  font-size: 15px;
+  -webkit-border-radius: 3px;
+  -moz-border-radius: 3px;
+  border-radius: 5px;
+  -moz-background-clip: padding;
+  -webkit-background-clip: padding-box;
+  background-clip: padding-box;
```
GJMcGowan, line 58: I feel like our sass compiler may be smart enough to add prefixing. Maybe I'm wrong though. It would be nice if we could add a preprocessor to add prefixes if not.

docs/global.css
```diff
@@ -116,10 +34,6 @@
...
 }
 
-.sectionBreaker {
-  margin: 30px 0px;
-  width: 80%;
-  border: 1px solid #e9e9e9;
-  height: 0px;
-  margin: 20px auto 20px;
+:global pre {
+  font-size: 0.95em;
```
GJMcGowan, line 38: Why `em` here and not elsewhere? There is a [discussion to be had](https://j.eremy.net/confused-about-rem-and-em/) as to whether we should be using (r)em's everywhere though.

docs/components/search/index.css
```diff
@@ -24,6 +24,12 @@
   left: calc(50% - 350px);
 }
 
+.searchBlock {
+  color: #6c6c6c;
+  font-weight: normal;
+  text-transform: none;
```
GJMcGowan, line 30: Do these styles do anything?

docs/components/page_links.tsx
```diff
@@ -11,11 +11,14 @@ interface IProps {
...
 export const PageLinks = (props: IProps) => (
-  <ul className={styles.pageLinks}>
-    {props.links.map(link => (
-      <li key={link.id}>
-        <a href={`#${link.id}`}>{link.text}</a>
-      </li>
-    ))}
-  </ul>
+  <>
+    {props.links?.length > 0 && <h5 className={styles.onThisPage}>On this page</h5>}
```
GJMcGowan, line 15: I wonder if we should just return null on the whole component if there are no links, or render a fallback or something. Seems weird to have an orphaned ul (as below)

docs/components/page_links.css
```diff
@@ -1,9 +1,32 @@
...
+.pageLinks > li > a {
+  color: var(--textSecondary);
+}
+
+.pageLinks > li > a:hover {
+  color: var(--textPrimary);
+}
+
+.onThisPage {
+  /* This the standard h5 so it should have been implemented elsewhere*/
```
GJMcGowan, line 21: So why was it implemented here? 😂

docs/components/footer.css
```diff
@@ -0,0 +1,20 @@
...
+  margin: 10px 0px 10px;
+}
+
+.sitemapMasonry > div {
+  display: inline-block;
+  width: 100%;
+  padding: 10px 0px;
+}
+
+.sitemapMasonry > div div {
```
GJMcGowan, line 13: wat

GJMcGowan: To clarify, I think you're selecting all divs within divs that are direct children of `sitemapMasonry`?

Styling like this is often quite brittle and prone to unexpected side effects (any divs within divs in this layout will have padding applied, often that will be wrong)

I would advise a more direct style on the class if possible

docs/components/code/highlight.global.scss
```diff
@@ -0,0 +1,198 @@
+div.hljs-filename {
```
GJMcGowan, line 1: I assume this is all custom styles for the syntax highlighter?

GJMcGowan: Why not use the syntax highlighter similarly to how we use it on the main app? [(e.g.)](https://github.com/dataform-co/dataform-co/blob/master/ts/app/components/code_viewers/code_view.tsx)

As in passing in styles?

tools/markdown-cms/tree.ts
```diff
@@ -61,4 +76,19 @@ export class Tree<T> implements ITree<T> {
       children: this.children.map(child => child.index())
     };
   }
+
+  public sort() {
+    this.children.sort((a: ITree<T>, b: ITree<T>) =>
```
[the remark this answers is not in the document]
lewish: This is just moved from another file. Not sure I can make it much more clear with code, but have added comments!

docs/components/search/index.css
GJMcGowan: I say this because I suspect that they don't, what happens if you remove them?

docs/global.css
GJMcGowan: 95% of the font size of the parent yes. My opinion isn't that strong either, just seems inconsistent.

docs/components/search/index.css
lewish: Yeah the default button font is tiny because of our custom buttons, this definitely does do something if I remove.

docs/pages/_docs.tsx
```diff
@@ -17,12 +18,27 @@ interface IQuery {
...
   current: ITree<IExtraAttributes>;
   version: string;
 }
 
+function MaybeCode(props: React.PropsWithChildren<{ className: string }>) {
+  if (props.className && props.className.startsWith("language")) {
+    const content = String(props.children).trim();
+    const lines = content.split("\n");
+    const firstLine = lines[0].trim();
+    const matches = firstLine.match(/(\/\/|\-\-)\s+(\S+\.\w+)/);
```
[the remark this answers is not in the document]
lewish: Oops missed, adding to other PR.

PR #766, opened 2020-05-22
Table row improvements

Ekrekr:
https://github.com/dataform-co/dataform-co/pull/6113


components/table.tsx
```diff
@@ -1,41 +1,38 @@
 import { HTMLTable } from "@blueprintjs/core";
-import * as React from "react";
-
 import * as styles from "df/components/table.css";
+import * as React from "react";
 
-export type TRow = Array<{
-  cells: Array<React.ReactElement | string>;
+export interface IRow {
+  cells?: Array<React.ReactElement | string>;
```
GJMcGowan, line 6: Why can you have a row without cells?

components/table.tsx
```diff
@@ -1,41 +1,38 @@
...
+export const Row = (
+  row: {
+    children: React.ReactElement;
+  } & IRow
+) => {
+  if (row.href || row.onClick) {
     return (
-      <a className={styles.tableRowLink} onClick={onClick} href={href}>
-        {children}
+      <a className={styles.tableRowLink} {...row}>
```
GJMcGowan, line 29: `...row` will add `colspans` to this tag, which is technically invalid. Easy way to remove is to extract colspans from row and then do `...rest`

components/table.tsx
```diff
@@ -55,12 +52,12 @@ export const Table = ({ columnWidths, headers, rows, className, condensed }: ITa
       </tr>
     </thead>
     <tbody>
-      {rows.map(({ href, cells, colspans }, rowIndex) => (
-        <Row key={rowIndex} href={href}>
+      {rows.map((row, rowIndex) => (
+        <Row key={rowIndex} {...row}>
           <>
-            {cells.map((cell, cellIndex) => (
+            {row.cells.map((cell, cellIndex) => (
```
GJMcGowan, line 58: if cells is optional now this should be `row?.cells`

components/table.tsx
Ekrekr: Ah you shouldn't be able to, nice spot

components/table.tsx
Ekrekr: Nice

PR #768, opened 2020-05-26
Docs Q2

lewish:
Previous PR + new content structure.
Redirects are in place for old URLs and I've gone through and checked all links with a link checker


lewish:
Sitemap - this is deliberate, although agree it's a lot, but it's so ctrl+f works.
Cards not sidebyside - can't for the love of me work out how to change this. something to do with the way columns/masonry works wher there are only 2 elements.
Old logo, yeayea, Guillaume is working on it :D
Weird icons - fixed.
Grey omnibar - fixed.

docs/global.css
```diff
@@ -37,3 +37,10 @@
 :global pre {
   font-size: 0.95em;
 }
+
+figcaption {
```
GJMcGowan, line 41: This won't do anything as it's targeting an html element, I believe you want the class?

lewish: No that is actually the intention! This is a real tag apparently, and Guillaume used it in the actual markdown files.

lewish: (I removed it before, that's why it messed up the spacing)

GJMcGowan: My god, you're right.

PR #774, opened 2020-05-27
Replace ExecutionAction.dependency_targets with ExecutionAction.transitive_inputs.

BenBirt:

### DIFF
--- a/api/commands/build.ts
+++ b/api/commands/build.ts
@@ -1,8 +1,9 @@
-import { Credentials } from "df/api/commands/credentials";
 import { prune } from "df/api/commands/prune";
 import { state } from "df/api/commands/state";
 import * as dbadapters from "df/api/dbadapters";
+import { actionsByStringifiedTarget } from "df/api/utils/graphs";
 import { adapters } from "df/core";
+import { IActionProto } from "df/core/session";
 import * as utils from "df/core/utils";
 import { dataform } from "df/protos/ts";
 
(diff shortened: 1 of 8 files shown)


api/utils/graphs.ts
```diff
@@ -0,0 +1,15 @@
+import { IActionProto } from "df/core/session";
+import { dataform } from "df/protos/ts";
+
+export function actionsByStringifiedTarget(compiledGraph: dataform.ICompiledGraph) {
```
lewish, line 4: The StringifiedMap class is open sourced now, and will arguably be more correct as it factors in ordering of keys, in case we somehow mess that up in future.

return new StringifiedMap(JSONObjectStringifier.create(), ...entries);

api/utils/graphs.ts
```diff
@@ -0,0 +1,15 @@
...
+  return new Map<string, IActionProto>(
```
lewish: The map type should be inferable, if you just omit the generic types altogether

api/commands/build.ts
```diff
@@ -13,88 +14,100 @@ export async function build(
...
   }
 
   public build(): dataform.ExecutionGraph {
-    if (utils.graphHasErrors(this.compiledGraph)) {
+    if (utils.graphHasErrors(this.prunedGraph)) {
       throw new Error(`Project has unresolved compilation or validation errors.`);
     }
 
-    const tableStateByTarget: { [targetJson: string]: dataform.ITableMetadata } = {};
+    const tableMetadataByTarget = new Map<string, dataform.ITableMetadata>();
```
lewish, line 46: ditto, may want to use stringified map here

api/commands/build.ts
```diff
@@ -104,16 +117,44 @@ export class Builder {
...
+  private getAllTransitiveInputs(
+    action: dataform.ITable | dataform.IOperation | dataform.IAssertion,
+    transitiveInputsByTarget: Map<string, dataform.ITarget[]>
+  ) {
+    const actionTargetKey = JSON.stringify(action.target);
+    if (!transitiveInputsByTarget.has(actionTargetKey)) {
+      let transitiveInputTargets = action.dependencyTargets;
+      for (const transitiveInputTarget of action.dependencyTargets || []) {
+        const transitiveInputAction = this.allActions.get(JSON.stringify(transitiveInputTarget));
+        if (
```
lewish, line 145: I think this could do with a comment. I think this is terminating the search if you hit a materialized object, but shouldn't we be checking that a dataform.Table is also not of type "view" ?

BenBirt: good catch!

PR #778, opened 2020-05-28
Allow configuring styles with SimpleCollapse.

BenBirt:


components/collapse.tsx
```diff
@@ -1,21 +1,26 @@
 import { Button, Collapse } from "@blueprintjs/core";
+import * as styles from "df/components/collapse.css";
 import { useState } from "react";
 import * as React from "react";
 
-import * as styles from "df/components/collapse.css";
-
 interface ISimpleCollapseProps {
   message: React.ReactNode;
+  style?: string;
```
lewish, line 8: this should be className

PR #783, opened 2020-05-29
Add controlled dialog

GJMcGowan:
Needed to pull controlled dialog into here to use it in navigator, also changes some props around

### DIFF
--- a/components/controlled_dialog.tsx
+++ b/components/controlled_dialog.tsx
@@ -7,13 +7,13 @@ import { Dialog } from "df/components/dialog";
 export interface IProps {
   title?: string;
   ref?: React.RefObject<ControlledDialog>;
-  button?: React.ReactElement<any>;
+  openButton?: React.ReactElement<Element>;
   onConfirm?: () => void;
   onClose?: (e: any) => void;
-  confirmButtonProps?: IButtonProps & React.ButtonHTMLAttributes<HTMLButtonElement>;
   cancelButtonText?: string;
+  confirmButtonProps?: IButtonProps & React.ButtonHTMLAttributes<HTMLButtonElement>;
   confirmHidden?: boolean;
-  ignore?: boolean;
+  disable?: boolean;
   width?: string;
 }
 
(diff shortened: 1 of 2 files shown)


components/controlled_dialog.tsx
```diff
@@ -0,0 +1,99 @@
...
+export interface IProps {
+  title?: string;
+  ref?: React.RefObject<ControlledDialog>;
+  button?: React.ReactElement<any>;
+  onConfirm?: () => void;
+  onClose?: (e: any) => void;
+  confirmButtonProps?: IButtonProps & React.ButtonHTMLAttributes<HTMLButtonElement>;
+  cancelButtonText?: string;
+  confirmHidden?: boolean;
+  ignore?: boolean;
```
lewish, line 16: This seems like a weird prop. shouldn't this be "disabled" or "openable" ?

components/controlled_dialog.tsx
```diff
@@ -0,0 +1,99 @@
...
+  width?: string;
+}
+
+interface IState {
+  isOpen: boolean;
+}
+
+export class ControlledDialog extends React.Component<IProps, IState> {
```
lewish, line 24: Hooks now plz! This is an easy one :)

components/controlled_dialog.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { CardActions } from "df/components/card";
+import { Dialog } from "df/components/dialog";
+
+export interface IProps {
+  title?: string;
+  ref?: React.RefObject<ControlledDialog>;
+  button?: React.ReactElement<any>;
+  onConfirm?: () => void;
+  onClose?: (e: any) => void;
+  confirmButtonProps?: IButtonProps & React.ButtonHTMLAttributes<HTMLButtonElement>;
```
lewish, line 13: If we are moving this here, let's maybe clean this up? All this props for the button is not very flexible, could replace with a `confirmButton` prop and let the user pass in a component when overriding. could do same for `cancelButton`

components/controlled_dialog.tsx
GJMcGowan: I was hoping you wouldn't say this, it's not easy because the files_list component attaches a ref to this component, and then uses that ref to call the showDialog function on this class, which sets state inside the class. It's gross but I didn't want to fix it. I guess I can fix it though.

components/controlled_dialog.tsx
GJMcGowan: I decided against this in the end. Practically every component that uses ControlledDialog wants to change at least one prop, and I don't like the idea of them all changing to pass in a whole button

PR #791, opened 2020-06-03
Add graph functions

GJMcGowan:
Connected to [this comment](https://github.com/dataform-co/dataform-co/pull/6248#discussion_r433682428)

### DIFF
--- a/api/utils/graphs.ts
+++ b/api/utils/graphs.ts
@@ -1,42 +1,21 @@
 import { JSONObjectStringifier, StringifiedMap } from "df/common/strings/stringifier";
-import { IActionProto } from "df/core/session";
 import { dataform } from "df/protos/ts";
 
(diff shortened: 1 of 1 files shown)


api/utils/graphs.ts
```diff
@@ -2,6 +2,31 @@ import { JSONObjectStringifier, StringifiedMap } from "df/common/strings/stringi
 import { IActionProto } from "df/core/session";
 import { dataform } from "df/protos/ts";
 
+const types = ["dataset", "assertion", "operation", "declaration"] as const;
+export type Type = typeof types[number];
+
+export interface IAction {
+  name: string;
+  type: Type;
```
BenBirt, line 10: as in the other PR, i still don't think we need this... `instanceof` guards should be enough.

TBH i think we should get rid of this interface altogether and just have an `allActions` function that concatenates all of the actions and returns the result.

PR #794, opened 2020-06-04
Add build file to //api/utils

GJMcGowan:


api/BUILD
```diff
@@ -4,13 +4,15 @@ load("@npm_bazel_typescript//:index.bzl", "ts_library")
 
 ts_library(
     name = "lib",
-    srcs = glob(["**/*.ts"]),
+    srcs = glob(["**/*.ts"],
+    exclude=["utils/**/*.*"]),
```
BenBirt, line 8: formatting here is weird, buildifier should be fixing it for you (`exclude` is an argument to the `glob` function)

PR #799, opened 2020-06-04
Contextual ops

Ekrekr:
Works with BQ and SQLDatawarehouse. Snowflake doesn't support multiple statements.

I'm not sure `Cursor` is the issue for redshift, or the one line limit, not really sure how to approach debugging it.


api/dbadapters/bigquery.ts
```diff
@@ -247,7 +247,7 @@ export class BigQueryDbAdapter implements IDbAdapter {
       const metadata = tableMetadataMap.get(action.target);
       const persistTable = dataform.PersistedTableMetadata.create({
         target: action.target,
-        lastUpdatedMillis: Long.fromString(metadata.lastModifiedTime),
+        lastUpdatedMillis: Long.fromString(metadata?.lastModifiedTime || "0"),
```
Ekrekr, line 250: When I ran the tests for the first time they threw because `metadata` was null; I think this is the right fix for that.

core/tasks.ts
```diff
@@ -49,6 +55,10 @@ export class Task {
     return this;
   }
 
+  public getStatement() {
+    return this.proto.statement;
+  }
```
Ekrekr, line 60: Once tasks are assembled you can't read the statement. I think you should be able to, so I added this, which made modifying `Tasks` post adapter assembly easier.

api/commands/build.ts
```diff
@@ -85,9 +86,23 @@ export class Builder {
...
+    }
+    if (table.bigquery?.hasOwnProperty("useContextualOps")) {
+      useContextualOps = table.bigquery.useContextualOps;
+    }
+    if (table.sqlDataWarehouse?.hasOwnProperty("useContextualOps")) {
+      useContextualOps = table.sqlDataWarehouse.useContextualOps;
+    }
+    if (useContextualOps) {
+      tasks = tasks.contextualize();
+    }
```
Ekrekr, line 105: I'm not sure this is the right place to do this. Is adding another parameter to `publishTasks` neater, or even adding the `useContextualOps` option to `runConfig`?

BenBirt:
I think there's a couple of high-level things here:
- let's not mark the issue as fixed, I think that if we can't do this for some warehouses we do need to have a reminder to update documentation when we launch this broadly
- I think there is some misunderstanding about the right place to put this option. We want this to be a temporary option a la: https://source.cloud.google.com/tada-analytics/github_dataform-co_dataform/+/master:protos/core.proto;l=23?q=ProjectConfig

Ekrekr:
> I think there's a couple of high-level things here:
> 
> * let's not mark the issue as fixed, I think that if we can't do this for some warehouses we do need to have a reminder to update documentation when we launch this broadly
> * I think there is some misunderstanding about the right place to put this option. We want this to be a temporary option a la: https://source.cloud.google.com/tada-analytics/github_dataform-co_dataform/+/master:protos/core.proto;l=23?q=ProjectConfig

- Have renamed the issue to not be BQ specific

- Have removed option to configure on a per file basis, instead only available via the project config

core/table.ts
```diff
@@ -225,6 +225,11 @@ export interface ITableConfig extends ITargetableConfig, IDocumentableConfig, ID
    * If configured, records with matching unique key(s) will be updated, rather than new rows being inserted.
```
BenBirt, line 225: please revert the changes in this file

core/tasks.ts
```diff
@@ -19,6 +19,12 @@ export class Tasks {
   public build() {
     return this.tasks.map(task => task.build());
   }
+
+  public contextualize() {
```
BenBirt, line 23:
```suggestion
  public concatenate() {
```

protos/core.proto
```diff
@@ -21,6 +21,7 @@ message ProjectConfig {
   string table_prefix = 11;
   
   bool use_run_cache = 10;
+  bool use_contextual_ops = 12;
```
BenBirt, line 24: I find this name a bit confusing. why not `use_single_query_per_action`?

api/commands/build.ts
```diff
@@ -85,9 +86,16 @@ export class Builder {
       throw new Error("Protected datasets cannot be fully refreshed.");
     }
 
+    let useContextualOps = false;
```
BenBirt, line 89: can we change this code to be consistent with `useRunCache` above?

tests/integration/bigquery.spec.ts
```diff
@@ -360,6 +360,29 @@ suite("@dataform/integration/bigquery", ({ before, after }) => {
...
+        postOps: ["postOps"],
+        target: { schema: "", name: "", database: "" }
+      };
+
+      const bqadapter = new BigQueryAdapter({ warehouse: "bigquery" }, "1.6.12");
+
+      const refresh = bqadapter
+        .publishTasks(table, { fullRefresh: true, useContextualOps: true }, { fields: [] })
+        .build();
+      expect(refresh.length).to.equal(1);
```
BenBirt, line 379: can we change these to assert equality against the actual query string?

tests/integration/sqldatawarehouse.spec.ts
```diff
@@ -152,23 +152,44 @@ suite("@dataform/integration/sqldatawarehouse", ({ before, after }) => {
...
+        postOps: ["postOps"],
+        target: { schema: "", name: "", database: "" }
+      };
+
+      const adapter = new SQLDataWarehouseAdapter({ warehouse: "sqldatawarehouse" }, "1.6.12");
+
+      const refresh = adapter
+        .publishTasks(table, { fullRefresh: true, useContextualOps: true }, { fields: [] })
+        .build();
+      expect(refresh.length).to.equal(1);
```
BenBirt, line 187: ditto

tests/integration/bigquery.spec.ts
```diff
@@ -360,6 +360,29 @@ suite("@dataform/integration/bigquery", ({ before, after }) => {
       expect(increment[increment.length - 2].statement).to.equal(table.postOps[0]);
       expect(increment[increment.length - 1].statement).to.equal(table.postOps[1]);
     });
+
+    test("contextual pre and post ops", async () => {
```
BenBirt, line 364: i'm not sure this is the right place for these tests; seems like they should be going in `api.spec.ts`

protos/core.proto
Ekrekr: is `use_single_task_per_action` is more correct? If we're changing task to query then I guess they're interchangeable

tests/integration/bigquery.spec.ts
Ekrekr: Sure, I was avoiding doing that in order to keep the tests more flexible

api/dbadapters/bigquery.ts
Ekrekr: > I don't understand why your change would affect this... I'm definitely a bit worried about this change, can you clarify under what condition(s) it happens?

Adding a new integration test file, the test will throw the first time from this line, saying that metadata is null and lastModifiedTime can't be read. Not needed to make tests pass, just figured I'd fix; have removed though so it can be investigated closer.

api/commands/build.ts
```diff
@@ -88,7 +89,20 @@ export class Builder {
 
     const tasks = table.disabled
       ? ([] as dataform.IExecutionTask[])
-      : this.adapter.publishTasks(table, this.runConfig, tableMetadata).build();
+      : this.adapter
+          .publishTasks(
+            table,
+            {
+              ...this.runConfig,
```
BenBirt, line 96: sorry, i think i wasn't clear enough. why not create `runConfig` *once*, in `build()`, before it's used anywhere, including both the overrides for `useRunCache` and `useSingleQueryPerAction`? then just set the `ExecutionGraph.runConfig` to that single `runConfig` (instead of repeatedly overriding the `runConfig`.

This would also be nicer because then `useSingleQueryPerAction` would be logged as part of the run (whereas right now it would not)

protos/core.proto
```diff
@@ -253,6 +256,7 @@ message Table {
   repeated string post_ops = 14;
   repeated string incremental_pre_ops = 28;
   repeated string incremental_post_ops = 29;
+  bool use_single_query_per_action = 33;
```
BenBirt, line 259: please remove this line

tests/api/api.spec.ts
```diff
@@ -696,6 +696,48 @@ suite("@dataform/api", () => {
         );
       });
     });
+
+    test("contextual pre and post ops", async () => {
```
BenBirt, line 700:
```suggestion
    test("useSingleQueryPerAction", async () => {
```

tests/api/api.spec.ts
```diff
@@ -696,6 +696,48 @@ suite("@dataform/api", () => {
...
+      [
+        {
+          warehouse: "bigquery",
+          statement: "preOps;create or replace table `database.schema.b` as query;postOps"
```
BenBirt, line 704:
```suggestion
          expectedQuery: "preOps;create or replace table `database.schema.b` as query;postOps"
```

tests/api/api.spec.ts
BenBirt, line 700: i would suggest moving this `test()` call *inside* the foreach loop. that way you can include the `warehouse` name as part of the test name, and you'll have singular tests for each warehouse.

api/commands/build.ts
Ekrekr: Ah I see, have done now

PR #801, opened 2020-06-05
provide documentation for using js with includes

will-misslin:
This PR simply expands the existing js file guide to include a section on referencing `includes` functions, macros and constants inside a .js file.

### DIFF
--- a/content/docs/guides/javascript/js-api.md
+++ b/content/docs/guides/javascript/js-api.md
@@ -116,6 +116,19 @@ The following methods and configuration options accept a function taking a `Cont
 - `preOps()`
 - `postOps()`
 
+## Referencing includes files
+
+Just like in `.sqlx` files, you can reference any `includes` function, macro or constant inside a `.js` file. However, the syntax differs slightly.
+
+For example, if you would like to reference `SERIVCE_NAME` and `SERVICE_ID` from the `includes/service` file, you must declare these constants using the JavaScript builtin `require` function.
+
+```js
+const {SERIVCE_NAME, SERVICE_ID} = require("../../includes/service");
+...
+```
+
+Once declared, these constants are available without any additional notation (i.e.. you don't need to wrap the constants in `${}`).
+
 ## Example: Dynamic dataset generation
 
 One of the most common use cases for using the JavaScript API is to perform a similar action repeatedly.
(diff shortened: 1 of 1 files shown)


content/docs/guides/javascript/js-api.md
```diff
@@ -116,6 +116,19 @@ The following methods and configuration options accept a function taking a `Cont
 - `preOps()`
 - `postOps()`
 
+## Referencing includes files
+
+Just like in `.sqlx` files, you can reference any `includes` function, macro or constant inside a `.js` file. However, the syntax differs slightly.
+
+For example, if you would like to reference `SERIVCE_NAME` and `SERVICE_ID` from the `includes/service` file, you must declare these constants using the JavaScript builtin `require` function.
```
BenBirt, line 123: The rules here are a little tricky:
- all top-level `includes` .js files are automatically imported, everywhere (i.e. if you have a top-level `includes.constants.js` file, a `constants` object is available for use in all files, whether .sqlx or .js, without any need for a `require`)
- all other `includes` .js files are not automatically imported, and wherever they are used, whether in .sqlx or .js files, they must be explicitly imported with `require()`

So it's more about the location of the file in `includes` than where you're using things exported by that file - does that make sense?

BenBirt, line 123: nit: mispelling of `SERVICE_NAME`, also would suggest you refer to `includes/service.js` just to make it clear it's a JS file you're referring to

content/docs/guides/javascript/js-api.md
```diff
@@ -116,6 +116,19 @@ The following methods and configuration options accept a function taking a `Cont
...
+
+```js
+const {SERIVCE_NAME, SERVICE_ID} = require("../../includes/service");
```
BenBirt, line 126: it's something we want to fix, but right now it's indeed frustrating that you need to put these `../` parts into `require` calls. i'd suggest though, only including a single one (thus assuming that this file is a top-level `definitions` file), or clarifying that this is a relative path between importer and importee.

(also this example may change as a result of the clarification I've included above, as this file - since it's top-level - *should* already be automatically imported)

content/docs/guides/javascript/js-api.md
will-misslin: Gotcha. Updated.

will-misslin: Well, that's embarrassing. Fixed!

PR #803, opened 2020-06-08
Vscode extension

GJMcGowan:
Fixes #6131

WIP on vscode extension for sqlx files


vscode/sql.grammar.json
```diff
@@ -0,0 +1,475 @@
+[
```
lewish, line 1: Pretty sure you can delete this file, and sql.plist, we don't actually use them

lewish: The SQLX grammar references the built in langauges directly

vscode/server.ts
```diff
@@ -0,0 +1,108 @@
...
+connection.onRequest("compile", async () => {
+  const _ = compileAndValidate();
+});
+
+documents.onDidSave(change => {
+  const _ = compileAndValidate();
+});
+
+async function compileAndValidate() {
+  const compileResult = await getProcessResult(exec("dataform compile --json"));
```
lewish, line 47: I don't quite understand how this works - you don't pass the project directory in, the PWD is probably not the one you want?

GJMcGowan: I don't understand how it works either. But it does 🤷‍♂️

PR #805, opened 2020-06-09
Contextual evaluations for BQ

Ekrekr:
Context: Evaluating tables rather than just queries will perform evaluations of exactly what will run; currently the evaluation could say something is fine, when in fact it will error on something that could have been checked during compilation. This update will also allow contextual pre and post ops validation.

To make this context work, evaluate instead needs to return an array of query evaluations so that each published table from the adapter can be evaluated separately.

I also renamed `QueryEvaluationResponse` to `QueryEvaluation` because it looks neater, particularly when used in conjunction with the closed source protos. I attached the query to the proto to make handling a query evaluation more clear once it's been made.

Closed source changes still pending.


api/commands/query.ts
```diff
@@ -30,11 +34,30 @@ export function run(
 
 export async function evaluate(
   dbadapter: dbadapters.IDbAdapter,
-  query: string,
-  compileConfig?: dataform.ICompileConfig
-): Promise<dataform.IQueryEvaluationResponse> {
-  const compiledQuery = await compile(query, compileConfig);
-  return await dbadapter.evaluate(compiledQuery);
+  queryStringOrTable: string | dataform.ITable,
```
BenBirt, line 37: i'm not sure this is the right place to do this.

IMO we should be moving away from `query.ts` (which is just a hacky way of injecting arbitrary SQL into the rest of a graph).

I'd expect to instead see us change the `IDbAdapter.evaluate` interface to do something like this. and when evaluating a table/assertion/operation, we'd call that interface (probably exposed somewhere higher level in `api/`), and this function would only really be used for 'selected query' evaluation.

(i would also expect the interface to change to allow all three action types, otherwise the caller has to branch on the action type which is being evaluated).

protos/core.proto
```diff
@@ -186,13 +186,14 @@ message QueryEvaluationError {
...
 
-message QueryEvaluationResponse {
+message QueryEvaluation {
   enum QueryEvaluationStatus {
     SUCCESS = 0;
     FAILURE = 1;
   }
   QueryEvaluationStatus status = 1;
   QueryEvaluationError error = 2;
+  string query = 3;
```
BenBirt, line 196: i'm not super clear on why this is useful? i would expect callers to `evaluate` single queries/tables/assertions/operations individually.

Ekrekr: This will be needed to make line numbers match in the closed source UI I believe. As the statement that is run against the warehouse will only be fully built in the backend (when the `evaluate` part of `@dataform/api` runs `publishTasks`), this will need to be send back.

Ekrekr:
Have made PR specific to BQ; as the API's `evaluate` isn't actually used anywhere currently (correct me if I'm wrong), it's OK to make a dramatic change to it.

components/dialog.tsx
```diff
@@ -14,6 +14,7 @@ export const Dialog = ({
     <BlueprintDialog
       className={styles.dialog}
       isCloseButtonShown={false}
+      title=""
```
Ekrekr, line 17: This prevents a console warning that I was getting spammed by; can move in to separate PR if you want.

core/adapters/bigquery.ts
```diff
@@ -16,7 +16,7 @@ export class BigQueryAdapter extends Adapter implements IAdapter {
   public publishTasks(
     table: dataform.ITable,
     runConfig: dataform.IRunConfig,
-    tableMetadata: dataform.ITableMetadata
+    tableMetadata?: dataform.ITableMetadata
```
Ekrekr, line 19: There's a slight issue with metadata; in order for the evaluation to be 100% accurate of evaluating what will run, metadata is required. This is annoying because acquiring metadata requires querying the entire warehouse (?), which obviously we don't want to happen every time a file is edited.

tests/integration/bigquery.spec.ts
```diff
@@ -308,6 +308,87 @@ suite("@dataform/integration/bigquery", ({ before, after }) => {
...
+            name: "example_incremental",
+            database: "dataform-integration-tests"
+          }
+        }),
+        { useSingleQueryPerAction: true }
+      );
+      expect(evaluation.status).to.equal(dataform.QueryEvaluation.QueryEvaluationStatus.SUCCESS);
+    });
+  });
+
```
Ekrekr, line 391: I haven't added a test for incremental yet

api/commands/build.ts
```diff
@@ -62,7 +62,7 @@ export class Builder {
       useSingleQueryPerAction:
         !this.prunedGraph.projectConfig?.hasOwnProperty("useSingleQueryPerAction") ||
         typeof this.prunedGraph.projectConfig?.useSingleQueryPerAction === "undefined"
-          ? this.prunedGraph.projectConfig.useSingleQueryPerAction
+          ? false
```
BenBirt, line 100: please merge with master, i made some fixes to this code recently

api/dbadapters/index.ts
```diff
@@ -26,7 +26,10 @@ export interface IDbAdapter {
       maxResults?: number;
     }
   ): Promise<IExecutionResult>;
-  evaluate(statement: string): Promise<dataform.IQueryEvaluationResponse>;
+  evaluate(
+    queryOrTable: string | dataform.ITable | dataform.IOperation | dataform.IAssertion,
```
BenBirt, line 30:
```suggestion
    queryOrAction: string | dataform.ITable | dataform.IOperation | dataform.IAssertion,
```

api/dbadapters/redshift.ts
```diff
@@ -68,12 +68,12 @@ export class RedshiftDbAdapter implements IDbAdapter {
     const statementWithExplain = `explain ${statement}`;
```
BenBirt, line 81: i'm not clear on how this code can still be compiling. you should need to change `evaluate` to match the new interface (and then evaluate the query(s) according to the type that is passed in)

api/dbadapters/bigquery.ts
```diff
@@ -91,22 +93,57 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
-        status: dataform.QueryEvaluationResponse.QueryEvaluationStatus.SUCCESS
-      });
-    } catch (e) {
-      return dataform.QueryEvaluationResponse.create({
-        status: dataform.QueryEvaluationResponse.QueryEvaluationStatus.FAILURE,
-        error: parseBigqueryEvalError(e)
-      });
+  public async evaluate(
+    query: string | dataform.Table | dataform.Operation | dataform.Assertion,
+    projectConfig?: dataform.IProjectConfig
```
BenBirt, line 103: AFAICT this is actually required now, otherwise the code won't work? seems like it should be required here and in the interface

api/dbadapters/bigquery.ts
```diff
@@ -91,22 +93,57 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+        if (query instanceof dataform.Table) {
+          executionTasks = coreAdapter
+            .publishTasks(
+              query,
+              { useSingleQueryPerAction: projectConfig?.useSingleQueryPerAction },
+              null
+            )
+            .build();
+        } else if (query instanceof dataform.Operation) {
+          executionTasks = query.queries.map(statement =>
```
BenBirt, line 120: OK, for this case i can see that we're going to need to return a list of evaluation responses, I'm fine with that API change!

tests/api/api.spec.ts
```diff
@@ -701,14 +701,6 @@ suite("@dataform/api", () => {
       {
         warehouse: "bigquery",
         expectedQuery: "preOps;create or replace table `database.schema.b` as query;postOps"
-      },
-      {
-        warehouse: "sqldatawarehouse",
```
BenBirt, line 731: does this test case need removing?

tests/integration/bigquery.spec.ts
```diff
@@ -346,6 +346,111 @@ suite("@dataform/integration/bigquery", { parallel: true }, ({ before, after })
     ]);
   });
 
+  suite("evaluate", async () => {
```
BenBirt, line 349: i suggest making this suite `parallel`

tests/integration/bigquery_project/dataform.json
```diff
@@ -2,5 +2,7 @@
   "warehouse": "bigquery",
   "defaultDatabase": "dataform-integration-tests",
   "defaultSchema": "df_integration_test",
-  "assertionSchema": "df_integration_test_assertions"
+  "assertionSchema": "df_integration_test_assertions",
+  "useRunCache": true,
```
BenBirt, line 6: please undo this

tests/integration/bigquery_project/dataform.json
```diff
@@ -2,5 +2,7 @@
...
+  "useSingleQueryPerAction": true
```
BenBirt, line 7: IMO i'd just set this in the tests themselves; that would be consistent with run caching

api/dbadapters/bigquery.ts
Ekrekr: Yeah pretty weird that that's allowed. Have updated to use new interface in other adapters, and added a throw if they're given things other than strings.

api/dbadapters/redshift.ts
Ekrekr: Yeah pretty weird that that's allowed. Have updated to use new interface in other adapters, and added a throw if they're given things other than strings.

tests/integration/bigquery.spec.ts
Ekrekr: Ooh

tests/integration/bigquery_project/dataform.json
Ekrekr: Nice

api/dbadapters/redshift.ts
Ekrekr: With TODOs

api/dbadapters/bigquery.ts
```diff
@@ -96,22 +98,57 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+        if (queryOrAction instanceof dataform.Table) {
+          executionTasks = coreAdapter
+            .publishTasks(
+              queryOrAction,
+              { useSingleQueryPerAction: projectConfig?.useSingleQueryPerAction },
+              null
+            )
+            .build();
+        } else if (queryOrAction instanceof dataform.Operation) {
+          executionTasks = queryOrAction.queries.map(statement =>
```
BenBirt, line 120: i think the last comment maybe was confusing? i mean that we should change (back to what you had originally) evaluate to return an array, so that cases like this make sense.

api/dbadapters/bigquery.ts
```diff
@@ -96,22 +98,57 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+  ) {
+    let executionTasks: dataform.ExecutionTask[];
+    if (typeof queryOrAction === "string") {
+      executionTasks = [dataform.ExecutionTask.create({ statement: queryOrAction })];
+    } else {
+      try {
+        const coreAdapter = new BigQueryAdapter(projectConfig, version);
+        if (queryOrAction instanceof dataform.Table) {
+          executionTasks = coreAdapter
+            .publishTasks(
```
BenBirt, line 113: just pinged you on chat, but this changes the semantics of evaluation a bit (sorry for just noticing this part).

i don't think we want to evaluate what is effectively an `ExecutionAction`, at least not in that part of the UI. I think a solution to that should be separate

api/dbadapters/bigquery.ts
Ekrekr: Sure, have updated

PR #808, opened 2020-06-09
Context ops docs

Ekrekr:
Fixes https://github.com/dataform-co/dataform-co/issues/6136

Final PR!


content/docs/guides/configuration.md
```diff
@@ -40,6 +40,32 @@ by changing the `defaultSchema` property to some other value. For example, to ch
...
+  set var = 'val';
+}
+select var as col;
+```
+
+is treated as if the `pre_operations` block wasn't there, becoming `declare var string; set var = 'val'; select var as col;`. However for Redshift or Snowflake, the pre_operations block would be executed as a separate query before the main block.
+
+To disable running all operations in the same context, place the following flag in your `dataform.json` file:
+
+```jsonkub
```
lewish, line 61: just json?

content/docs/guides/configuration.md
```diff
@@ -40,6 +40,32 @@ by changing the `defaultSchema` property to some other value. For example, to ch
...
+
+```sql
+pre_operations {
+  declare var string;
+  set var = 'val';
+}
+select var as col;
+```
+
+is treated as if the `pre_operations` block wasn't there, becoming `declare var string; set var = 'val'; select var as col;`. However for Redshift or Snowflake, the pre_operations block would be executed as a separate query before the main block.
```
lewish, line 57: nit: would be good to show where the create table expression get's injected.

content/docs/guides/configuration.md
```diff
@@ -40,6 +40,32 @@ by changing the `defaultSchema` property to some other value. For example, to ch
 
 [Assertions](assertions) are created inside a different schema as specified by the `assertionsSchema` property.
 
+### Running operations in context
+
+BigQuery and SQL Data Warehouse by default run all operations for a file in the same context by default. This functionality is not supported for Redshift or Snowflake.
```
lewish, line 45: This is useful for scripting, for example defining variables or UDFs in BigQuery before the create table statement.

PR #809, opened 2020-06-09
Fix run caching algorithm to check the status of all transitive inputs.

BenBirt:

high-level changes:
- change `state` to fetch state for *all* `CompiledGraph` actions (that have an output dataset), not just those included in the run, as this state may be required to run the run caching algorithm
- remove explicit requirement that `operations` never cache; instead replace this with `hermeticity` defaults on all actions (with `operations` defaulting to `NON_HERMETIC`)
- persist cache metadata rows for actions that are (a) hermetic, (b) included in the run, and (c) `SUCCESSFUL`/`CACHE_SKIPPED`
- overhaul BQ integration tests to (a) run separate test cases separately instead of one long test, and (b) run them all in parallel using schema suffixes


api/commands/run.ts
```diff
@@ -103,24 +118,29 @@ export class Runner {
...
-
-      // Currently, we don't support caching for operations (and any dependents)
-      await this.adapter.persistStateMetadata(
-        successfulActions.filter(action => action.type !== "operation")
+      await this.dbadapter.prepareStateMetadataTable();
+      await this.dbadapter.deleteStateMetadata(this.graph.actions);
+      await this.dbadapter.persistStateMetadata(
+        this.graph.actions.filter(executionAction => {
+          if (executionAction.hermeticity !== dataform.ActionHermeticity.HERMETIC) {
+            return false;
```
lewish, line 126: If an action goes from hermetic to non hermetic, will this cause issues due to cached state data that we never overwrite?

api/commands/run.ts
```diff
@@ -350,61 +374,94 @@ export class Runner {
...
-      lodash.isEqual(state.target, executionAction.target)
-    );
-    const tableMetadata = this.graph.warehouseState.tables.find(table =>
-      lodash.isEqual(table.target, executionAction.target)
-    );
-
-    if (!cachedState || !tableMetadata) {
+    const persistedTableMetadata = this.persistedStateByTarget.get(executionAction.target);
+    // tslint:disable-next-line: tsr-detect-possible-timing-attacks
+    if (hashExecutionAction(executionAction) !== persistedTableMetadata.definitionHash) {
```
lewish, line 394: timingSafeEqual is in open source now if you want it!

lewish: (Not bothered though, this doesn't seem like a risk here).

api/commands/build.ts
```diff
@@ -20,28 +19,31 @@ export async function build(
   runConfig: dataform.IRunConfig,
   dbadapter: dbadapters.IDbAdapter
 ) {
-  const prunedGraph = prune(compiledGraph, runConfig);
-  const stateResult = await state(prunedGraph, dbadapter);
```
lewish, line 24: As discussed - I think this could have a significant impact on performance of runs during development, e.g. projects with 500+ tables which will need to fetch a TON of metadata before doing anything.

Makes sense to remove the pruning code which was confusing, but perhaps the state computation should only fetch table metadata for the set of all transitive inputs + included actions to keep it snappy.

api/commands/run.ts
BenBirt: I don't think this is an issue, no.

If an action went to non-hermetic, we'll always run it regardless of what's in the cached state.

If it then returns to being non-hermetic, either: (a) if something else about its definition changed, the hash will break, or (b) if the table in the warehouse was modified, or any of its inputs was modified, it will not cache-skip, or (c) it must be in exactly the same state as it was when the cache record was originally written, and none of its inputs can have changed, thus it will (correctly) cache-skip.

PR #816, opened 2020-06-15
Contextual evaluations (for other warehouses, excl BQ)

Ekrekr:


api/dbadapters/index.ts
```diff
@@ -66,3 +68,74 @@ register("postgres", RedshiftDbAdapter);
...
+        );
+        if (concatenate) {
+          validationQueries.push({
+            query: concatenateQueries(tableQueries, queryModifier)
+          });
+        } else {
+          tableQueries.forEach(query => validationQueries.push({ query: queryModifier(query) }));
+        }
+      } else if (queryOrAction instanceof dataform.Operation) {
+        // TODO: The prefix method is a bit sketchy for operations. For example, after attaching the `eplain` prefix
```
Ekrekr, line 114: This issue already exists with the current method of evaluating operations, so is not a regression

api/dbadapters/sqldatawarehouse.ts
```diff
@@ -90,25 +95,30 @@ export class SQLDataWarehouseDBAdapter implements IDbAdapter {
...
-      await this.execute(`explain ${queryOrAction}`);
-      return [
-        dataform.QueryEvaluation.create({
-          status: dataform.QueryEvaluation.QueryEvaluationStatus.SUCCESS
-        })
-      ];
-    } catch (e) {
-      return [
-        dataform.QueryEvaluation.create({
+    // TODO: Using `explain` before declaring a variable is not valid in SQL Data Warehouse.
```
Ekrekr, line 98: It seems dry runs with variable declarations don't work. Know of any workarounds?

api/dbadapters/index.ts
BenBirt, line 114:
```suggestion
        // TODO: The prefix method is a bit sketchy for operations. For example, after attaching the `explain` prefix
```

BenBirt: does this not happen for all action types, though? e.g. if i have a `table` action that looks like this:

```
// some comment
select 1
```

api/dbadapters/index.ts
```diff
@@ -66,3 +68,74 @@ register("postgres", RedshiftDbAdapter);
...
+      } else if (queryOrAction instanceof dataform.Assertion) {
+        validationQueries.push({ query: queryModifier(queryOrAction.query) });
+      } else {
+        throw new Error("Unrecognized evaluate type.");
+      }
+    } catch (e) {
+      throw new ErrorWithCause(`Error building tasks for evaluation. ${e.message}`, e);
+    }
+  }
+  return validationQueries.filter(({ query }) => !!query);
```
BenBirt, line 140:
```suggestion
  return validationQueries.map(query => query.trim()).filter(({ query }) => !!query);
```

api/dbadapters/sqldatawarehouse.ts
BenBirt: I do not, but I wouldn't worry much about this edge case especially as we have extremely few Azure users

api/dbadapters/index.ts
```diff
@@ -66,3 +68,74 @@ register("postgres", RedshiftDbAdapter);
 register("redshift", RedshiftDbAdapter);
 register("snowflake", SnowflakeDbAdapter);
 register("sqldatawarehouse", SQLDataWarehouseDBAdapter);
+
+export function constructEvaluationFromQueryOrAction(
```
BenBirt, line 72:
```suggestion
export function collectEvaluationQueries(
```

Ekrekr:
> does this not happen for all action types, though?

Hmm, nice catch. Have moved comment to top of block and changed it a bit.

PR #817, opened 2020-06-15
Move to es2017 as default

lewish:
Currently compiling to es2015, which produces generator code instead of async/await, which makes the --async-stack-traces effectively a no-op.

Note that es2017 is pretty much supported in node 10 and above AFAICT.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -59,7 +59,7 @@ load("@bazel_gazelle//:def.bzl", "gazelle")
 # gazelle:proto_group go_package
 gazelle(name = "gazelle")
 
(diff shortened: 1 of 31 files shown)


lewish:
> Being supported in node 10 isn't the issue, the real issue is browser support. Our site already doesn't work for IE11 so this probably won't break it for them more. Googling around es2017 should be fine for all the other modern browsers. Just be aware there is a chance that some users will see breakages.
> 
> We should take a look at the size of our js bundle before and after - I would bet it's smaller afterwards!

Note that all this code will still go through webpack/babel which is what we should be using for client side targeting.

PR #819, opened 2020-06-16
Context eval tweaks

Ekrekr:
Removes wrapper code in evaluation response, and adds new line to joined task statements.


api/dbadapters/index.ts
```diff
@@ -71,8 +72,10 @@ register("redshift", RedshiftDbAdapter);
 register("snowflake", SnowflakeDbAdapter);
 register("sqldatawarehouse", SQLDataWarehouseDBAdapter);
 
+export type QueryOrAction = string | dataform.Table | dataform.Operation | dataform.Assertion;
```
lewish, line 75: nit: IMO use IQueryOrAction for types as well as interfaces, it makes it clear this is just a type, not a class

api/dbadapters/redshift.ts
```diff
@@ -77,27 +77,31 @@ export class RedshiftDbAdapter implements IDbAdapter {
...
+  public async evaluate(queryOrAction: QueryOrAction, projectConfig?: dataform.ProjectConfig) {
     const validationQueries = collectEvaluationQueries(queryOrAction, false, (query: string) =>
       !!query ? `explain ${query}` : ""
-    );
+    ).map((validationQuery, index) => ({ index, validationQuery }));
+    const validationQueriesWithoutWrappers = collectEvaluationQueries(queryOrAction, false);
+
     const queryEvaluations = new Array<dataform.IQueryEvaluation>();
-    for (const { query, incremental } of validationQueries) {
+    for (const { index, validationQuery } of validationQueries) {
```
lewish, line 87: This would be much faster if you run the queries in parallel, using Promise.all()

api/dbadapters/snowflake.ts
```diff
@@ -90,27 +90,31 @@ export class SnowflakeDbAdapter implements IDbAdapter {
...
+  public async evaluate(queryOrAction: QueryOrAction, projectConfig?: dataform.ProjectConfig) {
     const validationQueries = collectEvaluationQueries(queryOrAction, false, (query: string) =>
       !!query ? `select system$explain_plan_json($$${query}$$)` : ""
-    );
+    ).map((validationQuery, index) => ({ index, validationQuery }));
+    const validationQueriesWithoutWrappers = collectEvaluationQueries(queryOrAction, false);
+
     const queryEvaluations = new Array<dataform.IQueryEvaluation>();
-    for (const { query, incremental } of validationQueries) {
+    for (const { index, validationQuery } of validationQueries) {
```
lewish, line 100: Ditto

api/dbadapters/redshift.ts
Ekrekr: Hm, that will make pre, post and the main operation of the query all race, shouldn't the order be preserved?

api/dbadapters/index.ts
Ekrekr: I don't like that because you can't do some things with a type that you can do with an interface. Surely if we prefer using the `dataform.Table` protos now, that doesn't start with an `I` and acts as a type

Ekrekr:
Disable run caching tests with too many operations... azure decides to choose me as the deadlock victim!

PR #825, opened 2020-06-22
Add a method to the strings library

GJMcGowan:
Need to add a method for this https://github.com/dataform-co/dataform-co/pull/6799

### DIFF
--- a/common/strings/index.ts
+++ b/common/strings/index.ts
@@ -13,3 +13,7 @@ export function timingSafeEqual(a: string, b: string) {
 
   return result === 0;
 }
+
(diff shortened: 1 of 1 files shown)


common/strings/index.ts
```diff
@@ -13,3 +13,7 @@ export function timingSafeEqual(a: string, b: string) {
 
   return result === 0;
 }
+
```
lewish, line 16: Sorry but:
a) this is not tied in any way to open-source code
b) doesn't belong in the common libraries as it's dataform (web) specific!

PR #827, opened 2020-06-22
Remove depricated protos for comp/val errors

Ekrekr:
Related to https://github.com/dataform-co/dataform-co/pull/6803

Is `reserved` actually needed here?


lewish:
> Is `reserved` actually needed here?
Yup, as these are persisted in some cases

PR #830, opened 2020-06-24
Utils for coercing thrown values as errors

lewish:


common/errors/errors.ts
```diff
@@ -9,3 +9,24 @@ export class ErrorWithCause extends Error {
...
 }
+
+export function coerceAsError<T extends Error | any>(errorLike: T): T extends Error ? T : Error {
+  // If it's an error instance, return it.
+  if ((errorLike as any) instanceof Error) {
+    return errorLike as any;
+  }
+  // If it looks mostly like an error, return it.
+  if (!!errorLike.message && errorLike.stack) {
+    return errorLike as any;
```
BenBirt, line 20: propose constructing a real `Error` with all of the same fields as the `errorLike`, that way any `instanceof` checks further up the stack work nicely (and this function doesn't lie about its return type)

lewish: :+1:

PR #831, opened 2020-06-24
Deprecate validation errors

Ekrekr:


api/commands/compile.ts
```diff
@@ -50,11 +49,7 @@ export async function compile(
   const encodedGraphInBase64 = await CompileChildProcess.forkProcess().compile(compileConfig);
   const encodedGraphBytes = new Uint8Array(util.base64.length(encodedGraphInBase64));
   util.base64.decode(encodedGraphInBase64, encodedGraphBytes, 0);
-  const compiledGraph = dataform.CompiledGraph.decode(encodedGraphBytes);
-  return dataform.CompiledGraph.create({
-    ...compiledGraph,
-    graphErrors: validate(compiledGraph)
-  });
+  return dataform.CompiledGraph.decode(encodedGraphBytes);
```
BenBirt, line 52: my understanding is that you kept the UI code backwards-compatible with validation errors, is that right? if so, removing this part will mean that the backwards-compatible code never gets called. (that's fine with me, but an FYI to you)

core/session.ts
```diff
@@ -527,20 +539,147 @@ export class Session {
...
       }
       allCanonicalTargets.set(action.canonicalTarget, true);
     });
   }
 
+  private checkTableConfigValidity(tables: dataform.ITable[]) {
+    tables.forEach(table => {
+      // type
+      if (!!table.type && !TableType.includes(table.type as TableType)) {
+        const predefinedTypes = joinQuoted(TableType);
```
BenBirt, line 554: nit: just inline this?

BenBirt:
You should also mark the relevant field `[deprecated = true]`.

api/commands/compile.ts
Ekrekr: Ah yup, in that change I added a small bit of code where the compiled graph is retrieved that converts validation errors to compilation errors. I'll remove that soon after this.

PR #832, opened 2020-06-24
Enable CI for forked PRs

lewish:
Use the new dataform-public keychain and re-encrypt new low-risk secrets for integration testing databases/users. Have created new users and databases, and fixed a few things in integration tests for this.

Remove github default checks, and replace with two test and deploy triggers:
- Test - runs as dataform-public, can't do much except decode integration test credentials and use bazel cache, Triggers on ANY pull-request, internal or from a forked repo.
- Deploy - runs as tada-analytics, same as before, and deploys just the docs site currently and strictly only runs on the master branch after approval.


test_credentials/BUILD
```diff
@@ -6,30 +6,34 @@ gcloud_secret(
     name = "bigquery.json",
```
BenBirt, line 6: if everybody in the world can decrypt these secrets, why bother keeping them as encrypted files?

tests/integration/redshift.spec.ts
```diff
@@ -37,7 +37,11 @@ suite("@dataform/integration/redshift", ({ before, after }) => {
         dbadapter.execute(adapter.dropIfExists(assertion.target, "view"))
       )
     );
-    await dropFunctions.reduce((promiseChain, fn) => promiseChain.then(fn), Promise.resolve());
+    try {
+      await dropFunctions.reduce((promiseChain, fn) => promiseChain.then(fn), Promise.resolve());
+    } catch (e) {
+      // This seems to throw even if the tables don't exist.
```
BenBirt, line 43: is this `This seems to throw if the tables don't exist.`?

test_credentials/BUILD
lewish: I think there is a big difference in leaving plain unencrypted secrets on the indexed web from obfuscating them. We'll at least know when someone decodes them if they send a PR, as that's the only way to do it.

lewish:
> Question re: the Bazel cache - if anybody can now use it (or can escalate themselves to using it via Cloud Build), does that leave us at risk of people fetching things from the cache that they shouldn't be able to have access to? (e.g. secrets encoded using `tada-analytics` keychains)

Yup, currently. I'm planning to isolate these in buildfarm when we move to that.

PR #834, opened 2020-06-25
Overhaul run caching persistence code.

BenBirt:

- push queries inside `execute()` so that they have proper rate limiting etc
- push cancellation inside run cache persistence code, so that cancellation works correctly
- simplify `MERGE` query to an `INSERT` since all updateable entries must have previously been deleted from the cache
- cut `INSERT` statements into pieces small enough to fit inside BQ's max query size limit
- drop JSON-encoded `PersistedTableMetadata` protobuf column from cache state table (to save on `INSERT` query size)
- change cache state table key to be encoded with `JSONObjectStringifier`
- simplify `PersistedTableMetadata` (to save on `INSERT` query size)
- resurrect run caching integration test, this time allowing actions which should cache to *either* cache or execute successfully (to handle the case when run caching code encounters BQ rate limiting errors)

Also fix some CLI commands.


api/dbadapters/bigquery.ts
```diff
@@ -257,54 +248,82 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+CREATE TABLE IF NOT EXISTS \`${CACHED_STATE_TABLE_NAME}\` (
+  target STRING,
+  metadata_proto STRING
+)`,
+        options
+      );
+      // Before saving any new data, delete all entries for 'allActions'.
+      await this.execute(
+        `
+DELETE \`${CACHED_STATE_TABLE_NAME}\` WHERE target IN (${allActions
```
lewish, line 275: It does seem a shame to loose the atomicity of merge, why change it, is it just to simplify?

api/dbadapters/bigquery.ts
```diff
@@ -257,54 +248,82 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
-          }
-          return tableMetadataByTarget.get(transitiveInput);
-        })
-      });
-
-      const targetName = `${action.target.database}.${action.target.schema}.${action.target.name}`;
-
-      return buildQuery(targetName, persistTable);
-    });
+    const targetStringifier = JSONObjectStringifier.create<dataform.ITarget>();
```
lewish, line 261: sure we don't want to just write a custom stringifier for Target? this will write longer than necessary and hard to read columns, is it worth just writing something custom that is similar to getActionName?

api/dbadapters/bigquery.ts
```diff
@@ -257,54 +248,82 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
-    WHEN MATCHED THEN
-      UPDATE SET metadata_json = S.metadata_json,
-          metadata_proto = S.metadata_proto;`;
-    await this.runQuery(updateQuery);
+      // Save entries for 'actionsToPersist'.
+      const transitiveInputMetadataByTarget = await this.computeTransitiveInputMetadataMap(
+        actionsToPersist
+      );
+      const valuesTuples = actionsToPersist
+        // If we were unable to load metadata for the action's output dataset, or for any of the action's
```
lewish, line 286: this implies to me that the first time we create a table, we won't write a cache entry for it. This doesn't seem like what we want to happen?

api/dbadapters/bigquery.ts
```diff
@@ -257,54 +248,82 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+            transitiveInputMetadataByTarget.has(action.target) &&
+            action.transitiveInputs.every(transitiveInput =>
+              transitiveInputMetadataByTarget.has(transitiveInput)
+            )
+        )
+        .map(
+          action =>
+            `('${targetStringifier.stringify(action.target)}', '${encodePersistedTableMetadata({
+              target: action.target,
+              lastUpdatedMillis: transitiveInputMetadataByTarget.get(action.target)
```
lewish, line 299: I'll assume that my previous comment isn't correct then, as we must be fetching metadata after the run completes in this case or this wouldn't make any sense

api/dbadapters/bigquery.ts
```diff
@@ -257,54 +248,82 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+              definitionHash: hashExecutionAction(action),
+              transitiveInputTables: action.transitiveInputs.map(transitiveInput =>
+                transitiveInputMetadataByTarget.get(transitiveInput)
+              )
+            })}')`
+        );
+      // We have to split up the INSERT queries to get around BigQuery's query length limit.
+      while (valuesTuples.length > 0) {
+        let valuesToInsert = valuesTuples.pop();
+        while (valuesTuples.length > 0 && valuesToInsert.length < MAX_QUERY_LENGTH * 0.95) {
```
lewish, line 310: If this ever goes wrong, I imagine it will cause us a major headache. Seems like it wouldn't be hard to change this code to compute size iff we added the next set of values to the query and unpop (or just peek instead of popping immediately).

api/utils/run_cache.ts
```diff
@@ -17,16 +16,6 @@ export const decodePersistedTableMetadata = (protoString: string) => {
...
-
-export const buildQuery = (targetName: string, table: dataform.PersistedTableMetadata) => {
-  const encodedProtoString = encodePersistedTableMetadata(table);
-  const query = `SELECT 
-          '${targetName}' AS target_name,
-          '${JSON.stringify(table.toJSON())}' AS metadata_json,
-          '${encodedProtoString}' as metadata_proto`;
-  return query;
+export const encodePersistedTableMetadata = (table: dataform.IPersistedTableMetadata) => {
+  return Buffer.from(dataform.PersistedTableMetadata.encode(table).finish()).toString("base64");
```
lewish, line 20: r.e. what you said, is it worth us just writing a quick test somewhere to check that protobuf.js is at least consistent with itself on how it encodes protos? e.g. if we give it plain ojects with differently ordered keys, that this method produces the same buffer.

I'm worried now that we refactor some code and invalidate everyone's cache

api/dbadapters/bigquery.ts
```diff
@@ -257,54 +248,82 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+              transitiveInputMetadataByTarget.has(transitiveInput)
+            )
+        )
+        .map(
+          action =>
+            `('${targetStringifier.stringify(action.target)}', '${encodePersistedTableMetadata({
+              target: action.target,
+              lastUpdatedMillis: transitiveInputMetadataByTarget.get(action.target)
+                .lastUpdatedMillis,
+              definitionHash: hashExecutionAction(action),
```
lewish, line 301: A quick thought. Why do we store so much structure here, do we need it? I guess it's useful for understanding why something cached. Why not hash the entire thing - definition + (ordered) transitive inputs with last modified timestamps, then there's just one key to check?

Based on the PRD we aren't going to attempt to explain WHY an action wasn't cached. If we plan on doing that in the future though, it could be easier.

api/dbadapters/bigquery.ts
```diff
@@ -257,54 +248,82 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+          valuesToInsert += `, ${valuesTuples.pop()}`;
+        }
+        await this.execute(
+          `
+INSERT INTO \`${CACHED_STATE_TABLE_NAME}\` (target, metadata_proto) VALUES ${valuesToInsert}`,
+          options
+        );
+      }
+    } catch (e) {
+      if (e.message.includes("Exceeded rate limits")) {
```
lewish, line 320: I'd probably wrap this in String(e).includes - just in case

api/utils/run_cache.ts
lewish: Also feels like this should just have a unit test case in general

api/dbadapters/bigquery.ts
BenBirt: it wasn't really ever atomic, this was always happening, I've just moved the code around. (a) yes IMO it's simpler, (b) we need to separate out `INSERT` queries anyway due to query size limits, (c) IMO it's easier to reason about: everything is deleted *before* anything gets updated, and if the delete fails, nothing is updated.

api/dbadapters/bigquery.ts
BenBirt: no, after the run completes we read all necessary metadata before we write it back. see `computeTransitiveInputMetadataMap`.

api/utils/run_cache.ts
BenBirt: but we don't care whether a different encoding happens for `PersistedTableMetadata` - it's not used as a key, nor is it intended to be.

api/dbadapters/bigquery.ts
BenBirt: I have considered this, but I don't think it's a good idea. If we ever have to debug why something cached after the fact, having a hashed object isn't going to help at all. I have even considered un-hashing `ExecutionAction`, but didn't do it to avoid writing potentially very large rows to the cache.

api/dbadapters/bigquery.ts
BenBirt: honestly, my fear is that any custom stringifier will miss out any new fields we add in the future. but sure, done

PR #837, opened 2020-06-25
SQL utils package

lewish:
Publishes the classes under sql/ as a vanilla package that can be used directly in dataform projects, or non dataform code.
Replaces crossdb, as it's not very useful and moves it's one timestamp method into the sql class (And crossdb was only used in a few of our integration tests).


packages/@dataform/sql/BUILD
```diff
@@ -0,0 +1,54 @@
...
+)
+
+pkg_bundle_dts(
+    name = "bundle.d",
+    entry_point = "index.ts",
+    externals = [],
+    deps = [
+        ":sql",
+    ],
+)
```
Ekrekr, line 45: Nice

sql/index.ts
```diff
@@ -142,6 +142,19 @@ export class Sql {
     return `unix_millis(${timestamp})`;
   }
 
+  public timestampCurrentUTC() {
```
BenBirt, line 145:
```suggestion
  public currentUtcTimestamp() {
```

lewish:
> I'm slightly uncomfortable that this means we now need to (major-version) bump `version.bzl` for every API change in `sql` _or_ we now are making the promise not to make any backwards-incompatible changes to `sql` until we want to major-version bump `@dataform/core`.
> 
> Can we version `sql` separately?

Done!

sql/index.ts
lewish: Have moved all timestamp functions into a `timestamps` property and new class which makes them nicer to read.

tests/integration/sqldatawarehouse_project/definitions/example_incremental.sqlx
```diff
@@ -3,11 +3,11 @@ config {
...
 
 js {
-  const crossdb = require("@dataform/crossdb");
+  const sql = require("@dataform/sql");
 }
 
 ${ when(incremental(), "select * from (") }
 
-select ${crossdb.currentTimestampUTC()} as ts
+select ${sql().timestamps.currentUTC()} as ts
```
BenBirt, line 11:
```suggestion
select ${sql("mssql").timestamps.currentUTC()} as ts
```

PR #838, opened 2020-06-26
Move query evaluation collection to core

Ekrekr:
Required for https://github.com/dataform-co/dataform-co/pull/6887


Ekrekr:
Azure keeps choosing me as the deadlock victim, I'm not doing anything sketchy here

PR #840, opened 2020-06-29
Correctly serialize Errors across the subprocess boundary.

BenBirt:


api/vm/compile.ts
```diff
@@ -77,7 +73,11 @@ export function listenForCompileRequest() {
       const writeable = fs.createWriteStream(null, { fd: 4 });
       writeable.write(compiledResult, "utf8");
     } catch (e) {
-      process.send(e);
+      const serializableError = {};
+      for (const prop of Object.getOwnPropertyNames(e)) {
+        (serializableError as any)[prop] = e[prop];
+      }
+      process.send(serializableError);
```
GJMcGowan, line 80: This is real weird, but I guess I don't understand the underlying issue

BenBirt: This line is inside the compile child process, communicating errors back to the parent process. Unfortunately, `Error` classes serialize really poorly, `message` and `stack` don't make it across the IPC barrier. So I'm constructing a pseudo-`Error` (agreed that this is super weird) so that all fields serialize, then turning it back into an `Error` on the other side.

PR #843, opened 2020-06-30
Add number option to nav dimensions

Ekrekr:


core/column_descriptors.ts
```diff
@@ -84,6 +84,8 @@ export class ColumnDescriptors {
         return dataform.ColumnDescriptor.DimensionType.CATEGORY;
       case "timestamp":
         return dataform.ColumnDescriptor.DimensionType.TIMESTAMP;
+      case "number":
```
lewish, line 87: Sorry - can we call this "numerical" ? Feels slightly more accurate, as can capture floats, doubles, ints, etc.

protos/core.proto
```diff
@@ -224,6 +224,7 @@ message ColumnDescriptor {
     UNKNOWN_DIMENSION = 0;
     CATEGORY = 1;
     TIMESTAMP = 2;
+    NUMBER= 3;
```
lewish, line 227: same here 
also minor formatting issue

core/column_descriptors.ts
Ekrekr: I put `number` instead of `numerical` because we use `category` instead of `categorical`. I agree `numerical` sounds weird, but I can't figure out why

PR #850, opened 2020-07-02
Hide singleQueryPerAction

Ekrekr:
To confirm, the functionality of putting it in the `dataform.json` should still exist, but its existence should just be hidden to the end user?

### DIFF
--- a/content/docs/guides/configuration.md
+++ b/content/docs/guides/configuration.md
@@ -42,7 +42,7 @@ by changing the `defaultSchema` property to some other value. For example, to ch
 
 ### Running operations in context
 
-BigQuery and SQL Data Warehouse by default run all operations for a file in the same context. This functionality is not supported for Redshift or Snowflake.
+BigQuery and SQL Data Warehouse run all operations for a file in the same context. For Redshift or Snowflake, operations are run as separate queries.
 
 This is useful for scripting, for example defining variables or UDFs in BigQuery before the create table statement.
 
@@ -59,16 +59,6 @@ select var as col;
 is treated as if the `pre_operations` block wasn't there, becoming `declare var string; set var = 'val'; select var as col;`. For a table, SQL would be injected into the relevant place; the previous statement would be executed as `` declare var string; set var = 'val'; create or replace table `tada-analytics.dataform_data.silly_table` as select var as col; ``.
 
 For Redshift or Snowflake, the `pre_operations` block is always executed as a separate query before the main block.
-
-To disable running all operations in the same context, place the following flag in your `dataform.json` file:
-
-```json
-{
-  ...
-  "useSingleQueryPerAction": false,
-  ...
-}
-```
 
 ### Enable run caching to cut warehouse costs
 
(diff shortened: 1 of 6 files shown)


content/docs/guides/configuration.md
```diff
@@ -42,7 +42,7 @@ by changing the `defaultSchema` property to some other value. For example, to ch
 
 ### Running operations in context
```
BenBirt, line 43: we should probably move this to the docs which are specifically about `Table` and/or `Operation` now

content/docs/guides/operations.md
```diff
@@ -27,7 +27,10 @@ DROP VIEW IF EXISTS someschema.someview
 
 These statements will be run without modification against the warehouse. You can use warehouse specific commands in these files, such as BigQuery's DML or DDL statements, or Redshift/Postgres specific commands.
 
-Operations files behave very similarly to the statements provided to the `pre_operations { ... }` and `post_operations { ... }` blocks used when publishing datasets.
+<div className="bp3-callout bp3-icon-info-sign bp3-intent-none" markdown="1">
+  BigQuery and SQL Data Warehouse run all operations for a file in the same context; the executed SQL is created by joining all operations with a semi-colon <code>;</code>.
+  For Redshift or Snowflake, operations are run as separate queries.
```
BenBirt, line 32:
```suggestion
  For other warehouse types, operations are run as separate queries.
```

content/docs/guides/sqlx.md
```diff
@@ -92,9 +91,11 @@ Some examples can be found [here](datasets#referencing-other-datasets).
...
 - **Post-operations**: the same as pre-operations, but defined with `post_operations { }`, and runs after the main SQL.
 
-<div className="bp3-callout bp3-icon-info-sign" markdown="1">
-  By default, pre and post operations are run in the same context as the main query for BiqQuery and SQL Data Warehouse. To disable this, see <a href="./datasets">here</a>. This feature is not supported for Redshift or Snowflake.
-</div>
+BigQuery and SQL Data Warehouse run all operations for a file (such as pre and post operations) in the same context; the executed SQL is created by joining all operations with a semi-colon `;`. For Redshift or Snowflake, operations are run as separate queries.
+
+This is useful for scripting, for example defining variables or UDFs in BigQuery before the create table statement.
+
+For Redshift or Snowflake, the `pre_operations` block is always executed as a separate query before the main block, and the `post_operations` block as a separate query afterwards.
```
BenBirt, line 98: this isn't quite right, these blocks may contain multiple queries, each of which execute independently.

protos/core.proto
```diff
@@ -104,6 +106,7 @@ message RunConfig {
   bool use_run_cache = 8;
   bool disable_set_metadata = 9;
```
BenBirt, line 107: cleanup: would be nice if you could add the same comment here too

content/docs/guides/sqlx.md
Ekrekr: Fair enough, I'll get rid of the block because I feel it doesn't add much anyway

PR #859, opened 2020-07-08
Fix warehouse error names

Ekrekr:
Fixes https://github.com/dataform-co/dataform-co/issues/7007

Presumably the user knows which warehouse they're using

### DIFF
--- a/api/dbadapters/index.ts
+++ b/api/dbadapters/index.ts
@@ -48,7 +48,7 @@ export interface IDbAdapter {
 }
 
 export interface IDbAdapterClass<T extends IDbAdapter> {
-  create: (credentials: Credentials, warehouseType?: string) => Promise<T>;
+  create: (credentials: Credentials, warehouseType: string) => Promise<T>;
 }
 
 const registry: { [warehouseType: string]: IDbAdapterClass<IDbAdapter> } = {};
--- a/api/dbadapters/index.ts
+++ b/api/dbadapters/index.ts
@@ -1,5 +1,3 @@
-import { query } from "..";
-
 import { Credentials } from "df/api/commands/credentials";
 import { BigQueryDbAdapter } from "df/api/dbadapters/bigquery";
 import { RedshiftDbAdapter } from "df/api/dbadapters/redshift";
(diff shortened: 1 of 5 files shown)


api/dbadapters/redshift.ts
```diff
@@ -74,7 +74,7 @@ export class RedshiftDbAdapter implements IDbAdapter {
       if (options.includeQueryInError) {
         throw new Error(`Error encountered while running "${statement}": ${e.message}`);
       }
-      throw new ErrorWithCause(`Error executing Redshift query: ${e.message}`, e);
```
BenBirt, line 77: we shouldn't remove this, we added it for a reason (check blame if you like, but TL; DR: we wanted to clarify whether errors were in the warehouse, or dataform).

instead we should pass a constructor parameter to the DbAdapter that tells it which context it's being used for, and use that.

Ekrekr: Ah I see, have done this now

api/dbadapters/index.ts
```diff
@@ -50,7 +48,7 @@ export interface IDbAdapter {
 }
 
 export interface IDbAdapterClass<T extends IDbAdapter> {
-  create: (credentials: Credentials) => Promise<T>;
+  create: (credentials: Credentials, warehouseType?: string) => Promise<T>;
```
BenBirt, line 51:
```suggestion
  create: (credentials: Credentials, warehouseType: string) => Promise<T>;
```

`warehouseType` isn't optional, handling it is

api/dbadapters/redshift.ts
```diff
@@ -46,19 +51,22 @@ export class RedshiftDbAdapter implements IDbAdapter {
...
       };
       const queryExecutor = new PgPoolExecutor(clientConfig);
-      return new RedshiftDbAdapter(queryExecutor);
+      return new RedshiftDbAdapter(queryExecutor, { warehouseType });
     }
   }
 
-  private constructor(private queryExecutor: PgPoolExecutor, private sshTunnel?: SSHTunnelProxy) {}
+  private constructor(
+    private queryExecutor: PgPoolExecutor,
```
BenBirt, line 67: these should both be `readonly`

api/dbadapters/redshift.ts
```diff
@@ -74,7 +82,10 @@ export class RedshiftDbAdapter implements IDbAdapter {
       if (options.includeQueryInError) {
         throw new Error(`Error encountered while running "${statement}": ${e.message}`);
       }
-      throw new ErrorWithCause(`Error executing Redshift query: ${e.message}`, e);
+      throw new ErrorWithCause(
+        `Error executing ${this.options.warehouseType || ""} query: ${e.message}`,
```
BenBirt, line 86: make `warehouseType` non-optional, and remove the `|| ""`

PR #863, opened 2020-07-09
Add start open back to simple collapse

Ekrekr:
Prep for https://github.com/dataform-co/dataform-co/issues/7031


components/collapse.tsx
```diff
@@ -7,14 +7,16 @@ import * as styles from "df/components/collapse.css";
 interface ISimpleCollapseProps {
   message: React.ReactNode;
   className?: string;
+  startOpen?: boolean;
```
lewish, line 10: nit: I think the pattern for uncontrolled defaults is usually `defaultOpen`

PR #868, opened 2020-07-10
Fix docs link jumping

Ekrekr:
Fixes https://github.com/dataform-co/dataform-co/issues/6971

This frustrates me, the `::before` element is there, but it's not pushing the page down right. Any ideas?


Ekrekr:
Is this hacky enough for you now

PR #871, opened 2020-07-10
Add min, max, avg sql functions

lewish:


lewish:
> It really should be called `mean` in sql, not `avg`

Agree :rofl:

PR #872, opened 2020-07-10
Make it possible to disable operations and assertions.

BenBirt:
fixes https://github.com/dataform-co/dataform/issues/326

### DIFF
--- a/api/commands/build.ts
+++ b/api/commands/build.ts
@@ -115,15 +115,13 @@ export class Builder {
       throw new Error("Protected datasets cannot be fully refreshed.");
     }
 
-    const tasks = table.disabled
-      ? ([] as dataform.IExecutionTask[])
-      : this.adapter.publishTasks(table, runConfig, tableMetadata).build();
-
     return {
       ...this.toPartialExecutionAction(table),
       type: "table",
       tableType: table.type,
-      tasks,
+      tasks: table.disabled
+        ? []
+        : this.adapter.publishTasks(table, runConfig, tableMetadata).build(),
       hermeticity: table.hermeticity || dataform.ActionHermeticity.HERMETIC
     };
   }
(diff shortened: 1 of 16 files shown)


api/commands/build.ts
```diff
@@ -115,15 +115,13 @@ export class Builder {
...
-    const tasks = table.disabled
-      ? ([] as dataform.IExecutionTask[])
-      : this.adapter.publishTasks(table, runConfig, tableMetadata).build();
-
     return {
       ...this.toPartialExecutionAction(table),
       type: "table",
       tableType: table.type,
-      tasks,
+      tasks: table.disabled
```
lewish, line 122: This won't skip things like pushing metadata back to bigquery, but that's not the end of the world

core/operation.ts
```diff
@@ -111,6 +119,11 @@ export class Operation {
       : dataform.ActionHermeticity.NON_HERMETIC;
   }
 
+  public disabled() {
```
lewish, line 122: I wonder if we really want to encourage this syntax if the same can be achieved in the config block? I think at some point we should move to these APIs being fully object based, like:

```
publish({
  name: "x",
  query: ctx => ...
}
```

In other words I'd be tempted to leave this out for now

core/session.ts
```diff
@@ -103,20 +103,6 @@ export class Session {
         "Actions may only contain more than one SQL statement if they are of type 'operations'."
       );
     }
-    if (
```
lewish, line 106: Why remove all these?

BenBirt: your excess-property checks catch these - so it's just a double-error.

core/operation.ts
BenBirt: I kinda agree, but OTOH i do think API consistency is king and it costs us very little to have these, so I'd prefer to keep them

api/commands/build.ts
BenBirt: good catch, but actually the run code does correctly skip that code: https://source.cloud.google.com/tada-analytics/github_dataform-co_dataform/+/master:api/commands/run.ts;l=325?q=executeAction

PR #876, opened 2020-07-13
Fix proto enum collision, ad go, java proto build tests

lewish:


protos/BUILD
```diff
@@ -60,3 +60,15 @@ ts_proto_library(
         ":server_proto",
     ],
 )
+
+load("//testing:build_test.bzl", "build_test")
+
+build_test(
+    name = "build_test",
+    deps = [
```
BenBirt, line 68: should probably really have the proto_library rules here, the other ones are useful but not as necessary.

PR #877, opened 2020-07-13
Expose formatter for file content

Ekrekr:
Prep for https://github.com/dataform-co/dataform-co/issues/6084


BenBirt:
IMO we don't want to depend on all of :api in the app. I think we should be moving the formatting code into /sqlx, and depend on just that.

Ekrekr:
> IMO we don't want to depend on all of :api in the app. I think we should be moving the formatting code into /sqlx, and depend on just that.

Ok I'll do that in another PR

Issue #397, opened 2019-08-27
Make schema creation conditional on whether or not the schema currently exists


BenBirt:
The snowflake DB adapter now only creates new schemas when they don't exist. We should do the same for other warehouse types.

BenBirt:
dup of https://github.com/dataform-co/dataform/issues/689

PR #882, opened 2020-07-13
Make all DBAdapters only create schemas when the schema doesn't already exist.

BenBirt:
fixes https://github.com/dataform-co/dataform/issues/397


api/dbadapters/bigquery.ts
```diff
@@ -199,6 +199,11 @@ export class BigQueryDbAdapter implements IDbAdapter {
     return rows;
   }
 
+  public async schemas(database: string): Promise<string[]> {
+    const data = await this.getClient(database).getDatasets();
+    return data[0].map(dataset => dataset.id);
```
Ekrekr, line 204: Why `[0]`?

BenBirt:
method renamed!

api/dbadapters/bigquery.ts
BenBirt: Sadly BQ response interfaces are suuuuuper weird.

PR #890, opened 2020-07-16
Spinner component

Ekrekr:
For https://github.com/dataform-co/dataform-co/pull/7199

Note the default intent as primary


components/spinner.tsx
```diff
@@ -0,0 +1,29 @@
+import { IIntentProps, Intent, Spinner as BlueprintSpinner } from "@blueprintjs/core";
+import * as React from "react";
+
+import * as styles from "df/components/spinner.css";
+
+export type SpinnerSize = "large" | "standard" | "small";
+
+export interface ISpinnerProps extends IIntentProps {
+  size?: SpinnerSize;
+  value?: number;
```
GJMcGowan, line 10: What's the purpose in `value`?

components/spinner.scss
```diff
@@ -0,0 +1,7 @@
+.centered {
+  height: 100%;
```
GJMcGowan, line 2: May want width 100% as well?

components/spinner.tsx
Ekrekr: Value lets you set the percentage completion of the spinner. Instead of just extending the blueprint spinner props, I'm just adding in `value`, as `size` is different in the `ISpinnerProps` to the blueprint props (E.g. SPINNER.SIZE_LARGE is blueprint's style rather than string)

PR #896, opened 2020-07-20
Add props for additional buttons in controlled dialog

Ekrekr:
Required for https://github.com/dataform-co/dataform-co/pull/7201

Follows pattern of having buttons in the card actions section

### DIFF
--- a/components/controlled_dialog.tsx
+++ b/components/controlled_dialog.tsx
@@ -13,6 +13,7 @@ export interface IControlledDialogProps {
   cancelButtonText?: string;
   confirmButtonProps?: IButtonProps & React.ButtonHTMLAttributes<HTMLButtonElement>;
   confirmHidden?: boolean;
+  additionalButtons?: Array<IButtonProps & React.ButtonHTMLAttributes<HTMLButtonElement>>;
   disable?: boolean;
   width?: string;
 }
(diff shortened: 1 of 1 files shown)


components/controlled_dialog.tsx
```diff
@@ -92,6 +94,18 @@ export class ControlledDialog extends React.Component<IControlledDialogProps, IS
...
             />
+            {additionalButtons?.map(buttonProps => (
+              <Button
+                {...buttonProps}
+                className={`${Classes.POPOVER_DISMISS} ${buttonProps?.className || ""}`}
+                onClick={(e: React.MouseEvent<HTMLElement, MouseEvent>) => {
+                  if (!!buttonProps.onClick) {
+                    buttonProps.onClick(e);
+                  }
+                  this.setState({ isOpen: false });
```
GJMcGowan, line 105: Mmm, I feel like this is going to cause trouble in the future. This whole component is getting too generic, trying to do too much. Might be easier to just roll your own dialog for the thing you want this for.

If we do keep this approach, you might want the buttonProps to control whether `isOpen` gets set to false on button click.

Ekrekr: I see, is looking a bit complex; will update

PR #897, opened 2020-07-20
Add support for column level tags

lewish:


BenBirt:
This looks fine, but before we commit to adding something to the public API, let's agree on a design?

lewish:
> This looks fine, but before we commit to adding something to the public API, let's agree on a design?

I've made the tags property hidden so it won't show up in any documentation for now!

PR #904, opened 2020-07-22
Catch test query execution errors.

BenBirt:


api/commands/test.ts
```diff
@@ -14,10 +14,20 @@ async function runTest(
...
+  try {
+    [actualResults, expectedResults] = await Promise.all([
+      dbadapter.execute(testCase.testQuery, { maxResults: 1000 }),
+      dbadapter.execute(testCase.expectedOutputQuery, { maxResults: 1000 })
+    ]);
+  } catch (e) {
+    return {
+      name: testCase.name,
+      successful: false,
+      messages: [`Error thrown: ${e.message}.`]
```
GJMcGowan, line 28: stack?

BenBirt: In this case it *should* only really ever be errors coming straight from the DB, so I'm going with `message` to keep it short. But if we notice it's hiding stuff, we can definitely change this

PR #907, opened 2020-07-23
Dramatically simplify Redshift adapter using pg-query-stream.

BenBirt:


BenBirt:
I'm sure it'll be super safe

PR #917, opened 2020-07-30
Redshift fixes

### LINKED ISSUE #914: pg.Client - Cannot read property 'handleRowDescription' of null and Cannot read property 'name' of null
Tracking bug, possibly related: https://github.com/brianc/node-postgres/issues/1105, more details to follow.

lewish:
- Disable parallelism on the redshift test. It appears that running two things at the same time with the same adapter is causing flakiness, haven't worked out why yet. Bug: #918
- Change "close" handler to "end" handler - as this is what's in the docs and seems to be partly the cause of #914
- Change the error handler to dispose of the connection, so that it can't be re-used. It seems like errored connections do weird things that also cause some of the issues in #914, and as it shouldn't be that common a situation, shouldn't cause too many connections to be created
- Add some better error messages to integration tests

Fixes #914

### DIFF
--- a/tests/integration/redshift.spec.ts
+++ b/tests/integration/redshift.spec.ts
@@ -8,7 +8,7 @@ import { dataform } from "df/protos/ts";
 import { suite, test } from "df/testing";
 import { compile, getTableRows, keyBy } from "df/tests/integration/utils";
 
-suite("@dataform/integration/redshift", { parallel: true }, ({ before, after }) => {
+suite("@dataform/integration/redshift", { parallel: false }, ({ before, after }) => {
   const credentials = dfapi.credentials.read("redshift", "test_credentials/redshift.json");
   let dbadapter: dbadapters.IDbAdapter;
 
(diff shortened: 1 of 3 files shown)


tests/integration/redshift.spec.ts
```diff
@@ -180,8 +188,6 @@ suite("@dataform/integration/redshift", { parallel: true }, ({ before, after })
   suite("evaluate", async () => {
     test("evaluate from valid compiled graph as valid", async () => {
       const compiledGraph = await compile("tests/integration/redshift_project", "evaluate");
-      const executionGraph = await dfapi.build(compiledGraph, {}, dbadapter);
```
BenBirt, line 183: I don't understand this change:
- this build & run is required so that the tests work for a "clean" redshift (otherwise the evaluated queries fail, because the relevant tables don't exist)
- this shouldn't be executing the same run twice, by virtue of the `evaluate` schema suffix passed in on line 182 (all of the independent runs in this file should be using a different schema suffix to avoid exactly that problem)

lewish: Sorry should have read this in more depth.

I've filed another bug #918, for now I've turned parallelism off which also solves the problem.

PR #924, opened 2020-08-03
Release any pg.Clients that throw Errors which are not otherwise handled by the current query.

BenBirt:
for https://github.com/dataform-co/dataform/issues/914


api/dbadapters/redshift.ts
```diff
@@ -286,6 +286,7 @@ class PgPoolExecutor {
     client.on("error", err => {
       // tslint:disable-next-line: no-console
       console.error("pg.Client client error", err.message, err.stack);
+      client.release(err);
```
lewish, line 289: you may want a try catch around this and below on the other release calls now. You can't call release twice it seems, and I wouldn't be surprised if this would cause that to happen now which will cause more uncaughts

PR #929, opened 2020-08-05
Don't cause uncaught exceptions when column names are not unique

lewish:


api/dbadapters/redshift.ts
```diff
@@ -302,7 +302,12 @@ class PgPoolExecutor {
       options?.onCancel?.(() => query.destroy());
       query.on("data", (row: any) => {
         if (results.length < options.maxResults) {
-          verifyUniqueColumnNames((query as any).cursor._result.fields);
+          try {
+            verifyUniqueColumnNames((query as any).cursor._result.fields);
+          } catch (e) {
+            query.destroy();
+            reject(e);
```
BenBirt, line 309: `return`?

PR #939, opened 2020-08-12
BigQuery policy tag support

### LINKED ISSUE #938: Support setting BigQuery policy tags on columns
BigQuery policy tags can be used to control access to data at a column level.

Proposal:
- Extend the ColumnDescriptor interface to support `bigqueryPolicyTags` as a new field
- Add support for setting the column level policy tags in the BigQuery DB adapter, similarly to how we set column descriptions

```
config {
  type: "table",
  columns : {
    column1: "description 1",
    column2: {
      description: "description2",
      bigqueryPolicyTags: ["tag1", "tag2"]
    }
  }
}

### PR DESCRIPTION
**Pull request by lewish**
Also updates BigQuery libraries to latest and uses in-built types.

Fixes #938

### DIFF
--- a/protos/core.proto
+++ b/protos/core.proto
@@ -245,6 +245,10 @@ message ColumnDescriptor {
   string expression = 6;
 
   repeated string tags = 7;
+
+  // BigQuery specific settings.
+
+  repeated string bigqueryPolicyTags = 8;
 }
 
 message Table {
--- a/protos/core.proto
+++ b/protos/core.proto
@@ -248,7 +248,7 @@ message ColumnDescriptor {
 
   // BigQuery specific settings.
 
-  repeated string bigqueryPolicyTags = 8;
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: protos/core.proto [hunk 1]
```diff
@@ -245,6 +245,10 @@ message ColumnDescriptor {
   string expression = 6;
 
   repeated string tags = 7;
+
+  // BigQuery specific settings.
+
+  repeated string bigqueryPolicyTags = 8;
```
Ekrekr, line 251: Do you think it be worth extracting to `BigQueryColumnOptions` with this as a member, similar to how we use `BigQueryOptions` in actions?

lewish: Yeah Ben and I chatted about this, as we can't currently see any other future bigquery specific things, or other warehouse specific settings at a column level, we felt a prefixed field name made more sense in this case as the columns object is already getting quite deeply nested.

Issue #940, opened 2020-08-13
Always run downstream assertions when publishing tables/views


lewish:
Downstream assertions don't run if you run just a single action via the CLI or Dataform Web.

It makes sense that by default, when publishing a table we should run any assertions that are direct dependents of the action that is being run.

Proposal:
- When computing the execution graph during build and prune, make sure to include any direct dependents of included actions that are assertions

This feels like a sensible default behaviour and I can't think of any situations where we wouldn't want to do this, but in the future we can add an option such as `ignore_downstream_assertions` if we find there is a need to turn this behaviour off.

https://dataform.canny.io/feature-requests/p/include-assertions-by-default-when-publish-tableviewincremental-is-triggered

This is a significant change in behaviour, so bump minor version.

PR #944, opened 2020-08-13
Make prune keep assertions which are direct dependents of actions

Ekrekr:
Fixes https://github.com/dataform-co/dataform/issues/940

Implements [this](https://dataform.canny.io/feature-requests/p/include-assertions-by-default-when-publish-tableviewincremental-is-triggered) canny request

### DIFF
--- a/api/commands/prune.ts
+++ b/api/commands/prune.ts
@@ -82,6 +82,8 @@ function computeIncludedActionNames(
       const matchingDependencyNames = !!assertion.dependencies?.length
         ? utils.matchPatterns(assertion.dependencies, allActionNames)
         : [];
+      console.log("assertion.name", assertion.name);
+      console.log("matchingDependencyNames", matchingDependencyNames);
       if (!!matchingDependencyNames?.length) {
         includedActionNames.add(assertion.name);
       }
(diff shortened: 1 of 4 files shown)


version.bzl
```diff
@@ -1,3 +1,3 @@
 # NOTE: If you change the format of this line, you must change the bash command
 # in /scripts/publish to extract the version string correctly.
-DF_VERSION = "1.8.5"
+DF_VERSION = "1.8.6"
```
lewish, line 3: I think this should probably go to `1.9.0` as it's new functionality

api/commands/prune.ts
```diff
@@ -75,6 +75,17 @@ function computeIncludedActionNames(
...
       });
     }
+
+    // Add assertions which are direct dependents of actions.
+    const assertions = compiledGraph.assertions;
+    assertions.forEach(assertion => {
+      const matchingDependencyNames = !!assertion.dependencies?.length
+        ? utils.matchPatterns(assertion.dependencies, allActionNames)
+        : [];
+      if (!!matchingDependencyNames?.length) {
```
BenBirt, line 85: this seems incorrect: if the assertion depends on 2 actions, 1 of which isn't included in the run, IMO the assertion shouldn't run.

PR #955, opened 2020-08-19
Step indicator tweaks - work both in and out of CardActions block

Ekrekr:
Also uses appropriate css variable for color now

Screenshot 2020-08-19 at 12 19 20

Screenshot 2020-08-19 at 12 20 12

### DIFF
--- a/components/step_indicator.scss
+++ b/components/step_indicator.scss
@@ -1,0 +1,18 @@
+.stepIndicatorHolder {
+  display: flex;
+  justify-content: center;
+  margin-top: 20px;
+}
+
+.stepIconHolder {
+  width: 40px;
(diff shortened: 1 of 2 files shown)


components/step_indicator.scss
```diff
@@ -0,0 +1,18 @@
+.stepIndicatorHolder {
+  display: flex;
+  justify-content: center;
+  margin-top: 20px;
+}
+
+.stepIconHolder {
+  width: 40px;
+  height: 40px;
+  text-align: center;
```
GJMcGowan, line 10: If it's flex already, makes more sense to do `justify-content: center`.

GJMcGowan (approved):

PR #957, opened 2020-08-19
Update step_indicator.scss

Ekrekr:
Progresses https://github.com/dataform-co/dataform-co/issues/7627

This does actually need to be flex. Don't think it propagates down far enough?


GJMcGowan (approved):
I don't see how this could've worked otherwise, unless there was another class applying to the element.

PR #959, opened 2020-08-20
Always run auto assertions

Ekrekr:
I'm not quite sure where to test this from a higher level; the `examples.spec` already runs the assertion. Should it go in the warehouse specs?

### DIFF
--- a/api/commands/prune.ts
+++ b/api/commands/prune.ts
@@ -77,5 +77,12 @@ function computeIncludedActionNames(
     }
   }
 
+  // Add auto assertions
+  [...compiledGraph.assertions].forEach(assertion => {
+    if (includedActionNames.has(assertion.parent) && !includedActionNames.has(assertion.name)) {
+      includedActionNames.add(assertion.name);
+    }
+  });
+
   return includedActionNames;
 }
--- a/core/session.ts
+++ b/core/session.ts
@@ -284,7 +284,7 @@ export class Session {
     return newTable;
   }
 
-  public assert(name: string, query?: AContextable<string>): Assertion {
+  public assert(name: string, query?: AContextable<string>, parent?: string): Assertion {
     const assertion = new Assertion();
     assertion.session = this;
     utils.setNameAndTarget(this, assertion.proto, name, this.config.assertionSchema);
@@ -293,6 +293,9 @@ export class Session {
     }
     assertion.proto.fileName = utils.getCallerFile(this.rootDir);
     this.actions.push(assertion);
+    if (parent) {
+      assertion.proto.parent = parent;
+    }
     return assertion;
   }
 
--- a/core/table.ts
+++ b/core/table.ts
@@ -450,12 +450,6 @@ export class Table {
     newTags.forEach(t => {
       this.proto.tags.push(t);
     });
-    if (!!this.uniqueKeyAssertion) {
-      this.uniqueKeyAssertion.tags(value);
-    }
-    if (!!this.mergedRowConditionsAssertion) {
-      this.mergedRowConditionsAssertion.tags(value);
-    }
     return this;
   }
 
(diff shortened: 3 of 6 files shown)


api/commands/prune.ts
```diff
@@ -77,5 +77,12 @@ function computeIncludedActionNames(
     }
   }
 
+  // Add auto assertions
+  [...compiledGraph.assertions].forEach(assertion => {
+    if (includedActionNames.has(assertion.parent) && !includedActionNames.has(assertion.name)) {
```
BenBirt, line 82: you can remove `!includedActionNames.has(assertion.name)`, the point in a `Set` is that all this is handled for you

core/table.ts
```diff
@@ -450,12 +450,6 @@ export class Table {
     newTags.forEach(t => {
       this.proto.tags.push(t);
     });
-    if (!!this.uniqueKeyAssertion) {
```
BenBirt, line 453: you can remove `uniqueKeyAssertion` and `mergedRowConditionsAssertion` as object members now, they're no longer useful

core/session.ts
```diff
@@ -284,7 +284,7 @@ export class Session {
     return newTable;
   }
 
-  public assert(name: string, query?: AContextable<string>): Assertion {
+  public assert(name: string, query?: AContextable<string>, parent?: string): Assertion {
```
BenBirt, line 287: IMO this feels like a bad idea. I would suggest setting the `parent` field of the `Assertion`'s protobuf directly inside the `config()` method in `table.ts`. something like `assert(...).proto.parent = this.proto.name`

protos/core.proto
```diff
@@ -335,6 +335,9 @@ message Assertion {
   repeated string tags = 9;
 
   ActionDescriptor action_descriptor = 10;
+
+  // Only present for auto assertions.
+  string parent = 15;
```
BenBirt, line 340: I think we should use `Target` here.

note that this field needs to be fixed for any schema suffixes / table prefixes. See the code in `Session.alterActionName` for examples

BenBirt:
You should also bump the core version number.

core/session.ts
```diff
@@ -515,6 +518,12 @@ export class Session {
       action.dependencies = (action.dependencyTargets || []).map(dependencyTarget =>
         utils.targetToName(dependencyTarget)
       );
+
+      if ("parent" in action) {
+        if (!!action.parent) {
+          action.parent = newTargetByOriginalTarget.get(action.parent);
+        }
+      }
```
Ekrekr, line 526: Any idea why `hasOwnProperty` doesn't work here?

api/commands/prune.ts
```diff
@@ -77,5 +77,12 @@ function computeIncludedActionNames(
     }
   }
 
+  // Add auto assertions
+  [...compiledGraph.assertions].forEach(assertion => {
+    if (includedActionNames.has(utils.targetToName(assertion.parent))) {
```
BenBirt, line 82: for extra credit: change `includedActionNames` into a `StringifiedSet<dataform.ITarget>` so this stuff isn't necessary.

core/session.ts
BenBirt: I'd presume because the object prototype is where `parent` is defined, and thus the object inherits it (and so `hasOwnProperty` will return false for `parent`).

core/session.ts
```diff
@@ -515,6 +518,12 @@ export class Session {
       action.dependencies = (action.dependencyTargets || []).map(dependencyTarget =>
         utils.targetToName(dependencyTarget)
       );
+
+      if ("parent" in action) {
```
BenBirt, line 522: why not just change `IActionProto` to include an optional `parent` field, revert the changes to the method declaration, and then get rid of this?

protos/core.proto
```diff
@@ -335,6 +335,9 @@ message Assertion {
   repeated string tags = 9;
 
   ActionDescriptor action_descriptor = 10;
+
+  // Only present for auto assertions.
+  Target parent = 15;
```
BenBirt, line 340:
```suggestion
  Target parent_action = 15;
```

BenBirt:
you still need to bump the version number

core/session.ts
Ekrekr: Sure that seems neater

api/commands/prune.ts
Ekrekr: I got started on this, but a lot of things need to be transitioned from `name` to `Target`; will do in another PR

PR #960, opened 2020-08-20
Push table/view/column metadata to all warehouses that support it.

BenBirt:
for https://github.com/dataform-co/dataform-co/issues/7449

- azure doesn't support anything like this
- postgres has full support for table/view/table column/view column comments
- redshift and snowflake both support table/view comments, and table column comments, but not view column comments
- clean up BQ adapter to stop using concurrency limiter for non-warehouse query API calls (because this slows things down, causes flaky deadlock issues, and is unnecessary)
- clean up BQ adapter to issue all 'internal' queries using the `execute` API
- start using `svv_tables` in redshift adapter which is a combined view of 'real' tables/views and external tables


GJMcGowan (approved):
I totally read and understood all of this

PR #963, opened 2020-08-21
Make prune use targets rather than name

Ekrekr:
Name is still used in the wrapper function, and required for the tests

I made some progress removing the deprecated name instances but it's a pretty massive job; have that PR on the backburner.

### DIFF
--- a/api/commands/prune.ts
+++ b/api/commands/prune.ts
@@ -30,48 +30,51 @@ function computeIncludedActionNames(
     compiledGraph.assertions
   );
 
-  const allActionNames = allActions.map(n => n.name);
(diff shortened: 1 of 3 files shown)


lewish (commented):
I'm a bit confused about what we are trying to do here, the tests imply there is a change in behaviour, and from a code cleanup perspective this doesn't obviously look like a better state than where we were!

What problem are you trying to fix?

Ekrekr:
> I'm a bit confused about what we are trying to do here, the tests imply there is a change in behaviour, and from a code cleanup perspective this doesn't obviously look like a better state than where we were!
> 
> What problem are you trying to fix?

Name is deprecated, but prune uses it quite heavily. This starts some of the transition to using target instead.

PR #964, opened 2020-08-24
Add react router link to breadcrumb option

Ekrekr:
Intentionally haven't added `href` option because breadcrumbs should never point somewhere external

### DIFF
--- a/components/breadcrumb.tsx
+++ b/components/breadcrumb.tsx
@@ -14,13 +14,9 @@ export function BreadCrumb({ pathParts }: IBreadCrumbProps) {
       {pathParts.slice(0, -1).map((pathPart, index) => (
         <React.Fragment key={index}>
           <H6>
-            {!!pathPart.to ? (
-              <Link onClick={pathPart.onClick} to={pathPart.to}>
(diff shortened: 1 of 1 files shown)


components/breadcrumb.tsx
```diff
@@ -13,7 +14,13 @@ export function BreadCrumb({ pathParts }: IBreadCrumbProps) {
       {pathParts.slice(0, -1).map((pathPart, index) => (
         <React.Fragment key={index}>
           <H6>
-            <a onClick={pathPart.onClick}>{pathPart.label}</a>
+            {!!pathPart.to ? (
+              <Link onClick={pathPart.onClick} to={pathPart.to}>
+                {pathPart.label}
+              </Link>
+            ) : (
+              <a onClick={pathPart.onClick}>{pathPart.label}</a>
```
lewish, line 22: do you need this at all? looks like you could use link for both cases

lewish (approved):

components/breadcrumb.tsx
Ekrekr: Yup nice

PR #971, opened 2020-09-01
Bigquery label metadata

Ekrekr:
Fixes https://github.com/dataform-co/dataform-co/issues/7823

### DIFF
--- a/content/docs/warehouses/bigquery.md
+++ b/content/docs/warehouses/bigquery.md
@@ -66,6 +66,25 @@ In order to be able to query Google Sheets tables via BigQuery, you'll need to s
 - Find the email address of the service account through which you connected to Dataform. You can find this on the [Google Cloud IAM service accounts console page](https://console.cloud.google.com/iam-admin/serviceaccounts). If you're developing your Dataform project locally (as opposed to using Dataform Web), you can find the service accounts email in the `.df-credentials.json` file.
 - Share the Google sheet with the email address of the service account as you would a colleague, through the sheets sharing settings and make sure it has access.
 
+### Using BigQuery labels
+
+[BigQuery labels](https://cloud.google.com/bigquery/docs/labels-intro) are key-value pair that helps you organize your Google Cloud BigQuery resources. To use them in Dataform, add them to the config block as so:
+
+```sql
+config {
+  type: "table",
+  bigquery: {
+    labels: {
+      label1: "val1",
+      -- If the label name contains special characters, e.g. hyphens, then quote its name.
+      "label-name": "value"
+    }
+  }
+}
+
+select "test" as column1
+```
+
 ### Using different project_ids within the same project
 
 You can both read from and publish to two separate GCP project_ids within a single Dataform project. For example, you may have a project_id called `raw` that contains raw data loaded in your warehouse and a project_id called `analytics` in which you create data tables you use for analytics and reporting.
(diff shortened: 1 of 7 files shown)


core/table.ts
```diff
@@ -141,10 +141,17 @@ export interface IBigQueryOptions {
    * For more information, see our [incremental dataset docs](https://docs.dataform.co/guides/incremental-datasets).
    */
   updatePartitionFilter?: string;
+
+  /**
+   * Key-value pairs for [BigQuery labels](https://cloud.google.com/bigquery/docs/labels-intro).
+   *
+   * Both `labels: { "key": "value" }` and `labels: { key: "value" }` as part of the `bigquery` block are valid.
```
BenBirt, line 148: I'm not sure this is super helpful. I would suggest: `If the label name contains special characters, e.g. hyphens, then quote its name, e.g. labels: { "label-name": "value" }.`

tests/integration/bigquery_project/definitions/example_view.sqlx
```diff
@@ -6,6 +6,12 @@ config {
             description: "val doc",
             bigqueryPolicyTags: "projects/dataform-integration-tests/locations/us/taxonomies/800183280162998443/policyTags/494923997126550963"
         }
+    },
+    bigquery: {
```
BenBirt, line 10: there should be an accompanying change to the integration test assertions (there is one test case that specifically checks that table metadata stuff is set, just add to that).

BenBirt (changes requested):
thanks!

we also need to:
1) bump the major version
2) update docs

BenBirt (commented):
Not sure you saw the previous review message:

thanks!

we also need to:

bump the major version
update docs

content/docs/warehouses/bigquery.md
```diff
@@ -66,6 +66,25 @@ In order to be able to query Google Sheets tables via BigQuery, you'll need to s
 - Find the email address of the service account through which you connected to Dataform. You can find this on the [Google Cloud IAM service accounts console page](https://console.cloud.google.com/iam-admin/serviceaccounts). If you're developing your Dataform project locally (as opposed to using Dataform Web), you can find the service accounts email in the `.df-credentials.json` file.
 - Share the Google sheet with the email address of the service account as you would a colleague, through the sheets sharing settings and make sure it has access.
 
+### Using BigQuery labels
+
+[BigQuery labels](https://cloud.google.com/bigquery/docs/labels-intro) are key-value pair that helps you organize your Google Cloud BigQuery resources. To use them in Dataform, add them to the config block as so:
```
BenBirt, line 71:
```suggestion
[BigQuery labels](https://cloud.google.com/bigquery/docs/labels-intro) are key-value pairs that help you organize your Google Cloud BigQuery resources. To use them in Dataform, add them to the config block:
```

content/docs/warehouses/bigquery.md
```diff
@@ -66,6 +66,25 @@ In order to be able to query Google Sheets tables via BigQuery, you'll need to s
...
+
+```sql
+config {
+  type: "table",
+  bigquery: {
+    labels: {
+      label1: "val1",
+      -- If the label name contains special characters, e.g. hyphens, then quote its name.
```
BenBirt, line 79:
```suggestion
      /* If the label name contains special characters, e.g. hyphens, then quote its name. */
```

BenBirt (approved):

PR #972, opened 2020-09-02
Add postgres integration tests.

BenBirt:
The solution was to either build postgres in Bazel or run the postgres Docker image. I managed to get it to build inside Bazel, but it depends on far too many shared libraries, each of which would have to be explicitly listed to make it a runnable Bazel target. Instead I went with the Docker option, which has the downside of requiring a local Docker daemon to be running, but does work cleanly.

A nice side effect is that I found and fixed a small bug that prevented the use of `incremental` in postgres.

I have not yet branched either of the adapters, that will happen as necessary.

for https://github.com/dataform-co/dataform-co/issues/7449


tools/postgres/postgres_fixture.ts
```diff
@@ -0,0 +1,61 @@
+import { execSync } from "child_process";
+import * as dbadapters from "df/api/dbadapters";
+import { sleepUntil } from "df/common/promises";
+import { IHookHandler } from "df/testing";
+
+const USE_CLOUD_BUILD_NETWORK = !!process.env.USE_CLOUD_BUILD_NETWORK;
```
Ekrekr, line 6: Is there a `.env` file set up that's needed for this flag?

tests/integration/postgres.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+      { disableSslForTestsOnly: true }
+    );
+  });
+
+  test("run", { timeout: 60000 }, async () => {
+    const compiledGraph = await compile("tests/integration/postgres_project", "project_e2e");
+
+    const adapter = adapters.create(compiledGraph.projectConfig, compiledGraph.dataformCoreVersion);
+
+    // Redshift transactions are giving us headaches here. Drop tables sequentially.
```
Ekrekr, line 36: I guess changes like this are happening in the next PR?

Ekrekr (approved):
LGTM, dockerized solution looks very clean; maybe we should use it for more

tests/integration/postgres.spec.ts
BenBirt: yup, that is the plan

tools/postgres/postgres_fixture.ts
BenBirt: no - by default it's unset, and thus `USE_CLOUD_BUILD_NETWORK` will evaluate false. cloud build uses the `bazel_test` script, which sets it, thus `USE_CLOUD_BUILD_NETWORK` evaluates true.

PR #977, opened 2020-09-02
add a JS API incremental example

dwl285:


content/docs/examples/example-scripts.md
```diff
@@ -438,6 +438,17 @@ publish("example")
...
 
+### Creating incremental tables using the JS API
+
+```js
+// definitions/incremental_example.js
+
+publish("incremental_example", {
+  type: "incremental"
+}).query(ctx => `SELECT * FROM ${ctx.ref("other_table")}`)
+  .where(`where timestamp > (select max(date) from ${self()})`)
```
BenBirt, line 449: i wouldn't do this, we should be using `${when(incremental(), ...}`!

dwl285: Oh yeah, good point. Hadn't thought about that. We should update some other docs pages then I guess?

content/docs/examples/example-scripts.md
```diff
@@ -438,6 +438,18 @@ publish("example")
...
+### Creating incremental tables using the JS API
+
+```js
+// definitions/incremental_example.js
+
+publish("incremental_example", {
+  type: "incremental"
+}).query(ctx =>
+  `SELECT * FROM ${ctx.ref("other_table")}
+${ctx.when(ctx.incremental(),`where timestamp > (select max(date) from ${ctx.self()}`)}`)
```
BenBirt, line 450:
```suggestion
}).query(ctx => `
  SELECT * FROM ${ctx.ref("other_table")}
  ${ctx.when(ctx.incremental(),`WHERE timestamp > (SELECT MAX(date) FROM ${ctx.self()}`)}
`)
```

(slightly nicer spacing IMO, also i don't mind if we use CAPITAL_CASE or lower_case but we should be consistent)

BenBirt (approved):

Issue #980, opened 2020-09-03
Incremental merge is broken for Redshift


BenBirt:
https://dataform-users.slack.com/archives/CHMJK8MKM/p1599076187025600

We need to create temporary tables in the same session (connection) in which we use that temporary table: https://source.cloud.google.com/tada-analytics/github_dataform-co_dataform/+/master:core/adapters/redshift.ts;l=153?q=transaction&start=11

We already need to be running those queries inside a transaction (which would imply a single connection), I think now is the time to build that out.

Strawman proposal, very much open to discussion: I think we should change https://source.cloud.google.com/tada-analytics/github_dataform-co_dataform/+/master:protos/core.proto;l=402?q=core.proto to be a `oneof` that allows either a `string statement` or a new `Transaction transaction` (which would basically just be wrapper around `repeated string`). This would consequently require changes to the UI to display any statements in a `Transaction` correctly.

PR #982, opened 2020-09-03
Run all tasks in ExecutionActions using a single DB client (where applicable).

BenBirt:
fixes https://github.com/dataform-co/dataform/issues/980

- this only applies to redshift/postgres
- there is quite a lot of code surrounding execution of individual tasks; so I settled on the pattern of a method which takes a callback; the callback is called with a single client instance which is used for the duration of `ExecutionAction` execution
- with this, I could bring back transactionality to incremental merge

### DIFF
--- a/api/dbadapters/postgres.ts
+++ b/api/dbadapters/postgres.ts
@@ -1,7 +1,7 @@
 import * as pg from "pg";
 
 import { Credentials } from "df/api/commands/credentials";
-import { IDbAdapter } from "df/api/dbadapters/index";
+import { IDbAdapter, IDbClient } from "df/api/dbadapters/index";
 import { SSHTunnelProxy } from "df/api/ssh_tunnel_proxy";
 import { parseRedshiftEvalError } from "df/api/utils/error_parsing";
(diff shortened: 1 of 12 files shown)


api/dbadapters/postgres.ts
```diff
@@ -63,15 +63,33 @@ export class PostgresDbAdapter implements IDbAdapter {
...
-      }
-      throw new ErrorWithCause(`Error executing postgres query: ${e.message}`, e);
-    }
+    return await this.withClientLock(executor => executor.execute(statement, options));
+  }
+
+  public async withClientLock<T>(callback: (client: IDbClient) => Promise<T>) {
+    return await this.queryExecutor.withClientLock(
+      async client =>
+        await callback({
```
GJMcGowan, line 72: Do we really need to have a nested async await here?

GJMcGowan (approved):

api/dbadapters/postgres.ts
BenBirt: nice, cleaned up

PR #986, opened 2020-09-08
Add a better (than Yargs) flag library.

BenBirt:


common/flags/BUILD
```diff
@@ -0,0 +1,15 @@
+package(default_visibility = ["//visibility:public"])
+
+load("//tools:ts_library.bzl", "ts_library")
+
+ts_library(
```
lewish, line 5: as there are no tests (fair) you may want to add a `build_test` rule for this

common/flags/index.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  public static getRawFlagValue(flagName: string) {
+    return Flags.parsedArgv[flagName];
+  }
+
+  public static setRawFlagValueForTesting(flagName: string, value: string) {
+    Flags.parsedArgv[flagName] = value;
+  }
+
+  private static readonly parsedArgv = (() => {
```
lewish, line 42: What happens if this throws? Will it just immediately kill the process? Assume that's what we want it to, just hope it doesn't manifest as an uncaught exception or something weird

lewish (approved):

common/flags/index.ts
BenBirt: Yup, just tested, it does immediately kill the process.

PR #987, opened 2020-09-09
Hey Presto

Ekrekr:
For now there's a lot of TODOs as placeholders for things that need some more thought, and maybe design docs.


BenBirt (commented):
Happy to have this in, but can we hold off until we have the adapter ready to go?

Ekrekr:
> Happy to have this in, but can we hold off until we have the adapter ready to go?

Sure thing, just wanted to split it up a bit

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,228 @@
...
+          ...evaluationResponse,
+          incremental: validationQuery.incremental,
+          query: validationQueriesWithoutWrappers[index].query
+        })
+      );
+    }
+    return queryEvaluations;
+  }
+
+  public async catalogs(): Promise<string[]> {
```
BenBirt, line 101: should be `private`

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,228 @@
...
+    const result = await prestoExecute(this.client, "show catalogs");
+    const catalogs = flatten(result.data);
+    return catalogs;
+  }
+
+  public async schemas(): Promise<string[]> {
```
BenBirt, line 107: IMO this doesn't quite follow the spec, especially in conjunction with `createSchema` below. `catalog: string` should be an argument here, and then it should be used as the `catalog` in the `show schemas` query.

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,228 @@
...
+    let schemas: string[] = [];
+    await Promise.all(
+      (await this.catalogs()).map(async catalog => {
+        const result = await prestoExecute(this.client, `show schemas from ${catalog}`);
+        schemas = schemas.concat(flatten(result.data).map(schema => `${catalog}.${schema}`));
+      })
+    );
+    // TODO: Currently this returns `catalog.schema` rather than just schemas in order to
```
BenBirt, line 115: i don't really understand this TODO. I think that, for presto, `catalog` is what we consider `database`, and `schema` is 1:1 with `schema`. I don't understand why we're prefixing schema names with their catalog

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,228 @@
...
+    // conform to string return type. Should this be changed? In tables() it is converted to targets.
+    return schemas;
+  }
+
+  public async createSchema(catalog: string, schema: string): Promise<void> {
+    await prestoExecute(this.client, `create schema if not exists ${catalog}.${schema}`);
+  }
+
+  // TODO: This should take parameters to allow for retrieving from a specific catalog/schema.
```
BenBirt, line 124: while this is true... it's also true for BQ, snowflake, etc. if you want to keep this TODO, I think it should be moved to the interface.

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,228 @@
...
+    );
+    return tables;
+  }
+
+  public async table(target: dataform.ITarget): Promise<dataform.ITableMetadata> {
+    let columnsResult: IPrestoExecutionResult;
+    try {
+      columnsResult = await prestoExecute(this.client, `describe ${resolveTarget(target)}`);
+    } catch (e) {
+      // This probably failed because the table doesn't exist, so ignore as the information isn't necessary.
```
BenBirt, line 144: this 'probably' is a little scary, if it fails due to some other error (but the table does exist), we can easily end up e.g. corrupting incremental table state! can we do some string matching on the error at least?

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  public async preview(target: dataform.ITarget, limitRows: number = 10): Promise<any[]> {
+    const { rows } = await this.execute(
+      `select * from ${target.database}.${target.schema}.${target.name} limit ${limitRows}`
+    );
+    return rows;
+  }
+
+  public async persistedStateMetadata(): Promise<dataform.IPersistedTableMetadata[]> {
+    // TODO.
```
BenBirt, line 168: nit: let's be consistent with the other DB adapters, rm these 'TODO's (which may never get done) and write `// Unimplemented.`

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,228 @@
...
+): Promise<IPrestoExecutionResult> {
+  return new Promise((resolve, reject) => {
+    const result: IPrestoExecutionResult = {};
+    client.execute({
+      query: statement,
+      cancel: () => {
+        return false;
+      },
+      state: (error: any, queryId: string, stats: Presto.IPrestoClientStats) => {
+        result.error = error;
```
BenBirt, line 197: IMO: if an error is ever set, we should be immediately `reject`-ing this promise, as that conforms to the DB adapter interface spec. if it really is `any`, then use `coerceAsError`. i don't think the `IPrestoExecutionResult` interface should have an `error` field, IMO it's a bit confusing.

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,228 @@
...
+        stats: Presto.IPrestoClientStats
+      ) => {
+        result.error = error;
+        result.data = data;
+        result.columns = columns;
+        result.stats = stats;
+      },
+      success: (error: any, stats: Presto.IPrestoClientStats) => {
+        if (!!error) {
+          reject(error);
```
BenBirt, line 218: `return` after this

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,228 @@
...
+        result.columns = columns;
+      },
+      data: (
+        error: any,
+        data: Presto.PrestoClientColumnDatum[],
+        columns: Presto.IPrestoClientColumnMetaData[],
+        stats: Presto.IPrestoClientStats
+      ) => {
+        result.error = error;
+        result.data = data;
```
BenBirt, line 212: this isn't right, the API docs say that `data` may be called >1 times so this just overwrites previous data fetches.

BenBirt, line 212: please use `LimitedResultSet`, in conjunction with an `isCancelled` boolean variable returned by `cancel`. (this is important to stop bad queries from crashing DF web backends).

common/arrays/arrays.ts
```diff
@@ -1,5 +1,7 @@
 export function flatten<T>(nestedArray: T[][]) {
```
BenBirt, line 1: we shouldn't even need this function any more, we can use `Array.flat()` now.

core/adapters/index.ts
```diff
@@ -73,13 +75,16 @@ const requiredSQLDataWarehouseProps: Array<keyof dataform.ISQLDataWarehouse> = [
   "password",
   "database"
 ];
+// TODO: Add SSL.
```
BenBirt, line 78: this TODO probably belongs in the protobuf.

core/adapters/presto.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    if (table.type === "incremental") {
+      throw new Error("Incremental table types are not currently supported for Presto.");
+    } else {
+      tasks.add(Task.statement(this.createOrReplace(table)));
+    }
+
+    this.postOps(table, runConfig, tableMetadata).forEach(statement => tasks.add(statement));
+
+    if (runConfig.useSingleQueryPerAction) {
```
BenBirt, line 41: side note: we should probably get rid of this option now that it's been defaulted on for a long while!

core/adapters/presto.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  public createOrReplace(table: dataform.ITable) {
+    // TODO: Not all connectors support replacing/dropping tables, but DROP is supported.
+    // A solution should be thought about for handling presto connections to see whether the target
+    // catalog allows dropping. For now default to creating the table.
+    return this.createTable(table);
+  }
+
+  public mergeInto() {
+    // TODO: merge isn't supported by Presto. Instead a solution could be formed by using a temporary
```
BenBirt, line 88: this should throw until it's supported.

core/table.ts
```diff
@@ -153,6 +153,39 @@ export interface IBigQueryOptions {
...
   strictKeysOf<IBigQueryOptions>()(["partitionBy", "clusterBy", "updatePartitionFilter", "labels"]);
 
+/**
+ * Options for creating tables within Presto projects.
+ */
+export interface IPrestoOptions {
+  /**
+   * The key with which to partition the table. Typically the name of a timestamp or date column.
+   *
+   * For more information, read the partiton by docs for the connection you are using.
```
BenBirt, line 163:
```suggestion
   * For more information, read the partitioning documentation for the Presto connection in use.
```

core/table.ts
```diff
@@ -153,6 +153,39 @@ export interface IBigQueryOptions {
...
+   */
+  partitionBy?: string[];
+
+  /**
+   * The key with which to partition the table. Typically the name of a timestamp or date column.
+   *
+   * For more information, read the partiton by docs for the connection you are using.
```
BenBirt, line 170: ditto

core/table.ts
```diff
@@ -153,6 +153,39 @@ export interface IBigQueryOptions {
...
+   */
+
+  /**
+   * If catalog and schema are specified in the request, then leading prefixes
+   *
+   * can be ignored, e.g. "catalog.schema.table" can be referred to as just "table".
+   */
+  catalog?: string;
```
BenBirt, line 178: i don't think we should have these - let's align `database` (part of any action `config` block) with `catalog`, and we already have `schema` there too.

protos/core.proto
```diff
@@ -167,6 +167,13 @@ message SQLDataWarehouseOptions {
   string distribution = 1;
 }
 
+message PrestoOptions {
+  string catalog = 1;
+  string schema = 2;
+  repeated string partition_by = 3;
+  // TODO: Add map of configurations to different presto connections.
```
BenBirt, line 174: for my context: can you clarify what you mean here / how it would work? my presumption was that all that config lives inside Presto itself, such that if i query against some catalog, Presto uses whatever connection config it needs to for that catalog.

protos/profiles.proto
```diff
@@ -52,3 +52,15 @@ message SSHTunnel {
...
 }
+
+// TODO: Add TLS/SSL details.
+message Presto {
+  string host = 1;
+  int32 port = 2;
+  string user = 3;
+  // TODO: Catalog and schema are required until this PR gets merged,
+  // but should then be deprecated.
+  // https://github.com/tagomoris/presto-client-node/pull/40
```
BenBirt, line 63:
```suggestion
  // TODO: Deprecate these fields once https://github.com/tagomoris/presto-client-node/pull/40 is merged.
```

readme.md
```diff
@@ -34,6 +34,7 @@ Compile hundreds of data models in under a second using SQLX. SQLX extends your
 * Redshift
 * Postgres
 * Azure SQL data warehouse
+* Presto (pre-alpha)
```
BenBirt, line 37:
```suggestion
* Presto (under development)
```

tests/integration/presto.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+import { PrestoFixture, prestoTestCredentials } from "df/tools/presto/presto_fixture";
+
+suite("@dataform/integration/presto", { parallel: true }, ({ before, after }) => {
+  const _ = new PrestoFixture(1234, before, after);
+
+  let dbadapter: dbadapters.IDbAdapter;
+
+  before("create adapter", async () => {
+    dbadapter = await dbadapters.create(prestoTestCredentials, "presto", {
+      disableSslForTestsOnly: true
```
BenBirt, line 17: this isn't actually used, rm

tests/integration/presto.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    const actionMap = keyBy(executedGraph.actions, v => v.name);
+    expect(Object.keys(actionMap).length).eql(1);
+
+    for (const actionName of Object.keys(actionMap)) {
+      expect(actionMap[actionName].status).equals(dataform.ActionResult.ExecutionStatus.SUCCESSFUL);
+    }
+  });
+
+  test("catalog inspection", { timeout: 60000 }, async () => {
```
BenBirt, line 38: not sure this test is super useful (it seems to be more about investigating the docker image!). rm?

tools/presto/presto_fixture.ts
```diff
@@ -0,0 +1,68 @@
+import * as presto from "presto-client";
+
+import { execSync } from "child_process";
+import * as dbadapters from "df/api/dbadapters";
+import { prestoExecute } from "df/api/dbadapters/presto";
+import { sleep, sleepUntil } from "df/common/promises";
+import { IHookHandler } from "df/testing";
+
+const USE_CLOUD_BUILD_NETWORK = !!process.env.USE_CLOUD_BUILD_NETWORK;
```
BenBirt, line 9: can you refactor this, and the other fixture, to use a proper flag (shared in some common location) now that we have support for them in open-source?

tools/presto/presto_fixture.ts
```diff
@@ -0,0 +1,68 @@
...
+const DOCKER_CONTAINER_NAME = "presto-df-integration-testing";
+// We remap this to something more unique.
+const PRESTO_SERVE_PORT = 8080;
+
+export const prestoTestCredentials = {
```
BenBirt, line 14:
```suggestion
export const TEST_CREDENTIALS = {
```

tests/integration/presto.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+
+import * as dfapi from "df/api";
+import * as dbadapters from "df/api/dbadapters";
+import { dataform } from "df/protos/ts";
+import { suite, test } from "df/testing";
+import { compile, keyBy } from "df/tests/integration/utils";
+import { PrestoFixture, prestoTestCredentials } from "df/tools/presto/presto_fixture";
+
+suite("@dataform/integration/presto", { parallel: true }, ({ before, after }) => {
+  const _ = new PrestoFixture(1234, before, after);
```
BenBirt, line 11: `1234` should reference the port set in the presto fixture file, instead of being hardcoded here

tools/presto/presto_fixture.ts
```diff
@@ -0,0 +1,68 @@
...
+          `--name ${DOCKER_CONTAINER_NAME}`,
+          "-d",
+          `-p ${port}:${PRESTO_SERVE_PORT}`,
+          USE_CLOUD_BUILD_NETWORK ? "--network cloudbuild" : "",
+          "bazel/tools/presto:presto_image"
+        ].join(" ")
+      );
+
+      const dbadapter = await dbadapters.create({ ...prestoTestCredentials, port }, "presto", {
+        disableSslForTestsOnly: true
```
BenBirt, line 50: rm, not used

tools/presto/presto_fixture.ts
```diff
@@ -0,0 +1,68 @@
...
+
+      // Block until presto is ready to accept requests.
+      await sleepUntil(async () => {
+        try {
+          await dbadapter.execute("select 1");
+          return true;
+        } catch (e) {
+          return false;
+        }
+      }, 500);
```
BenBirt, line 61: half a second so sloooow, `100`?

tools/presto/presto_fixture.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export const prestoTestCredentials = {
+  host: "127.0.0.1",
+  port: 1234,
+  user: "df-integration-tests",
+  catalog: "memory",
+  schema: "default"
+};
+
+export class PrestoFixture {
```
BenBirt, line 22: looks like a lot of this can be common-ified into a `DockerFixture`?

BenBirt (changes requested):

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,228 @@
...
+  private client: Presto.Client;
+
+  private constructor(credentials: Credentials, options?: { concurrencyLimit?: number }) {
+    this.prestoCredentials = credentials as dataform.IPresto;
+    this.client = new Presto.Client(this.prestoCredentials as Presto.IPrestoClientOptions);
+  }
+
+  public async execute(
+    statement: string,
+    // TODO: These execute options should actually allow any of Presto.IPrestoClientExecuteOptions.
```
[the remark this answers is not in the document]
Ekrekr: Agreed, on looking closer the other fields don't look very useful

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,228 @@
...
+    const queryEvaluations = new Array<dataform.IQueryEvaluation>();
+    for (const { index, validationQuery } of validationQueries) {
+      let evaluationResponse: dataform.IQueryEvaluation = {
+        status: dataform.QueryEvaluation.QueryEvaluationStatus.SUCCESS
+      };
+      try {
+        await prestoExecute(this.client, validationQuery.query);
+      } catch (e) {
+        evaluationResponse = {
+          // TODO: Parse the error with line number.
```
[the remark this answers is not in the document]
Ekrekr: Presto actually returns the line number itself which is nice!

api/dbadapters/presto.ts
Ekrekr: My thinking for having it public was for external users of `@dataform/api`, being able to query catalogs for Presto could be useful. What do you think?

api/dbadapters/presto.ts
Ekrekr: I made it an optional parameter so the default is to show everything, WDYT?

api/dbadapters/presto.ts
Ekrekr: I'm not sure what my thinking was here any more, I think this should throw in all cases

api/dbadapters/presto.ts
Ekrekr: SGTM

api/dbadapters/presto.ts
Ekrekr: Have updated it to `concat` data instead. I'm wondering of columns does the same thing. Having a full project for the integration test should catch  any weird behaviour here.

Ekrekr: I think I've done this right, it's starting to get a bit complex

common/arrays/arrays.ts
Ekrekr: Doesn't seem to be available? Isn't in any of our code base too.

protos/core.proto
Ekrekr: So in your initial connection you can specify a catalog and a schema if desired, then if you query `select * from table` it'll do the same as `select * from catalog.schema.table`. Currently catalog and schema are required by the client (but they shouldn't be), and the plan is for connections to **not have catalog or schema set in the connection at all**, so that there's no confusion about arbitrary table names.

 Having schema and catalog in the config seems like a nice way to allow users to have cross catalog files (default for if there's no catalog or schema defined), or specify a catalog and schema within the config if they don't want to have to write out the full target name for each table.

Ekrekr: If just catalog is provided then `select * from schema.table` works too

tests/integration/presto.spec.ts
Ekrekr: Once the test project is a bit bigger and `tables` can be tested after the project has been run then this can be removed in favor of that; for now it's already stopped me breaking a few things!

tools/presto/presto_fixture.ts
Ekrekr: I added Flags.string(), but I don't really understand where the value gets set

tools/presto/presto_fixture.ts
Ekrekr: If there comes a third docker image then will do!

protos/core.proto
Ekrekr: The current `database` and `schema` should work fine for this though, so there's no need to add presto specific proto options for them

api/dbadapters/presto.ts
Ekrekr: Oh I see now, when we run a project we call describe on the table in case there's useful metadata. Have added a check for the exact error code of the client

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,259 @@
...
+              query: statement,
+              cancel: () => {
+                // The presto client cancels executions if this returns true.
+                return isCancelled;
+              },
+              state: (error, queryId, stats) => {
+                if (cancelIfError(error)) {
+                  return;
+                }
+                result.queryId = queryId;
```
BenBirt, line 77: while this stuff may be useful in the future, i'm wondering if it's useful right now to keep `IPrestoExecutionResult` around at all. it's a lot of code to keep around when we immediately throw away all the non-`data` fields.

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,259 @@
...
+              const result = await this.execute(`show tables from ${catalog}.${targetSchema}`);
+              targets = targets.concat(
+                flatten(result.rows).map(table =>
+                  dataform.Target.create({
+                    database: catalog,
+                    schema: targetSchema,
+                    name: table as string
+                  })
+                )
+              );
```
BenBirt, line 187:
```suggestion
              targets.push(...flatten(result.rows).map(table =>
                  dataform.Target.create({
                    database: catalog,
                    schema: targetSchema,
                    name: table as string
                  })
              );
```

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,259 @@
...
+    try {
+      // TODO: In the future, the connection won't specify the catalog (database) and schema to allow
+      // moving data between tables, in which case target resolution should throw if both are not present.
+      columnsResult = await this.execute(`describe ${resolveTarget(target)}`);
+    } catch (e) {
+      // If the table doesn't exist just return empty metadata.
+      // In this circumstance the error will be of type Presto.IPrestoClientError.
+      if (e.errorName === "TABLE_NOT_FOUND") {
+        return dataform.TableMetadata.create({});
+      }
```
BenBirt, line 206: bug: this should re-throw `e` if the above condition is not true.

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,259 @@
...
+    return dataform.TableMetadata.create({
+      // TODO: Add missing table metadata items.
+      target,
+      fields
+    });
+  }
+
+  public async preview(target: dataform.ITarget, limitRows: number = 10): Promise<any[]> {
+    const { rows } = await this.execute(
+      `select * from ${target.database}.${target.schema}.${target.name} limit ${limitRows}`
```
BenBirt, line 224:
```suggestion
      `select * from ${resolveTarget(target)} limit ${limitRows}`
```

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,259 @@
...
+    options?: { limit: number }
+  ): Promise<dataform.ITableMetadata[]> {
+    // Unimplemented.
+    return [];
+  }
+
+  private async catalogs(): Promise<string[]> {
+    const result = await this.execute("show catalogs");
+    const catalogs: string[] = flatten(result.rows);
+    return catalogs;
```
BenBirt, line 257:
```suggestion
    return flatten(result.rows);
```

tests/integration/presto.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+  });
+
+  test("table target failure", async () => {
+    try {
+      await dbadapter.table({
+        database: "memory",
+        schema: "information_schema",
+        name: "nonexistant_table"
+      });
+      expect(false);
```
BenBirt, line 68:
```suggestion
      fail("...");
```

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,259 @@
...
+    let columnsResult: IExecutionResult;
+    try {
+      // TODO: In the future, the connection won't specify the catalog (database) and schema to allow
+      // moving data between tables, in which case target resolution should throw if both are not present.
+      columnsResult = await this.execute(`describe ${resolveTarget(target)}`);
+    } catch (e) {
+      // If the table doesn't exist just return empty metadata.
+      // In this circumstance the error will be of type Presto.IPrestoClientError.
+      if (e.errorName === "TABLE_NOT_FOUND") {
+        return dataform.TableMetadata.create({});
```
BenBirt, line 205: this should really return `null`, as per the other warehouses (otherwise various logic in the `build` and `run` steps will break in certain circumstances)

tests/integration/presto.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+  test("evaluation failure", async () => {
+    const failedEvaluation = await dbadapter.evaluate("select\nz\nas\nx");
+    expect(failedEvaluation.length).to.equal(1);
+    expect(failedEvaluation[0].status).to.equal(
+      dataform.QueryEvaluation.QueryEvaluationStatus.FAILURE
+    );
+    expect(failedEvaluation[0].error.errorLocation.line).to.equal(2);
+  });
+
+  test("table target failure", async () => {
```
BenBirt, line 61: i don't understand this test case. the code in the DB adapter suggests that no error should be thrown, and yet this catches one.

BenBirt (changes requested):
please remove `asd.tsx` from this PR?

api/dbadapters/presto.ts
```diff
@@ -0,0 +1,259 @@
...
+          new Promise((resolve, reject) => {
+            const result: IPrestoExecutionResult = { data: [] };
+            const allRows = new LimitedResultSet({
+              rowLimit: options.rowLimit,
+              byteLimit: options.byteLimit
+            });
+            const cancelIfError = (error: Presto.IPrestoClientError) => {
+              if (!!error) {
+                reject(error);
+                return (isCancelled = true);
```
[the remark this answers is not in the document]
Ekrekr: Good point, changed to rejectIfError

api/dbadapters/presto.ts
Ekrekr: Sure, let's just ignore the other stuff for now

tests/integration/presto.spec.ts
Ekrekr: Yep it's wrong!

api/dbadapters/presto.ts
```diff
@@ -70,44 +62,27 @@ export class PrestoDbAdapter implements IDbAdapter {
...
+                resolve(allRows.rows);
               },
-              error: cancelIfError
+              error: rejectIfError
             });
           })
       })
       .promise();
-    return { rows: prestoResult.data, metadata: {} };
+    return { rows: rows as any[], metadata: {} };
```
BenBirt, line 85: this `as any[]` is a code smell, we're just overriding TS's type checking. if you change `new Promise` to `new Promise<any[]>`, this shouldn't be necessary

tests/integration/presto.spec.ts
```diff
@@ -1,3 +1,4 @@
+import { fail } from "assert";
```
BenBirt, line 1: now not used?

common/arrays/arrays.ts
BenBirt: It's available, it's just that the open-source `tsconfig.json` doesn't include an up-to-date lib. If you change the relevant line in `tsconfig.json` to `"lib": ["es2019", "dom"]` (which would then match what we have in closed-source, where we use e.g. `flatMap`), then you can use it.

BenBirt (approved):

Ekrekr:
> It's available, it's just that the open-source tsconfig.json doesn't include an up-to-date lib. If you change the relevant line in tsconfig.json to "lib": ["es2019", "dom"] (which would then match what we have in closed-source, where we use e.g. flatMap), then you can use it.

Will do in follow up PR

Also deprecate `runConfig.useSingleQueryPerAction`

BenBirt:
Figured out what is going on. It's down to the behaviour of Docker networking:
1. when containers are running on the same user-defined bridge network (as they are in cloud build), they "effectively expose all ports to each other" (https://docs.docker.com/network/bridge/)
2. however, if you want to talk to a container on the default bridge network from the "outside world" (as in our tests running locally), you need to "publish" a port mapping (which is what we do with the `-p` flag)

When running locally, the integration test connects to port `2345`. This works fine, since `2345` has been published to the external world using the `-p` flag.

When running in cloud build (or locally, it's possible to simulate all this locally), the test code is running in a sibling container to the Presto container, both running inside a user-defined `cloudbuild` Docker bridge network. In this context, publishing port `2345` has no visible effect to the test code, because it's not in "the outside world", it's a sibling container to the Presto container, on the same network.

The reason all this works for Postgres is because the published port is the same as Postgres' serving port (`5432`). So locally it connects using the published port, and in cloud build it connects using the "real" port.

The reason this doesn't work for Presto is because we try to publish `8080` (Presto's default serving port) as `2345`. So locally it connects using the published port, but in cloud build the code tries to connect to `2345` which nothing is serving on.

I think we should:
- always use the same port number when publishing one, i.e. `-p X:X` where `X` is the port number
- reconfigure Presto to use a non-`8080` port number (I managed this, I need to clean up the change as configuring Presto is a right pain)

Ekrekr:
Why individual file mounting is working locally but not on cloud build:

The error `ERROR: [Errno 21] Is a directory: '/etc/presto/jvm.config'` is being given. The error message isn't clear about why it's a directory; the `jvm.config` file hasn't been found when trying to mount the file, so a directory is created instead.

PR #989, opened 2020-09-10
Vscode [Now publishing]

GJMcGowan:


vscode/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
lewish, line 1: I don't think you need to repeat this here if you aren't including it in the package itself.
If you intended to included it, you'll need to add it to the packager data deps

vscode/BUILD
```diff
@@ -32,3 +32,15 @@ sh_binary(
         ":vscode-sources",
     ],
 )
+
+sh_binary(
+    name = "packager",
+    srcs = ["packager.sh"],
+    data = [
+        ":language-configuration.json",
+        ":dataform_logo.png",
```
lewish, line 41: Is this from an existing file in repo?

If so, rather than copying the file, you can use the bazel rule in tools/common to copy the file from where it was to here

vscode/package.json
```diff
@@ -1,13 +1,26 @@
...
     "workspaceContains:**/*.sqlx"
   ],
   "main": "extension.js",
+  "dependencies": {
+    "vscode-languageclient": "^6.1.3",
+    "vscode-languageserver": "^6.1.1",
+    "vscode-languageserver-textdocument": "^1.0.1"
+  },
   "contributes": {
     "commands": [
```
lewish, line 25: Can we clean the commands a bit? I'd suggest:

title: "Compile project"
category: "Dataform"

vscode/LICENSE
lewish: (you can also use copy_file like for the image)

lewish (commented):
Can you remove the .vsix file from the git repo plz?

vscode/BUILD
GJMcGowan: It is not, I had to make a logo that was the correct size

vscode/LICENSE
GJMcGowan: So I don't think I do need to include this in the package data deps, because it's appearing on the marketplace site when I publish: https://marketplace.visualstudio.com/items/dataform.dataform/license

GJMcGowan: In fact I think what's happening here is that the `vsce publish` command I run takes some info that isn't from the packaged bundle, just from the directory it's in. So I think I need this file here (unless/until we write a bazel publish command)

lewish:
Can you remove the VSIX file from the repo plz?

vscode/package.json
```diff
@@ -1,19 +1,32 @@
...
       {
         "command": "dataform.compile",
-        "title": "Recompile dataform",
-        "category": "cli"
+        "title": "Compile project",
+        "category": "dataform"
```
lewish, line 29: Everything else seems to capitalize, so let's do `Dataform`

vscode/package.json
```diff
@@ -1,19 +1,32 @@
 {
   "name": "dataform",
```
lewish, line 2: nit: also seems to be mostly capitalized, so let's do that

vscode/package.json
```diff
@@ -1,19 +1,32 @@
...
-  "version": "0.0.1",
+  "categories": [
```
lewish, line 3: Can we add a description?

```
Syntax highlighting, compilation, and intellisense for Dataform and SQLX projects.
```

lewish (approved):

vscode/package.json
GJMcGowan: The name is required to be lowercased, we can add a displayName if we want

PR #1004, opened 2020-09-21
Remove primitiveDeprecated

Ekrekr:
Progresses https://github.com/dataform-co/dataform-co/issues/7904

Follow on from https://github.com/dataform-co/dataform-co/pull/7934

### DIFF
--- a/api/dbadapters/bigquery.ts
+++ b/api/dbadapters/bigquery.ts
@@ -513,10 +513,12 @@ function convertField(field: TableField): dataform.IField {
   };
   if (field.type === "RECORD" || field.type === "STRUCT") {
     result.struct = dataform.Fields.create({
-      fields: field.fields.map(innerField => convertField(innerField))
+      fields: field.fields.map(innerField => convertField(innerField)),
+      type: convertStructuredType(field.type)
     });
+  } else {
+    result.primitive = convertFieldType(field.type);
   }
-  result.primitive = convertFieldType(field.type);
   return dataform.Field.create(result);
 }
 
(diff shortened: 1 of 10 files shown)


api/dbadapters/bigquery.ts
```diff
@@ -515,10 +515,8 @@ function convertField(field: TableField): dataform.IField {
     result.struct = dataform.Fields.create({
       fields: field.fields.map(innerField => convertField(innerField))
     });
-  } else {
-    result.primitiveDeprecated = field.type;
-    result.primitive = convertFieldType(field.type);
   }
+  result.primitive = convertFieldType(field.type);
```
lewish, line 519: primitive and struct fields are in a oneof, so setting this will erase the struct information.

The point of this was that record/struct isn't a primitive type, and depending on if primitive or struct is set, you know that type of field. I would personally suggest keeping it this way as lots of code expects this to be the case.

Ekrekr: Ah I didn't see that `oneof`! I'll think of a different way of differentiating

Ekrekr: Putting a type inside fields seems to be a nice way

lewish (commented):
Sorry but I'm really not a fan of this. It's duplicate data, duplicate data is the enemy!

Store information once. You already have the information in this proto, based on the value of the oneof, adding another type field just gives us a potential route to these things not matching and causing bugs down the line.

I acknowledge oneofs in protobuf.js are ugly, but IMO this isn't the solution to that

lewish (commented):
Also worth pointing out that I don't think there should be both record and struct types, they are the same thing

Ekrekr:
> Also worth pointing out that I don't think there should be both record and struct types, they are the same thing

Ok let's change this to just remove `primitiveDeprecated` then

lewish (approved):

Issue #966, opened 2020-08-26
[Redshift] intermittent conflict transactions if multiple views feed into another view


lewish:
With a dependency graph such as:

```
A
   \
     C
   /
B
```
Where C depends on A and B, and all are views - the drop statements for A and B can run into conflicts, as they both run `drop cascade`, and will both be triggered at pretty much exactly the same time.

Cascade is required as C may have a schema binding to the source views, by using the `bind: true` option.

I believe a workaround for this situation is to turn C into a `table`.

Pending some more exploration to look at workarounds.

BenBirt:
Related issue, we should consider when fixing this: https://github.com/dataform-co/dataform-co/issues/7798

BenBirt:
- Some thoughts:
    - Remove CASCADE: not workable as long as ‘bind’ is an option, because it makes it impossible to drop tables/views before re-creating them.
    - Remove ‘bind’: workable. Users may create bound views outside of DF that depend on DF dataset, but that’s up to them to correctly manage.
    - Only run CASCADE when a dataset has dependents with ‘bind: true’: I don’t think this will reduce the occurrence of conflicts, because CASCADE should already only be considering such datasets.
    - Try to run DROP CASCADE commands in a mutually exclusive way: this is really hard to implement, and basically impossible to guarantee.
        - Hard to implement:
            - It’s hard to have complete knowledge of all dependent views. We could rely on ‘bind: true’ for this, but (a) it’s versioned (so if somebody removes ‘bind’, we won’t know to DROP CASCADE for the previously-existing view), and (b) we don’t necessarily have complete knowledge of view dependencies, because we only know about things that users `ref` (i.e. does not include explicitly named inputs or - worse - `resolve` inputs). However, we could determine the data at build-time, by querying Redshift to know which views are dependent on which datasets.
            - We’d then have to somehow make execution of potentially-conflicting statements mutually exclusive. This would require retaining metadata in the ExecutionGraph, explicitly marking DROP statements as such. We’d then have to ensure that no two conflicting DROP statements run concurrently, or more simply never run two DROP statements in parallel.
        - Impossible to guarantee:
            - Concurrent runs (on different parts of the graph) may run conflicting DROP statements.
    - Recommend people use tables instead of bound views

BenBirt:
[...]

PR #1008, opened 2020-09-23
Remove support for bound views in Redshift.

BenBirt:
fixes https://github.com/dataform-co/dataform/issues/966
fixes https://github.com/dataform-co/dataform-co/issues/7798

Once this is published, I'll announce it on Slack. On the plus side, anybody using `bind` will get a compilation error when they upgrade, so nothing will silently change for them.


version.bzl
```diff
@@ -1,3 +1,3 @@
 # NOTE: If you change the format of this line, you must change the bash command
 # in /scripts/publish to extract the version string correctly.
-DF_VERSION = "1.10.0"
+DF_VERSION = "1.10.1"
```
lewish, line 3: Would suggest this is `1.11` at least

lewish (approved):

PR #1013, opened 2020-09-28
Snowflake error parsing fix & test

Ekrekr:
Fixes https://github.com/dataform-co/dataform-co/issues/7723

Correct me if I'm wrong:
* Our SQL code preview component line and column number start at 1.
* BQ's column and line number start at 1.
* Snowflakes column number starts at 0 but line number start at 1.


tests/api/api.spec.ts
```diff
@@ -1366,6 +1367,28 @@ postOps`
       expect(result.actions[0].tasks[0].errorMessage).to.match(/cancelled/);
     });
   });
+
+  suite("error_parsing", () => {
```
BenBirt, line 1371: why not move this case to each of the integration tests, then we'll be testing against the actual data that snowflake/BQ/whatever returns?

BenBirt (approved):

PR #1015, opened 2020-09-30
Publish a better CLI Docker image.

BenBirt:
- include Git, which is important for Git-based NPM deps
- publish a `latest` tag


Ekrekr (approved):
Lewis probably has an opinion on not using slim

PR #1062, opened 2020-10-22
Compilation error on trailing semicolons in SQL statements in the core adapter

Ekrekr:
Fixes https://github.com/dataform-co/dataform-co/issues/8400

### DIFF
--- a/core/assertion.ts
+++ b/core/assertion.ts
@@ -12,6 +12,7 @@ import {
   resolvableAsTarget,
   setNameAndTarget,
   strictKeysOf,
+  throwIfTrailingSemicolonInQuery,
   toResolvable
 } from "df/core/utils";
 import { dataform } from "df/protos/ts";
@@ -101,6 +102,7 @@ export class Assertion {
   }
 
   public query(query: AContextable<string>) {
+    throwIfTrailingSemicolonInQuery(this.session, query.toString());
     this.contextableQuery = query;
     return this;
   }
--- a/core/assertion.ts
+++ b/core/assertion.ts
@@ -12,8 +12,8 @@ import {
   resolvableAsTarget,
   setNameAndTarget,
   strictKeysOf,
-  throwIfTrailingSemicolonInQuery,
-  toResolvable
+  toResolvable,
+  validateQueryString
 } from "df/core/utils";
 import { dataform } from "df/protos/ts";
 
@@ -102,7 +102,6 @@ export class Assertion {
   }
 
   public query(query: AContextable<string>) {
-    throwIfTrailingSemicolonInQuery(this.session, query.toString());
     this.contextableQuery = query;
     return this;
   }
@@ -160,8 +159,8 @@ export class Assertion {
   public compile() {
     const context = new AssertionContext(this);
 
-    const appliedQuery = context.apply(this.contextableQuery);
-    this.proto.query = appliedQuery;
+    this.proto.query = context.apply(this.contextableQuery);
+    validateQueryString(this.session, this.proto.query);
 
     return this.proto;
   }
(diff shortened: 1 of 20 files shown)


core/assertion.ts
```diff
@@ -101,6 +102,7 @@ export class Assertion {
   }
 
   public query(query: AContextable<string>) {
+    throwIfTrailingSemicolonInQuery(this.session, query.toString());
```
BenBirt, line 105: that `.toString()` is not going to work for anybody passing in an `AContextable` function.

this check should probably instead happen inside `.compile()`, after we have the pure, post-compilation query string.

core/utils.ts
```diff
@@ -219,3 +219,12 @@ export function checkExcessProperties<T>(
     );
   }
 }
+
+export function throwIfTrailingSemicolonInQuery(session: Session, query: string) {
```
BenBirt, line 223: nit: IMO this should not be called `throw`, since it doesn't.

`validateQueryString`?

core/utils.ts
```diff
@@ -219,3 +219,12 @@ export function checkExcessProperties<T>(
...
+  query = query.trim();
```
BenBirt, line 224: nit: just inline this

core/utils.ts
```diff
@@ -219,3 +219,12 @@ export function checkExcessProperties<T>(
...
+  if (query.charAt(query.length - 1) === ";") {
+    session.compileError(
+      "Semi-colons are not allowed at the end of SQLX table or assertion statements"
```
BenBirt, line 227: you should really wrap this up in an `Error` so that `compileError` can investigate the stacktrace and correctly attach the file which causes the error.

`session.compileError(new Error("..."));`

BenBirt, line 227: this error should also probably not reference `SQLX` specifically, as users may be using the plain JS API.

core/utils.ts
Ekrekr: This was on purpose, otherwise I have to call `.trim()` twice, which seems more messy

core/operation.ts
```diff
@@ -193,6 +193,7 @@ export class Operation {
 
     const appliedQueries = context.apply(this.contextableQueries);
     this.proto.queries = typeof appliedQueries === "string" ? [appliedQueries] : appliedQueries;
+    this.proto.queries.forEach(query => validateQueryString(this.session, query));
```
Ekrekr, line 196: I added this for operations too, but it's not actually needed is it, because we never append SQL to the end of operations? WDYT?

BenBirt: yeah personally, I liked the way you had it originally - leave operations as-is, because we never futz with them.

core/utils.ts
```diff
@@ -219,3 +219,14 @@ export function checkExcessProperties<T>(
     );
   }
 }
+
+export function validateQueryString(session: Session, query: string) {
+  query = query.trim();
+  if (query.charAt(query.length - 1) === ";") {
```
BenBirt, line 225: personally i like:

```suggestion
  if (query.trim().slice(-1)[0] === ";") {
```

but up to you

core/utils.ts
```diff
@@ -219,3 +219,14 @@ export function checkExcessProperties<T>(
...
+    session.compileError(
+      new Error(
+        "Semi-colons are not allowed at the end of SQL statements undergoing contextualization, as this can break appended adapter specific SQL"
```
BenBirt, line 228:
```suggestion
        "Semi-colons are not allowed at the end of SQL statements"
```

BenBirt (approved):

BenBirt:
you also need to bump the version

Issue #997, opened 2020-09-16
It should be possible to escape ${


BenBirt:
Context: https://dataform-users.slack.com/archives/CHMJK8MKM/p1600255751059600

BenBirt:
For the reasons listed in https://github.com/dataform-co/dataform/pull/1064, I suspect this is something we actually don't want to do - or at least risks being able to undo what is arguably a mistake in SQLX.

For now, users will just have to write `${"${"}` which isn't too bad.

PR #1064, opened 2020-10-23
Make it possible to escape dollar-brace in SQLX.

BenBirt:
fixes https://github.com/dataform-co/dataform/issues/997

- treat `\${` as plain text, i.e. the parser should not jump to a new state when seeing a `${` preceded by a `\`
- when escaping `\`, only escape if it is not followed by a `${`, i.e. `\${` should be left as-is

### DIFF
--- a/sqlx/tslint.json
+++ b/sqlx/tslint.json
@@ -1,7 +1,8 @@
 {
   "extends": ["../tslint.json"],
   "rules": {
+    "tsr-detect-non-literal-fs-filename": false,
     "tsr-detect-possible-timing-attacks": false,
-    "tsr-detect-non-literal-fs-filename": false
+    "tsr-detect-unsafe-regexp": false
   }
 }
--- a/core/compilers.ts
+++ b/core/compilers.ts
@@ -263,20 +263,20 @@ function createEscapedStatements(nodes: Array<string | SyntaxTreeNode>) {
   return results;
 }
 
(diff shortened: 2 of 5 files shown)


sqlx/tslint.json
```diff
@@ -1,7 +1,8 @@
 {
   "extends": ["../tslint.json"],
   "rules": {
+    "tsr-detect-non-literal-fs-filename": false,
     "tsr-detect-possible-timing-attacks": false,
-    "tsr-detect-non-literal-fs-filename": false
+    "tsr-detect-unsafe-regexp": false
```
Ekrekr, line 6: Just to confirm, this is OK because the SQLX stuff is run inside the VM sandbox?

Ekrekr (approved):
This might not work for cases with `\\${`, where the double backslash should actually be interpreted as a backslash, but it's not important I think

BenBirt:
I'm going to hold off merging this, I need to think about it a bit.

I'm concerned that this (as you say) makes it impossible to write `\\${` - or more correctly, what should that even mean?

We may have made a mistake in the language in the decision to automatically allow `\` to be interpreted literally, perhaps we should be forcing the user to manually escape it if they want a backslash, i.e. if they want a literal `\` then maybe they should be writing `\\`. And this PR may compound that mistake, making it harder to go back on that decision...

BenBirt:
I'm closing this for now.

PR #1069, opened 2020-10-26
Add protobuf utils to open source. Use them to clone RunResult before passing it to listeners.

BenBirt:
for https://github.com/dataform-co/dataform-co/issues/8280

### DIFF
--- a/api/commands/run.ts
+++ b/api/commands/run.ts
@@ -4,6 +4,7 @@ import Long from "long";
 import * as dbadapters from "df/api/dbadapters";
 import { retry } from "df/api/utils/retry";
 import { hashExecutionAction } from "df/api/utils/run_cache";
+import { deepClone } from "df/common/protos";
 import { timingSafeEqual } from "df/common/strings";
 import {
   JSONObjectStringifier,
@@ -99,7 +100,7 @@ export class Runner {
     this.eEmitter.setMaxListeners(0);
   }
 
-  public onChange(listener: (graph: dataform.IRunResult) => Promise<void> | void): Runner {
+  public onChange(listener: (graph: dataform.IRunResult) => void): Runner {
     this.changeListeners.push(listener);
     return this;
   }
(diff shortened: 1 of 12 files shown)


core/gen_index.ts
```diff
@@ -1,12 +1,14 @@
...
 
 export function genIndex(base64EncodedConfig: string): string {
-  const encodedGraphBytes = new Uint8Array(util.base64.length(base64EncodedConfig));
-  util.base64.decode(base64EncodedConfig, encodedGraphBytes, 0);
-  const config = dataform.GenerateIndexConfig.decode(encodedGraphBytes);
+  let config;
+  try {
+    config = decode(dataform.GenerateIndexConfig, base64EncodedConfig);
+  } catch (e) {
+    throw new Error("base64EncodedConfig :::" + base64EncodedConfig);
```
lewish: error message could be more useful

api/commands/run.ts
```diff
@@ -99,7 +100,7 @@ export class Runner {
     this.eEmitter.setMaxListeners(0);
   }
 
-  public onChange(listener: (graph: dataform.IRunResult) => Promise<void> | void): Runner {
+  public onChange(listener: (graph: dataform.IRunResult) => void): Runner {
```
lewish, line 103: This is quite a meaningful change, I'm sure we discussed it a lot in the past, it will change behaviour quite a bit! Should be fine as we debounce (I think??) these calls ourselves just wanted to check. Does mean runs should go a bit faster in the future.

One thing that could go wrong is, the way we use a debouncer, it will only write the change if no updates happen for some period of time. That could me runs with lots of fast queries and no breaks in updates could mean we don't write new results back as frequently?

lewish (approved):

core/gen_index.ts
BenBirt: sorry, this was me debugging, it's gone now

api/commands/run.ts
BenBirt: So closed source never actually returns a `Promise`, and hasn't for a very long time, so this is actually a no-op.

Runs with fast queries shouldn't be a problem, my debouncer always writes out a value after a specified period (where the start of that period is determine by the last write time): https://source.cloud.google.com/tada-analytics/github_dataform-co_dataform-co/+/master:ts/job-service/runs.ts;l=422?q=Debouncer

PR #1073, opened 2020-10-28
Hide snowflake hidden schemas in search

Ekrekr:
Fixes https://github.com/dataform-co/dataform-co/issues/8418

Information schemas aren't being hidden by the search query for the Snowflake API adapter, though they are hidden in the tableList query.

Can be replicated using our test snowflake project with this search:

Screenshot 2020-10-28 at 11 31 11

### DIFF
--- a/api/dbadapters/snowflake.ts
+++ b/api/dbadapters/snowflake.ts
@@ -208,7 +208,6 @@ where LOWER(table_schema) != 'information_schema'`,
           and tables.table_name = columns.table_name
         where tables.table_catalog ilike :1 or tables.table_schema ilike :1 or tables.table_name ilike :1 or tables.comment ilike :1
           or columns.column_name ilike :1 or columns.comment ilike :1
-          and LOWER(tables.table_schema) != 'information_schema' and LOWER(tables.table_schema) != 'pg_catalog' and LOWER(tables.table_schema) != 'pg_internal'
         group by 1, 2, 3
       ) where LOWER(table_schema) != 'information_schema'
        `,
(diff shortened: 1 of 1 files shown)


BenBirt (approved):

api/dbadapters/snowflake.ts
```diff
@@ -203,13 +203,18 @@ where LOWER(table_schema) != 'information_schema'
...
-         or columns.column_name ilike :1 or columns.comment ilike :1
-       group by 1, 2, 3
+      `select * from (
+        select tables.table_catalog as table_catalog, tables.table_schema as table_schema, tables.table_name as table_name
+        from information_schema.tables as tables
+        left join information_schema.columns as columns on tables.table_catalog = columns.table_catalog and tables.table_schema = columns.table_schema
+          and tables.table_name = columns.table_name
+        where tables.table_catalog ilike :1 or tables.table_schema ilike :1 or tables.table_name ilike :1 or tables.comment ilike :1
+          or columns.column_name ilike :1 or columns.comment ilike :1
+          and LOWER(tables.table_schema) != 'information_schema' and LOWER(tables.table_schema) != 'pg_catalog' and LOWER(tables.table_schema) != 'pg_internal'
```
BenBirt, line 213: hold up, this is weird - why are we excluding _postgres_ stuff in a snowflake DB?

BenBirt (changes requested):

api/dbadapters/snowflake.ts
Ekrekr: Nice catch! When I see something weird to do with Postgres I naturally assume it's exactly how it should be.

Looking around, I think that's not meant to be there; have taken it out of both list and search. I imagine the SQL for `tables()` was copied other from either redshift or postgres without removing this bit.

BenBirt (approved):

PR #1079, opened 2020-11-03
Fix BigQuery location

Ekrekr:
Fixes https://github.com/dataform-co/dataform-co/issues/8598

Functionality breaking change from bumping BQ NodeJS client to 5.3.0 to 5.4.0

From PR: https://github.com/googleapis/nodejs-bigquery/pull/864/files# for issue https://github.com/googleapis/nodejs-bigquery/issues/857

Location needs to be set in client options.

Location *may* also need to be passed in during the actual request, but hopefully not.


lewish (commented):
Thanks for fix!
This is the kind of thing that is going to be hard to keep on top of without a test.

Can we add one? Alternatively, just move all our integration tests to a non US location (you'll need to change the test credentials) which should be fairly easy.

We should really move BQ location into a project setting soon!

Ekrekr:
After this we'll safe to bump to 5.4, as [suggested by Stephen](https://github.com/googleapis/nodejs-bigquery/issues/888#issuecomment-722468680). Could do it in this PR if you'd prefer!

BenBirt:
I think doing it in this PR makes sense! Make sure that the upgrade to `5.4.0` fails on `master`, but succeeds on this PR?

sql/integration.spec.ts
```diff
@@ -13,7 +13,7 @@ suite("builders", { parallel: true }, ({ before, after }) => {
 
   before("create bigquery", async () => {
     bigquery = await dbadapters.create(
-      dfapi.credentials.read("bigquery", "test_credentials/bigquery.json"),
+      { ...dfapi.credentials.read("bigquery", "test_credentials/bigquery.json"), location: "EU" },
```
Ekrekr, line 16: This makes it more accurate to what we do in the web app. Previously we stored a location in the credentials, whereas what we actually do in the app is have people upload credentials (often with no location), then set the location or override it in the UI. This approach makes sense because it makes it easier to change location.

BenBirt (approved):

sql/integration.spec.ts
lewish: I don't think this is true? The location setting in the UI maps into the credentials object. https://github.com/dataform-co/dataform/blob/master/protos/profiles.proto#L44

For clarity, there are 2 types of credentials here:

- BigQuery service account credentials (emails, client secrets, billing project-id, no location info)
- Dataform BigQuery credentials (BQ service account credentials + location and project-id)

PR #1081, opened 2020-11-04
Fix getCallerFile().

BenBirt:
- make unable-to-find-caller-file `Error` actually throw; previously this could never happen: the issue is not that the returned `stack` is empty, but that the loop doesn't find a matching entry - the loop pops stuff off the stack, which is why we thought it sometimes came out empty
- when throwing an `Error` to figure out the current stack, override v8's default `stackTraceLimit` of `10`


Ekrekr (approved):
Nice find!

PR #1082, opened 2020-11-04
Move BigQuery integration tests to EU

Ekrekr:
Related to https://github.com/dataform-co/dataform-co/issues/8598

In prep for https://github.com/dataform-co/dataform/pull/1079


tests/integration/bigquery.spec.ts
```diff
@@ -15,7 +15,10 @@ suite("@dataform/integration/bigquery", { parallel: true }, ({ before, after })
   let dbadapter: BigQueryDbAdapter;
 
   before("create adapter", async () => {
-    dbadapter = (await dbadapters.create(credentials, "bigquery")) as BigQueryDbAdapter;
+    dbadapter = (await dbadapters.create(
+      { ...credentials, location: "EU" },
```
lewish, line 19: why do you need to do this if it's in the credentials file already?

lewish (approved):

tests/integration/bigquery.spec.ts
Ekrekr: Credentials uploaded to the web UI don't have location specified, so I've taken them out of the encoded credentials file too to make the tests closer to our actual use case

PR #1084, opened 2020-11-05
Stop validating pre/post ops. Start validating incremental queries.

BenBirt:
fixes https://github.com/dataform-co/dataform-co/issues/8591


Ekrekr (approved):
I think the issue isn't quite right; our adapter automatically adds a semi-colon if there isn't one there, so it's OK to not have a semi-colon.

But as we don't append SQL to pre or post operations there isn't anything wrong with getting rid of this either.

PR #1086, opened 2020-11-05
More sql functions

dwl285:
sql.timestamps.diff
sql.windowFunction
sql.surrogateKey

### DIFF
--- a/sql/index.ts
+++ b/sql/index.ts
@@ -145,16 +145,10 @@ export class Sql {
     if (this.dialect === "standard") {
       return this.asString(`farm_fingerprint(concat(${columnsAsStrings}))`);
     }
-    if (
-      this.dialect === "postgres" ||
-      this.dialect === "redshift" ||
-      this.dialect === "snowflake"
-    ) {
-      return this.asString(`md5(concat(${columnsAsStrings}))`);
-    }
     if (this.dialect === "mssql") {
       return this.asString(`hashbytes("md5", (concat(${columnsAsStrings})))`);
     }
+    return this.asString(`md5(concat(${columnsAsStrings}))`);
   }
 
   // window function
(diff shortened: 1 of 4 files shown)


sql/index.ts
```diff
@@ -132,12 +132,123 @@ export class Sql {
...
   }
 
+  // Surrogate keys.
+
+  public surrogateKey(columnNames: string[]) {
+    const columnsAsStrings = columnNames.map(id => this.asString(id)).join(`,`);
+    if (this.dialect === "standard") {
+      return this.asString(`farm_fingerprint(concat(${columnsAsStrings}))`);
+    }
+    if (
```
lewish, line 148: you could just leave this as the default, so special case standard, mssql, then afterwards just return this expression and remove the if.

sql/index.ts
```diff
@@ -132,12 +132,123 @@ export class Sql {
...
+      return `${name}(${value} ${ignoreNulls ? `ignore nulls` : ``}) over (${
+        windowSpecification.partitionFields ? `partition by ${partitionFieldsAsString}` : ``
+      } ${windowSpecification.orderFields ? `order by ${orderFieldsAsString}` : ``} ${
+        windowSpecification.frameClause ? windowSpecification.frameClause : ``
+      })`;
+    }
+
+    if (this.dialect === "redshift") {
+      // For some window functions in Redshift, a frame clause is required
+      const requiresFrame = [
```
lewish, line 189: AFAICT these lists are identical here and for postgres, so pull it out into a constant before this `if` and you can use it in both places

sql/integration.spec.ts
```diff
@@ -178,6 +178,116 @@ suite("builders", { parallel: true }, ({ before, after }) => {
...
+          ),
+          millisecond: sql.timestamps.diff(
+            "millisecond",
+            sql.timestamps.fromMillis("previous_millisecond"),
+            sql.timestamps.fromMillis("millis")
+          )
+        });
+
+        const result: any = await execute(query);
+        expect(result[0].day).equals(1);
```
lewish, line 284: you can do:

```
expect(result[0]).deep.equals({
    day: 1,
    minute: 1,
    ...
  });
```

lewish (approved):
Couple of small things otherwise LGTM!

Note that we don't have any tests for your postgres logic, as we don't actually set up a postgres test environment. Assume that's not a priority but we can add it at some point if you need.

PR #1089, opened 2020-11-09
Refactor run caching API.

BenBirt:
- no longer store persisted cache state in the warehouse; instead, the user must pass pairs of successful `ExecutionAction` and `ActionResult` protos obtained from previous run executions
- simplify `Runner` warehouse state management down to a single `Map` seeded in the `Runner` constructor and updated throughout run execution
- append new data to `ActionResult` protobufs (to be used in the run caching algorithm on the next execution of the associated `ExecutionAction`

Some nice side effects of this change:
- we can now immediately support run caching for Snowflake
- BigQuery run cache integration tests are no longer "optimistic", because we no longer rely on flaky BQ insert logic to store cache entries


api/commands/run.ts
```diff
@@ -413,36 +364,35 @@ export class Runner {
...
+    }
+    if (newMetadata) {
+      this.warehouseStateByTarget.set(action.target, newMetadata);
+      actionResult.postExecutionTimestampMillis = newMetadata.lastUpdatedMillis;
+      this.notifyListeners();
+    } else {
+      this.warehouseStateByTarget.delete(action.target);
+    }
+
+    if (actionResult.status === dataform.ActionResult.ExecutionStatus.RUNNING) {
```
lewish, line 394: This seems like a strange bit of code, do we not know this earlier?

tests/integration/bigquery.spec.ts
```diff
@@ -80,12 +80,19 @@ suite("@dataform/integration/bigquery", { parallel: true }, ({ before, after })
...
       await cleanWarehouse(compiledGraph, dbadapter);
 
-      // Drop the meta schema
-      await dbadapter.dropSchema("dataform-integration-tests", "dataform_meta");
-
       // Run the project.
       let executionGraph = await dfapi.build(compiledGraph, {}, dbadapter);
-      let executedGraph = await dfapi.run(dbadapter, executionGraph).result();
+      let runResult = await dfapi.run(dbadapter, executionGraph).result();
+      const previouslyExecutedActions = runResult.actions
```
lewish, line 86: it would be nice if the API returned something that could be passed immediately back in to the next run, rather than having each caller need to re-implement something like this. Not sure what that is exactly, but could be possible?

lewish (approved):
Very clean!

api/commands/run.ts
BenBirt: This is here to prevent us setting metadata if the action has already `FAILED`. I'll add a comment

tests/integration/bigquery.spec.ts
BenBirt: We can consider a new API going forward, but IMO the reality is that any caller who wants to use run caching is going to have to store some state _somewhere_, and so this code is going to look different for each such case. (e.g. Our closed source code doesn't look like this.)

PR #1094, opened 2020-11-09
add string agg function

dwl285:


sql/integration.spec.ts
```diff
@@ -241,6 +241,30 @@ suite("builders", { parallel: true }, ({ before, after }) => {
         expect(result[0].first_value).equals(1);
       });
 
+      if (false) {
```
dwl285, line 244: @lewish there is a quirk in redshift that you can't call listagg unless a user-defined table is included in the SQL. I was thinking we'd just skip redshift for this test, wdyt?

If so, how can I check whether it's redshift here?

lewish: Line 44 - this is all wrapped in a big for loop, so you can just do:

```
if (name === "redshift") { ... }
```

Could probably refactor "name" to "dialect".

If the function doesn't work for Redshift, I'd throw an error in the actual method to say that, e.g.:

```
if (this.dialect === "redshift") {
  throw new Error("String aggregation is not currently supported for Redshift.");
}
```

However, I'd question the usefulness of a function in here that doesn't support the 3 major dialects!

lewish (approved):

lewish (approved):

PR #1111, opened 2020-11-20
Fixes a bug where the compiler would return random stuff when asked to compile an empty string

GJMcGowan:


api/commands/query.ts
```diff
@@ -44,7 +44,7 @@ export async function compile(
   compileConfig?: dataform.ICompileConfig
 ): Promise<string> {
   // If there is no project directory, no need to compile the script.
-  if (!compileConfig || !compileConfig.projectDir) {
+  if (!compileConfig || !compileConfig.projectDir || query === "") {
```
BenBirt, line 47: i have a strong preference for fixing the bug as close to its source as possible; that must be somewhere inside `core` or inside the `returnOverride` code below. but let's please first figure out what is actually happening

core/gen_index.ts
```diff
@@ -26,8 +26,9 @@ export function genIndex(base64EncodedConfig: string): string {
     dataform.ProjectConfig.create(config.compileConfig.projectConfigOverride).toJSON()
   );
 
-  const returnValue = !!config.compileConfig.query
-    ? `(function() {
+  const returnValue =
```
BenBirt, line 29: urgh, i see the issue now, thanks for debugging.

the problem with this fix is that it's reliant on protobufjs mis-implementing the proto3 spec. that empty `query` field _should not be set_, and it's just luck that it is set.

can we instead a boolean field to `CompileConfig` called `compile_single_query` or something, to properly indicate that there is a query to be compiled (even if it's empty)?

BenBirt (approved):

PR #1114, opened 2020-11-23
Include 'database' in Targets returned from BigQueryDbAdapter.tables(...).

BenBirt:
fixes https://github.com/dataform-co/dataform-co/issues/8807

### DIFF
--- a/api/dbadapters/bigquery.ts
+++ b/api/dbadapters/bigquery.ts
@@ -15,6 +15,7 @@ import { LimitedResultSet } from "df/api/utils/results";
 import { coerceAsError } from "df/common/errors/errors";
 import { collectEvaluationQueries, QueryOrAction } from "df/core/adapters";
 import { dataform } from "df/protos/ts";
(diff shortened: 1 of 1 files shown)


api/dbadapters/bigquery.ts
```diff
@@ -167,7 +177,11 @@ export class BigQueryDbAdapter implements IDbAdapter {
           : metadata.type === "VIEW"
           ? dataform.TableMetadata.Type.VIEW
           : dataform.TableMetadata.Type.UNKNOWN,
-      target,
+      target: {
+        database: metadata.tableReference.projectId,
+        schema: metadata.tableReference.datasetId,
+        name: metadata.tableReference.tableId
+      },
```
Ekrekr, line 184: Could be worth adding a throw in if the input target and this derived one don't match, otherwise we could end up with a feedback loop of consistently referencing the wrong table when `table` is called.

Ekrekr (approved):

PR #1116, opened 2020-11-25
Fix database action overrides

Ekrekr:
Fixes https://github.com/dataform-co/dataform-web-tracking/issues/282

Issue was actually happening when schema was put as well as database in an action's config; the schema was resetting the database back to the default database as the database proto option wasn't set

Was also happening with assertions, not just operations. Also did some tidying too while at it.

### DIFF
--- a/core/assertion.ts
+++ b/core/assertion.ts
@@ -152,7 +152,13 @@ export class Assertion {
   }
 
   public schema(schema: string) {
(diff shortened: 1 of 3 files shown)


BenBirt (commented):
You need to bump the version

BenBirt (approved):

PR #1120, opened 2020-11-26
Make it possible to compile single queries that use ref/resolve's 'rest' argument form.

BenBirt:
fixes https://github.com/dataform-co/dataform-co/issues/8923

### DIFF
--- a/core/session.ts
+++ b/core/session.ts
@@ -19,6 +19,7 @@ import {
 } from "df/core/table";
 import * as test from "df/core/test";
 import * as utils from "df/core/utils";
+import { toResolvable } from "df/core/utils";
 import { version as dataformCoreVersion } from "df/core/version";
 import { dataform } from "df/protos/ts";
 
@@ -216,7 +217,8 @@ export class Session {
     }
   }
 
-  public resolve(ref: Resolvable): string {
+  public resolve(ref: Resolvable | string[], ...rest: string[]): string {
+    ref = toResolvable(ref, rest);
     const allResolved = this.findActions(utils.resolvableAsTarget(ref));
     if (allResolved.length > 1) {
       this.compileError(new Error(utils.ambiguousActionNameMsg(ref, allResolved)));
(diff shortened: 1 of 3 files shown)


tests/api/api.spec.ts
```diff
@@ -835,6 +835,18 @@ suite("@dataform/api", () => {
         "select 1 from `tada-analytics.df_integration_test.example_view`"
       );
     });
+    test("bigquery_example with explicit schema", async () => {
+      const compiledQuery = await query.compile(
+        'select 1 from ${ref("df_integration_test", "example_view")}',
+        {
+          projectDir: "examples/common_v1",
+          projectConfigOverride: { warehouse: "bigquery", defaultDatabase: "tada-analytics" }
```
Ekrekr, line 843:
```suggestion
          projectConfigOverride: { warehouse: "bigquery", defaultDatabase: "tada-analytics", defaultSchema: "not_a_schema" }
```

I think, otherwise this isn't testing the override actually happens?

core/session.ts
```diff
@@ -216,7 +217,8 @@ export class Session {
     }
   }
 
-  public resolve(ref: Resolvable): string {
+  public resolve(ref: Resolvable | string[], ...rest: string[]): string {
```
Ekrekr, line 220: Shouldn't the string here not be pluralised?
```suggestion
  public resolve(ref: Resolvable | string, ...rest: string[]): string {
```

Ekrekr (approved):

tests/api/api.spec.ts
BenBirt: no override is happening, it's just explicitly referencing a schema in `ref`. 

note that without the bugfix, this test does indeed fail.

core/session.ts
BenBirt: hmm, no.

we need to accept any of the following:
1) a `Resolvable`, i.e. a single string corresponding to the target's `name`; OR the full `Target`
2) an explicit array, i.e. `["db", "schema", "name"]`
3) that array but without the square brackets, i.e. using the `rest` parameter

PR #1122, opened 2020-11-26
Add full support for SQLX compilation to single-query compile requests.

BenBirt:
fixes https://github.com/dataform-co/dataform-co/issues/8925

A bit of explanation:
- in order to properly support various JS constructs, we really need to lex the query string, which is why the generated `index.js` file now calls into `compilers.ts`
- the generated `index.js` file then "calls" the compiled result inside a dynamically-constructed `Function`. (note that this is not strictly necessary, we could just `eval` the result, but `Function` makes it easier to handle multiple lines of query code)
- this means that it's possible to have full support for arbitrary JavaScript, and even `js {}` blocks
- note that the majority of the changes in `compilers.ts` is just moving code around, specifically everything inside `extractSqlxParts`; only `compileStandaloneSqlxQuery` is new

**NOTE**: this PR introduces a backwards-incompatible change. Certain kinds of specific queries that compiled before will no longer be compilable (until the user upgrades `@dataform/core` past this change); specifically, those that contain literal backslashes or backticks. I can't really see a way around this. On the plus side, this PR fully moves single-query compilation inside `@dataform/core`, so that future changes can be more safe across the `api` <-> `core` boundary.

### DIFF
--- a/core/compilers.ts
+++ b/core/compilers.ts
@@ -4,7 +4,10 @@ import { TableContext } from "df/core/table";
 import * as utils from "df/core/utils";
 import { SyntaxTreeNode, SyntaxTreeNodeType } from "df/sqlx/lexer";
 
-export function compile(code: string, path: string) {
+export function compile(code: string, path?: string) {
+  if (!path) {
+    return compileStandaloneSqlxQuery(code);
+  }
   if (path.endsWith(".sqlx")) {
     return compileSqlx(SyntaxTreeNode.create(code), path);
   }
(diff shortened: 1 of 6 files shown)


core/compilers.ts
```diff
@@ -4,7 +4,10 @@ import { TableContext } from "df/core/table";
 import * as utils from "df/core/utils";
 import { SyntaxTreeNode, SyntaxTreeNodeType } from "df/sqlx/lexer";
 
-export function compile(code: string, path: string) {
+export function compile(code: string, path?: string) {
+  if (!path) {
```
lewish, line 8: Can we just call the correct method from the caller? Changing the logic based on the presence of path or not is a bit strange to me

lewish (approved):

PR #1124, opened 2020-11-26
Add databases and schema options for multiple databases to the API adapter, introduce schema proto

Ekrekr:
Progresses https://github.com/dataform-co/dataform-web-tracking/issues/258

General tidying to allow showing schemas in multi-database warehouses

I think schemas should return partial targets in the future instead of just strings, so that it's clear which database they come from


api/dbadapters/bigquery.ts
```diff
@@ -232,9 +232,26 @@ export class BigQueryDbAdapter implements IDbAdapter {
     return rows;
   }
 
-  public async schemas(database: string): Promise<string[]> {
-    const data = await this.getClient(database).getDatasets();
-    return data[0].map(dataset => dataset.id);
+  public async databases(): Promise<string[]> {
+    throw new Error("BigQuery does not support multiple databases in the same connection");
```
BenBirt, line 236: i'm not sure this error message makes sense in the context of BQ.

it's not that it's got anything to do with the 'connection' - any connection can theoretically 'connect' to any BQ project, given the right permissions.

presumably we should just make the right API call to fetch all projects that this auth'd account has access to? https://cloud.google.com/bigquery/docs/reference/rest/v2/projects/list

api/dbadapters/index.ts
```diff
@@ -38,7 +38,8 @@ export interface IDbAdapter extends IDbClient {
     projectConfig?: dataform.IProjectConfig
   ): Promise<dataform.IQueryEvaluation[]>;
 
-  schemas(database: string): Promise<string[]>;
+  databases(): Promise<string[]>;
+  schemas(databases?: string | string[]): Promise<string[]>;
```
BenBirt, line 42: do we need the overload? can we just always require an array, for simplicity's sake?

BenBirt, line 42: IMO: now is the time to introduce a `Schema` type (which has `database` and `schema` fields)

api/dbadapters/bigquery.ts
```diff
@@ -232,9 +232,26 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+
+  public async schemas(database?: string | string[]): Promise<string[]> {
+    if (database === undefined) {
+      database = await this.databases();
+    }
+    if (typeof database === "string") {
+      database = [database];
+    }
+
+    let schemas: string[] = [];
```
BenBirt, line 247: nit: make this `const`, and use `schemas.push(...data);`

(throughout the PR)

api/dbadapters/postgres.ts
```diff
@@ -231,7 +231,11 @@ export class PostgresDbAdapter implements IDbAdapter {
     return rows;
   }
 
-  public async schemas(): Promise<string[]> {
+  public async databases(): Promise<string[]> {
+    throw new Error("Postgres does not support multiple databases in the same connection");
+  }
+
+  public async schemas(database?: string | string[]): Promise<string[]> {
```
BenBirt, line 238:
```suggestion
  public async schemas(): Promise<string[]> {
```

api/dbadapters/redshift.ts
```diff
@@ -209,7 +209,11 @@ export class RedshiftDbAdapter implements IDbAdapter {
     return rows;
   }
 
-  public async schemas(): Promise<string[]> {
+  public async databases(): Promise<string[]> {
+    throw new Error("Redshift does not support multiple databases in the same connection");
+  }
+
+  public async schemas(database?: string | string[]): Promise<string[]> {
```
BenBirt, line 216:
```suggestion
  public async schemas(): Promise<string[]> {
```

BenBirt (changes requested):

api/dbadapters/bigquery.ts
Ekrekr: Ah I see. This is pretty big to do I think, because credentials aren't user scoped, and we don't have a system for setting that up at the moment.

For now I'll just make the error message more clear.

protos/core.proto
```diff
@@ -144,6 +144,11 @@ message CompileConfig {
   string return_override = 5;
 }
 
+message Schema {
```
Ekrekr, line 147:
```suggestion
message SchemaTarget {
```
WDYT?

tests/api/api.spec.ts
```diff
@@ -1116,16 +1116,15 @@ select \${foo}
       expect(dataform.RunResult.create(cleanTiming(await runner.execute().result()))).to.deep.equal(
         EXPECTED_RUN_RESULT
       );
-      verify(mockedDbAdapter.createSchema("database", "schema1")).once();
-      verify(mockedDbAdapter.createSchema("database2", "schema2")).once();
+      verify(mockedDbAdapter.createSchema(anything())).twice();
```
Ekrekr, line 1119: Is there a way to catch match protos here so don't need to make it more ambiguous?

protos/core.proto
BenBirt: yeah `SchemaTarget` seems better to me, to align with the `Target` naming.

protos/core.proto
```diff
@@ -144,6 +144,11 @@ message CompileConfig {
...
+  string schema = 1;
+  string database = 2;
```
BenBirt, line 149: this should be before `schema` as in `Target`

api/dbadapters/bigquery.ts
```diff
@@ -232,41 +232,57 @@ export class BigQueryDbAdapter implements IDbAdapter {
     return rows;
   }
 
-  public async schemas(database: string): Promise<string[]> {
-    const data = await this.getClient(database).getDatasets();
-    return data[0].map(dataset => dataset.id);
+  public async databases(): Promise<string[]> {
+    throw new Error(
+      "The Dataform BigQuery API adapter does not (yet) support multiple databases in the same connection"
```
BenBirt, line 237: GH won't let me respond to the earlier comment chain (ugh), but in response: *all* GCP credentials are user-scoped. a service account is still a "user", it's just not a human.

i'm not going to block on adding `databases()` support for BQ connections, but it should be fairly simple, you should be able to issue the API call pretty easily right here.

tests/api/api.spec.ts
BenBirt: have you tried both `{database: ..., schema: ...}` and `dataform.Schema.create({ database: ..., schema: ...})` forms?

BenBirt (changes requested):

PR #1129, opened 2020-11-27
Add initial implementation of protoc-gen-ts.

BenBirt:


protobufts/descriptor.proto
```diff
@@ -0,0 +1,909 @@
+// Protocol Buffers - Google's data interchange format
```
lewish, line 1: Are these modified at all? If not, can we not pull them in directly via bazel rather than adding to source?

lewish: Like how we do this for e.g. `annotations.proto`

lewish (approved):
Approving for now but will take a longer pass when I can!

protobufts/descriptor.proto
BenBirt: nice suggestion, cleaned up a lot!

PR #1134, opened 2020-12-01
Move cache client to open source

Ekrekr:


common/protos/proto_utils.ts
```diff
@@ -0,0 +1,33 @@
+import { util } from "protobufjs";
+
+export class ProtoUtils {
```
BenBirt, line 3: https://source.cloud.google.com/tada-analytics/github_dataform-co_dataform/+/master:common/protos/index.ts?q=f:protos%20lang:ts

common/strings/stringifier.ts
```diff
@@ -282,3 +284,18 @@ export class StringifiedSet<T> implements Set<T> {
     })();
   }
 }
+
+export class ProtoStringifier<T> implements IStringifier<T> {
```
BenBirt, line 288: this probably more cleanly belongs inside https://source.cloud.google.com/tada-analytics/github_dataform-co_dataform/+/master:common/protos/index.ts?q=f:protos%20lang:ts too.

Ekrekr: I did that at first too, but it results in circular dependencies. Shouldn't bother for now I think

common/protos/index.ts
```diff
@@ -50,3 +50,35 @@ function fromBase64(value: string): Uint8Array {
   util.base64.decode(value, buf, 0);
   return buf;
 }
+
+export class ProtoUtils {
```
BenBirt, line 54: sorry, i think i wasn't clear. this file already contains all of these utilities, i don't think we need `ProtoUtils` at all.

common/strings/stringifier.ts
BenBirt: wait, why? that would imply that `strings/stringifier.ts` somehow depends on `protos/index.ts`, it really shouldn't, how does it?

Ekrekr: My bad, was making protos depend on itself, not strings

BenBirt (approved):

PR #1136, opened 2020-12-02
Downgrade handy redis to 1.8.3

Ekrekr:
Solution to fixing https://github.com/dataform-co/dataform-co/pull/9012

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "google-protobuf": "^3.6.1",
     "grpc": "^1.24.0",
     "grpc-web-client": "^0.5.0",
-    "handy-redis": "^2.0.2",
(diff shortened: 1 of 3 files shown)


BenBirt (approved):
I would've been happy if you'd just upped the closed-source version too :P

Ekrekr:
> I would've been happy if you'd just upped the closed-source version too :P

It broke a bunch of things, this seems easier!

Ekrekr:
> @Ekrekr I maintain handy-redis and happened to see this - was there an issue you found?

No, but thanks for reaching out. We maintain a duorepo for the majority of our code, and need package versions to match across the two. Downgrading was just quicker than upgrading the other repo.

PR #1139, opened 2020-12-07
Add validation (mostly bounds checking) on numeric field inputs.

BenBirt:


protoc-gen-ts/runtime/validation.ts
```diff
@@ -0,0 +1,89 @@
+import Long from "long";
+
+const TWO_TO_31 = Math.pow(2, 31);
+const NEGATIVE_TWO_TO_31 = -TWO_TO_31;
```
Ekrekr, line 4:
```suggestion
const MAX_32_BIT_INT = Math.pow(2, 31);
const MIN_32_BIT_INT = -TWO_TO_31;
```
To follow pattern of floats? Or have I interpreted this wrong

Ekrekr (approved):

protoc-gen-ts/runtime/validation.ts
BenBirt: No your naming makes sense, with a minor edit - will make!

PR #1146, opened 2020-12-11
Fix mac vm piping

Ekrekr:
Fixes https://github.com/dataform-co/dataform-co/issues/9099

Finally found a combination that works. I don't know why this works in particular but the others don't. What I really don't understand is why I can't have the parent process receive data using a socket rather than `this.childProcess.stdio[4]`.


api/commands/compile.ts
```diff
@@ -94,7 +95,13 @@ export class CompileChildProcess {
       // Handle UTF-8 string chunks returned by the child process.
       const pipe = this.childProcess.stdio[4];
       const chunks: Buffer[] = [];
-      pipe.on("data", (chunk: Buffer) => chunks.push(chunk));
+      // Typing issues here are from the 4th item in stdio (pipe) when developing on a mac.
```
BenBirt, line 98: i think you can: `import { Readable } from 'stream';`, and `pipe as Readable`

BenBirt (changes requested):

BenBirt:
You also need to bump the version so we can ship a new CLI package.

api/commands/compile.ts
```diff
@@ -92,9 +94,14 @@ export class CompileChildProcess {
       this.childProcess.on("message", (e: Error) => reject(coerceAsError(e)));
 
       // Handle UTF-8 string chunks returned by the child process.
-      const pipe = this.childProcess.stdio[4];
+      const pipe = this.childProcess.stdio[4] as Readable;
       const chunks: Buffer[] = [];
-      pipe.on("data", (chunk: Buffer) => chunks.push(chunk));
+      pipe?.on("readable", () => {
+        let buffer: any;
```
BenBirt, line 100: does this have to be `any` and not `Buffer`?

BenBirt (approved):

PR #1147, opened 2020-12-15
Fix IPC piping readable for mac CLI for larger compiled graphs

Ekrekr:
Fixes https://github.com/dataform-co/dataform-co/issues/9099

### DIFF
--- a/api/commands/compile.ts
+++ b/api/commands/compile.ts
@@ -97,10 +97,14 @@ export class CompileChildProcess {
       const pipe = this.childProcess.stdio[4] as Readable;
       const chunks: Buffer[] = [];
       pipe?.on("readable", () => {
-        let buffer: Buffer = pipe.read();
-        while (buffer) {
-          chunks.push(buffer);
+        let firstRead = true;
+        let buffer: Buffer;
+        while (buffer || firstRead) {
+          firstRead = false;
           buffer = pipe.read();
+          if (!!buffer) {
+            chunks.push(buffer);
+          }
         }
       });
 
--- a/api/commands/compile.ts
+++ b/api/commands/compile.ts
@@ -97,14 +97,10 @@ export class CompileChildProcess {
       const pipe = this.childProcess.stdio[4] as Readable;
       const chunks: Buffer[] = [];
       pipe?.on("readable", () => {
-        let firstRead = true;
-        let buffer: Buffer;
-        while (buffer || firstRead) {
-          firstRead = false;
+        let buffer: Buffer = pipe.read();
+        while (buffer) {
+          chunks.push(buffer);
           buffer = pipe.read();
-          if (!!buffer) {
-            chunks.push(buffer);
-          }
         }
       });
 
--- a/api/vm/compile.ts
+++ b/api/vm/compile.ts
@@ -1,12 +1,12 @@
 import * as net from "net";
 import * as path from "path";
+import { promisify } from "util";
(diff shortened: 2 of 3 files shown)


api/commands/compile.ts
```diff
@@ -97,10 +97,14 @@ export class CompileChildProcess {
       const pipe = this.childProcess.stdio[4] as Readable;
       const chunks: Buffer[] = [];
       pipe?.on("readable", () => {
-        let buffer: Buffer = pipe.read();
-        while (buffer) {
-          chunks.push(buffer);
+        let firstRead = true;
+        let buffer: Buffer;
+        while (buffer || firstRead) {
```
Ekrekr, line 102: Assignment in while loop not allowed, so can't do `while(buffer = pipe.read())`

api/vm/compile.ts
```diff
@@ -69,9 +69,20 @@ export function compile(compileConfig: dataform.ICompileConfig) {
...
+      let splitCompiledResult: string[] = [];
+      while (compiledResult.length > 0) {
+        splitCompiledResult.push(compiledResult.substring(0, 2 ** 12));
+        compiledResult = compiledResult.substring(2 ** 12);
+      }
+      for (let chunk of splitCompiledResult) {
+        writer.write(chunk, "utf8");
+        // This is needed for an unclear reason.
+        // tslint:disable-next-line: no-console
+        console.log();
```
Ekrekr, line 84: Any ideas why?

api/commands/compile.ts
```diff
@@ -97,10 +97,14 @@ export class CompileChildProcess {
...
+          firstRead = false;
           buffer = pipe.read();
+          if (!!buffer) {
+            chunks.push(buffer);
+          }
```
BenBirt, line 107:
```suggestion
        let buffer: Buffer = pipe.read();
        while (buffer) {
          chunks.push(buffer);
          buffer = pipe.read();
        }
```

api/vm/compile.ts
BenBirt: are you sure the actual problem isn't that you're just not waiting for the write to complete? `socket.write` takes a callback which is called when the write completes, seems like you should `await` / `promisify`: https://nodejs.org/api/net.html#net_socket_write_data_encoding_callback

api/commands/compile.ts
Ekrekr: Reading when there's nothing in the pipe throws an error; the second read can read when there's only one thing to read

Ekrekr:
Yup I was entirely wrong about the limit, can just pass the process exit in the callback 🤷‍♂️

api/vm/compile.ts
```diff
@@ -71,15 +71,14 @@ export function listenForCompileRequest() {
     try {
       const compiledResult = compile(compileConfig);
       const writer = new net.Socket({ fd: 4 });
-      writer.write(compiledResult);
+      writer.write(compiledResult, () => process.exit());
```
BenBirt, line 74: i think we should avoid this, it feels like a code smell. just `await` `promisify`?

api/commands/compile.ts
BenBirt: i think my code is equivalent to yours

api/vm/compile.ts
Ekrekr: Promisify doesn't seem to be valid on `writer.write`... or in any case I couldn't get it to work. How about this alternative?

BenBirt (approved):

PR #1152, opened 2020-12-17
Fix error piping from VM when using mac CLI

Ekrekr:
[This PR](https://github.com/dataform-co/dataform/commit/2b3ef4c1b28ba0e129b52fe629539ae74e55d689) caused a conflict between socket on fd:4 and process.send.

Instead swap to just using `process.send` for all IPC.

### DIFF
--- a/api/commands/compile.ts
+++ b/api/commands/compile.ts
@@ -89,22 +89,17 @@ export class CompileChildProcess {
 
   public async compile(compileConfig: dataform.ICompileConfig) {
     const compileInChildProcess = new Promise<string>(async (resolve, reject) => {
-      let encodedCompiledGraph: string;
-      // Handle any Error caused by spawning the child process, or sent directly from the child process.
       this.childProcess.on("error", (e: Error) => reject(coerceAsError(e)));
+
       this.childProcess.on("message", (messageOrError: string | Error) => {
         if (typeof messageOrError === "string") {
-          encodedCompiledGraph = messageOrError;
-          return;
+          resolve(messageOrError);
         }
         reject(coerceAsError(messageOrError));
       });
 
-      // When the child process closes all stdio streams, return the compiled result.
       this.childProcess.on("close", exitCode => {
-        if (exitCode === 0) {
-          resolve(encodedCompiledGraph);
-        } else {
+        if (exitCode !== 0) {
           reject(new Error(`Compilation child process exited with exit code ${exitCode}.`));
         }
       });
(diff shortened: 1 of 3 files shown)


api/vm/compile.ts
```diff
@@ -69,25 +68,20 @@ export function compile(compileConfig: dataform.ICompileConfig) {
...
+    let writer: net.Socket;
     try {
       const compiledResult = compile(compileConfig);
-      const writer = new net.Socket({ fd: 4 });
-      writer.write(compiledResult, (err) => {
+      writer = new net.Socket({ fd: 4 });
+      writer.write(compiledResult, (err: any) => {
         if (err) {
-          handleError(err);
+          writer?.end(`Error compiling in child process: ${err?.message}`);
```
BenBirt, line 78: can you clarify what exactly the conflict is, how it arises, etc?

this fix, i.e. multiplexing the response in a very unstructured way, feels like a bad idea / something that is going to bite us. it also loses any `Error` fields.

Ekrekr: Having `process.send` present in the code (it doesn't have to be called) after or before socket initialisation causes the `uv_pipe_open` conflict error. Happy for alternatives, but can't find a decent way to send the entire error as is; `writer.destroy` won't send it.

api/vm/compile.ts
```diff
@@ -69,25 +69,16 @@ export function compile(compileConfig: dataform.ICompileConfig) {
 
 export function listenForCompileRequest() {
   process.on("message", (compileConfig: dataform.ICompileConfig) => {
-    const handleError = (e: any) => {
+    try {
+      const compiledResult = compile(compileConfig);
+      process.send(compiledResult, () => process.exit());
```
BenBirt, line 74: remove all of the `process.exit` calls

BenBirt (commented):
you need to bump the version

api/commands/compile.ts
```diff
@@ -89,26 +89,17 @@ export class CompileChildProcess {
...
-      const chunks: Buffer[] = [];
-      pipe?.on("readable", () => {
-        let buffer: Buffer = pipe.read();
-        while (buffer) {
-          chunks.push(buffer);
-          buffer = pipe.read();
+
+      this.childProcess.on("message", (messageOrError: string | Error) => {
+        if (typeof messageOrError === "string") {
+          resolve(messageOrError);
```
BenBirt, line 96: bring back the `return` after this, it's not really valid to both `resolve` and `reject` a single `Promise`

Ekrekr: Ah I thought resolve and reject returned

BenBirt (approved):

Issue #1169, opened 2021-01-18
Bazel rule helm_template - allow usage of `--include-crds`


AleksanderGondek:
Direct link to Bazel rule: [link](https://github.com/dataform-co/dataform/blob/master/tools/helm/helm_template.bzl)

**Feature request short description**:
Add ability to use `--include-crds` flag in helm_template rule.

**Long description**:
Helm 3 encourages using special subdirectory (`crd/`) for handling CRDs in charts ([source](https://helm.sh/docs/topics/charts/)) - unless explicitly told to do so, the `helm template` command will not add the CRDs to its output.

The helm_template rule should allow for passing the aforementioned flag to turn CRD templating on/off. 

**Proposal**:
```python
load("//tools/helm:helm_template.bzl", "helm_template")

...

helm_template(
  name = "helm-chart",
  ... 
  include_crds = True,
)
```

PR #1170, opened 2021-01-18
Bazel rule helm_template - support '--include-crds'

AleksanderGondek:
Implements #1169 .

Introduces support for materializing CRDs defined in helm chart.
Dedicated rule attribute ("include_crds") controls whether to pass
'--include-crds' flag to helm template execution - by default it is set
to false.


BenBirt (approved):
Thanks!

PR #1172, opened 2021-01-20
Changed "Analytics" to "Reporting"

DolanP:
Changed "Analytics" to "Reporting" to match what is shown in the folder_example.png image.


BenBirt (approved):
Thanks!

PR #1161, opened 2021-01-13
Sandbox api

Ekrekr:

PR #1176, opened 2021-01-26
Consolidate worker bundle and make development and releases use the same binary output

Ekrekr:
Follow on from https://github.com/dataform-co/dataform/pull/1171
Precursor to https://github.com/dataform-co/dataform/pull/1161

Swaps to using a single entry point for compilation by @dataform/api, when packaged and when run in Bazel


tests/cli/BUILD
```diff
@@ -14,6 +14,7 @@ ts_test_suite(
     data = [
         ":node_modules",
         "//packages/@dataform/core:package_tar",
+        "//sandbox/vm:worker_bundle",
```
lewish, line 17: This makes your tests pass, but the CLI itself isn't going to actually work, plz remove!

tests/api/BUILD
```diff
@@ -16,6 +16,7 @@ ts_test_suite(
         "//examples/invalid_dataform_json:node_modules",
         "//examples/never_finishes_compiling:files",
         "//examples/never_finishes_compiling:node_modules",
+        "//sandbox/vm:worker_bundle",
```
lewish, line 19: This shouldn't be here, it should come through the dep on `//api`

tests/integration/BUILD
```diff
@@ -10,6 +10,7 @@ ts_test_suite(
         exclude = ["sqldatawarehouse.spec.ts"],
     ),
     data = [
+        "//sandbox/vm:worker_bundle",
```
lewish, line 13: Ditto - This shouldn't be here, it should come through the dep on `//api`

tests/cli/BUILD
lewish: And see my other comment for fix

api/commands/compile.ts
```diff
@@ -65,19 +65,17 @@ export async function compile(
...
-        return require.resolve("../../sandbox/vm/compile_loader");
-      }
-    };
-    const forkScript = findForkScript();
+    let workerBundle: string;
+    try {
+      workerBundle = require.resolve("df/sandbox/vm/worker_bundle");
+    } catch (e) {
+      // The bundled CLI doesn't allow the `df` prefix; the source file is packaged directly.
+      workerBundle = "./sandbox/vm/worker_bundle";
```
lewish, line 73: This won't work in the CLI, when you install `@dataform/cli` on a machine, this file won't exist.

This should be: `require.resolve("./worker_bundle")`

Additionally, you need to make sure the worker bundle js file get's copied to the CLI package. When the CLI runs on a normal machine it's executing inside a file such as:

`node_modules/@dataform/cli/bundle.js`

To make your require call work, you need to make sure there is also the file:

`node_modules/@dataform/cli/worker_bundle.js`

This may be as simple as keeping it as a dep of the CLI pkg_npm_tar rule, but you may also need to copy the file to the folder first. Let me know if that doesn't work on it's own.

Ekrekr: Copying worked, extracting shows this now!

```
$ tar -xvf bazel-out/k8-py2-fastbuild/bin/packages/@dataform/cli/package.tgz
package/
package/worker_bundle
package/bundle.js
package/package.json
```

tests/api/BUILD
Ekrekr: I've have a play with this, and some alternative structures. I think you're right, it should come through from `//api`. However we can't include it as a `dep` of `//api` because the worker bundle is plain javascript and has no types, so we have to put it as `data`. From this, I think that the real issue here is therefore that `ts_test_suite` does not include transitive `data` dependencies from it's typescript `dep` rules, and so we should just include the explicit `worker_bundle` `data` rule in `//tests/api` and `//tests/integration`. WDYT?

Ekrekr: The tests actually pass without the worker bundle being in `//api`, but I think we need that there for the CS

lewish: Sounds good, yup this is fine! Getting it out of the CLI test was the important bit :)

packages/@dataform/cli/BUILD
```diff
@@ -82,21 +80,21 @@ pkg_bundle(
...
-        ":worker_bundle",
     ],
 )
+
+# Worker bundle needs to be copied over for packaging, as deps files are only
+# included if they're in this folder.
+genrule(
+    name = "copy_bundle",
+    srcs = ["//sandbox/vm:worker_bundle"],
+    outs = ["worker_bundle"],
```
lewish, line 97: Hmmm, the output folder you pasted before - should this not be `outs = ["worker_bundle.js"]` ? I know it's executable, but it's a bit more clear if it's a JS file.

api/commands/compile.ts
```diff
@@ -65,19 +65,17 @@ export async function compile(
...
-        return workerBundlePath;
-      } catch (e) {
-        return require.resolve("../../sandbox/vm/compile_loader");
-      }
-    };
-    const forkScript = findForkScript();
+    let workerBundle: string;
+    try {
+      // The bundled CLI packages the worker_bundle directly.
+      workerBundle = require.resolve("./worker_bundle");
```
lewish, line 71: Given your changes now, I suspect you can get rid of the try catch and just use this line for both cases? `resolve("./worker_bundle")` should work in both scenarios!

packages/@dataform/cli/BUILD
lewish: I'm actually surprised the CLI tests pass without a `.js` extension on the file.

lewish (approved):
Can you plz double check that this new CLI package works? We've broken it so many times in the past  and it's a hard one to test :D

- Uninstall @dataform/cli globally with NPM
- Install @dataform/cli globally from the tar that is generated by packages/@dataform/cli:package
- Check that you can run `dataform compile` while in any normal dataform project

packages/@dataform/cli/BUILD
Ekrekr: Ah nice catch. I'm getting rid of the `#!/usr/bin/env node` prepended too, because that's not necessarily where the sandbox should be executing it from

api/commands/compile.ts
Ekrekr: I think it's neater (/more bazely) to have this than have multiple genrules copying the bundle to the correct places in order to replicate the CLI bundle locality

Ekrekr:
> Can you plz double check that this new CLI package works? We've broken it so many times in the past and it's a hard one to test :D
> 
> * Uninstall @dataform/cli globally with NPM
> * Install @dataform/cli globally from the tar that is generated by packages/@dataform/cli:package
> * Check that you can run `dataform compile` while in any normal dataform project

Can't global install on my dev machine so I did this slightly differently. For futue reference (assuming starting directory is the dataform repo):

```
$ cd ..
$ git clone https://github.com/dataform-co/dataform_demo_project
$ npm i ../dataform/bazel-out/k8-py2-fastbuild/bin/packages/@dataform/cli/package.tgz
$ node_modules/@dataform/cli/bundle.js compile
```

Checks out!

PR #1178, opened 2021-01-27
Merge project config & overrides correctly.

BenBirt:
related to https://github.com/dataform-co/dataform-web-tracking/issues/287

This isn't actually the cause of that bug, because the broken merge that bug is referring to happens at a higher level, but I'm fixing this up for completeness.

### DIFF
--- a/api/commands/compile.ts
+++ b/api/commands/compile.ts
@@ -1,8 +1,8 @@
-import deepmerge from "deepmerge";
 import * as fs from "fs";
 import * as path from "path";
 
 import { ChildProcess, fork } from "child_process";
+import deepmerge from "deepmerge";
 import { validWarehouses } from "df/api/dbadapters";
 import { coerceAsError, ErrorWithCause } from "df/common/errors/errors";
 import { decode } from "df/common/protos";
--- a/api/commands/compile.ts
+++ b/api/commands/compile.ts
@@ -1,3 +1,4 @@
+import deepmerge from "deepmerge";
 import * as fs from "fs";
 import * as path from "path";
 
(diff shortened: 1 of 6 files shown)


api/commands/compile.ts
```diff
@@ -129,7 +123,7 @@ export class CompileChildProcess {
   }
 }
 
-export const checkDataformJsonValidity = (dataformJsonParsed: { [prop: string]: string }) => {
+export const checkDataformJsonValidity = (dataformJsonParsed: { [prop: string]: any }) => {
```
GJMcGowan, line 126: seems a shame to lose the typing

core/gen_index.ts
```diff
@@ -62,7 +62,7 @@ projectConfig.schemaSuffix = "${
   }" || projectConfig.schemaSuffix;
 
 // Merge in general project config overrides.
-projectConfig = { ...projectConfig, ...${projectOverridesJsonString} };
+projectConfig = { ...projectConfig, ...${projectOverridesJsonString}, vars: { ...projectConfig.vars, ...${projectOverridesJsonString}.vars } };
```
GJMcGowan, line 65: Why not deepmerge here as well?

GJMcGowan (approved):
Same comment about the library as on the other PR.

api/commands/compile.ts
BenBirt: the typing isn't actually correct :P

core/gen_index.ts
BenBirt: This is effectively the index.js that we generate for each DF project. Plumbing through dependencies on other NPM packages at this stage is a bit of a pain, while this isn't ideal, I think it's the best option we have for now.

Issue #1175, opened 2021-01-25
Be able to pass parameters to dataform run command


Altons:
As I data engineer I would like to be able to pass parameters to a pipeline in bigquery as

`dataform run --tags countryReport --vars {country:"UK", department:"marketing"}`

which will inject the parameters in my `sqlx` as

```sql
select * from 
 ${ref("report_mart")}
 where country = ${vars.country} and department =  ${vars.department}
```

and compiled version would look like this:

```sql
select * from 
 `my_project.my_dataset.report_mart`
 where country = "UK" and department =  "marketing"
```

This will enable the creation/update of  tables reusing the same logic.

BenBirt:
I don't love the JSON blob model because bash makes it a bit of a pain to use. I'd suggest `--var foo=bar --var baz=qux --vars=foos=bars,bazs=quxs`

PR #1184, opened 2021-02-05
Add vars option to CLI

Ekrekr:
Fixes https://github.com/dataform-co/dataform/issues/1175

Delimited vars with colons rather than more equals because I think it looks more clear.

### DIFF
--- a/api/commands/compile.ts
+++ b/api/commands/compile.ts
@@ -28,14 +28,6 @@ export async function compile(
   // Resolve the path in case it hasn't been resolved already.
   path.resolve(compileConfig.projectDir);
 
-  // Create an empty projectConfigOverride if not set.
-  compileConfig = { projectConfigOverride: {}, ...compileConfig };
-
-  compileConfig.projectConfigOverride.vars = {
-    ...compileConfig.vars,
-    ...compileConfig.projectConfigOverride.vars
-  };
-
   // Schema overrides field can be set in two places, projectConfigOverride.schemaSuffix takes precedent.
   if (compileConfig.schemaSuffixOverride) {
     compileConfig.projectConfigOverride = {
@@ -48,7 +40,7 @@ export async function compile(
     // check dataformJson is valid before we try to compile
     const dataformJson = fs.readFileSync(`${compileConfig.projectDir}/dataform.json`, "utf8");
     const projectConfig = JSON.parse(dataformJson);
-    checkDataformJsonValidity(deepmerge(projectConfig, compileConfig.projectConfigOverride));
+    checkDataformJsonValidity(deepmerge(projectConfig, compileConfig.projectConfigOverride || {}));
   } catch (e) {
     throw new ErrorWithCause(
       `Compilation failed. ProjectConfig ('dataform.json') is invalid: ${e.message}`,
--- a/api/commands/compile.ts
+++ b/api/commands/compile.ts
@@ -31,6 +31,11 @@ export async function compile(
   // Create an empty projectConfigOverride if not set.
   compileConfig = { projectConfigOverride: {}, ...compileConfig };
 
+  compileConfig.projectConfigOverride.vars = {
+    ...compileConfig.vars,
+    ...compileConfig.projectConfigOverride.vars
+  };
+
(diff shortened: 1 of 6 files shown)


lewish:
Two reasons why I still think JSON is a worth considering:

- Dynamic types - JSON can capture numbers, booleans and strings, not just strings
- Automation - I imagine a common use case would be to pull vars from file, which would be easy if JSON

I don't think bash/json is that big of an issue, `--vars '{ "field": "value"}'` for simple cases, `jq` for more complex ones.

BenBirt:
IMO, if we want to support JSON, let's just fix the CLI to understand environments. That's actually the way you'd want to implement those complex use-cases anyway.

api/commands/compile.ts
```diff
@@ -31,6 +31,11 @@ export async function compile(
   // Create an empty projectConfigOverride if not set.
   compileConfig = { projectConfigOverride: {}, ...compileConfig };
 
+  compileConfig.projectConfigOverride.vars = {
+    ...compileConfig.vars,
+    ...compileConfig.projectConfigOverride.vars
+  };
```
BenBirt, line 37: we can simplify this block with `deepmerge` (which is happily already imported

cli/index.ts
```diff
@@ -420,7 +452,8 @@ export function runCli() {
           print("Compiling...\n");
           const compiledGraph = await compile({
             projectDir: argv["project-dir"],
-            schemaSuffixOverride: argv["schema-suffix"]
+            schemaSuffixOverride: argv["schema-suffix"],
+            vars: parseVarOptions(argv.vars)
```
BenBirt, line 456: you should use `option.coerce` in the `option` declaration (search above)

cli/index.ts
```diff
@@ -149,6 +149,35 @@ const jsonOutputOption: INamedOption<yargs.Options> = {
   }
 };
 
+const varsOption: INamedOption<yargs.Options> = {
+  name: "vars",
+  option: {
+    describe:
+      "Variables to inject that are exposed at compile and runtime via `dataform.projectConfig.vars.myVariableName`.\n" +
+      "They should take the structure of '--vars=someKey:someValue,someOtherKey:someOtherValue'.",
```
BenBirt, line 157: the newline in the middle of this description is a bit odd!

also i do think we should use `=` rather than `:`

cli/index.ts
```diff
@@ -328,19 +357,22 @@ export function runCli() {
             }
           },
           schemaSuffixOverrideOption,
-          jsonOutputOption
+          jsonOutputOption,
+          varsOption
```
BenBirt, line 361: you haven't added this to the list of options for `run` and `test`

protos/core.proto
```diff
@@ -142,6 +142,8 @@ message CompileConfig {
   // Deprecated.
   string schema_suffix_override = 2;
   string return_override = 5;
+
+  map<string, string> vars = 8;
```
BenBirt, line 146: i don't understand why we need this... it's already inside `project_config_override`

tests/cli/cli.spec.ts
```diff
@@ -56,7 +56,13 @@ select 1 as test
 
     // Compile the project using the CLI.
     const compileResult = await getProcessResult(
-      execFile(nodePath, [cliEntryPointPath, "compile", projectDir, "--json"])
+      execFile(nodePath, [
+        cliEntryPointPath,
+        "compile",
```
BenBirt, line 61: shall we include vars for the `run` (dry run) case below, too? (would catch situations like the bug in this PR)

BenBirt:
You'll also want to increment the version number (major version IMO)

cli/index.ts
Ekrekr: How come? It'd be weird if running `dataform compile` worked, but `dataform run` didn't because you couldn't inject a variable where it was required. It might also be useful for people wanting to reference a different dataset during testing rather than compilation, for example

BenBirt: I think you misread/misunderstood my comment? I'm just saying that you haven't added this option to `run` or `test`, even though you depend upon it in those commands.

Ekrekr: Yup lol

BenBirt (approved):

PR #1185, opened 2021-02-05
bigquery.md modification proposal

RantoB:
I think the DB where the table will published to is the one overriden in config. And the read table comes from the defaultDatabase from dataform.json.


BenBirt (approved):
Thanks!

PR #1197, opened 2021-03-02
Remediate prismjs Dependabot alert.

BenBirt:
https://github.com/dataform-co/dataform/security/dependabot/yarn.lock/prismjs/open


GJMcGowan (approved):
dunno what this library even does

PR #1206, opened 2021-04-08
Bump gcloud version to 318.0.0

Ekrekr:
Fixes issues with `AttributeError: module 'importlib' has no attribute 'util'` when running on system using Python 3.9+.

This does now spit out the error

```
Error in atexit._run_exitfuncs:
...
ValueError: Could not find Python executable.
```

but it doesn't have an effect on actual running as it's an exit func


BenBirt (commented):
Why not set the version directly in https://github.com/dataform-co/dataform-co/blob/master/WORKSPACE#L478? (It'll save you a PR)

Ekrekr:
We need it to be bumped in open source too, for that to work on our linux machines

lewish (approved):

PR #1210, opened 2021-04-22
Enabling Application Default Credentials for BigQuery

mescanne:
This does the following -
 - Prompt the user whether they want to use application default credentials or JSON key
 - If ADC, then only the projectId is requested (for billing) and credentials are an empty string
 - On BigQuery() connection creation, credentials are left as empty string (use ADC) and otherwise JSON parsed.

This has been tested successfully by hand using ADC. The unit tests fail due to permission denied decrypting a key.

### DIFF
--- a/api/dbadapters/bigquery.ts
+++ b/api/dbadapters/bigquery.ts
@@ -312,18 +312,15 @@ export class BigQueryDbAdapter implements IDbAdapter {
   private getClient(projectId?: string) {
     projectId = projectId || this.bigQueryCredentials.projectId;
     if (!this.clients.has(projectId)) {
-      let bigQueryOptions = {
-        projectId,
-        scopes: EXTRA_GOOGLE_SCOPES,
-        location: this.bigQueryCredentials.location,
-        credentials: <any>this.bigQueryCredentials.credentials,
-      };
-      if (bigQueryOptions.credentials !== "") {
-        bigQueryOptions.credentials = JSON.parse(this.bigQueryCredentials.credentials);
-      }
       this.clients.set(
         projectId,
-        new BigQuery(bigQueryOptions)
+        new BigQuery({
+          projectId,
+          scopes: EXTRA_GOOGLE_SCOPES,
+          location: this.bigQueryCredentials.location,
+          credentials: this.bigQueryCredentials.credentials &&
+                       JSON.parse(this.bigQueryCredentials.credentials)
+        })
       );
     }
     return this.clients.get(projectId);
--- a/api/dbadapters/bigquery.ts
+++ b/api/dbadapters/bigquery.ts
@@ -312,14 +312,18 @@ export class BigQueryDbAdapter implements IDbAdapter {
   private getClient(projectId?: string) {
     projectId = projectId || this.bigQueryCredentials.projectId;
     if (!this.clients.has(projectId)) {
+      let bigQueryOptions = {
+        projectId,
+        scopes: EXTRA_GOOGLE_SCOPES,
+        location: this.bigQueryCredentials.location,
+        credentials: <any>this.bigQueryCredentials.credentials,
+      };
+      if (bigQueryOptions.credentials !== "") {
+        bigQueryOptions.credentials = JSON.parse(this.bigQueryCredentials.credentials);
+      }
       this.clients.set(
         projectId,
-        new BigQuery({
-          projectId,
-          credentials: JSON.parse(this.bigQueryCredentials.credentials),
-          scopes: EXTRA_GOOGLE_SCOPES,
-          location: this.bigQueryCredentials.location
-        })
+        new BigQuery(bigQueryOptions)
       );
     }
     return this.clients.get(projectId);
--- a/cli/credentials.ts
+++ b/cli/credentials.ts
@@ -19,10 +19,9 @@ export function getBigQueryCredentials(): dataform.IBigQuery {
     "JSON Key"
   ]);
   if (isApplicationDefaultOrJSONKeyIndex === 0)  {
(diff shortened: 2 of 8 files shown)


cli/credentials.ts
```diff
@@ -5,6 +5,27 @@ import { actuallyResolve } from "df/cli/util";
...
+  }
+  const isApplicationDefaultOrJSONKeyIndex = selectionQuestion("Do you wish to use Application Default Credentials or JSON Key:", [
+    "ADC (default)",
+    "JSON Key"
+  ]);
+  if (isApplicationDefaultOrJSONKeyIndex === 0)  {
+    const projectId = question("Enter your billing project:");
+    return {
+      projectId: projectId,
+      credentials: "",
```
BenBirt, line 25: nit: you can just leave this unset

cli/credentials.ts
```diff
@@ -5,6 +5,27 @@ import { actuallyResolve } from "df/cli/util";
...
+  let location = locationIndex === 0 ? "US" : "EU";
+  if (locationIndex === 2) {
+    location = question("Enter the location's region name (e.g. 'asia-south1'):");
+  }
+  const isApplicationDefaultOrJSONKeyIndex = selectionQuestion("Do you wish to use Application Default Credentials or JSON Key:", [
+    "ADC (default)",
+    "JSON Key"
+  ]);
+  if (isApplicationDefaultOrJSONKeyIndex === 0)  {
+    const projectId = question("Enter your billing project:");
```
BenBirt, line 22:
```suggestion
    const projectId = question("Enter your billing project ID:");
```

cli/credentials.ts
BenBirt, line 25: can you leave a comment on the relevant protobuf field indicating that if it's unset, this implies that ADC should be used instead? https://github.com/dataform-co/dataform/blob/master/protos/profiles.proto#L46

api/dbadapters/bigquery.ts
```diff
@@ -312,14 +312,18 @@ export class BigQueryDbAdapter implements IDbAdapter {
   private getClient(projectId?: string) {
     projectId = projectId || this.bigQueryCredentials.projectId;
     if (!this.clients.has(projectId)) {
+      let bigQueryOptions = {
```
BenBirt, line 315: nit: should be `const` (fields of `const` objects are still mutable)

api/dbadapters/bigquery.ts
```diff
@@ -312,14 +312,18 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+        projectId,
+        scopes: EXTRA_GOOGLE_SCOPES,
+        location: this.bigQueryCredentials.location,
+        credentials: <any>this.bigQueryCredentials.credentials,
```
BenBirt, line 319: i don't really understand this part - the field is optional, why set it in this case at all?

perhaps the compiler is complaining that you're setting a field that doesn't exist, in which case, i propose either of the following:

preferred:
```
new BigQuery({
  projectId,
  credentials: this.bigQueryCredentials.credentials && JSON.parse(this.bigQueryCredentials.credentials),
  scopes: EXTRA_GOOGLE_SCOPES,
  location: this.bigQueryCredentials.location
})
```

or: 
```
const bigqueryOptions: BigQueryOptions = {
  projectId,
  scopes: EXTRA_GOOGLE_SCOPES,
  location: this.bigQueryCredentials.location,
}
// etc
```

BenBirt (changes requested):
This is awesome, thank you very much!

This will require me to push new package versions, so would you mind updating `version.bzl` to `1.19.0`?

(Optionally) it would also be great to update our docs, specifically:
- https://docs.dataform.co/getting-started-tutorial/set-up#generate-warehouse-credentials (probably needs clarifying that ADC can be used too, maybe linking to some relevant docs on cloud.google.com)
- https://docs.dataform.co/warehouses/bigquery#authentification (wow we need to fix this 'authentification' word that we've invented)

content/docs/warehouses/bigquery.md
```diff
@@ -1,12 +1,20 @@
 ---
 title: Google BigQuery
-subtitle: Authentification, configuration options, and content for BigQuery.
+subtitle: Authentication, configuration options, and content for BigQuery.
```
BenBirt, line 3: please merge with `master`, i've fixed this throughout our docs

content/docs/warehouses/bigquery.md
```diff
@@ -1,12 +1,20 @@
...
 
-## Authentification
+## Authentication
 
-Dataform will connect to BigQuery using a service account. You’ll need to create a service account from your Google Cloud Console and assign it permissions to access BigQuery.
+Dataform will connect to BigQuery using Application Default Credentials or using a service account.
+
+### Application Default Credentials
+
+If using Application Default Credentials ensure that the service account or user has `BigQuery Admin` role or equivalent. (Dataform requires access to create queries and list tables.) Read <a target="_blank" rel="noopener" href="https://cloud.google.com/iam/docs/granting-roles-to-service-accounts#granting_access_to_a_service_account_for_a_resource">this</a> if you need help.
```
BenBirt, line 13:
```suggestion
If using Application Default Credentials ensure that the service account or user has `BigQuery Admin` role or equivalent. (Dataform requires access to create datasets and list tables.) Read <a target="_blank" rel="noopener" href="https://cloud.google.com/iam/docs/granting-roles-to-service-accounts#granting_access_to_a_service_account_for_a_resource">this</a> if you need help.
```

BenBirt (approved):

BenBirt (approved):

BenBirt:
Now published in `1.19.0`.

PR #1252, opened 2021-08-09
Add a new "core execution" interface for simplified sandbox <> sandboxee communication

lewish:
This continues to simplify the steps required to compile a Dataform project in a sandbox and make it easier to add new compiler wrappers.

Main compilation is a setting and won't be used by default as it may change again before we release it.

The new flow for a sandbox wrapper to compile a project:

- Register and use the compiler function provided by `require("@dataform/core").compiler` for compiling SQLX files
- Generate a list of all files within the project directory and add it to the `CompileConfig`.
- Generate a  base64 encoded `MainConfig` and invoke `require("@dataform/core").main` with it
- Parse the returned base64 encoded `MainResult` proto

MainConfig and MainResult allows for additional actions other than `compile` to be added through the same code path, relying purely on protos to pass through the args and the response, with a future `build` command in mind.

Listing all file paths reduces interdependent logic between the compiler wrapper and the compilee code, so we can e.g support compilation of SQLX includes in the future without further compiler wrapper changes, or support other directory types or paths other than `definitions` and `includes`.

Next steps: Push two things that currently happen during compiler wrapper compilation into the versioned dataform/core code:
- Merging of project config overrides with the dataform.json file
- Validation of the project configuration
https://github.com/dataform-co/dataform/blob/master/api/commands/compile.ts#L43


core/session.ts
```diff
@@ -434,13 +433,13 @@ export class Session {
...
 
+    utils.throwIfInvalid(compiledGraph, dataform.CompiledGraph.verify);
     return compiledGraph;
   }
 
   public compileToBase64() {
     const compiledGraph = this.compile();
-    utils.throwIfInvalid(compiledGraph, dataform.CompiledGraph.verify);
-    return encode(dataform.CompiledGraph, compiledGraph);
+    return encode64(dataform.CompiledGraph, compiledGraph);
```
BenBirt, line 442: `return encode64(dataform.CompiledGraph, this.compile());`

BenBirt (commented):
Looks good, but can we discuss the naming of "main"? I realise it seems nitty, but it seems very abstract / hard to talk about / communicate, and I'm wondering if we can do better?

lewish:
> Looks good, but can we discuss the naming of "main"? I realise it seems nitty, but it seems very abstract / hard to talk about / communicate, and I'm wondering if we can do better?

Renamed CoreExecutionConfig and CoreExecutionResult as discussed, left the method as "main".

BenBirt (approved):

PR #1257, opened 2021-08-19
Remove @dataform/core's dependency on NodeJS

lewish:
- Make sure there are no known NodeJS built-ins being used when bundling the core package with rollup.
- Add a replacement for our usage of `basename` and `sep` from the `path`  built in module that works outside of Node environments.

### DIFF
--- a/core/utils.ts
+++ b/core/utils.ts
@@ -1,5 +1,3 @@
-import * as path from "path";
-
 import { adapters } from "df/core";
 import { Assertion } from "df/core/assertion";
 import { Resolvable } from "df/core/common";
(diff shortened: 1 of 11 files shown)


core/utils.ts
```diff
@@ -9,20 +7,30 @@ import { IActionProto, Session } from "df/core/session";
 import { Table } from "df/core/table";
 import { dataform } from "df/protos/ts";
 
+const pathSeperator = (() => {
+  if (typeof process !== "undefined") {
+    return process.platform === "win32" ? "\\" : "/";
+  }
+  return "/";
+})();
+
```
BenBirt, line 16:
```suggestion
const PATH_SEPARATOR = process?.platform === "win32" ? "\\" : "/";
```

BenBirt (approved):

core/utils.ts
lewish: A very reasonable suggestion! And what I first tried, however, it does not work:

image

Issue #1246, opened 2021-07-23
Make compilation timeout configurable via CLI flag


BenBirt:

BenBirt:
`--timeout` flag added in version `1.20.0`. You can pass arbitrary duration strings, e.g. `--timeout=10s` or `--timeout=5m`.

sukhamritboparai:
Thanks a-lot, this was precious.

Murry-Ovo:
@BenBirt is there any way to customise the timeout flag via the web UI? We are still frequently (once or twice a week) getting the error message on our scheduled runs:
`Unable to compile project: Unable to compile project: Compilation timed out`

BenBirt:
Unfortunately not - we (Dataform) need to retain control of compilation timeouts in the webapp.

If you're hitting compilation timeouts, the best things to do are:
- remove dead code
- consider splitting your project into smaller units
- consider wider use of the JavaScript API (using which avoids SQLX -> JavaScript transpilation)

PR #1259, opened 2021-08-20
Implement a --timeout flag for the CLI. Make the CLI options much more type-safe.

BenBirt:
fixes https://github.com/dataform-co/dataform/issues/1246

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -2,6 +2,7 @@ import * as fs from "fs";
 import * as glob from "glob";
 import * as path from "path";
 import yargs from "yargs";
+import parseDuration from "parse-duration";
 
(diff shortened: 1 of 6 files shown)


cli/index.ts
```diff
@@ -347,23 +388,25 @@ export function runCli() {
           },
           schemaSuffixOverrideOption,
           jsonOutputOption,
-          varsOption
+          varsOption,
+          timeoutOption
```
lewish, line 392: You might want to add this to the run command too?

lewish (approved):

cli/index.ts
BenBirt: whoops, thanks

PR #1288, opened 2021-11-16
Refactor core.proto into multiple files

lewish:
The intention of this is to keep core.proto itself relatively clean, only containing protos that are directly related to core-execution / compilation APIs (and therefore need to be kept backwards compatible).


core/adapters/index.ts
```diff
@@ -114,11 +112,15 @@ register("sqldatawarehouse", SQLDataWarehouseAdapter);
 
 export type QueryOrAction = string | dataform.Table | dataform.Operation | dataform.Assertion;
 
+export interface IValidationQuery {
```
Ekrekr, line 115: Why don't we want this in the proto anymore?

Ekrekr (approved):

core/adapters/index.ts
lewish: It's not actually used anywhere that needs to be serialized, it's only used in code, so I've moved it to a typescript interface.

PR #1303, opened 2022-01-26
Fix quadratic-time increase in compile time as project size grows.

BenBirt:


lewish (commented):
Annoyingly complicated but I don't see a better way, LGTM!

lewish (approved):

mahlats:
Hallo @BenBirt is the compile time now linear?

BenBirt:
As far as we are aware, yes.

PR #1306, opened 2022-02-09
Stop vm2 upgrading to a broken version

GJMcGowan:


Ekrekr (approved):
Isn't this a downgrade? To clarify, 3.9.4 is broken, 3.9.5 is OK?

GJMcGowan:
`3.9.6` is broken and the `^3.9.4` was allowing an upgrade to it. `3.9.5` is safe - tests pass for it and fail for `3.9.6`.

So I figured we'd upgrade to the latest safe version and then lock.

PR #1327, opened 2022-03-16
bigquery options

stankiewicz:
this includes:
- require_partition_filter
- partition_expiration_days

which are common options to set in partitoned data models.
At this moment it requires additional step (manual gcloud) to patch the table. 

this change is not patching/updating the existing tables, it only affect create table DDL.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "@zeit/next-css": "1.0.1",
     "algoliasearch": "^3.32.1",
     "analytics-node": "^3.5.0",
-    "axios": "^0.21.1",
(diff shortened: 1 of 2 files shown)


astorrs:
@stankiewicz @lewish Is this being incorporated in a new PR? This flushed out the missing pieces for BQ partitioning.

stankiewicz:
hi, 
1. I had to clean up my repo - never do PR on master ;) as it makes other PR impossible. Created new  pr #1332
2. There is wider change related to passing options so probably #1332  won't be merged.

PR #1332, opened 2022-04-15
Additional BigQuery Options

stankiewicz:
this includes:

require_partition_filter
partition_expiration_days
which are common options to set in partitoned data models.
At this moment it requires additional step (manual gcloud) to patch the table.

this change is not patching/updating the existing tables, it only affect create table DDL.


stankiewicz:
@canberkkoparal can I have update on this? thanks!

canberkkoparal:
Implementation LGTM
You should add validity checking [here](https://github.com/dataform-co/dataform/blob/99ba7ace2c79561f85046929ba733f7d915856ae/core/session.ts#L736), throwing an error if target is a view or if there is no partitionBy.
Also can you add compilation tests: both [passing](https://github.com/dataform-co/dataform/blob/99ba7ace2c79561f85046929ba733f7d915856ae/tests/core/core.spec.ts#L569) and [failing](https://github.com/dataform-co/dataform/blob/99ba7ace2c79561f85046929ba733f7d915856ae/tests/core/core.spec.ts#L475)
Finally you can add even more testing by writing [integration](https://github.com/dataform-co/dataform/blob/master/tests/integration/bigquery.spec.ts) and [example](https://github.com/dataform-co/dataform/blob/master/tests/api/examples.spec.ts) tests but those are not critical.

protos/core.proto
```diff
@@ -73,6 +73,9 @@ message BigQueryOptions {
   repeated string cluster_by = 2;
   string update_partition_filter = 3;
   map<string, string> labels = 4;
+  int32 partition_expiration_days = 5;
+  bool require_partition_filter = 6;
+
```
canberkkoparal, line 78: nit: delete extra empty line

canberkkoparal (changes requested):

stankiewicz:
I can't merge it (don't have permissions)  - @canberkkoparal @lewish can you merge it?

PR #1337, opened 2022-05-11
expose dependants in cli

stankiewicz:
There is a need to expose dependants options in CLI. It's already implemented in builder.


stankiewicz:
@lewish can we expose dependants recalculation in cli?

lewish (approved):
It is unfortunate that we called the original CLI flag `nclude-deps`!

PR #1342, opened 2022-05-26
Add new parameters runnerNotificationPeriodMillis to the run API

nmohseny-atz:
Adding a new parameters to the run API:

runnerNotificationPeriodMillis (default: 5000): This optional parameter takes in the number of milliseconds the runner should wait before pushing the next run's result in order to reduce noise.

This is useful when we to stream and visualize a run's result as each action begins running.

### DIFF
--- a/api/commands/run.ts
+++ b/api/commands/run.ts
@@ -22,7 +22,7 @@ export interface IExecutedAction {
   executionAction: dataform.IExecutionAction;
   actionResult: dataform.IActionResult;
 }
-// dummy comment
+
 export function run(
   dbadapter: dbadapters.IDbAdapter,
   graph: dataform.IExecutionGraph,
--- a/api/commands/run.ts
+++ b/api/commands/run.ts
@@ -22,7 +22,7 @@ export interface IExecutedAction {
   executionAction: dataform.IExecutionAction;
   actionResult: dataform.IActionResult;
 }
-//dummy comment
+
 export function run(
   dbadapter: dbadapters.IDbAdapter,
   graph: dataform.IExecutionGraph,
--- a/api/commands/run.ts
+++ b/api/commands/run.ts
@@ -2,7 +2,7 @@ import EventEmitter from "events";
 import Long from "long";
 
 import * as dbadapters from "df/api/dbadapters";
-import { Flags } from "df/common/flags";
+import { Flags, SingleValueFlag } from "df/common/flags";
 import { retry } from "df/common/promises";
 import { deepClone, equals } from "df/common/protos";
 import { StringifiedMap, StringifiedSet } from "df/common/strings/stringifier";
@@ -10,10 +10,6 @@ import { targetsAreEqual, targetStringifier } from "df/core/targets";
 import { dataform } from "df/protos/ts";
 
 const CANCEL_EVENT = "jobCancel";
-
-const flags = {
-  runnerNotificationPeriodMillis: Flags.number("runner-notification-period-millis", 5000)
-};
 
 const isSuccessfulAction = (actionResult: dataform.IActionResult) =>
   actionResult.status === dataform.ActionResult.ExecutionStatus.SUCCESSFUL ||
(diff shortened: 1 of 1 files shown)


api/commands/run.ts
```diff
@@ -42,7 +42,10 @@ export class Runner {
   private readonly runResult: dataform.IRunResult;
   private readonly changeListeners: Array<(graph: dataform.IRunResult) => void> = [];
   private readonly eEmitter: EventEmitter;
-
+  private readonly flags: {
```
lewish, line 45: You want to keep the flags at the top level, and then use them as defaults - is probably a cleaner way to do this! (And keep your new flag added)

api/commands/run.ts
```diff
@@ -27,9 +25,11 @@ export interface IExecutedAction {
 export function run(
   dbadapter: dbadapters.IDbAdapter,
   graph: dataform.IExecutionGraph,
-  partiallyExecutedRunResult: dataform.IRunResult = {}
+  partiallyExecutedRunResult: dataform.IRunResult = {},
+  runnerNotificationPeriodMillis: number = RUNNER_NOTIFICATION_PERIOD_MILLIS,
```
lewish, line 29: rather than using these consts as defaults, use the flag values as defaults, e.g:

runnerNotificationPeriodMillis: number = flags.runnerNotificationPeriodMillis.get()

(and you can get rid of the consts entirely)

api/commands/run.ts
```diff
@@ -70,12 +75,18 @@ export class Runner {
...
+    this.flags = {
+      runnerNotificationPeriodMillis: Flags.number("runner-notification-period-millis", runnerNotificationPeriodMillis),
+      displayRunningStatus: Flags.boolean("display-running-status", displayRunningStatus)
+    };
     this.executedActionTargets = new StringifiedSet(
       targetStringifier,
-      this.runResult.actions
-        .filter(action => action.status !== dataform.ActionResult.ExecutionStatus.RUNNING)
-        .map(action => action.target)
+      !!this.flags.displayRunningStatus.get() ?
```
lewish, line 84: better to express this inside the filter:

```
this.runResult.actions
          .filter(action => this.flags.displayRunningStatus.get() || action.status !== dataform.ActionResult.ExecutionStatus.RUNNING)
```

(I actually think you got the flag the wrong way around here!)

api/commands/run.ts
```diff
@@ -56,7 +59,9 @@ export class Runner {
   constructor(
     private readonly dbadapter: dbadapters.IDbAdapter,
     private readonly graph: dataform.IExecutionGraph,
-    partiallyExecutedRunResult: dataform.IRunResult = {}
+    partiallyExecutedRunResult: dataform.IRunResult = {},
+    runnerNotificationPeriodMillis: number = RUNNER_NOTIFICATION_PERIOD_MILLIS,
```
lewish, line 63: these should both have `private readonly` infront of them, then you can use them below (see other comment)

api/commands/run.ts
```diff
@@ -133,7 +144,7 @@ export class Runner {
 
   private notifyListeners() {
     if (
-      Date.now() - flags.runnerNotificationPeriodMillis.get() <
+      Date.now() - this.flags.runnerNotificationPeriodMillis.get() <
```
lewish, line 147: You don't want to reference either of these flags in this class anywhere anymore, use e.g `this.runnerNotificationPeriodMillis` instead (after adding the `private readonly` bit above)

api/commands/run.ts
```diff
@@ -10,8 +10,7 @@ import { targetsAreEqual, targetStringifier } from "df/core/targets";
 import { dataform } from "df/protos/ts";
 
 const CANCEL_EVENT = "jobCancel";
-
-const flags = {
+let flags = {
```
lewish, line 13: can keep this const

lewish (approved):

lewish:
Note that the build is failing due to:

```
ERROR: /workspace/api/commands/run.ts:13:5 - Identifier 'flags' is never reassigned; use 'const' instead of 'let'.
```
If you fix this, should be good to submit!

PR #1353, opened 2022-07-21
Require defaultLocation to be set in BigQuery

lewish:
- Requires setting --default-location when using dataform-init
- Throws a compilation error if location is not set


lewish:
> Want to increment the version number now?

Done

PR #1366, opened 2022-08-24
Update @dataform/core to never return a graph with non-unique action names.

diasdauletov:


core/session.ts
```diff
@@ -445,6 +445,8 @@ export class Session {
       );
     }
 
+    this.removeNonUniqueActionsFromCompiledGraph(standardActions, compiledGraph)
```
BenBirt, line 448: I think this should come immediately after the call to `alterActionName` - ideally we would stop mutating the graph before running any of the `check` methods.

core/session.ts
```diff
@@ -866,6 +868,34 @@ export class Session {
       );
     });
   }
+
+  private getNonUniqueTargets(targets: dataform.ITarget[]): StringifiedSet<dataform.ITarget> {
```
BenBirt, line 872: this doesn't need to be a method on `Session` - either move it outside of the class or just define it as a function inside `removeNonUniqueActionsFromCompiledGraph`

core/session.ts
```diff
@@ -866,6 +868,34 @@ export class Session {
...
+    const nonUniqueCanonicalTargets = this.getNonUniqueTargets(actions.map(action => action.canonicalTarget));
+
+    function isUniqueAction(action: IActionProto): boolean {
+      return !nonUniqueTargets.has(action.target) && !nonUniqueCanonicalTargets.has(action.canonicalTarget);
+    }
+
+    compiledGraph.tables = compiledGraph.tables.filter(isUniqueAction);
+    compiledGraph.operations = compiledGraph.operations.filter(isUniqueAction);
+    compiledGraph.declarations = compiledGraph.declarations.filter(isUniqueAction);
+    compiledGraph.assertions = compiledGraph.assertions.filter(isUniqueAction);
```
BenBirt, line 897: this looks good, but we should also be including a compilation error - with the relevant actions' filenames' - for each removed action.

BenBirt (changes requested):

core/session.ts
diasdauletov: [checkActionNameUniqueness](https://github.com/dataform-co/dataform/blob/main/core/session.ts#:~:text=private-,checkActionNameUniqueness,-(actions%3A) and [checkCanonicalTargetUniqueness](https://github.com/dataform-co/dataform/blob/main/core/session.ts#:~:text=private-,checkCanonicalTargetUniqueness,-(actions%3A) detect non-unique actions and each of them include N - 1 corresponding compilation errors (N - number of occurrences of a duplicate action). So if we remove duplicate actions before these check methods starts they will not detect any duplicate actions. Should I attach those errors when removing duplicate actions from the graph and remove these check methods?

BenBirt: Hmm, good point, I'd missed that! Either way is OK with me, it's probably a bit more obvious to the reader to do as you suggest, but also happy with the code as-is!

BenBirt (changes requested):

tests/core/core.spec.ts
```diff
@@ -310,9 +310,9 @@ suite("@dataform/core", () => {
         schema: "schema"
       });
       const graph = session.compile();
-      expect(graph.graphErrors.compilationErrors.map(error => error.message)).deep.equals([
+      expect(graph.graphErrors.compilationErrors.map(error => error.message)).deep.equals(Array(2).fill(
```
diasdauletov, line 313: [check uniqueness](https://github.com/dataform-co/dataform/blob/e42c57bce81221bc36e92a66c0bf107244b63ddf/core/session.ts#L597) methods haven't been including all duplicate actions. Because the first duplicate action wasn't detecting.

BenBirt (approved):

PR #1401, opened 2022-11-23
Enable `main` compilation by default for all @dataform/core versions >= `2.0.4`.

BenBirt:


api/commands/compile.ts
```diff
@@ -40,6 +41,18 @@ export async function compile(
     );
   }
 
+  if (compileConfig.useMain === null || compileConfig.useMain === undefined) {
+    try {
+      const packageJson = JSON.parse(
+        fs.readFileSync(`${compileConfig.projectDir}/package.json`, "utf8")
+      );
+      const dataformCoreVersion = packageJson.dependencies["@dataform/core"];
+      compileConfig.useMain = semver.gte(dataformCoreVersion, "2.0.4");
```
lewish, line 50: does semver.gte work with package.json specifiers like `^2.0.4`? Might be worth checking!

BenBirt: good catch. I am now using `semver.subset` (https://www.npmjs.com/package/semver), which I believe should capture all cases.

lewish (approved):

PR #1406, opened 2022-12-06
Make table level assertions have the same tags as the table

Ekrekr:
(with some automatic formatting fixes...)

### DIFF
--- a/core/table.ts
+++ b/core/table.ts
@@ -604,6 +604,8 @@ export class Table {
     newTags.forEach(t => {
       this.proto.tags.push(t);
     });
+    this.uniqueKeyAssertions.forEach(assertion => assertion.tags(value));
+    this.rowConditionsAssertion?.tags(value);
     return this;
   }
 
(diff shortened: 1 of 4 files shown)


core/table.ts
```diff
@@ -664,7 +671,8 @@ export class Table {
       uniqueKeys.forEach((uniqueKey, index) => {
         const uniqueKeyAssertion = this.session.assert(
           `${this.proto.target.schema}_${this.proto.target.name}_assertions_uniqueKey_${index}`,
-          ctx => this.session.adapter().indexAssertion(ctx.ref(this.proto.target), uniqueKey)
+          ctx => this.session.adapter().indexAssertion(ctx.ref(this.proto.target), uniqueKey),
+          this.proto.tags
```
BenBirt, line 675: this only works for when the caller calls `assertions(...)` *after* calling `tags(...)`.

consider the case where the caller adds another tag to the table *after* adding assertions. if you look at the equivalent for `disabled`, you will see that `disabled(...)` copies that through to any attached assertions - you need to do the same for tags.

BenBirt (changes requested):

core/table.ts
Ekrekr: Done. I would edit a test for this case, but I don't see any pre-existing that test this order of events

BenBirt (approved):
Please update `version.bzl` to `2.1.0` and I will push a new version.

Ekrekr:
> Please update `version.bzl` to `2.1.0` and I will push a new version.

Done.

PR #1417, opened 2023-01-05
Remove DB tools dependence on `/api`

Ekrekr:


BenBirt (commented):
I don't follow the intent of this change. What is it you are actually trying to do? @dataform/api doesn't actually exist.

Ekrekr:
> I don't follow the intent of this change. What is it you are actually trying to do? @dataform/api doesn't actually exist.

Fair point, what I really mean is moving `/api` inside of `/cli` now that `@dataform/api` is not it's own entity.

Aside from this PR, the main change before doing this is that rather than having `tests/api` which covers functionality of the @dataform/api, we test functionality that is relevant to the CLI. All part of removing dead code and simplifying maintainability.

image

BenBirt:
I think I still don't really follow.

I think there is almost certainly value in maintaining these library functions (whether called `api` or whatever) separately from the CLI - not least from a testing perspective.

It's hard to know whether this is a good change without actually knowing what the destination is.

Ekrekr:
> value in maintaining these library functions

This is the bit I'm interested in, could you give some examples? If we're OK with not supporting a packaged javascript client then why not minimise it's exposure to the rest of the repo too?

Note that although the VSCode extension has it as a build dependency, it doesn't actually use `/api` at all.

BenBirt (approved):

PR #1421, opened 2023-01-06
Fix bazel build //... for our distro by bumping versions

Ekrekr:
and quick vscode extension fix


.bazelversion
```diff
@@ -1 +1 @@
-3.5.0
```
lewish, line 1: I am suprised this works, but if the tests are passing, that's awesome :D

lewish (approved):

.bazelversion
Ekrekr: it breaks at `6.0.0` because docker rules can't support that, but seems fine locally!

Cloud build tests time out on the PR, but it looks like it's just because postgres flaked - will retry.

PR #1422, opened 2023-01-06
Remove sql file support

Ekrekr:


lewish:
You can remove SQL from here too: https://github.com/dataform-co/dataform/blob/main/sandbox/vm/compile.ts#L47

lewish (approved):

version.bzl
```diff
@@ -1,3 +1,3 @@
 # NOTE: If you change the format of this line, you must change the bash command
 # in /scripts/publish to extract the version string correctly.
-DF_VERSION = "2.1.0"
```
BenBirt, line 3: hm. should we now be making a `main_v2` to be able to push bugfixes (security fixes) to v2 users?

Ekrekr: Fair point, more effort than it's worth currently - will hold off submitting this

Ekrekr:
Actually done in https://github.com/dataform-co/dataform/pull/1564

PR #1426, opened 2023-01-10
Remove docs content

Ekrekr:
Content is hosted from the `main_v1` branch


BenBirt (commented):
Before we delete this, do we actually have a plan for open-source documentation?

Ekrekr:
> Before we delete this, do we actually have a plan for open-source documentation?

Because this content isn't being served from this branch I think it's confusing to have it around - and little of the current content will be re-used, because it's duplicate with what is available on https://cloud.google.com/dataform/docs. Roadmap:

1. Keep current docs site running as is until legacy deprecation.
2. In the meantime, create a new set of content specifically for open source in this branch - in markdown to start with.
3. Host it via GitHub pages.

Outlined in https://docs.google.com/document/d/1FekFye75nT0I9LxmzRcd8iZbQeo2eiZWmjq13qcDnk4/edit#heading=h.5wtvzv51zb08

Ekrekr:
> users (who do occasionally send us docs fixes/updates) will have no idea where/how to do that

This is my point - if they submit changes to the docs site here currently then they don't get published, because of `main_v1`.

We should be aiming for the use of our legacy docs site to decrease anyway - the links on our main readme now point to the Dataform on GCP docs https://github.com/dataform-co/dataform#data-modeling-with-dataform, and there's a big "send feedback" button for suggesting changes there.

Have added a note to the main readme to clarify this to readers.

PR #1432, opened 2023-01-25
Added schema function

a2wd:
* Added schema functions to each of `ICommonContext`, `Table`, `Operation` and `Assertion`.
* Added corresponding tests
* Refactored the resolve method to de-dupe common logic in normalizing the `Schema`/`Database` and `Table` names.
* Removed redundant calls to the `toResolvable` method

### DIFF
--- a/core/session.ts
+++ b/core/session.ts
@@ -245,15 +245,9 @@ export class Session {
         ...resolved.proto.target,
         database:
           resolved.proto.target.database &&
-          this.adapter().normalizeIdentifier(
-            `${resolved.proto.target.database}${this.getDatabaseSuffixWithUnderscore()}`
-          ),
-        schema: this.adapter().normalizeIdentifier(
-          `${resolved.proto.target.schema}${this.getSchemaSuffixWithUnderscore()}`
-        ),
-        name: this.adapter().normalizeIdentifier(
-          `${this.getTablePrefixWithUnderscore()}${resolved.proto.target.name}`
-        )
+          this.normalizeDatabase(resolved.proto.target.database),
+        schema: this.normalizeSchema(resolved.proto.target.schema),
+        name: this.normalizeTable(resolved.proto.target.name),
       });
     }
     // TODO: Here we allow 'ref' to go unresolved. This is for backwards compatibility with projects
@@ -266,14 +260,10 @@ export class Session {
         utils.target(
           this.adapter(),
           this.config,
-          this.adapter().normalizeIdentifier(`${this.getTablePrefixWithUnderscore()}${ref}`),
-          this.adapter().normalizeIdentifier(
-            `${this.config.defaultSchema}${this.getSchemaSuffixWithUnderscore()}`
-          ),
+          this.normalizeTable(ref),
+          this.normalizeSchema(this.config.defaultSchema),
           this.config.defaultDatabase &&
-            this.adapter().normalizeIdentifier(
-              `${this.config.defaultDatabase}${this.getDatabaseSuffixWithUnderscore()}`
-            )
+            this.normalizeDatabase(this.config.defaultDatabase),
         )
       );
     }
@@ -281,12 +271,9 @@ export class Session {
       utils.target(
         this.adapter(),
         this.config,
-        this.adapter().normalizeIdentifier(`${this.getTablePrefixWithUnderscore()}${ref.name}`),
-        this.adapter().normalizeIdentifier(`${ref.schema}${this.getSchemaSuffixWithUnderscore()}`),
-        ref.database &&
-          this.adapter().normalizeIdentifier(
-            `${ref.database}${this.getDatabaseSuffixWithUnderscore()}`
-          )
+        this.normalizeTable(ref.name),
+        this.normalizeSchema(ref.schema),
+        ref.database && this.normalizeTable(ref.database),
       )
     );
   }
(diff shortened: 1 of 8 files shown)


tests/core/core.spec.ts
```diff
@@ -887,6 +887,37 @@ suite("@dataform/core", () => {
         });
       }
     );
+
+    test("schema without suffix", () => {
+      const session = new Session(path.dirname(__filename), TestConfigs.redshift);
+      session.publish("test", { type: "table" })
+        .query(ctx => `select "${ctx.schema()}"`);
```
Ekrekr, line 894: Nit: can just be `ctx => ctx.schema()`

Ekrekr (commented):
Nice, looks great to me! But wait for ben to check.

Probably worth doing a patch version bump https://github.com/dataform-co/dataform/blob/main/version.bzl

Nit: do the refactoring part in a separate CL next time

core/session.ts
```diff
@@ -444,6 +431,21 @@ export class Session {
...
+    return this.adapter().normalizeIdentifier(
+      `${database}${this.getDatabaseSuffixWithUnderscore()}`);
+  }
+
+  public normalizeSchema(schema: string): string {
+    return this.adapter().normalizeIdentifier(
+      `${schema}${this.getSchemaSuffixWithUnderscore()}`);
+  }
+
+  public normalizeTable(table: string): string {
```
BenBirt, line 444: nit: i think "normalize" isn't quite the right word to use here. elsewhere in the code, "normalizing" an identifier means: to translate (usually involves escaping, casing, whatever) this string to something which can be used as an identifier in the target warehouse.

here we *are* normalizing, sure, but we're doing more than that - we are suffixing/prefixing as necessary.

i would suggest "finalize", i.e. `finalizeDatabase`

BenBirt, line 444: this should be `normalizeName`. i realise that "name" as a term really sucks, but it is what the entire framework (and other code which uses the framework) uses to indicate the last part of a target, which could be a table, view, or any other materializable object - or might not even be materializable, in the case of `operations`!

tests/core/core.spec.ts
```diff
@@ -887,6 +887,37 @@ suite("@dataform/core", () => {
         });
       }
     );
+
+    test("schema without suffix", () => {
```
BenBirt, line 891: generally for suffix/suffix-less tests we do some table-driven stuff, i.e. something like

```
for ({config, expectedTarget of [{config: TestConfigs.redshift, expectedTarget: "schema.test"}, ...]) {
  test("schema with config " + JSON.stringify(config), () => {
    // test
  }
}
```

BenBirt (commented):
Please bump the minor version in `version.bzl`.

tests/core/core.spec.ts
a2wd: Thanks, done

core/session.ts
a2wd: Sounds good, done - thanks.

a2wd: Ack. Have updated alongside the `normalize` => `finalize` change.

a2wd:
> Nice, looks great to me! But wait for ben to check.
> 
> Probably worth doing a patch version bump https://github.com/dataform-co/dataform/blob/main/version.bzl
> 
> Nit: do the refactoring part in a separate CL next time

Wasn't sure on the way we prefer to do this (eg. this is all in separate commits, but one PR). Will keep to separate PRs in the future, especially considering we squash on merge.

tests/core/core.spec.ts
a2wd: Thanks, updated

tests/core/core.spec.ts
```diff
@@ -1,13 +1,13 @@
...
 import * as compilers from "df/core/compilers";
-import { Session } from "df/core/session";
-import { targetAsReadableString } from "df/core/targets";
-import { dataform } from "df/protos/ts";
-import { suite, test } from "df/testing";
-import { asPlainObject } from "df/tests/utils";
+import {Session} from "df/core/session";
+import {targetAsReadableString} from "df/core/targets";
+import {dataform} from "df/protos/ts";
+import {suite, Test, test} from "df/testing";
```
BenBirt, line 9: do you need this `Test` import?

tests/core/core.spec.ts
```diff
@@ -931,6 +946,17 @@ suite("@dataform/core", () => {
       ).deep.equals(["schema.operate-1"]);
       expect(graph.operations[1].queries).deep.equals(['select * from "schema"."operate-1"']);
     });
+
+    [{testConfig: TestConfigs.redshift, finalizedSchema: 'schema'}, {testConfig: TestConfigs.redshiftWithSuffix, finalizedSchema: 'schema_suffix'}].forEach(({testConfig, finalizedSchema}) => {
```
BenBirt, line 950: seems bigger than line limit. ideally, prettier should be auto-formatting your code, can you look into that?

BenBirt (approved):

tests/core/core.spec.ts
a2wd: Did speak with Elias about that, he said we use the auto format in VS code, but that didn't touch line length. I did notice we have other tests with an `array.forEach` that goes past the standard 100 character width, so left that.

I suggest we add some global format file/tool (eg. a `tslint`) and update the contributing.md too, to mention using this. WDYT?

BenBirt: We do use tslint in closed-source, I'm a little surprised it's not enforced here (as part of presubmit tests).

tests/core/core.spec.ts
a2wd: No, removed.

BenBirt (approved):

tests/core/core.spec.ts
a2wd: Will look into adding this.

PR #1450, opened 2023-02-13
Adds a flag 'track' to allow users to configure analytics with flags

GJMcGowan:
Usage: `dataform [command] --track=false`

### DIFF
--- a/cli/analytics.ts
+++ b/cli/analytics.ts
@@ -17,6 +17,7 @@ export const noTrackOption: INamedOption<yargs.Options> = {
 const analytics = new Analytics("eR24ln3MniE3TKZXkvAkOGkiSN02xXqw");
 
 let currentCommand: string;
+let allowAnonymousAnalytics: boolean;
 
 export async function maybeConfigureAnalytics(noTrack: boolean) {
   const settings = await getConfigSettings();
@@ -25,13 +26,11 @@ export async function maybeConfigureAnalytics(noTrack: boolean) {
     return;
   }
   if (settings.allowAnonymousAnalytics !== undefined) {
+    allowAnonymousAnalytics = settings.allowAnonymousAnalytics;
     return;
   }
   if (noTrack) {
-    await upsertConfigSettings({
-      allowAnonymousAnalytics: false,
-      anonymousUserId: uuidv4()
-    });
+    allowAnonymousAnalytics = false;
     return;
   }
 
@@ -47,12 +46,13 @@ Would you like to opt-in to anonymous usage and error tracking?`,
     allowAnonymousAnalytics: optInResponse,
     anonymousUserId: uuidv4()
   });
+  allowAnonymousAnalytics = optInResponse;
 }
 
 export async function trackCommand(command: string) {
   currentCommand = command;
   const config = await getConfigSettings();
-  if (!config.allowAnonymousAnalytics) {
+  if (!allowAnonymousAnalytics) {
     return;
   }
   await new Promise(resolve => {
@@ -74,7 +74,7 @@ export async function trackCommand(command: string) {
 
 export async function trackError() {
   const config = await getConfigSettings();
-  if (!config.allowAnonymousAnalytics) {
+  if (!allowAnonymousAnalytics) {
     return;
   }
   await new Promise(resolve => {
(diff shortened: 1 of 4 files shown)


Ekrekr (commented):
IMO the CLI option should always be able to override the local setting. Proposal:
* Remove default value of false
* Update function parameter type to `noTrack?:`
* Rather than `if (noTrack)` check for `if (noTrack != null | undefined)` (whatever yargs argv not populated gives)
* Move this check to the top of the function

cli/analytics.ts
```diff
@@ -28,12 +47,13 @@ Would you like to opt-in to anonymous usage and error tracking?`,
...
+  allowAnonymousAnalytics = optInResponse;
 }
 
 export async function trackCommand(command: string) {
   currentCommand = command;
   const config = await getConfigSettings();
-  if (!config.allowAnonymousAnalytics) {
+  if (!allowAnonymousAnalytics) {
     return;
   }
```
Ekrekr, line 58: Move to start of function

cli/analytics.ts
```diff
@@ -55,7 +75,7 @@ export async function trackCommand(command: string) {
 
 export async function trackError() {
   const config = await getConfigSettings();
-  if (!config.allowAnonymousAnalytics) {
+  if (!allowAnonymousAnalytics) {
     return;
   }
```
Ekrekr, line 80: ditto, move to start

cli/analytics.ts
```diff
@@ -1,21 +1,40 @@
...
   if (settings.allowAnonymousAnalytics !== undefined) {
+    allowAnonymousAnalytics = settings.allowAnonymousAnalytics;
     return;
   }
+  if (noTrack) {
+    allowAnonymousAnalytics = false;
+    return;
+  }
```
Ekrekr, line 36:
```
if (track != undefined) {
    allowAnonymousAnalytics = track;
}
```

and move to top.

cli/analytics.ts
```diff
@@ -1,21 +1,40 @@
...
 import Analytics from "analytics-node";
-import { getConfigSettings, getConfigSettingsPath, upsertConfigSettings } from "df/cli/config";
-import { ynQuestion } from "df/cli/console";
-import { v4 as uuidv4 } from "uuid";
+import {getConfigSettings, getConfigSettingsPath, upsertConfigSettings} from "df/cli/config";
+import {ynQuestion} from "df/cli/console";
+import {INamedOption} from "df/cli/yargswrapper";
+import {v4 as uuidv4} from "uuid";
+
+export const noTrackOption: INamedOption<yargs.Options> = {
```
Ekrekr, line 9: IMO rename to track, makes logic simpler. Otherwise `--no-track=false` is a double negative.

Also it shows that it overrides the default settings more clear.

cli/analytics.ts
```diff
@@ -1,21 +1,40 @@
...
 
 const analytics = new Analytics("eR24ln3MniE3TKZXkvAkOGkiSN02xXqw");
 
 let currentCommand: string;
+let allowAnonymousAnalytics: boolean;
 
-export async function maybeConfigureAnalytics() {
+export async function maybeConfigureAnalytics(noTrack: boolean) {
   const settings = await getConfigSettings();
   // We should only ask if users want to track analytics if they are in an interactive terminal;
```
Ekrekr, line 25: Nit: It feels like we should be reading the settings once rather than in every function, and overriding the tracking field (once) after it is read.

But this is a bit extra in the context of this PR.

cli/analytics.ts
GJMcGowan: I have moved to top, and added some extra logic for when `--track=true` and no settings.json exists.

cli/analytics.ts
GJMcGowan: Renamed

cli/analytics.ts
GJMcGowan: I have switched to Yet Another Global Variable

cli/analytics.ts
```diff
@@ -1,21 +1,46 @@
...
+};
 
 const analytics = new Analytics("eR24ln3MniE3TKZXkvAkOGkiSN02xXqw");
 
 let currentCommand: string;
+let allowAnonymousAnalytics: boolean;
+let anonymousUserId: string;
 
-export async function maybeConfigureAnalytics() {
+export async function maybeConfigureAnalytics(track: boolean) {
```
Ekrekr, line 23: `track?: boolean`

Ekrekr (approved):

PR #1453, opened 2023-02-14
Added enum for table's type field

kolina:
New enum field `enum_type` was added to `Table` proto message. The old field `type` marked deprecated. For backward compatibilty both fields are filled in the code when creating `CompliedGraph` representation.

### DIFF
--- a/api/commands/build.ts
+++ b/api/commands/build.ts
@@ -94,7 +94,7 @@ export class Builder {
     return {
       ...this.toPartialExecutionAction(table),
       type: "table",
-      tableType: table.type,
+      tableType: utils.tableEnumTypeToString(utils.tableTypeFromProto(table, true)),
       tasks: table.disabled
         ? []
         : this.adapter.publishTasks(table, runConfig, tableMetadata).build(),
--- a/api/commands/build.ts
+++ b/api/commands/build.ts
@@ -52,6 +52,7 @@ export class Builder {
       prunedGraph.projectConfig,
       prunedGraph.dataformCoreVersion || "1.0.0"
     );
+    prunedGraph.tables.forEach(utils.setOrValidateTableEnumType);
   }
 
   public build(): dataform.ExecutionGraph {
@@ -94,7 +95,7 @@ export class Builder {
     return {
       ...this.toPartialExecutionAction(table),
       type: "table",
-      tableType: utils.tableEnumTypeToString(utils.tableTypeFromProto(table, true)),
+      tableType: utils.tableTypeEnumToString(table.enumType),
       tasks: table.disabled
         ? []
         : this.adapter.publishTasks(table, runConfig, tableMetadata).build(),
--- a/cli/console.ts
+++ b/cli/console.ts
@@ -1,5 +1,6 @@
 import { IInitResult } from "df/api/commands/init";
 import { prettyJsonStringify } from "df/api/utils";
+import { tableEnumTypeToString, tableTypeFromProto } from "df/core/utils";
 import { dataform } from "df/protos/ts";
 import * as readlineSync from "readline-sync";
 
@@ -115,7 +116,7 @@ export function printCompiledGraph(graph: dataform.ICompiledGraph, verbose: bool
       writeStdOut(`${graph.tables.length} dataset(s):`);
       graph.tables.forEach(compiledTable => {
         writeStdOut(
-          `${datasetString(compiledTable.target, compiledTable.type, compiledTable.disabled)}`,
+          `${datasetString(compiledTable.target, tableEnumTypeToString(tableTypeFromProto(compiledTable, true)), compiledTable.disabled)}`,
           1
         );
       });
--- a/cli/console.ts
+++ b/cli/console.ts
@@ -1,6 +1,6 @@
 import { IInitResult } from "df/api/commands/init";
 import { prettyJsonStringify } from "df/api/utils";
-import { tableEnumTypeToString, tableTypeFromProto } from "df/core/utils";
+import { setOrValidateTableEnumType, tableTypeEnumToString } from "df/core/utils";
 import { dataform } from "df/protos/ts";
 import * as readlineSync from "readline-sync";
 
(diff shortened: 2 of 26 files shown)


Ekrekr (commented):
This is actually quite complex, I shouldn't have labelled it "good first issue" - sorry. Nice work picking it up so quickly though!

Looking at this, I think we could potentially implement this without ever breaking backwards compatibility. Rather than introducing `enum_type`, we can read the SQLX config and capitalise the string value for `type` - which would turn existing SQLX table definitions to be interpreted as the enum correctly. This would also remove the need for the conversions in `core/utils.ts`.

CCing in @BenBirt because although I think this is possible, I'm not sure where the place to do it is.

Ekrekr:
> This is actually quite complex, I shouldn't have labelled it "good first issue" - sorry. Nice work picking it up so quickly though!
> 
> Looking at this, I think we could potentially implement this without ever breaking backwards compatibility. Rather than introducing `enum_type`, we can read the SQLX config and capitalise the string value for `type` - which would turn existing SQLX table definitions to be interpreted as the enum correctly. This would also remove the need for the conversions in `core/utils.ts`.
> 
> CCing in @BenBirt because although I think this is possible, I'm not sure where the place to do it is.

Have chatted to Ben, I was confused thinking that this proto being derived from SQLX files directly when it's in fact just to do with the compiled graph. Will review again!

api/commands/build.ts
```diff
@@ -94,7 +94,7 @@ export class Builder {
     return {
       ...this.toPartialExecutionAction(table),
       type: "table",
-      tableType: table.type,
+      tableType: utils.tableEnumTypeToString(utils.tableTypeFromProto(table, true)),
```
Ekrekr, line 97: To match `type` above, can't this just be `dataform.TableType.TABLE`

cli/console.ts
```diff
@@ -115,7 +116,7 @@ export function printCompiledGraph(graph: dataform.ICompiledGraph, verbose: bool
       writeStdOut(`${graph.tables.length} dataset(s):`);
       graph.tables.forEach(compiledTable => {
         writeStdOut(
-          `${datasetString(compiledTable.target, compiledTable.type, compiledTable.disabled)}`,
+          `${datasetString(compiledTable.target, tableEnumTypeToString(tableTypeFromProto(compiledTable, true)), compiledTable.disabled)}`,
```
Ekrekr, line 119: `compiledTable.type`?

core/adapters/base.ts
```diff
@@ -133,7 +134,7 @@ from (${query}) as insertions`;
     let preOps = table.preOps;
     if (
       semver.gt(this.dataformCoreVersion, "1.4.8") &&
-      table.type === "incremental" &&
+      tableTypeFromProto(table, true) === dataform.TableType.INCREMENTAL &&
```
Ekrekr, line 137: `table.enumType`

Ekrekr (commented):
IMO:
* When creating/loading compiled graphs, we should enforce `type` and `enumType` to alway be their equivalent type.
* If both fields are specified but of different equivalent types, we should throw.
* When performing type checks in the code, we should always use `enumType` now.

 This will preserve backwards compatibility, but also make it easier when we remove `type` in favour of `enumType` entirely when we need to - and also remove the need for most of the utils conversions!

core/utils.ts
```diff
@@ -246,3 +246,33 @@ export function throwIfInvalid<T>(proto: T, verify: (proto: T) => string) {
...
+    default: {
+      if (throwIfUnknown) {
+        throw new Error(`Unexpected table type: ${type}`);
+      }
+      return dataform.TableType.UNKNOWN_TYPE;
+    }
+  }
+}
+
+export function tableTypeFromProto(table: dataform.ITable, throwIfUnknown: boolean) {
```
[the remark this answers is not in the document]
kolina: We fill `type` field here for `Table` https://github.com/dataform-co/dataform/blob/b25786b3a950921dbd58bc71236a0f83b8a0879f/core/table.ts#L482. We do not want to throw right here to add compile error later https://github.com/dataform-co/dataform/blob/b25786b3a950921dbd58bc71236a0f83b8a0879f/core/session.ts#L571.

api/commands/build.ts
kolina: No, `type` corresponds to `TableMetadata.Type` https://github.com/dataform-co/dataform/blob/f61bcedfd4d84f667a711c414fa868eafa8bcc07/protos/execution.proto#L181, it's slightly different from `typleType` options from `TableType` here https://github.com/dataform-co/dataform/blob/b25786b3a950921dbd58bc71236a0f83b8a0879f/protos/core.proto#L162

cli/console.ts
kolina: Changed to string representation of `compiledTable.enumType`.

kolina:
Simplified code by filling `enumType` when loading compiled graphs (and validating that it's consistent with `type`) in cli commands. A later code uses `enumType` then.

kolina:
`Adapter` subclasses for warehouses do not perform conversion for `enumType` field and assume that it's present. I changed a code for integration tests accordingly.

core/session.ts
```diff
@@ -570,7 +570,9 @@ export class Session {
   private checkTableConfigValidity(tables: dataform.ITable[]) {
     tables.forEach(table => {
       // type
-      if (!!table.type && !TableType.includes(table.type as TableType)) {
+      try {
+        setOrValidateTableEnumType(table);
+      } catch (e) {
```
Ekrekr, line 575: The error caught here could be from `are not equivalent` which would be misleading.

I think you can just remove this check altogether because `tableTypeStringToEnum` throws for it!

Ekrekr (approved):
Nice, great bit of cleanup!

core/session.ts
kolina: Good point, replaced with checking `enumType` for  `UNKNOWN_TYPE` value.

Issue #1295, opened 2021-12-14
Override database config is not applied to Auto generated assertions


newuniverse:
About this https://docs.dataform.co/guides/assertions#auto-generated-assertions .

I am using `dataform run --vars` option to embed a new database(GCP project_id) into each sqlx files to
override the default_database  in my CI/CD.

Not from code base but from behavior base, I found that override is not applied to auto generated assertions.

Is it an expected  behavior or just a bug?

For example: 

`dataform.json`
```
{
    "warehouse": "bigquery",
    "defaultSchema": "dataform",
    "assertionSchema": "dataform_assertions",
    "defaultDatabase": "my-dev-project"
}

```

`some_query.sqlx`
```
config {
  type: "table",
  database: dataform.projectConfig.vars.embed_my_prd_database,
  assertions: {
    uniqueKey: ["user_id"],
    nonNull: ["user_id", "customer_id"],
    rowConditions: [
      'signup_date is null or signup_date > "2019-01-01"',
      'email like "%@%.%"'
    ]
  }
}
select ...
```

If you run `dataform run --actions=some_query --vars=embed_my_prd_database=my-prd-project`

expect: success

but I got followings because I was using SA credential of `my-prd-project` not of `my-dev-project`
```
Assertion failed:  dataform_assertions.some_query_assertions_uniqueKey_0
  > 
  >       create or replace view `my-dev-project.dataform_assertions.some_query_assertions_uniqueKey_0` as
```

Ekrekr:
A more elegant solution that we have planned for this is CLI flags for `dataform.json` parameters.

newuniverse:
I'm looking forward to it!

Ekrekr:
I think this boils down to the [config not being contextable](https://github.com/dataform-co/dataform/blob/08be45fd193ddc856e69c8590a330024051eb7bc/core/compilers.ts#L75).

The var is for `config` is correctly defined in the raw string there:

```
🚀 ~ file: compilers.ts:55 ~ compileSqlx ~ config: {
  type: "table",
  database: dataform.projectConfig.vars.var1,
}
```

But because `config` is not contextable [during publish(https://github.com/dataform-co/dataform/blob/08be45fd193ddc856e69c8590a330024051eb7bc/core/session.ts#L155), dataform ends up being undefined:

```
🚀 ~ file: table.ts:274 ~ Table ~ config ~ config: { name: 'file', type: 'table', database: undefined }
```

[...]

PR #1454, opened 2023-02-14
Add project config override cli options

Ekrekr:
Fixes https://github.com/dataform-co/dataform/issues/1295

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -1,7 +1,7 @@
 import * as fs from "fs";
 import * as glob from "glob";
 import * as path from "path";
-import yargs from "yargs";
+import yargs, { string } from "yargs";
 
 import * as chokidar from "chokidar";
 import { build, compile, credentials, init, install, run, table, test } from "df/api";
@@ -133,23 +133,6 @@ const includeDependentsOption: INamedOption<yargs.Options> = {
   }
 };
 
-const schemaSuffixOverrideOption: INamedOption<yargs.Options> = {
-  name: "schema-suffix",
-  option: {
-    describe: "A suffix to be appended to output schema names."
-  },
-  check: (argv: yargs.Arguments<any>) => {
-    if (
-      argv[schemaSuffixOverrideOption.name] &&
-      !/^[a-zA-Z_0-9]+$/.test(argv[schemaSuffixOverrideOption.name])
-    ) {
-      throw new Error(
-        `--${schemaSuffixOverrideOption.name} should contain only alphanumeric characters and/or underscores.`
-      );
-    }
-  }
-};
-
 const credentialsOption: INamedOption<yargs.Options> = {
   name: "credentials",
   option: {
@@ -177,24 +160,6 @@ const jsonOutputOption: INamedOption<yargs.Options> = {
   }
 };
 
-const varsOptionName = "vars";
-const varsOption: INamedOption<yargs.Options> = {
-  name: varsOptionName,
-  option: {
-    describe: `Variables to inject via '--${varsOptionName}=someKey=someValue,a=b', referenced by \`dataform.projectConfig.vars.someValue\`.`,
-    type: "string",
-    default: null,
-    coerce: (rawVarsString: string | null) => {
-      const variables: { [key: string]: string } = {};
-      rawVarsString?.split(",").forEach(keyValueStr => {
-        const [key, value] = keyValueStr.split("=");
-        variables[key] = value;
-      });
-      return variables;
-    }
-  }
-};
-
 const timeoutOption: INamedOption<yargs.Options> = {
   name: "timeout",
   option: {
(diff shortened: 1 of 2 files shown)


cli/index.ts
```diff
@@ -802,3 +760,153 @@ export function runCli() {
     yargs.showHelp();
   }
 }
+
+class ProjectConfigOverride {
```
BenBirt, line 764: this structure seems really odd to me, IMO. i am not sure why we are (partially? why are some options both here and outside?) moving options into this class. I buy the need for a utility function which can construct a final project config "override", but the rest feels quite weird.

cli/index.ts
```diff
@@ -206,8 +171,53 @@ const timeoutOption: INamedOption<yargs.Options> = {
   }
 };
 
-const defaultDatabaseOptionName = "default-database";
-const defaultLocationOptionName = "default-location";
+const defaultDatabaseOption: INamedOption<yargs.Options> = {
+  name: "default-database",
+  option: {
+    describe: "The default database to use. For BigQuery, this is a Google Cloud Project ID.",
```
BenBirt, line 177: we should clarify (for all of these) that they override any settings in `dataform.json`

cli/index.ts
```diff
@@ -206,8 +171,53 @@ const timeoutOption: INamedOption<yargs.Options> = {
...
+    type: "string"
+  },
+  check: (argv: yargs.Arguments<any>) => {
+    if (!argv[warehouseOption.name]) {
+      return;
+    }
+    if (
+      argv[defaultDatabaseOption.name] &&
```
BenBirt, line 185: this is necessarily true given the above condition

cli/index.ts
```diff
@@ -206,8 +171,53 @@ const timeoutOption: INamedOption<yargs.Options> = {
...
+    ) {
+      throw new Error(
+        `The --${defaultDatabaseOption.name} flag is only used for BigQuery and Snowflake projects.`
+      );
+    }
+    if (!argv[defaultDatabaseOption.name] && argv[warehouseOption.name] === "bigquery") {
+      throw new Error(
+        `The --${defaultDatabaseOption.name} flag is required for BigQuery projects.`
+      );
+    }
```
BenBirt, line 196: why are we requiring this? we should continue defaulting to whatever is in `dataform.json` (throughout)!

BenBirt (changes requested):

cli/index.ts
Ekrekr: Namespacing is the goal, to differentiate things like `defaultDatabaseOption` and `ProjectConfigOverride.defaultDatabaseOverrideOption` (the latter has override in the name, though the name of the flag is the same).

The idea is that `ProjectConfigOverride` applies to options where `dataform.json` is expected to already exist.

cli/index.ts
Ekrekr: Thread above - this is for init, so does not override anything, and is a required flag in this case.

cli/index.ts
Ekrekr: `argv[warehouseOption.name]` isn't required for options other than `init`

cli/index.ts
BenBirt: I guess I think it is really weird to define two different flags for the same name. Yes I get that they are used in different contexts, but the _flag_, i.e. its name, is the same entity. If one use-case requires more validation than the other, then put it at the callsite?

sukhamritboparai:
hi team, when is this PR merge or feature available?

cli/index.ts
Ekrekr: That makes sense - have de-duped the flags.

Have avoided the misleading `If unset, the value from dataform.json is used.` in `$ dataform init` by overriding the description at init instead.

Ekrekr:
Examples:

```
$ bazel run //packages/@dataform/cli:bin init -- --help
...
dataform init <warehouse> [project-dir]

Create a new dataform project.

Positionals:
  warehouse    The project's data warehouse type. If unset, the value from dataform.json is used.  [required] [choices: "bigquery", "presto", "postgres", "redshift", "snowflake", "sqldatawarehouse"]
  project-dir  The Dataform project directory.  [default: "."]

Options:
  --help              Show help  [boolean]
  --version           Show version number  [boolean]
  --default-database  The default database to use. For BigQuery, this is a Google Cloud Project ID.  [string]
  --default-location  The default BigQuery location to use. See https://cloud.google.com/bigquery/docs/locations for supported values.
  --skip-install      Whether to skip installing NPM packages.  [default: false]
```

```
$ bazel run //packages/@dataform/cli:bin run -- --help
...
dataform run [project-dir]

Run the dataform project's scripts on the configured data warehouse.

Positionals:
  project-dir  The Dataform project directory.  [default: "."]

Options:
  --help                Show help  [boolean]
  --version             Show version number  [boolean]
  --dry-run             If set, built SQL is not run against the data warehouse and instead is printed to the console.  [boolean]
  --run-tests           If set, the project's unit tests are required to pass before running the project.  [boolean]
  --full-refresh        Forces incremental tables to be rebuilt from scratch.  [boolean] [default: false]
  --actions             A list of action names or patterns to run. Can include '*' wildcards.  [array]
  --tags                A list of tags to filter the actions to run.  [array]
  --include-deps        If set, dependencies for selected actions will also be run.  [boolean]
  --include-dependents  If set, dependents (downstream) for selected actions will also be run.  [boolean]
  --credentials         The location of the credentials JSON file to use.  [default: ".df-credentials.json"]
  --json                Outputs a JSON representation of the compiled project.  [boolean] [default: false]
  --timeout             Duration to allow project compilation to complete. Examples: '1s', '10m', etc.  [string] [default: null]
  --warehouse           The project's data warehouse type. If unset, the value from dataform.json is used.  [choices: "bigquery", "presto", "postgres", "redshift", "snowflake", "sqldatawarehouse"]
  --default-database    The default database to use. For BigQuery, this is a Google Cloud Project ID. If unset, the value from dataform.json is used.  [string]
  --default-schema      Override for the default schema name. If unset, the value from dataform.json is used.
  --default-location    The default BigQuery location to use. See https://cloud.google.com/bigquery/docs/locations for supported values. If unset, the value from dataform.json is used.
  --assertion-schema    Default assertion schema. If unset, the value from dataform.json is used.
  --vars                Override for variables to inject via '--vars=someKey=someValue,a=b', referenced by `dataform.projectConfig.vars.someValue`. If unset, the value from dataform.json is used.  [string] [default: null]
  --database-suffix     Default assertion schema. If unset, the value from dataform.json is used.
  --schema-suffix       A suffix to be appended to output schema names. If unset, the value from dataform.json is used.
  --table-prefix        Adds a prefix for all table names. If unset, the value from dataform.json is used.
```

Issue #1398, opened 2022-11-10
Running docker image without a command produces a very confusing error message


BenBirt:
```
$ docker run dataformco/dataform:latest 
Dataform encountered an error: Unknown argument: node
```

(Note that passing a command does work, e.g. `docker run dataformco/dataform:latest help`.)

PR #1456, opened 2023-02-14
Fix docker cmd for image

Ekrekr:
The docker CMD for the node image was `node`, so when no arguments we were running `dataform node`. This change makes the default `dataform help`.

Also move loads to top of file - bazel build linter was complaining.

Fixes https://github.com/dataform-co/dataform/issues/1398

### DIFF
--- a/cli/BUILD
+++ b/cli/BUILD
@@ -1,6 +1,10 @@
-package(default_visibility = ["//visibility:public"])
-
 load("//tools:ts_library.bzl", "ts_library")
+load("//tools/common:copy.bzl", "copy_file")
+load("@io_bazel_rules_docker//container:container.bzl", "container_image", "container_push")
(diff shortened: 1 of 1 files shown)


cli/BUILD
```diff
@@ -51,11 +50,10 @@ container_run_and_commit(
     ],
 )
 
-load("@io_bazel_rules_docker//container:container.bzl", "container_image", "container_push")
-
 container_image(
     name = "image_with_dataform_entrypoint",
     base = ":image_with_dataform_cli_commit.tar",
+    cmd = ["help"],
```
BenBirt, line 56: this would probably be more future-proof if it was just `""`, that way it will always do the default CLI command

(yes it has the same result, but if we change the default we'll only need to do it in one place)

Ekrekr: I tried this with `[""]` which didn't work as expected, but `[]` does seem to work! Updated.

kolina (approved):

Issue #1458, opened 2023-02-15
Dataform CLI outputs an AWS SDK message


wintermi:
The following is output when executing the Dataform CLI:
```
(node:25544) NOTE: The AWS SDK for JavaScript (v2) will be put into maintenance mode in 2023.

Please migrate your code to use AWS SDK for JavaScript (v3).
For more information, check the migration guide at https://a.co/7PzMCcy
(Use `node --trace-warnings ...` to show where the warning was created)
```

Ekrekr:
A workaround for now is to set the Node environment variable: `NODE_NO_WARNINGS=1`

karszawa:
`AWS_SDK_JS_SUPPRESS_MAINTENANCE_MODE_MESSAGE=1` also works, but only for the AWS message.

https://github.com/aws/aws-sdk-js/blob/39e7e8a2b96374ec8e9573c8aaf96267d85d5175/lib/maintenance_mode_message.js#L31

PR #1459, opened 2023-02-15
Disable docker node warnings to prevent futzing of docker output

Ekrekr:
This was putting

```
(node:25544) NOTE: The AWS SDK for JavaScript (v2) will be put into maintenance mode in 2023.

Please migrate your code to use AWS SDK for JavaScript (v3).
For more information, check the migration guide at https://a.co/7PzMCcy
(Use `node --trace-warnings ...` to show where the warning was created)
```

in the output - fine for most things, but could break workflows using things like `--json`

Fixes https://github.com/dataform-co/dataform/issues/1458

### DIFF
--- a/cli/BUILD
+++ b/cli/BUILD
@@ -55,6 +55,9 @@ container_image(
     base = ":image_with_dataform_cli_commit.tar",
     cmd = [],
     entrypoint = ["dataform"],
+    env = {
+        "NODE_NO_WARNINGS": "1",
+    },
     tags = [
         "no-remote",
     ],


BenBirt:
So obviously this works for users who use the Docker image, but what about the CLI more generally? Can we explore capturing the warning more generically, for example with https://nodejs.org/api/process.html#event-warning?

Ekrekr:
I can't get this to work atm, `require` seems to operate in a different scope to what's currently running
https://github.com/dataform-co/dataform/blob/09447c49818c3f517de9054cf70ea19e225a3dc1/api/dbadapters/snowflake.ts#L27

Can see what would otherwise work in the bundle 
image

I'm fine with non-docker users having to pass the `--no-warnings` flag to Node.

BenBirt:
I don't really follow, what does the snowflake SDK have to do with it? why can't we add the warning handler before any import happens?

Ekrekr:
`process.env.NODE_NO_WARNINGS = "1";` doesn't work either sadly

Will try to figure out how to bump our node version

Ekrekr:
No longer relevant, no more AWS

Issue #1466, opened 2023-03-10
Dataform cli filtering union of tags and not intersection


AnOtterGithubUser:
I have noticed that running dataform with the --tags flag runs any action which has at least one tag mentioned in the flag. The resulting behaviour is a union of actions.
Ex: `dataform run --tags A --tags B` runs any action with tag A or tag B
I would like to run the intersection of actions. Ex: `dataform run --tags A --tags B` would run actions with tag A AND tag B.      
Is there any way to do it with the cli ? If not, it would be a nice feature
Thanks

BenBirt:
There's no way of doing this, no. Feature request understood, though!

AnOtterGithubUser:
Thanks for the quick reply ! I can give it a try

PR #1467, opened 2023-03-11
[#1466] Filter intersection of tags

AnOtterGithubUser:
Hello,
I added a new flag `--include-all-tags` that works with `--tags` options to get the behavior requested in #1466    
Default flag value is set to false to keep existing behavior unchanged

### DIFF
--- a/api/commands/prune.ts
+++ b/api/commands/prune.ts
@@ -47,6 +47,7 @@ function computeIncludedActionNames(
 
   const hasActionSelector = runConfig.actions?.length > 0;
   const hasTagSelector = runConfig.tags?.length > 0;
+  const hasIncludeAllTagsSelector = runConfig.includeAllTags;
 
   // If no selectors, return all actions.
   if (!hasActionSelector && !hasTagSelector) {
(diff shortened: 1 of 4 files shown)


protos/execution.proto
```diff
@@ -10,6 +10,7 @@ option go_package = "github.com/dataform-co/dataform/protos/dataform";
 message RunConfig {
   repeated string actions = 1;
   repeated string tags = 5;
+  bool include_all_tags = 10;
```
Ekrekr, line 13: "include all" is ambiguous in semantic meaning - IMO something like `filterTagsAsIntersection` would be better as it's more explicit.

tests/api/api.spec.ts
```diff
@@ -395,6 +400,23 @@ suite("@dataform/api", () => {
...
+      const actionNames = [
+        ...prunedGraph.tables.map(action => targetAsReadableString(action.target)),
+        ...prunedGraph.operations.map(action => targetAsReadableString(action.target))
+      ];
+      expect(actionNames).includes("schema.op_e");
+      expect(actionNames).includes("schema.tab_a");
+      expect(actionNames).not.includes("schema.op_a");
+      expect(actionNames).not.includes("schema.op_b");
+      expect(actionNames).not.includes("schema.op_c");
+      expect(actionNames).not.includes("schema.op_d");
```
Ekrekr, line 417: This would be better as `expect(actionNames).deep.equals(["schema.op_e", "schema.op_a"]);`

api/commands/prune.ts
```diff
@@ -64,9 +65,18 @@ function computeIncludedActionNames(
 
   // Determine actions selected with --tag option and update applicable actions
   if (hasTagSelector) {
-    allActions
-      .filter(action => action.tags.some(tag => runConfig.tags.includes(tag)))
-      .forEach(action => includedActionNames.add(targetAsReadableString(action.target)));
+    // Keep actions wich include every tags in --tag option
```
Ekrekr, line 68:
```suggestion
    // Select actions which match the intersection of the set of tags.
```

api/commands/prune.ts
```diff
@@ -64,9 +65,18 @@ function computeIncludedActionNames(
...
+    if (hasIncludeAllTagsSelector) {
+      allActions
+        .filter(action => runConfig.tags.every(tag => action.tags.includes(tag)))
+        .forEach(action => includedActionNames.add(targetAsReadableString(action.target)));
+    }
+    // Keep actions with include at least one tag in --tag option
```
Ekrekr, line 74:
```suggestion
    // Select actions which match the union of the set of tags.
```

Ekrekr (changes requested):
Sorry that this went under the radar for so long! It just needs a sync and a few nits, then we can get this in.

zsomborkincses:
It would be great if this feature was added!

apilaskowski:
Closing, as new PR was proposed base on this, and no movement in either.

PR #1490, opened 2023-05-28
Update sql-formatter & specify SQL language according to warehouse

### LINKED ISSUE #1489: Updating sql-formatter will resolve several formatting issues
Would you like to update [sql-formatter](https://www.npmjs.com/package/sql-formatter)?

Currently, dataform CLI depends on version `^2.3.3`, which is about 4 years old.
The latest version is `12.2.1`.

By updating, we can resolve the following issues related to formatting:

- #1070
  - A fix has been merged in [sql-formatter-org/sql-formatter#603](https://github.com/sql-formatter-org/sql-formatter/pull/603). (I did it)
- #1077
  - The current version of sql-formatter can handle this.
- #1444
  - This requires to fix the lexer of this repo as well. (so updating doesn't resolve this)

## Handling dialects in sql-formatter

The current version of sql-formatter have a feature for "dialects" specific to different database products.
Now sql-formatter covers all SQL dialects supported warehouse by Dataform.
https://github.com/sql-formatter-org/sql-formatter/blob/master/docs/language.md

By passing a language value corresponding to the warehouse setting in Dataform to the formatter, we can expect more appropriate formatting. 
But if we use the neutral default `sql` that means StandardSQL, formatting for UNNEST and QUALIFY will not work well, and I think it need to refer to the warehouse value in `dataform.json`.

I have also confirmed that the formatting result changes in some cases. The files under `examples/formatter/` seems to assume BigQuery, so if we format as BigQuery,

- The space after `IF`, `IFNULL` disappears
- A space is added after `UNNEST`
- A line break is added after `ALTER TABLE {name}`

While sql-formatter updating brings differences in the formatting results, QUALIFY is a frequently used clause, and named arguments are broken by current formatter. The update makes the format command practical.

Could you take a look at the pull request I'm going to send here? You may use only some of the commits.

pokutuna:
resolves #1489


sqlx/format.ts
```diff
@@ -1,4 +1,3 @@
-import * as crypto from "crypto";
```
pokutuna, line 1: an unused import

sqlx/format.ts
```diff
@@ -16,11 +15,13 @@ const JS_BEAUTIFY_OPTIONS: JsBeautifyOptions = {
 
 const MAX_SQL_FORMAT_ATTEMPTS = 5;
 
-export function format(text: string, fileExtension: string) {
+export type SqlLanguage = sqlFormatter.SqlLanguage;
```
pokutuna, line 18: Adding a dependenncy to `WarehouseType (df/core/adapters)` looks unwelcome, so I took the sql-formatter argument type here. But it may seem smarter to define `Record<WarehouseType, SqlLanguage>` here.

cli/index.ts
```diff
@@ -715,14 +724,28 @@ export function runCli() {
...
+            const dataformJson = fs.readFileSync(path.resolve(argv[projectDirMustExistOption.name], "dataform.json"), 'utf8');
+            const projectConfig = JSON.parse(dataformJson);
+            warehouse = projectConfig.warehouse;
+          } catch {
+            throw new Error('Could not parse dataform.json');
+          }
+          if (!dbadapters.validWarehouses.includes(warehouse)) {
+            throw new Error("Unrecognized 'warehouse' setting in dataform.json");
+          }
+          const language = warehouseSqlLanguageMap[warehouse as WarehouseType];
```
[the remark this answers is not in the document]
BenBirt: this is probably the best way for now to be honest!

cli/index.ts
```diff
@@ -715,14 +724,28 @@ export function runCli() {
         positionalOptions: [projectDirMustExistOption],
         options: [trackOption],
         processFn: async argv => {
+          let warehouse: string | undefined;
+          try {
+            const dataformJson = fs.readFileSync(path.resolve(argv[projectDirMustExistOption.name], "dataform.json"), 'utf8');
+            const projectConfig = JSON.parse(dataformJson);
+            warehouse = projectConfig.warehouse;
+          } catch {
+            throw new Error('Could not parse dataform.json');
```
BenBirt, line 733: include the thrown error message?

sqlx/format.ts
BenBirt: agreed - while it is not ideal to add a dependency to `core` here, I think it's still probably cleaner to use `WarehouseType` in `format`'s API, and hide `sql-formatter` types from the caller (and so move `Record<WarehouseType, SqlLanguage>` here, as you say).

BenBirt (changes requested):

cli/index.ts
pokutuna: Does this fix look good?
70338e0a06

sqlx/format.ts
pokutuna: Thanks for reviewing! I agree that moving the definitions around would look cleaner.

But adding a dependency in sqlx/BUILD to core, it causes a circular dependency.
```
pokutuna@instance-1:~/dataform$ bazel test //tests/sqlx/...
ERROR: /home/pokutuna/dataform/sqlx/BUILD:5:11: in ts_library rule //sqlx:sqlx: cycle in dependency graph:
    //tests/sqlx:format.spec (9bf3c46410944eee39af2a199012f4cb7dbe9d88374c707ffb94070f4c0045b9)
    //tests/sqlx:tests (15696f0530617362d5668c59388c0b9089022219dd20111c87cd2937b947b8fe)
.-> //sqlx:sqlx (15696f0530617362d5668c59388c0b9089022219dd20111c87cd2937b947b8fe)
|   //core:core (15696f0530617362d5668c59388c0b9089022219dd20111c87cd2937b947b8fe)
`-- //sqlx:sqlx (15696f0530617362d5668c59388c0b9089022219dd20111c87cd2937b947b8fe)
ERROR: Analysis of target '//tests/sqlx:format.spec' failed; build aborted
```

How should this be resolved?
If I cut out them to `//core/adapters` as a new build target, the dependency will not cycle.
- core -> sqlx
- core -> core/adapters
- sqlx -> core/adapters

Or it could move only some definitions to common/warehouses.
(I'm not familiar with bazel and I'm speaking from a feeling )

I feel these ideas make more changes, so if I need to create a new BUILD, I will do it in another separate PR. After that, I rebase this.
Can you give me some direction on how to fix it?

cli/index.ts
BenBirt: no need to use `ErrorWithCause` unless you really want to; I think the message is enough in this case!

sqlx/format.ts
BenBirt: I think the issue is that the `//sqlx:sqlx` BUILD rule is too large, i.e. it's pulling in both lexing/parsing AND formatting.

I'd suggest cutting `//sqlx:sqlx` into two rules: `format` and `sqlx`. `format` will depend on `sqlx`, and only include `format.ts` in its `srcs`. `sqlx` will include all other files (you can use `exclude` to do this: https://bazel.build/reference/be/functions#glob).

Then we'll have:
`//sqlx:format` -> `//sqlx:sqlx`
`//sqlx:format` -> `//core:core`
`//core:core` -> `//sqlx:sqlx`

i.e. no cycles.

pokutuna: Thanks for the suggestion, I made another PR.
#1499

cli/index.ts
pokutuna: I've reverted it & include only the message.
https://github.com/dataform-co/dataform/pull/1490/commits/9a613148c4258fcf6e50041506e41a5f1d53eb9f

sqlx/format.ts
BenBirt: Great, thanks. Merge that in here and update this PR and I will approve!

(Also, if you increment the minor version in `version.bzl`, I will publish a new CLI version with your changes.)

cli/index.ts
```diff
@@ -715,14 +715,31 @@ export function runCli() {
...
+              warehouse = projectConfig.warehouse;
+            } catch (e) {
+              throw new Error(`Could not parse dataform.json: ${e.message}`);
+            }
+            if (!isWarehouseType(warehouse)) {
+              throw new Error("Unrecognized 'warehouse' setting in dataform.json");
+            }
+            return warehouse;
+          };
+          const warehouse = readWarehouseConfig();
```
pokutuna, line 732: I defined a immediate function because it is convenient for typing and scoping

pokutuna:
I've merged the main branch & update codes.
Thanks for giving me the experience of bump minor version.

However, CI is failing.
As I can't see the results in the Cloud Build, could you inform me what the issue might be?
Alternatively, if it's an easy fix, would it be possible for a core member to handle it?

I initially thought it might be a lint issue and made corrections accordingly, but it seems I was mistaken. 9d797db
(I'm unsure of how to execute tslint with bazel! It would be helpful if it were included in CONTRIBUTING.md).

#1499 passes, but the merge commit f27fb73c0 is failing. Maybe it's not a problem with this pull request.

core/adapters/index.ts
```diff
@@ -41,6 +41,10 @@ export enum WarehouseType {
   SQLDATAWAREHOUSE = "sqldatawarehouse"
 }
 
+export function isWarehouseType(input: any): input is WarehouseType {
```
BenBirt, line 44: `input: string`?

sqlx/format.ts
```diff
@@ -15,13 +16,26 @@ const JS_BEAUTIFY_OPTIONS: JsBeautifyOptions = {
 
 const MAX_SQL_FORMAT_ATTEMPTS = 5;
 
-export type SqlLanguage = sqlFormatter.SqlLanguage;
+const WAREHOUSE_LANGUAGE_MAP: Record<WarehouseType, sqlFormatter.SqlLanguage> = {
+  [WarehouseType.BIGQUERY]:"bigquery",
```
BenBirt, line 20: nit: seems unformatted

BenBirt (approved):

BenBirt:
The build failure is indeed a lint issue. You can see how we run tslint here: https://github.com/dataform-co/dataform/blob/main/scripts/cloudbuild/bazel_test#L6

core/adapters/index.ts
pokutuna: It may take string, since this does type narrowing using user-defined type guard, I think it would be better to take any or unknown.

pokutuna:
Thanks for telling me how to run lint.
I should have noticed myself to check cloudbuild.yaml.

PR #1499, opened 2023-06-06
Separate sqlx build target into sqlx and format

pokutuna:
This change is from the conversation here: https://github.com/dataform-co/dataform/pull/1490#discussion_r1208628703

`//sqlx` includes two functionalities: a lexer for sqlx files and a formatter.
In #1490, I worked on improving the formatter, but encountered a circular dependency.

```
//core -> //sqlx (using lexer)
//sqlx -> //core (using adapters)
```

As @BenBirt suggested, I will separate the `sqlx` target into two parts to solve the circular deps.

With this PR, the dependencies are as follows
```
//sqlx:format -> //sqlx:sqlx
//sqlx:format -> //core:core
//core:core -> //sqlx:sqlx
```


core/BUILD
```diff
@@ -26,7 +26,8 @@ ts_library(
         "//common/protos",
         "//common/strings",
         "//protos:ts",
-        "//sqlx",
+        "//sqlx:sqlx",
+        "@npm//@types/node",
```
pokutuna, line 30: `core` used `@types/node` transitively through `//sqlx`.
 Now `//sqlx:sqlx` is a pure lexer and does not depend on `@types/node`, so I have added here.

Specifically, core need it in places like this
https://github.com/dataform-co/dataform/blob/69cad94ab695d7a0f11d1d227226cc519ea43298/core/utils.ts#L92

Issue #1483, opened 2023-05-05
External contributors cannot pass the test commands in contributing.md


pokutuna:
This is a guide to run tests provided in contributing.md

https://github.com/dataform-co/dataform/blob/f58bbddf599ba1efe3c1d927213c3209a4b8681b/contributing.md#L35-L43

However, it fails with the following output:
```
ERROR: /home/pokutuna/dataform/test_credentials/BUILD:5:14: Action test_credentials/bigquery.json failed: (Exit 1): gcloud failed: error executing command external/gcloud_sdk/bin/gcloud kms decrypt '--ciphertext-file=test_credentials/bigquery.json.enc' '--plaintext-file=bazel-out/k8-py2-fastbuild/bin/test_credentials/bigquery.json' ... (remaining 4 arguments skipped)
 ERROR: (gcloud.kms.decrypt) PERMISSION_DENIED: Request had insufficient authentication scopes.
 - '@type': type.googleapis.com/google.rpc.ErrorInfo
   domain: googleapis.com
   metadata:
     method: google.cloud.kms.v1.KeyManagementService.Decrypt
     service: cloudkms.googleapis.com
   reason: ACCESS_TOKEN_SCOPE_INSUFFICIENT
```

This command runs all tests except those depending on secrets. But the tests under the `tests/cli/` also rely on Cloud KMS.
As a result, external contributors cannot run these tests.

This line is depending on a secret.
https://github.com/dataform-co/dataform/blob/f58bbddf599ba1efe3c1d927213c3209a4b8681b/tests/cli/BUILD#L17

Ekrekr:
Thanks for pointing this out!

PR #1505, opened 2023-06-21
Enable formatting for triple-quoted strings

### LINKED ISSUE #1444: `dataform format` breaks triple-quoted strings in BigQuery
BigQuery SQL supports [triple-quoted strings](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#literals) wrapped in `'''`.

When running `dataform format` against a SQLX file containing a triple-quoted string, the quotes are formatted like the below example:

(original)
```sql
'''
select 1
'''
```

(after formatting) 
```sql
''
'
select 1
'
''
```

BigQuery fails when executing the query after this formatting change.

pokutuna:
resolves #1444 

In #1489, I worked on improving the formatter but didn't include support for multiline strings using triple-quotes as it required a little bigger changes in the sqlx lexer.

In this pull request, I have added new states and tokens to the sqlx lexer to handle triple-quoted multiline strings.

While the changes might look complex, what it does is simple:
- Define a minimal subset of functionality based on how existing quotes are handled.
  - With triple-quoted strings, there's no need to worry about escaping quotes.
- During formatting, if it's a multi-line string, don't format within it and split by lines.
  - Need to keep internal line breaks and whitespaces intact.


tests/sqlx/format.spec.ts
```diff
@@ -146,7 +146,56 @@ WHERE
...
+
+post_operations {
+  select
+    """1""" as inner_sql
+}
 `);
     });
   });
+
+  suite("formatter todos", () => {
```
pokutuna, line 180: I realized that previously, whether using single or triple quotes, template strings inside string literals could not be formatted. I made it like this because there was no way to represent todo tests.

pokutuna:
I can't see the CI results, but followings are passed in my env.
- `$ bazel test --build_tests_only -- ... -tests/integration/...` (except #1483 )
-  `$ bazel build @npm//tslint/bin:tslint && bazel-bin/external/npm/tslint/bin/tslint.sh --project .`

tests/sqlx/format.spec.ts
```diff
@@ -146,7 +146,56 @@ WHERE
...
+    test("TODO format tempalte string in a string", async () => {
```
BenBirt, line 181: nit: `template`

pokutuna:
Thank you for reviewing. I've fixed typos.

desbo:
Thank you for the fix @pokutuna!

PR #1536, opened 2023-08-29
Sandbox2 wrapper

lewish:
Adds a Sandbox2 wrapper for compilation.
- Change the way we communicate between child processes to use unix domain sockets
- Add new webpacked worker bundle, and unbundled vm2 dependencies
- Add Sandbox2 wrapper script for running the worker bundle inside node
- Update tests to test both sandbox2 and non-sandbox2 mode

Disable tests for CLI (which is now broken in this branch), and integration tests that have a package dependency issue on this older branch.


sandbox/compile_executor.cc
```diff
@@ -0,0 +1,252 @@
...
+                      .BuildOrDie();
+
+    sandbox2::Sandbox2 s2(std::move(executor), std::move(policy));
+
+    // If the sandbox program fails to start, return early.
+    if (!s2.RunAsync())
+    {
+        auto result = s2.AwaitResultWithTimeout(
+            absl::Seconds(TIMEOUT_SECS + FALLBACK_TIMEOUT_DELAY));
+        LOG(ERROR) << "sandbox failed to start: " << result->ToString();
```
Ekrekr, line 239: Are things logged to stderr readable from the Javascript? I'm slightly confused by the socket usage above.

sandbox/worker/worker.ts
```diff
@@ -0,0 +1,3 @@
+import { compileAndSend } from "df/sandbox/vm/compile";
+
+compileAndSend(process.argv[2], process.argv[3]);
```
Ekrekr, line 3: nit: newline

scripts/cloudbuild/bazel_test
```diff
@@ -3,7 +3,7 @@ set -e
 
 # Run tslint.
 bazel run @nodejs//:yarn
-bazel build @npm//tslint/bin:tslint && bazel-bin/external/npm/tslint/bin/tslint.sh --project .
+# bazel build @npm//tslint/bin:tslint && bazel-bin/external/npm/tslint/bin/tslint.sh --project .
```
Ekrekr, line 6: nit: re-add or rm

api/commands/compile.ts
```diff
@@ -40,67 +45,70 @@ export async function compile(
...
+
+  server.listen(socketPath);
+
+  await CompileChildProcess.forkProcess(socketPath, { ...compileConfig, useMain: false }, useSandbox2).timeout(compileConfig.timeoutMillis || 5000);
 
-  return decode64(dataform.CompiledGraph, result);
+  await promisify(server.close.bind(server))();
+
+  if (result.startsWith("ERROR:")) {
+    throw coerceAsError(JSON.parse(result.substring(6)));
```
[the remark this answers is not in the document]
lewish: Yup. This is obviously hacky as hell, but if you look at the compileAndSend function of sandbox/vm/compile.ts, you'll see this is guaranteed.

sandbox/compile_executor.cc
lewish: They are, but we don't read them. But this might be useful if we do run into errors, that they will be in our logs at least.
Basically, all we care about is the process return code for the sandbox. The logs are ignored, and all communication back to the parent parent process goes indirectly via the socket.

sandbox/compile_executor.cc
```diff
@@ -0,0 +1,252 @@
...
+        return EXIT_FAILURE;
+    }
+
+    auto result = s2.AwaitResultWithTimeout(
+        absl::Seconds(TIMEOUT_SECS + FALLBACK_TIMEOUT_DELAY));
+
```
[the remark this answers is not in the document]
lewish: Added this check below

sandbox/compile_executor.cc
Ekrekr: Fair enough, if we need to add errors based off of the sandbox2 final result code we can do that after

Ekrekr (approved):

PR #1540, opened 2023-09-15
windows compile fix

mathlr:
I've been using Dataform on Windows machine since v1.x.x and after a very early version (which I don't remember now), my compile stopped working. Exactly 1 year ago I tried the slack channel for some support with no reply, so I debugged bundle.js and coded myself a quick fix to continue using, and with every Dataform update, I had to "patch" bundle.js for it to continue to work for me.
Recently, when I upgraded to 2.6.4 (Fix `dataform compile` on Windows), my own patch stopped working as well, so I went back on debugging again to code my own fix.

I am NOWHERE NEAR a javascript dev, so I apologize in advance for any misconception. What I found was:
The code doesn't need to know which system you are in (and therefore which folder separator to use) during the execution, but only in the compiling step.

For this:
-  I rolled back the pathSeperator to the default "/" where it is system agnostic; and
-  In the main function, I coded a local function to get the baseFilename that would use the default "/" as the separator.

### DIFF
--- a/core/main.ts
+++ b/core/main.ts
@@ -47,14 +47,16 @@ export function main(encodedCoreExecutionRequest: string): string {
   // We delay attaching them to `global` until after all have been required, to prevent
   // "includes" files from implicitly depending on other "includes" files.
   const topLevelIncludes: {[key: string]: any} = {};
+  function localBaseFilename(fullPath) {
+    return fullPath.split("/").slice(-1)[0].split(".")[0];};
   compileRequest.compileConfig.filePaths
-    .filter(path => path.startsWith(`includes${utils.pathSeperator}`))
-    .filter(path => path.split(utils.pathSeperator).length === 2) // Only include top-level "includes" files.
+    .filter(path => path.startsWith(`includes/`))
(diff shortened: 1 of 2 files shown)


BenBirt:
Thanks for this!

Apologies for the flaky Windows support. I have found it tricky to get a working development environment set up for it.

I am surprised that `2.6.4` doesn't work, though. Can you please verify that:
1) you are using `>=2.6.4` @dataform/cli (`npm i -g @dataform/cli`)
2) you are using `>=2.6.4` @dataform/core (update `package.json` and run `dataform install` or `npm i` - they are equivalent)

mathlr:
Thanks for the reply @BenBirt 
I was running an old version for the CLI, and now bundle.js reads correctly the folder tree structure. I totally misunderstood the concept on the interaction between core and cli packages.
Sorry for this hooky mistake.
No more self patches for me :)

BenBirt:
Excellent, glad we could get this fixed for you!

In general you shouldn't **need** to worry about interdependencies between the core and cli packages - by design we intentionally keep them cross-functional - but sometimes this isn't possible, unfortunately.

PR #1550, opened 2023-10-17
Remove non bigquery warehouses

Ekrekr:
b/305915462


core/session.ts
```diff
@@ -586,127 +572,6 @@ export class Session {
         );
       }
 
-      // materialized
-      if (!!table.materialized) {
```
lewish, line 590: I don't think you want to remove this, it's still a valid check

elyisu: Hi @Ekrekr ! Does this mean we are dropping support for non bigquery data warehouses from dataform for good?

Ekrekr: We're still experimenting here, nothing is official until we announce it on https://cloud.google.com/dataform/docs/release-notes.

We want to improve performance, quality and quantity of features for the BigQuery offering - and removing this huge amount of code that requires maintenance, testing, and backwards compatibility, will facilitate this.

Following this, the current plan is to:

* Keep maintaining current functionality under version 2, following the 12 month GCP deprecation policy.

* Remove adapters in version 3 (note that is the merge target of this PR).

Feel free to drop me a message at [eliaskassell@google.com](mailto:eliaskassell@gmail.com) - I'd love to hear your opinion. I'd like to keep this PR uncluttered!

PR #1551, opened 2023-10-17
Fixing bug for target name not escaped

maverickjoy:
Fixing bug for filenames not escaped in compiled graph target and canonical target : b/281045260
Testing: Could not add integration test file with a backslash because build rules doesn't allow it, but have tested manually the object: https://screenshot.googleplex.com/B4jrhURyKiyxozB

### DIFF
--- a/core/compilers.ts
+++ b/core/compilers.ts
@@ -32,7 +32,7 @@ function compileTableSql(code: string, path: string) {
   );
 
   return `
-  publish("${utils.baseFilename(path)}").query(ctx => {
+  publish("${utils.getEscapedFileName(path)}").query(ctx => {
     ${functionsBindings.join("\n")}
     ${js}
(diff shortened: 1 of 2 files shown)


Ekrekr:
Can we not test this in the `name` part of `sqlxConfig` https://github.com/dataform-co/dataform/blob/main/tests/core/backslashes-act-literally.js.test

maverickjoy:
> Can we not test this in the `name` part of `sqlxConfig` https://github.com/dataform-co/dataform/blob/main/tests/core/backslashes-act-literally.js.test

This is a post operation after extracting the paths and file names, our root cause of error has happened before this.

Ekrekr:
Ah fair point! LGTM

Ekrekr (approved):

maverickjoy:
Thank You

PR #1552, opened 2023-10-18
Minify @dataform/core and bundle all dependencies

lewish:
Reduces total node_modules uncompressed down from 9.4Mb down to ~432Kb (20x reduction)
Reduces compressed node_modules size down from 1.57Mb to 83Kb
Reduces total node_modules files from 381 to 4

Running through our performance tests for Dataform on GCP, this appears to increase performance of compilation significantly (around 3x faster).

While this is a bit of a faux pas for a normal NPM package, @dataform/core is not a normal NPM package. It is rarely used with other interacting NPM modules, it is usually the only package installed. Additionally, it is often used in contexts where performance of installing and executing the library are important, or where bandwidth, file operations can be limited.

Functional changes:
- No longer generates types (a `bundle.d.ts` file)


BenBirt (approved):
Slightly scary that we don't understand the speedup, but hey. Ship it!

Ekrekr (approved):

PR #1557, opened 2023-10-26
Consolidate cli package

Ekrekr:
Move the Dataform API and VM packages under the CLI, as that's what uses them


lewish (approved):
Nice, this was a common source of confusion!

PR #1558, opened 2023-10-26
Remove formatter test files from examples

Ekrekr:
They're really bad examples of SQLX.

For testing the `formatFile` function, write a file to the temporary Bazel directory instead.

Also make the formatter test structure truly googley.


sqlx/BUILD
```diff
@@ -17,17 +17,33 @@ ts_library(
...
-        ["format.ts"],
-    ),
+    srcs = [
+        "format.ts"
+    ],
     deps = [
         "//:modules-fix",
         "//common/errors",
-        "//sqlx:sqlx",
+        "//sqlx:lexer",
```
lewish, line 26: can just be :lexer

lewish:
FWIW, for changes like this (and actually your last PR) there isn't really reason to batch these up into v3 branch, as they are no-ops, just cleaning code. I'd consider just merging them into the main branch instead.

lewish (approved):

Ekrekr:
> FWIW, for changes like this (and actually your last PR) there isn't really reason to batch these up into v3 branch, as they are no-ops, just cleaning code. I'd consider just merging them into the main branch instead.

Good shout! Done.

PR #1559, opened 2023-10-27
Improve example projects

Ekrekr:
* Import BigQuery stackoverflow example https://github.com/dataform-co/dataform-example-project-bigquery/tree/master
* Move examples which are actually just tests into the tests/api folder, which is what runs them


examples/stackoverflow_reporter/package.json
```diff
@@ -0,0 +1,5 @@
+{
+    "dependencies": {
+        "@dataform/core": "2.3.0"
```
lewish, line 3: I don't think you need this file? But if you do, make the version latest

lewish (approved):

PR #1564, opened 2023-11-01
Deprecate support for old type `.sql` files

Ekrekr:


tests/api/projects/backwards_compatibility/BUILD
lewish, line 1: I think we should keep this test, but I would remove all the old SQL files, and bump the core version to something more recent that we do want to continue to support (probably 2.0.0) and add some basic SQLX files in here.

The purpose of this test is partially to make sure we don't break the compilation "contract" between the old versions of core and the new versions of the CLI, which is still valuable IMO.

Ekrekr: Fair enough! Done.

PR #1576, opened 2023-11-14
Add initial action config proto definitions

Ekrekr:


protos/core.proto
```diff
@@ -291,3 +291,87 @@ message CompileExecutionRequest {
...
+  bool disabled = 1;
+
+  // Declares that this `operations` action creates a dataset which should be
+  // referenceable using the `ref` function.
+  bool has_output = 2;
+}
+
+// Configuration options for BigQuery table declarations.
+// TODO(ekrekr): remove this empty message, if explicit action type without a
+// `type` field proves to be impossible / ugly.
```
BenBirt, line 367: fwiw: i am much more comfortable keeping this message around now that we no longer have a `type` field - it has semantic purpose now.

BenBirt (approved):

protos/core.proto
Ekrekr: Nice - I want to make sure that all use cases do end up being OK without `type` while I play around with this a bit, which is why I've left this comment.

PR #1580, opened 2023-11-16
Read workflow settings

Ekrekr:
This means:
* Implementing a way of loading YAML as JS objects.
* Converting these to project configs.
* Adding tests for `main`, as that's where this loading and conversion happens.


core/compilers.ts
```diff
@@ -1,10 +1,23 @@
...
 
-export function compile(code: string, path: string) {
+export function compile(code: string, path: string): string {
   if (path.endsWith(".sqlx")) {
     return compileSqlx(SyntaxTreeNode.create(code), path);
   }
+  if (path.endsWith(".yaml")) {
+    try {
+      const yamlAsJson = JSON.stringify(loadYaml(code) as { [key: string]: number });
+      return `exports.asJson = () => (${yamlAsJson})`;
```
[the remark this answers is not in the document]
Ekrekr: ikr, can you imagine the amount of sandboxing we have to put around this

core/compilers.ts
```diff
@@ -1,10 +1,23 @@
...
 import { SyntaxTreeNode, SyntaxTreeNodeType } from "df/sqlx/lexer";
 
-export function compile(code: string, path: string) {
+export function compile(code: string, path: string): string {
   if (path.endsWith(".sqlx")) {
     return compileSqlx(SyntaxTreeNode.create(code), path);
   }
+  if (path.endsWith(".yaml")) {
+    try {
+      const yamlAsJson = JSON.stringify(loadYaml(code) as { [key: string]: number });
```
BenBirt, line 12: instead of `as` just use `@types/js-yaml`?

core/compilers.ts
BenBirt, line 13: why bother with the function? why not return the JSON string directly?

core/index.ts
```diff
@@ -19,4 +20,4 @@ const supportedFeatures = [dataform.SupportedFeatures.ARRAY_BUFFER_IPC];
 
 // These exports constitute the public API of @dataform/core.
 // Changes to these will break @dataform/api, so take care!
-export { compiler, main, session, supportedFeatures, version };
+export { compiler, main, readWorkflowSettings, session, supportedFeatures, version };
```
BenBirt, line 23: if you are exposing this to call it from the CLI and/or closed-source... i would strongly recommend _not_ doing that.

it exposes the caller to hacked versions of `@dataform/core` which could do anything in `readWorkflowSettings`. yes, it can still be run in a sandbox, but why expose yourself to a hacked version when the caller could just read the YAML directly and trust that it reads what is **actually** there?

core/workflow_settings.ts
```diff
@@ -0,0 +1,46 @@
+import { dataform } from "df/protos/ts";
+
+export function readWorkflowSettings(projectDir: string): dataform.ProjectConfig {
+  let workflowSettingsFileExists = false;
+  let workflowSettingsPath: string | undefined;
+  try {
+    workflowSettingsPath = require.resolve(projectDir + "/workflow_settings.yaml");
+    workflowSettingsFileExists = true;
+  } catch (e) {}
```
BenBirt, line 9:
```
function maybeRequire(file: string): unknown? {
  try {
    return require(file);
  } catch (e) { return undefined }
}
```

```
  dataformJson = maybeRequire("dataform.json");
  workflowSettingsJson = maybeRequire("workflow_settings.yaml");
  if (dataformJson && workflowSettingsJson) { throw ... }
  // etc
```

core/workflow_settings.ts
```diff
@@ -0,0 +1,46 @@
+import { dataform } from "df/protos/ts";
+
+export function readWorkflowSettings(projectDir: string): dataform.ProjectConfig {
```
BenBirt, line 3: i don't think this code should be aware of `projectDir` at all. (and indeed reading `dataform.json` previously it was not!)

one of the whole reasons for the sandbox is to be able to root execution in that directory - require should (does) always start with that as root.

BenBirt (changes requested):

core/compilation_sql/BUILD
```diff
@@ -0,0 +1,13 @@
+load("//tools:ts_library.bzl", "ts_library")
```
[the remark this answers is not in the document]
Ekrekr: I'm a bit confused, there is an `index.ts` image

This file could not be in this directory, but then the `tslint.json` would apply to the whole of core: `"tsr-detect-sql-literal-injection": false`

core/compilers.ts
Ekrekr: NodeVM wraps it in a bunch of boilerplate, which makes that not work due to the wrapper

image

Putting an exports seemed like the clean way to do it

core/index.ts
Ekrekr: I don't think the security 

* CLI callers are always potentially exposed to hacked versions of dependencies in their `node_modules`, I don't see why this makes it any more or less secure.
* Closed source will run it using the same sandbox as `main` is invoked from.

> the caller could just read the YAML directly

I don't want to do this in two places. It means reading the YAML, converting it to JSON, and then validating that the proto is correct. Doing it in one place makes it a lot more consistent and less work.

core/workflow_settings.ts
Ekrekr: Thanks, much cleaner

core/workflow_settings.ts
Ekrekr: Done.

core/compilation_sql/BUILD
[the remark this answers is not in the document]
BenBirt: Ah, sorry, I didn't realise it was already there (expected it to be in this PR).

core/index.ts
BenBirt: Yes, agreed, there are no (immediate) security implications here - I didn't say that.

The point is that it is useful to be able to read files and trust what you read. If a caller (outside of `@dataform/core`) wants to read `workflow_settings.yaml`, presumably it wants to read what is _actually there_, not (potentially) something else entirely.

I feel pretty strongly we should not be using this function outside of `@dataform/core`, for that reason alone.

core/compilers.ts
BenBirt: Fair enough!

core/index.ts
Ekrekr: I can sort of buy that, but it would be weird for a user to call `main` from a different Node context to `readWorkflowSettings`.

I've removed the export for now anyway, and will see if I can find a workaround for restricting the warehouse, that doesn't require calling that function first or changing the request to the compiler.

BenBirt (approved):

PR #1581, opened 2023-11-20
Validate workflow settings fields

Ekrekr:
Include a generic best effort Protobuf field validator.

### DIFF
--- a/common/protos/index.ts
+++ b/common/protos/index.ts
@@ -14,6 +14,33 @@ export interface IProtoClass<IProto, Proto> {
   fromObject(obj: { [k: string]: any }): Proto;
 }
 
+// ProtobufJS's native verify method does not check that only defined fields are present.
+// TODO(ekrekr): swap to Typescript protobuf library rather than having to do this.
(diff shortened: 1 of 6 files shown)


common/protos/index.ts
```diff
@@ -14,6 +14,33 @@ export interface IProtoClass<IProto, Proto> {
   fromObject(obj: { [k: string]: any }): Proto;
 }
 
+// ProtobufJS's native verify method does not check that only defined fields are present.
+// TODO(ekrekr): swap to Typescript protobuf library rather than having to do this.
+export function verifyObjectMatchesProto<Proto>(
```
BenBirt, line 19: i hate this, let's please switch

BenBirt (approved):

common/protos/index.ts
Ekrekr: I did some digging into this, and the alternatives which can validate (https://www.npmjs.com/package/@protobuf-ts/plugin, https://www.npmjs.com/package/protoc-gen-ts) are significantly slower. More details in b/305915462#comment18.

PR #1585, opened 2023-11-22
Move concurrency and action retries from project config proto to the CLI

Ekrekr:
As they are not valid options for `workflow_settings.yaml`

### DIFF
--- a/tests/api/api.spec.ts
+++ b/tests/api/api.spec.ts
@@ -1044,7 +1044,6 @@ suite("@dataform/api", () => {
           rows: [],
           metadata: {
             bigquery: {
-              actionRetryLimit: 1,
               jobId: "abc",
               totalBytesBilled: Long.fromNumber(0),
               totalBytesProcessed: Long.fromNumber(0)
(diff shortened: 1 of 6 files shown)


canberkkoparal (changes requested):
LGTM, but you also need to update the cli test

tests/api/api.spec.ts
```diff
@@ -1036,17 +1036,15 @@ suite("@dataform/api", () => {
...
-        };
+        const NEW_TEST_GRAPH = RUN_TEST_GRAPH;
         when(mockedDbAdapter.createSchema(anyString(), anyString())).thenResolve(null);
         when(
           mockedDbAdapter.execute(NEW_TEST_GRAPH.actions[0].tasks[0].statement, anything())
         ).thenResolve({
           rows: [],
           metadata: {
             bigquery: {
+              actionRetryLimit: 1,
```
Ekrekr, line 1047: This doesn't do what I thought it did, will think about the fix

Ekrekr: Fixed, the option just needed to be passed into the runner instead.

canberkkoparal (approved):

PR #1587, opened 2023-11-22
Add support for publishing pre-release/next packages (and stage 2.8.0-beta.1)

lewish:
If we publish an alpha, beta package we don't want to mark our docker image or our npm packages as `latest` as they will then get installed by accident.

Both get marked with the `next` tag so you only get them if you explicitly ask for them.

### DIFF
--- a/scripts/publish
+++ b/scripts/publish
@@ -15,14 +15,24 @@ fi
 
 bazel test //... --build_tests_only -- -tests/integration:snowflake.spec
 
-bazel run packages/@dataform/cli:package.publish
-bazel run packages/@dataform/core:package.publish
+VERSION=$(cat version.bzl | grep DF_VERSION | awk '{ print $3 }' | sed "s/\"//g")
+
+TAG=next
+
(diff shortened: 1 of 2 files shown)


scripts/publish
```diff
@@ -15,14 +15,24 @@ fi
...
 bazel run cli:push
 
-VERSION=$(cat version.bzl | grep DF_VERSION | awk '{ print $3 }' | sed "s/\"//g")
 git tag $VERSION
 git push origin $VERSION
 
 docker pull dataformco/dataform:$VERSION
-docker tag dataformco/dataform:$VERSION dataformco/dataform:latest
+docker tag dataformco/dataform:$VERSION dataformco/dataform:$TAG
 docker push dataformco/dataform:latest
```
BenBirt, line 38: presumably we do not want to do this if `TAG=next`?

BenBirt (approved):

scripts/publish
lewish: Ah yup, have replaced with $TAG

PR #1592, opened 2023-11-27
Update version of snowflake-sdk to 1.9.1

claydiffrient:
This is to get an updated dependency on axios so that dependabot security alerts stop happening.


BenBirt (approved):

BenBirt:
Thanks @claydiffrient!

PR #1595, opened 2023-11-29
Add basic actions.yaml and notebook support

Ekrekr:


core/compilers.ts
```diff
@@ -18,6 +18,12 @@
       throw e;
     }
   }
+  if (path.endsWith(".ipynb")) {
+    const notebookAsJson = JSON.parse(code);
+    // TODO(ekrekr): strip notebook cell outputs.
+    // TODO(ekrekr): base64 encode the notebook as a string instead.
+    return `exports.asBase64String = () => \`${JSON.stringify(notebookAsJson)})\``;
```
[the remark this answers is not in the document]
Ekrekr: shock and horror

core/actions/notebook.ts
```diff
@@ -0,0 +1,35 @@
+import { verifyObjectMatchesProto } from "df/common/protos";
+import { IActionBuilder } from "df/core/actions";
+import { Session } from "df/core/session";
+import { dataform } from "df/protos/ts";
+
+/**
+ * @hidden
```
GJMcGowan, line 7: This? https://typedoc.org/tags/hidden

So your intention is to hide the class from docs until you're ready, is that right?

core/main.ts
```diff
@@ -102,3 +104,53 @@ export function main(coreExecutionRequest: Uint8Array | string): Uint8Array | st
 
   return dataform.CoreExecutionResponse.encode(coreExecutionResponse).finish();
 }
+
+function loadActionConfigs(session: Session, filePaths: string[]) {
+  filePaths
+    .filter(path => path.startsWith(`definitions${utils.pathSeperator}`))
+    .filter(path => path.endsWith("actions.yaml"))
```
GJMcGowan, line 111: Why two filters instead of `path.startsWith && path.endsWith`?

GJMcGowan (approved):
yuge

core/actions/notebook.ts
Ekrekr: We don't expose these action classes anyways, because users don't access them directly https://github.com/dataform-co/dataform/blob/a4a0a6086a7baa0c6a694dada3895608766a8c9e/core/table.ts#L381

core/main.ts
Ekrekr: Done.

core/main.ts
```diff
@@ -102,3 +104,54 @@ export function main(coreExecutionRequest: Uint8Array | string): Uint8Array | st
...
+      actionConfigs.actions.forEach(actionConfig => {
+        const { fileExtension, fileNameAsTargetName } = extractActionDetailsFromFileName(
+          actionConfig.fileName
+        );
+        if (!actionConfig.target?.name) {
+          if (!actionConfig.target) {
+            actionConfig.target = {};
+          }
+          actionConfig.target.name = fileNameAsTargetName;
+        }
```
BenBirt, line 135: IMO this would be clearer as:

```
if (!actionConfig.target) {
  actionConfig.target = {};
}
if (!actionConfig.target.name) {
  actionConfig.target.name = fileNameAsTargetName;
}
```

core/main.ts
```diff
@@ -102,3 +104,54 @@ export function main(coreExecutionRequest: Uint8Array | string): Uint8Array | st
...
+          // TODO(ekrekr): throw an error if any non-notebook configs are given.
+          // TODO(ekrekr): add test for nice errors if file not found.
+          const notebookContents = nativeRequire(actionConfig.fileName).asBase64String();
+          session.notebookAction(dataform.ActionConfig.create(actionConfig), notebookContents);
+        }
+      });
+    });
+}
+
+function extractActionDetailsFromFileName(
```
BenBirt, line 147: this feels very related to a lot of stuff in `utils.ts`, and should probably be there

core/main.ts
```diff
@@ -102,3 +104,54 @@ export function main(coreExecutionRequest: Uint8Array | string): Uint8Array | st
...
+  path: string
+): { fileExtension: string; fileNameAsTargetName: string } {
+  // TODO(ekrekr): make filename in actions.yaml files not need the `definitions` prefix. In
+  // addition, actions.yaml in nested directories should prefix file imports with their path.
+  const fileName = path.split("/").slice(-1)[0];
```
BenBirt, line 152: bug: should be using separator

core/session.ts
```diff
@@ -193,6 +195,14 @@ export class Session {
     }
   }
 
+  public notebookAction(notebookConfig: dataform.ActionConfig, notebookContents: string): Notebook {
```
BenBirt, line 198: this is part of our public API, and should follow that consistently, i.e. `notebook()`

core/utils.ts
```diff
@@ -189,6 +186,10 @@ export function setNameAndTarget(
   overrideDatabase?: string
 ) {
   action.target = target(session.config, name, overrideSchema, overrideDatabase);
+  if (action.config) {
+    action.config.target = target(session.canonicalConfig, name, overrideSchema, overrideDatabase);
+    return;
+  }
```
BenBirt, line 192: i'm confused. why do we need this / what's it for / should it really be here, as opposed to the callsite?

PR #1596, opened 2023-12-01
Stop using ProtobufJS's verify method (it doesn't do much)

Ekrekr:


common/protos/index.ts
```diff
@@ -15,21 +15,31 @@ export interface IProtoClass<IProto, Proto> {
 }
 
 // ProtobufJS's native verify method does not check that only defined fields are present.
-// TODO(ekrekr): swap to Typescript protobuf library rather than having to do this.
+// TODO(ekrekr): swap to Typescript protobuf library rather than having to do this; however using TS
+// libraries currently available would incur a significant performance hit b/305915462#comment18.
```
a2wd, line 19: Is there an external bug you can point to or replicate or else can you add this as a github issue?

Ekrekr: I will just remove the specific reference.

common/protos/index.ts
```diff
@@ -39,6 +49,7 @@ export function verifyObjectMatchesProto<Proto>(
   }
```
a2wd, line 49: Little hard to see in review, is the indentation right here?

a2wd (approved):

PR #1605, opened 2023-12-07
Add licenses as prefixes to output bundles

Ekrekr:


packages/@dataform/core/BUILD
```diff
@@ -51,21 +51,27 @@ pkg_json(
         "//:package.json",
         "//packages/@dataform:package.layer.json",
     ],
-    main = "bundle.js",
+    main = "bundle.js.no_license",
```
lewish, line 54: I don't see why you need this change?

packages/@dataform/cli/BUILD
```diff
@@ -85,11 +85,23 @@ pkg_bundle(
     ],
 )
 
+add_license_header_to_file(
+    name="bundle_with_license",
+    from_file=":bundle_no_license",
+    to_file=":bundle.js",
```
lewish, line 91: these aren't labels so you don't want : in there, that's probably why tests are failing

packages/@dataform/cli/BUILD
```diff
@@ -85,11 +85,23 @@ pkg_bundle(
...
+)
+
+add_license_header_to_file(
+    name="worker_bundle_with_license",
+    from_file=":worker_bundle_no_license",
+    to_file=":worker_bundle.js",
```
lewish, line 97: ditto

packages/@dataform/core/BUILD
```diff
@@ -51,21 +51,27 @@ pkg_json(
...
 )
 
-copy_file(
-    name = "core_proto",
-    src = "//protos:core.proto",
-    out = "core.proto",
+add_license_header_to_file(
+    name="core_proto_with_license",
+    from_file="//protos:core.proto",
+    to_file=":core.proto",
```
lewish, line 61: ditto

packages/@dataform/core/BUILD
```diff
@@ -51,21 +51,27 @@ pkg_json(
...
+)
+
+add_license_header_to_file(
+    name="bundle_with_license",
+    from_file=":bundle",
+    to_file=":bundle.js",
```
lewish, line 67: ditto

packages/@dataform/core/BUILD
Ekrekr: Done.

packages/@dataform/cli/BUILD
Ekrekr: Tests were also failing because the Node shebang for the CLI was getting in the way; have moved license to below it.

lewish (approved):

PR #1606, opened 2023-12-07
Action builders and path utils fixes

Ekrekr:
Including comments from https://github.com/dataform-co/dataform/pull/1595.

### DIFF
--- a/core/utils.ts
+++ b/core/utils.ts
@@ -177,8 +177,8 @@ export function target(
 ): dataform.ITarget {
   return dataform.Target.create({
     name,
-    schema: schema || config.defaultSchema,
-    database: database || config.defaultDatabase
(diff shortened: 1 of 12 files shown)


core/utils.ts
```diff
@@ -6,34 +6,40 @@ import { Resolvable } from "df/core/common";
 import { IActionProto, Session } from "df/core/session";
 import { dataform } from "df/protos/ts";
 
-export const pathSeperator = (() => {
-  if (typeof process !== "undefined") {
-    return process.platform === "win32" ? "\\" : "/";
-  }
-  return "/";
-})();
+export class Path {
```
BenBirt, line 9: using a `class` as a namespace is really Java-y.

just use normal TypeScript style, `import * as path from "...path.ts"`

BenBirt (approved):

PR #1610, opened 2023-12-11
Add version to workflow settings

Ekrekr:


core/workflow_settings.ts
```diff
@@ -54,6 +55,18 @@ function verifyWorkflowSettingsAsJson(workflowSettingsAsJson: object): dataform.
     // tslint:disable-next-line: no-string-literal
     projectConfig.warehouse = "bigquery";
   }
+
+  if (!projectConfig.version) {
```
lewish, line 59: I'm not really following this.

If the project is manually installing @dataform/core, I don't think we should set this field with a default. The version being undefined in the projectConfig is actually useful to know!

The validation check - also not sure why we need this. This would only happen in the case that the version specified in workflow_settings was not automatically installed by the CLI or Dataform on GCP. Neither should be possible!

Ekrekr: Can you elaborate on what the version being undefined in the projectConfig useful for? I think all cases are covered by the version not being exported by core.

For the validation check, if neither should be possible for either the CLI or Dataform on GCP, then why not check this here!

The Dataform Core interface currently allows all of the `package.json`, the `workflow_settings.yaml` file, and the currently installed Dataform Core version under `node_modules` to be different. By enforcing an opinion on where the version should actually come from, it makes the UX waaay simpler, and prevents every Dataform consumer from having its own implementation.

An example UX:
* Dataform user has an Airflow configuration, using the new `workflow_settings.yaml`.
* They check in a package.json, because they want core to be installed via `npm install`.
* Someone bumps the `workflow_settings.yaml`, but it has no effect.
* With this CL, it will be clear that just bumping the version in `workflow_settings.yaml` does nothing.

protos/core.proto
```diff
@@ -23,6 +23,8 @@ message ProjectConfig {
   string schema_suffix = 7;
   string table_prefix = 11;
 
+  string version = 17;
```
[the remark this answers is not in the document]
Ekrekr: Done, good point

core/workflow_settings.ts
Ekrekr: Also if someone has a `package.json` with multiple other packages specified, then this works well with `workflow_settings.yaml` automatically interpreting the Dataform Version installed.

lewish: ProjectConfig is a proto representation of workflow_settings.yaml or dataform.json. But you're using it to track useful info about the compilation, but we already track the version of dataform/core used for compilation here: https://github.com/dataform-co/dataform/blob/main/protos/core.proto#L296

So I think in general we should not be writing to `ProjectConfig` except to define static default values, which is all we to it before this CL as far as I'm aware.

In terms of knowing if it's undefined or not - this is kind of useful as it tells you where the version came from - was it actually defined in workflow_settings and therefore installed automatically, or, did we use a user installed version of dataform/core that already existed inside node_modules (probably from package.json)

I'm OK with the equality check 👍

protos/core.proto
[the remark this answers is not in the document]
lewish: IMO calling it `version` makes sense if that's what we are calling the field in the YAML representation, and as we already have a `dataform_core_version` field that is specified in `CompiledGraph`, given the ProjectConfig is a proto representation of that file.

core/workflow_settings.ts
lewish: > Someone bumps the workflow_settings.yaml, but it has no effect.

This should never be the case, as we are always going to give `version` in workflow settings precedence right? Or if there is a conflict and both are specified, we will throw an error. Either way it will not be ignored.

protos/core.proto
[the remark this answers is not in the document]
BenBirt: I don't really follow the point about `CompiledGraph`. I don't understand why we can't have two different fields, in two different contexts, that mean different things.

`version` is still confusing. Again, if I see `version` in a workflows YAML file, I am going to assume that this somehow refers to the version _of that workflow_, not the version of the code that interprets it.

core/workflow_settings.ts
BenBirt: I agree with Lewis.

We're never going to _ignore_ the case where the version is defined both in workflow settings and package.json (I would personally argue for throwing an error, as this seems like a very easy path to lead to user confusion otherwise!).

We shouldn't be treating `ProjectConfig` as an _output_.

Ekrekr: Ok I think I can buy not setting it with the current version if unset. The exact UX for this will become more clear with the internal changes and tests for them.

It sounds like there's agreement now that this is a good place for the validation check? `projectConfig.dataformCoreVersion !== version`

protos/core.proto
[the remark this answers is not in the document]
Ekrekr: My original thinking was to deprecate specifying the version in https://github.com/dataform-co/dataform/blob/main/protos/core.proto#L296, but because we're not writing any output to ProjectConfig I don't think there's harm in having the Dataform Core version duplicated in some scenarios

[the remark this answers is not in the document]
BenBirt: So again - this is _not_ duplication. One is a user request, the other is what actually happened. The two may not be the same!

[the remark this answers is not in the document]
Ekrekr: I dislike the idea that could happen, but it makes sense

lewish (approved):

BenBirt (approved):

PR #1628, opened 2023-12-29
Add basic support for actions.yaml reading SQL files as operations

Ekrekr:
The Javascript context in the SQL file is preserved.

The javascript API for adding operations to the session is extended to accept the full action config proto as an option, instead of just the name.

### DIFF
--- a/core/actions/operation.ts
+++ b/core/actions/operation.ts
@@ -1,5 +1,5 @@
 import { verifyObjectMatchesProto } from "df/common/protos";
-import { IActionBuilder } from "df/core/actions";
+import { ActionBuilder } from "df/core/actions";
 import { ColumnDescriptors } from "df/core/column_descriptors";
 import {
   Contextable,
@@ -61,7 +61,7 @@ export const IIOperationConfigProperties = strictKeysOf<IOperationConfig>()([
 /**
  * @hidden
  */
-export class Operation implements IActionBuilder<dataform.Operation> {
+export class Operation extends ActionBuilder<dataform.Operation> {
   // TODO(ekrekr): make this field private, to enforce proto update logic to happen in this class.
   public proto: dataform.IOperation = dataform.Operation.create();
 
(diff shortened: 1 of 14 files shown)


core/actions/operation.ts
```diff
@@ -71,8 +71,26 @@ export class Operation extends ActionBuilder<dataform.Operation> {
   // We delay contextification until the final compile step, so hold these here for now.
   private contextableQueries: Contextable<ICommonContext, string | string[]>;
 
-  constructor(session?: Session) {
+  constructor(session?: Session, config?: dataform.ActionConfig) {
     super(session);
+    if (!session || !config) {
```
tatianaschmidt, line 76: if one is set and the other is not, the set one will just be ignored? that seems misleading

core/main.ts
```diff
@@ -135,12 +136,28 @@ function loadActionConfigs(session: Session, filePaths: string[]) {
...
+        // TODO(ekrekr): add a test for nice errors if files are not found.
+
         if (fileExtension === "ipynb") {
-          // TODO(ekrekr): throw an error if any non-notebook configs are given.
-          // TODO(ekrekr): add test for nice errors if file not found.
           const notebookContents = nativeRequire(actionConfig.fileName).asBase64String();
           session.notebook(dataform.ActionConfig.create(actionConfig), notebookContents);
         }
+
+        if (fileExtension === "sql") {
```
tatianaschmidt, line 147: just checking, in the case of action config we only expect sql nor sqlx?

core/main_test.ts
```diff
@@ -617,7 +617,65 @@ actions:
...
-              name: "note"
+              name: "notebook"
+            }
+          }
+        ])
+      );
+    });
+  });
+
+  suite("SQL actions", () => {
```
tatianaschmidt, line 628: another test for a non-operation failing

tatianaschmidt (changes requested):

core/main_test.ts
Ekrekr: My next PRs will add support for the other action types so I'm not bothering

core/main.ts
Ekrekr: Yep! SQLX is still handled here https://github.com/dataform-co/dataform/blob/cab21959a1857f964b113e94d9a2d3b8fbc51c0a/core/main.ts#L83

core/actions/operation.ts
Ekrekr: If either are unset, then the constructor returns early, because all of the methods in the constructor will throw otherwise.

Longer term the plan is to make `config` private, and require session and config to be defined - but for backwards compatibility with the existing Javascript API session methods (e.g. https://github.com/dataform-co/dataform/blob/cab21959a1857f964b113e94d9a2d3b8fbc51c0a/core/session.ts#L280) I've allowed them to be optional.

tatianaschmidt (approved):

PR #1634, opened 2024-01-03
Cleanup: remove action hermeticity

Ekrekr:
Although this option is currently documented https://cloud.google.com/dataform/docs/reference/dataform-core-reference#itableconfig, it doesn't have an effect anywhere.

IIRC from past discussions of lessons learned, we would do this differently if done today.

### DIFF
--- a/cli/api/commands/build.ts
+++ b/cli/api/commands/build.ts
@@ -84,8 +84,7 @@ export class Builder {
       tableType: utils.tableTypeEnumToString(table.enumType),
       tasks: table.disabled
         ? []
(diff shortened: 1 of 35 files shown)


GJMcGowan (approved):

BenBirt:
Can we discuss this change, please?

Action hermeticity _does_ have an effect - in terms of compilation errors. The intention was to prevent users from adding implicit dependencies (i.e. which are not understood by the compiler, and thus do not make their way into the DAG).

The reason for _that_ was (a) correctness, and (b) to enable something like run caching (or - as an extension - run-on-dependency-change).

I'm not sure removing this is a good idea.

PR #1636, opened 2024-01-04
Make filenames defined in action config files be treated as relative to the action config file

Ekrekr:


core/main.ts
```diff
@@ -147,29 +137,54 @@ function loadActionConfigs(session: Session, filePaths: string[]) {
           actionConfig.target.name = fileNameAsTargetName;
         }
 
+        // Users define file paths relative to action config path, but internally and in the
+        // compiled graph they are absolute paths.
+        actionConfig.fileName =
+          actionConfigsPath.substring(0, actionConfigsPath.length - "actions.yaml".length) +
```
kolina, line 143: NIT: put `actions.yaml` in some constant.

kolina (approved):

core/main.ts
Ekrekr: Fair shout, I'll do this in a later follow up PR with all names

PR #1637, opened 2024-01-04
Better errors for invalid declarations, add more test coverage

Ekrekr:


core/compilers.ts
```diff
@@ -23,7 +23,7 @@ export function compile(code: string, path: string): string {
   if (path.endsWith(".yaml")) {
     try {
       const yamlAsJson = JSON.stringify(loadYaml(code));
-      return `exports.asJson = () => (${yamlAsJson})`;
+      return `exports.asJson = ${yamlAsJson}`;
```
BenBirt, line 26: why not just `exports = ${yamlAsJson}`?

core/compilers.ts
```diff
@@ -33,8 +33,7 @@ export function compile(code: string, path: string): string {
   }
   if (path.endsWith(".ipynb")) {
     const notebookAsJson = stripNotebookOutputs(JSON.parse(code), path);
-    // TODO(ekrekr): base64 encode the notebook as a string instead.
-    return `exports.asBase64String = () => \`${JSON.stringify(notebookAsJson)}\``;
+    return `exports.asBase64String = \`${JSON.stringify(notebookAsJson)}\``;
```
BenBirt, line 36: this is not a base64 string

core/compilers.ts
Ekrekr: When setting `exports = ...` directly, the return value of `require()` always ends up being `{}`.

The wrapper looks like:

```
(function (exports, require, module, __filename, __dirname) { "use strict"; <return code here> })
```

core/compilers.ts
Ekrekr: Done.

core/compilers.ts
```diff
@@ -33,8 +33,7 @@ export function compile(code: string, path: string): string {
   }
   if (path.endsWith(".ipynb")) {
     const notebookAsJson = stripNotebookOutputs(JSON.parse(code), path);
```
BenBirt, line 35: I realise this is unrelated to this PR but: I think this is the wrong place for this to be happening! IMO output stripping should be happening in the `compile()` method of a NotebookAction class - following the same pattern for resolving/compiling SQLX/JS code into SQL - not here.

core/compilers.ts
```diff
@@ -33,8 +33,7 @@ export function compile(code: string, path: string): string {
...
-    // TODO(ekrekr): base64 encode the notebook as a string instead.
-    return `exports.asBase64String = () => \`${JSON.stringify(notebookAsJson)}\``;
+    return `exports.asString = \`${JSON.stringify(notebookAsJson)}\``;
```
BenBirt, line 36: IMO this should be returned as JSON, I think that would be a lot cleaner. But I won't block.

BenBirt (approved):

core/compilers.ts
Ekrekr: Done.

core/compilers.ts
Ekrekr: Fair point, it's not really a string! Done.

core/compilers.ts
BenBirt: It occurred to me: the right way to do this would be:

module.exports = fileAsJson

PR #1627, opened 2023-12-28
VSCode: avoid errors upon completion requests

barna-emarsys:
While authoring a `.sqlx` file, the VS Code Dataform extension keeps showing error messages about failed completion requests. They read:
```[Error - 18:53:55] Request textDocument/completion failed.
  Message: Unhandled method textDocument/completion
  Code: -32601
```

Here's a demonstration:
dataform_compile_crash_demostration

As far as I see, the issue is caused by the Dataform LSP declaring itself as a completion provider, but never providing a method that provides completions. I could not find any part of the source code that handles completions (the implemented hooks are `onInitialized`, `onRequest("compile")`, `onDidSave` and `onDefinition`). It also seems to me the declarative parts of a language extension can not be responsible for providing completions.

The extension lists as its features: syntax highlighting, compilation and go to definition. All those features continue to work without the LSP declaring itself as a completion provider.

This PR resolves this issue by removing the completion provider declaration.

PS: I have two more PR's ready with suggested features, but I didn't want to overload you with 3 separate PR's at once. 
They would:
- add a configuration setting for specifying the location of the dataform folder within the project
- add a configuration setting to disable compile on save

Are you open to reviewing those as well?

Issue #1226, opened 2021-05-25
Configure Dataform directory in Visual Studio Code Extension 


jansphil:
Hi all,

Thanks for providing the Visual Studio Code Extension. However when the Dataform directory is in a subfolder inside the project, the extension does not work, since it expects the `dataform.json` file to be in the root.

There could be a configuration property to be able to define the Dataform subdirectory.

Thanks!

mokshasoul:
bumping this issue

GJMcGowan:
This issue is now resolved, please update to version 0.0.13 and apply the new config https://github.com/dataform-co/dataform/pull/1639

psvaiter:
Since it's resolved, what is the configuration property to define the Dataform subdirectory? I couldn't find it in the PR.

p13rr0m:
There is the compiler options in VS Code that sets the project-dir argument:

![image](https://github.com/user-attachments/assets/ff2abd51-4082-4846-940a-9ca3a1d4121f)

![image](https://github.com/user-attachments/assets/8d34852c-a235-4a68-af04-bfbfcfc90e63)

PR #1639, opened 2024-01-07
VSCode: Add ability to configure compile command, add option to skip compilation on save

bmagyarkuti:
As I've discussed with @GJMcGowan in https://github.com/dataform-co/dataform/pull/1627, this PR introduces two new configration options to the VS Code dataform extension.

1. `dataform.compileOnSave` allows the user to disable the current automatic behavior of re-compiling the entire project each time an `sqlx` file is saved. I'd like to introduce this option because the notification that pops up each time compilation has finished is rather annoying. Rather than introduce the option to just silence the notification, I propose introducing the ability to disable auto-compile, as project compilation is indeed an expensive and taxing process that shouldn't just happen silently. Another reason users might want to disable the auto-compile option is to save system resources, for example when they're working on battery or on a slower computer. The default setting is the same as the current behavior (auto-compile is enabled).

2. `dataform.compilerOptions` lets the user inject any compiler options into the shell command that the extension uses to compile the project (either automatically or through the contributed `dataform.compile` vs code command). The shell command is `dataform compile --json`, and the injected settings are appended after the end of this command. The default setting (no additional commands) does not change the current behavior. On my team, we would use this command to set a custom directory (this would be a `dist` directory that contains generated `js` files. We need to transpile because our project is in Typescript). I think the introduction of this setting resolves the following open issue: https://github.com/dataform-co/dataform/issues/1226.

Attached I'm also including the [compiled vsix](https://github.com/dataform-co/dataform/files/13854379/dataform-package.vsix.zip).

(I'm making this PR from a new account. This also resolves all the issues we've had with the Dataform extension on our team. I'm not planning to submit more PR's to the extension.)

### DIFF
--- a/vscode/package-lock.json
+++ b/vscode/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "dataform",
-  "version": "0.0.11",
+  "version": "0.0.13",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "dataform",
-      "version": "0.0.11",
+      "version": "0.0.13",
       "dependencies": {
         "vscode-languageclient": "^6.1.3",
         "vscode-languageserver": "^6.1.1",
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -28,13 +28,18 @@
       {
         "title": "Dataform",
         "properties": {
+          "dataform.compileOnSave": {
+            "type": "boolean",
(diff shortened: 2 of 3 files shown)


GJMcGowan (approved):
Looks great, thank you for your contribution!

I'm going to go ahead and merge, then I'll update store version 👍

Issue #1646, opened 2024-01-14
Typescript Typings for @dataform/core


bmagyarkuti:
Before https://github.com/dataform-co/dataform/pull/1552 removed Typescript typings from the `@dataform/core` package, TS projects using Dataform got type declarations for `@dataform/core`'s API out of the box. Since version 2.8.0 shipped this PR, the API of `@dataform/core` is untyped. My team uses Dataform in a TS project. For now, we've manually included the `2.7` types in our project, but this requires us to manually maintain those types going forward. It would be very helpful if Dataform kept providing types for the `dataform/core` package, like it has since https://github.com/dataform-co/dataform/commit/f6e3ce60a5d628cf6ecb91cd828638d10bf10de5. 

There are two ways for a package to provide types:
1) By bundling the declarations file with the package.
2) By separately publishing the types.

Typescript [recommends](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) the first approach, but the second approach would also solve our problem and I'd greatly appreciate either solution.

I'll also gladly help implement either of these solutions.

Also, if there are no plans to provide types for this package, even just a negative answer helps us greatly. That way, we can decide to either invest in building out TS support for our Dataform packages  internally (by setting up a pipeline that compiles the .d.ts files from the open-source repo), or move our Dataform projects to vanilla JS.

PR #1647, opened 2024-01-14
Include type declarations with @dataform/core

bmagyarkuti:
A suggested, trivial resolution for https://github.com/dataform-co/dataform/issues/1646.

I understand the typings were removed in https://github.com/dataform-co/dataform/pull/1552 to minimize the installed footprint of the `@dataform/core` package.

While re-adding the declarations file increases the bundle size, the size reductions introduced by #1552 would remain very similar. Before the introduction of minification, the total `node_module` size was 9.4MB. After the minifaction PR, it was ~432KB, and with the `d.ts` file re-added it would increase back to ~675KB. Instead of the original ~20x reduction, we just get a ~16x reduction. This increase can not impact compile performance, as the types file is not used during run time, so that performance gain remains.

This is just a suggestion. As I mentioned in the issue, my goal is to get type declarations for the `@dataform/core` package.

I'm very happy to assist with any solution that achieves this, including publishing type declarations in a separate package, possibly to the Definitely Typed repository.

I'd like to nominate @lewish as reviewer, since he introduced #1552 .

### DIFF
--- a/packages/@dataform/core/BUILD
+++ b/packages/@dataform/core/BUILD
@@ -61,10 +61,26 @@ copy_file(
     out = "core.proto",
 )
 
+pkg_bundle_dts(
+    name = "bundle.d",
+    entry_point = "index.ts",
+    externals = [
+      "protobufjs",
+      "tarjan-graph",
+      "semver",
+      "moo",
+      "long",
+    ],
+    deps = [
+        ":core",
+    ],
+)
+
 pkg_npm_tar(
     name = "package",
     deps = [
         ":bundle",
+        ":bundle.d",
         ":package.json",
         ":core.proto",
     ],


BenBirt:
Just some added context: it turns out that https://github.com/dataform-co/dataform/pull/1552 did not actually speed up compilation time (that was a red herring). But we originally did it to speed up _package installation_ time (and that certainly did work).

My 2c: clearly some users need typings, so we have two options: bring them back, or export a separate package including types. I think I'd rather bring back typings into the same package - it's simple, easy, and is pretty low-cost. We can always change our minds later!

lewish:
I'm OK with this! The files will rarely be loaded for most users, and it's not that big an increase.

It is slightly weird that the types are generated via rollup and the code via webpack, but I don't obviously see why that would cause any issues.

lewish (approved):

bmagyarkuti:
Thanks for the quick response! I'm so glad we get to continue using Dataform with TS (❁´◡`❁).

PR #1649, opened 2024-01-15
Update Dataform CLI npm installs to best-effort

Ekrekr:
General UX:

|                      | **dataformCoreVersion present**                          | **no dataformCoreVersion present**                  |
|----------------------|------------------------------------------------------|-------------------------------------------------|
| **package.json present** | ✅ Install from dataformCoreVersion in workflow_settings.yaml | ✅ Install from @dataform/core version in package.json                |
| **no package.json**      | ✅ Install from dataformCoreVersion in workflow_settings.yaml                     | ❌ Throw error about no version definition present |

### DIFF
--- a/cli/api/commands/install.ts
+++ b/cli/api/commands/install.ts
@@ -22,17 +22,6 @@ export async function install(projectPath: string, skipInstall?: boolean) {
   let dataformCoreVersion = readDataformCoreVersionIfPresent(workflowSettingsPath);
 
   if (dataformCoreVersion) {
-    if (fs.existsSync(packageJsonPath)) {
-      throw new Error(
-        "dataformCoreVersion cannot be defined in workflow_settings.yaml when a package.json is present"
-      );
-    }
-    if (fs.existsSync(packageLockJsonPath)) {
-      throw new Error(
-        "dataformCoreVersion cannot be defined in workflow_settings.yaml when a package-lock.json is present"
-      );
-    }
-
     // If there are other packages already in the package.json, specifying a specific package to
     // install will trigger the other packages to be installed too.
     installCommand += ` @dataform/core@${dataformCoreVersion}`;
--- a/cli/api/commands/install.ts
+++ b/cli/api/commands/install.ts
@@ -1,11 +1,62 @@
+import * as fs from "fs";
+import { load as loadYaml, YAMLException } from "js-yaml";
 import * as path from "path";
 import { promisify } from "util";
 
 import * as childProcess from "child_process";
+import { dataform } from "df/protos/ts";
 
-export async function install(projectDir: string, skipInstall?: boolean) {
+export async function install(projectPath: string, skipInstall?: boolean) {
   if (skipInstall) {
     return;
   }
-  await promisify(childProcess.exec)("npm i --ignore-scripts", { cwd: path.resolve(projectDir) });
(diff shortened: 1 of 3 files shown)


BenBirt:
I think the top-left of the table has incorrect semantics. I think that if both files exist, `workflow_settings.yaml` should _not_ contain a version. Otherwise things get way too confusing.

Ekrekr:
> I think the top-left of the table has incorrect semantics. I think that if both files exist, `workflow_settings.yaml` should _not_ contain a version. Otherwise things get way too confusing.

This issue with that in the CLI is that compilation isn't stateless, so we don't delete `package.json` after an install. So the first time installing works if there's dataformCoreVersion in workflow settings, but then subsequent installs will fail. From playing around with it, doing this seems to work just fine!

A workaround could be making it stateless, and updating the package manager to `pnpm` instead.

Let's decide on this in the UX review though, to avoid doing work that might not be needed.

BenBirt:
OK, but we did already discuss this point!

cli/api/commands/install.ts
```diff
@@ -1,11 +1,55 @@
...
+
+  let installCommand = "npm i --ignore-scripts";
+
+  // Core's readWorkflowSettings method cannot be used for this because Core assumes that
+  // `require` can read YAML files directly.
+  const dataformCoreVersion = readDataformCoreVersionIfPresent(workflowSettingsPath);
+
+  if (dataformCoreVersion) {
+    // If there are other packages already in the package.json, specifying a specific package to
+    // install will trigger the other packages to be installed too.
```
BenBirt, line 26: are you sure about this? it seems very surprising to me, and it does not seem to be in the docs: https://docs.npmjs.com/cli/v10/commands/npm-install

Ekrekr:
> OK, but we did already discuss this point!

I was surprised too, but it happened when I tried it!

cli/api/commands/install.ts
Ekrekr: I was surprised too, but it happened when I tried it!

Ekrekr:
> OK, but we did already discuss this point!

> In closed source, the Dataform Core version can be specified in either the `package.json` or the `workflow_settings.yaml` file.

It's true - but is a pretty terrible UX when using it in practice.

In any case, the code fix if we do decide to do this is trivial.

BenBirt:
OK - I'll say:
1) The UX nastiness is pretty resolvable. For one, it's a once-off thing (when migrating from workflow_settings.yaml to package.json). But if even that is too much, we can always make the CLI help you out with that migration.
2) Even if we do want to allow the version to be configured in two places, I would *strongly* argue that the preferred location should be package.json, not least because that's what all other NPM tooling would actually look at.

BenBirt (approved):

Ekrekr:
> I would strongly argue that the preferred location should be package.json

I agree with this, extending it even to be the *only* location that open source packages configure versions - it's a bit weird having two ways of doing things.

> once-off thing

I'm not sure that's true, because it's fairly common to run the Dataform CLI as part of CI/CD in a GitHub action from PRs - which gets run on the version of code on Google Cloud.

I'll have a think about the different user journeys, but atm this is the most versatile option!

BenBirt:
But GCP should follow the same logic: i.e. if both files provide versions, it should throw. Therefore: in all environments, either you are setting it in one place or the other, but never both.

PR #1650, opened 2024-01-15
Make examples work without modification, by adding dataformCoreVersion

Ekrekr:
Also:
* Add a test to make sure it doesn't get out of sync.
* Move version to the common directory, as it is used in several packages already.


common/BUILD
```diff
@@ -0,0 +1,21 @@
...
+    },
+    template = "version.ts.tmpl",
+)
+
+ts_library(
+    name = "current_version",
+    srcs = [
+        ":version.ts",
+    ]
+)
```
BenBirt, line 21: nit: either output `current_version.ts` from the above rule, or simply rename this to `version` and change the rule above to `version_tmpl_expansion`

examples/examples_test.ts
```diff
@@ -1,10 +1,30 @@
...
 
 import { compile } from "df/cli/api";
+import { version } from "df/common/version";
+import { dataform } from "df/protos/ts";
 import { suite, test } from "df/testing";
 
 suite("examples", () => {
+  test("ensure matching dataformCoreVersion", async () => {
+    // To make examples code work without modification, dataformCoreVersion needs to be given. This
+    // test ensures that the version doesn't get out of sync with that of current development.
```
BenBirt, line 14: i don't think i really understand this. if you want the version to stay in sync, then you should generate `workflow_settings.yaml` (with `version` as an input) in bazel.

BenBirt (changes requested):

examples/examples_test.ts
Ekrekr: I really hate it when I'm trying out a new piece of software, I make a copy of an example project, and then it doesn't work.

This way the example projects will work exactly as they are, and is one of the reasons I got rid of the BUILD files from within them too - I think this is worth the cost of bumping the version in a couple additional places.

Ekrekr: Also this means that the example projects will work literally in our e2e tests too

BenBirt: I think this just indicates an overloading between _examples_ and _tests_.

I want tests to be as trivial to manage as possible. They should not fail with an error message saying that I need to go in multiple places and update the version (which is going to happen, a lot).

If we want examples, and you want to manually manage them such that they can be checked out in full - then sure, let's have them! - but they are separate things, and should not block e.g. version upgrades.

Ekrekr: I think there's a significant difference between examples that are tests, and examples that are *tested*!

It's one of the advantages of having the examples in our monorepo - we can be guaranteed that example projects will work.

Having to bump the version here is no different to bumping the version in https://github.com/dataform-co/dataform-example-project-bigquery (which we should be doing) - except that we're actually going to do it.

BenBirt: These are two separate things.

You can test the examples, without forcing version bumps to increase version numbers in various different places. If you want, you can write automation to update the example project dependency version numbers (just as we used to have for `dataform-co`).

I think that forcing us to remember to increase the version in a number of different places is just TOIL. Realistically, it won't survive beyond a couple of months - we'll either remove the requirement that the examples use the most recent version, or more likely just generate it in Bazel.

Ekrekr: ^ going with this

PR #1652, opened 2024-01-17
Remove dead resolve code

Ekrekr:


Ekrekr:
I don't think it makes sense to have the `supportSqlFileCompilation` in v3, it overloads the loading of SQL files from with actionConfigs

core/session.ts
```diff
@@ -244,34 +240,8 @@ export class Session {
...
-    return this.compilationSql().resolveTarget(
-      utils.target(
-        this.config,
-        this.finalizeName(ref.name),
-        this.finalizeSchema(ref.schema),
-        ref.database && this.finalizeName(ref.database)
-      )
-    );
+    this.compileError(new Error(`Could not resolve ${JSON.stringify(ref)}`));
+    return "unresolved";
```
BenBirt, line 244: please continue to return an empty string, as we do above for similar cases.

BenBirt:
> I don't think it makes sense to have the `supportSqlFileCompilation` in v3, it overloads the loading of SQL files from with actionConfigs

I agree - it is part of the reason I wanted to the merge to happen

BenBirt (approved):

PR #1656, opened 2024-01-22
Update proto to use a separate compiled graph, with UX review changes

Ekrekr:


protos/core.proto
```diff
@@ -260,127 +250,422 @@ message CompiledGraph {
...
-    AssertionConfig assertion = 8;
-    OperationConfig operation = 9;
-    DeclarationConfig declaration = 10;
-    NotebookConfig notebook = 11;
+    // Pre/post operations.
+    repeated string pre_ops = 10;
+    repeated string post_ops = 11;
+
+    // Common BigQuery options.
+    BigQueryOptions bigquery = 12;
```
lewish, line 338: tempting to flatten these. Or, I would at least rename them - they are really "table options".

protos/core.proto
```diff
@@ -260,127 +250,422 @@ message CompiledGraph {
...
+    // timestamp or the date column. See
+    // https://cloud.google.com/dataform/docs/partitions-clusters.
+    string partition_by = 1;
+
+    // The keys by which to cluster partitions by. See
+    // https://cloud.google.com/dataform/docs/partitions-clusters.
+    repeated string cluster_by = 2;
+
+    // SQL-based filter for when incremental updates are applied.
+    string update_partition_filter = 3;
```
lewish, line 280: In a BQ only world, this could move to IncrementalTable

protos/core.proto
```diff
@@ -260,127 +250,422 @@ message CompiledGraph {
...
-  repeated Target dependency_targets = 3;
+    // If set to true, this action will not be executed. However, the action may
+    // still be depended upon.
+    bool disabled = 7;
 
-  // Filename to load the action contents from.
-  string file_name = 4;
+    // Describes a struct, object or record in a dataset that has nested
+    // columns.
+    ActionDescriptor action_descriptor = 8;
```
lewish, line 328: This contains description and column descriptions,  I don't think you need them below? or vice versa

protos/core.proto
```diff
@@ -260,127 +250,422 @@ message CompiledGraph {
...
+
+    // Common BigQuery options.
+    BigQueryOptions bigquery = 12;
+
+    // Applies the materialized view optimization, see
+    // https://cloud.google.com/bigquery/docs/materialized-views-intro.
+    bool materialized = 13;
+
+    // Description of the view.
+    string description = 14;
```
lewish, line 392: ditto - it's in ActionDescriptor, don't need it twice

lewish: And elsewhere below

protos/core.proto
```diff
@@ -260,127 +250,422 @@ message CompiledGraph {
...
 
-// Configuration options for BigQuery table declarations.
-// TODO(ekrekr): remove this empty message, if explicit action type without a
-// `type` field proves to be impossible / ugly.
-message DeclarationConfig {}
+message CompileExecutionRequest {
+  CompileConfig compile_config = 1;
+}
+message CompileExecutionResponse {
+  CompiledGraph compiled_graph = 1;
```
lewish, line 658: nit: I'd say the oneof goes here. It's still a "compile execution response" in either case

protos/core.proto
Ekrekr: Flattening to the specific action types makes sense to me, particularly looking at logic like https://github.com/dataform-co/dataform/blob/f14c861488ea04e7862995aa8ee82a9657824867/core/session.ts#L615

BenBirt:
Let me know once Lewis approves and I'll take a look then? Don't want to waste your/my time reviewing stuff that is being changed

protos/core.proto
Ekrekr: Ah yep! That was the whole point of me flattening the fields under the actions

lewish (approved):

protos/core.proto
```diff
@@ -260,24 +267,420 @@ message CompiledGraph {
...
+    string database_suffix = 7;
+    string schema_suffix = 8;
+    string table_prefix = 9;
+  }
+
+  WorkflowSettings workflow_settings = 2;
+
+  GraphErrors graph_errors = 3;
+
+  string dataform_core_version = 4;
```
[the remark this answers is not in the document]
Ekrekr: Yes, but I don't think there should be a scenario where the requested dataform version doesn't equal the resolve dataform version - we'd throw an error in these cases. So IMO this field should be deleted, in favor of the WorkflowSettings field, which can be backfilled even if not specified.

protos/core.proto
```diff
@@ -260,24 +267,420 @@ message CompiledGraph {
...
+  message ColumnDescriptor {
+    // The identifier for the column, using multiple parts for nested records.
+    repeated string path = 1;
+
+    // A text description of the column.
+    string description = 2;
+
+    // A list of tags for this column which will be applied if supported by the
+    // warehouse.
+    repeated string tags = 3;
```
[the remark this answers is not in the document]
Ekrekr: AFAICT no, have removed

protos/core.proto
```diff
@@ -260,24 +267,414 @@ message CompiledGraph {
...
+
+    // Targets (non-canonical) of actions that this action is dependent on.
+    repeated Target dependency_targets = 3;
+
+    // Path to the source file that the contents of the action is loaded from.
+    string source = 4;
+
+    // Path that the config of the action is loaded from.
+    // This option cannot be manually specified in action configs.
+    string config_path = 5;
```
BenBirt, line 299: I'm a bit confused about these two. I *think* `source` would be `foo.sqlx`, `bar.js`, or `baz.sql`, and `config_path` would only be set to `...actions.yaml`. Is that correct?

I'd like to more tightly define this. I'd propose something like:

```
message Source {
  string declaration_file = 1;
  string source_file = 2;
}
```

And I'd propose that *both* fields are *always* set (sometimes - as with SQLX and JS actions - to the same value).

protos/core.proto
```diff
@@ -260,24 +267,414 @@ message CompiledGraph {
...
+    // The SQL query.
+    string query = 8;
+
+    // The query that is applied when the table already exists.
+    // This option cannot be manually specified in action configs.
+    string incremental_query = 9;
+
+    // Pre/post operations.
+    repeated string pre_ops = 10;
+    repeated string post_ops = 11;
```
BenBirt, line 441: nit: AFAIUI this is _also_ being used to define `actions.yaml`'s schema. if so, I think we should expand these to `pre_operations` etc

protos/core.proto
```diff
@@ -260,24 +267,414 @@ message CompiledGraph {
...
+    Target canonical_target = 2;
+
+    // Targets (non-canonical) of actions that this action is dependent on.
+    repeated Target dependency_targets = 3;
+
+    // Path to the source file that the contents of the action is loaded from.
+    string source = 4;
+
+    // Path that the config of the action is loaded from.
+    // This option cannot be manually specified in action configs.
```
BenBirt, line 605: I'm confused by these comments. is it because you intend these to be used in `actions.yaml` too?

if so... i know you won't like this, but I think this is confusing. i know you're trying to use the same protos in as many places as possible... but i think the cure is worse than the disease (of having multiple definitions).

i think it's confusing to have some fields (e.g. `config_path`) that cannot be set, but which exist in the proto. IMO it should be possible to generate help docs for `actions.yaml` schema without having to specify that some fields may not be set. (and be forced to check that those fields are not set in our code, etc)

protos/core.proto
```diff
@@ -371,16 +770,12 @@ message OperationConfig {
...
-// Configuration options for Vertex AI notebooks.
+// TODO(ekrekr): remove this.
 message NotebookConfig {
   // If set to true, this action will not be executed. However, the action may
   // still be depended upon.
   bool disabled = 1;
-
-  // TODO(ekrekr): add a notebook runtime field definition.
-}
+}
```
BenBirt, line 781: trailing newline plz

protos/core.proto
Ekrekr: Yes, I think this makes a lot more sense.

I don't think it's that confusing, it's like putting an `[output_only = true]` option - which we do in the cloud API (there's just no proto field for this supported by default).

Having separate configs means:
* Users will have to look at the behavior of both the config option, and the outputted field - which in most cases will be identical. It's a LOT of information for a new user already, this would make it worse rather than better that the current situation.
* Hundreds more lines of proto definitions, requiring identical comments of the fields.
* Hundreds more lines of copying these fields from the config definitions to the action definitions.

protos/core.proto
Ekrekr: Delaying this decision until a later version, as discussed in chat

protos/core.proto
Ekrekr: Moving to separate config protos, from discussion

protos/core.proto
```diff
@@ -289,99 +279,374 @@ enum SupportedFeatures {
...
+  }
 
-  // Filename to load the action contents from.
-  string file_name = 4;
+  message TableConfig {
+    // Target is the unique action identifier.
+    Target target = 1;
+
+    // Targets of actions that this action is dependent on.
+    repeated Target dependency_targets = 2;
```
Ekrekr, line 338: Should these be strings?

protos/core.proto
```diff
@@ -289,99 +289,375 @@ enum SupportedFeatures {
   ARRAY_BUFFER_IPC = 1;
 }
 
+// Workflow Settings Config defines the contents of the `workflow_settings.yaml`
+// configuration file.
+message WorkflowSettingsConfig {
```
BenBirt, line 294: why not just `WorkflowSettings`?

protos/core.proto
```diff
@@ -289,99 +289,375 @@ enum SupportedFeatures {
...
+  // The desired dataform core version to compile against.
+  string dataform_core_version = 1;
+
+  // Required. The default database (Google Cloud project ID).
+  string default_database = 2;
```
BenBirt, line 299: should we consider changing this to `gcp_project` now?

protos/core.proto
```diff
@@ -289,99 +289,375 @@ enum SupportedFeatures {
...
+
+  // Required. The default schema (BigQuery dataset ID).
+  string default_schema = 3;
```
BenBirt, line 302: ditto, should we consider `default_dataset` (similar comments below)?

protos/core.proto
```diff
@@ -289,99 +289,375 @@ enum SupportedFeatures {
...
+  // Optional. The suffix that should be appended to all database (Google Cloud
+  // project ID) names.
+  string database_suffix = 7;
+
+  // Optional. The suffix that should be appended to all schema (BigQuery
+  // dataset ID) names.
+  string schema_suffix = 8;
+
+  // Optional. The prefix that should be prepended to all table names.
+  string table_prefix = 9;
```
BenBirt, line 325: should this be `object_prefix`? `action_prefix`?

protos/core.proto
```diff
@@ -289,99 +289,375 @@ enum SupportedFeatures {
   ARRAY_BUFFER_IPC = 1;
 }
 
+// Workflow Settings Config defines the contents of the `workflow_settings.yaml`
```
BenBirt, line 292: nit: do you want to put these new messages in `config_files.proto` or something?

protos/core.proto
BenBirt: for me: no

protos/core.proto
```diff
@@ -289,99 +289,375 @@ enum SupportedFeatures {
...
+    // A list of user-defined tags with which the action should be labeled.
+    repeated string tags = 4;
+
+    // If set to true, this action will not be executed. However, the action can
+    // still be depended upon. Useful for temporarily turning off broken
+    // actions.
+    bool disabled = 5;
+
+    // The SQL query.
+    string query = 6;
```
BenBirt, line 366: surely this should be in a `oneof` with `filename`?

Ekrekr: Let's just not support the query field until we know that's something we want to do long-term

protos/configs.proto
```diff
@@ -0,0 +1,418 @@
...
+    string name = 1;
+
+    // The dataset (schema) of the declaration.
+    string dataset = 2;
+
+    // The Google Cloud project (database) of the declaration.
+    string project = 3;
+
+    // Targets of actions that this action is dependent on.
+    repeated Target dependency_targets = 4;
```
BenBirt, line 370: declarations never (cannot) have dependencies

protos/configs.proto
```diff
@@ -0,0 +1,418 @@
...
+
+    // A list of user-defined tags with which the action should be labeled.
+    repeated string tags = 5;
```
BenBirt, line 373: or tags

protos/configs.proto
```diff
@@ -0,0 +1,418 @@
...
+
+    // Description of the declaration.
+    string description = 6;
```
BenBirt, line 376: whyno column descriptions

PR #1657, opened 2024-01-22
Make v3 JS API act the same as v2

Ekrekr:


core/session.ts
```diff
@@ -377,6 +302,15 @@ export class Session {
     return newTest;
   }
 
+  public notebook(target: dataform.ITarget): Notebook {
```
BenBirt, line 305: should this not have the same/similar parameters as the other functions? (`declaration` is a bit special, i would consider it separately)

BenBirt (approved):

PR #1671, opened 2024-02-06
Improve generated workflow_settings.yaml field order

Ekrekr:


cli/api/commands/init.ts
```diff
@@ -51,20 +51,14 @@ export async function init(
     dirsCreated.push(projectDir);
   }
 
-  const workflowSettings = dataform.WorkflowSettings.create({
+  // The order that fields are set here is preserved in the outputted yaml.
```
BenBirt, line 54:
```suggestion
  // The order that fields are set here is preserved in the written yaml.
```

BenBirt (approved):

PR #1672, opened 2024-02-06
Remove JS context from SQL files

Ekrekr:
Update uniqueKey field description while here.


core/compilers.ts
```diff
@@ -37,10 +37,7 @@ export function compile(code: string, path: string): string {
   }
   if (Path.fileExtension(path) === "sql") {
     const { sql } = extractSqlxParts(SyntaxTreeNode.create(code));
```
BenBirt, line 39: i think this is pretty heavyweight, just for the sake of some escaping. i think you should just

```
const escapedCode = code.replace('`', '\\`')}.replace('${', '\\${');
return `exports.query = `${escapedCode}`;`;
```

BenBirt (changes requested):
Surely to be semantically equivalent you should also add `dependencyTargets: ...` in the `actions.yaml` for `extreme_weather_programming` actions?

core/compilers.ts
Ekrekr: Wait, why `.replace('${', '\\${')`? I thought the chat meant no JS in the SQL at all, not just no contextables

Ekrekr:
> Surely to be semantically equivalent you should also add `dependencyTargets: ...` in the `actions.yaml` for `extreme_weather_programming` actions?

Done

BenBirt (approved):

PR #1673, opened 2024-02-06
Fix config targets

Ekrekr:


core/main_test.ts
```diff
@@ -1116,6 +1116,53 @@ actions:
...
+        fs.writeFileSync(path.join(projectDir, `definitions/${filename}`), "SELECT 1");
+      });
+      fs.writeFileSync(
+        path.join(projectDir, `definitions/notebook.ipynb`),
+        EMPTY_NOTEBOOK_CONTENTS
+      );
+
+      const result = runMainInVm(coreExecutionRequestFromPath(projectDir));
+
+      expect(result.compile.compiledGraph.graphErrors.compilationErrors).deep.equals([]);
```
BenBirt, line 1164: this seems like a pretty weak assertion - dependency targets could be completely ignored by compilation, and this would still be true

BenBirt (approved):
I'll be honest, I am very confused about the current state of `table.ts` and friends. But the tests are passing, and presumably we can always clean up / refactor later.

core/main_test.ts
Ekrekr: Have added more tests

core/utils.ts
```diff
@@ -253,3 +253,15 @@ export function extractActionDetailsFromFileName(
...
+
+export function actionConfigToCompiledGraphTarget(
+  // The target interface is used here because Action configs contain all the fields of action
+  // config targets, even if they are not strictly target objects.
+  actionConfigTarget: dataform.ActionConfig.ITarget
+): dataform.Target {
+  return dataform.Target.create({
+    name: actionConfigTarget.name,
+    schema: actionConfigTarget.dataset,
+    database: actionConfigTarget.project
```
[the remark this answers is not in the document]
Ekrekr: This would add the dataset and project fields to the proto object, which even though they're not fields, is valid in protobufjs objects

Ekrekr:
> I'll be honest, I am very confused about the current state of `table.ts` and friends. But the tests are passing, and presumably we can always clean up / refactor later.

The class constructor is the builder for action configs - it's large at the moment because I want to split out view and incrementalTable into separate classes (soon).

PR #1675, opened 2024-02-15
Lazy / stateless installation

Ekrekr:
This more closely mimics what closed source does.


GJMcGowan:
Giving this to Ben as he'll understand better and Elias is away next couple of days anyway

Ekrekr:
> Giving this to Ben as he'll understand better and Elias is away next couple of days anyway

You're a changed man

cli/api/commands/install.ts
```diff
@@ -6,37 +6,88 @@ import { promisify } from "util";
...
-    throw new Error(
-      "dataformCoreVersion must be specified either in workflow_settings.yaml or via a package.json"
+  if (!fs.existsSync(packageJsonPath)) {
+    throw new Error(MISSING_CORE_VERSION_ERROR);
+  }
+
+  await promisify(childProcess.exec)("npm i --ignore-scripts", { cwd: resolvedProjectPath });
+}
+
+export async function runInstallIfWorkflowSettingsDataformCoreVersion(
```
BenBirt, line 33: I don't think this function belongs here. It's only used by compilation - so let's keep it there! I think the best thing to do would be to extract out common code - really just execution of `npm i` - to somewhere like cli/api/utils.ts.

cli/api/commands/install.ts
```diff
@@ -6,37 +6,88 @@ import { promisify } from "util";
...
+      packageLockJsonPath,
+      `{
+  "requires": true,
+  "dependencies": {
+    "@dataform/core": {
+      "version": "${workflowSettingsDataformCoreVersion}"
+    }
   }
+}`
+    );
```
BenBirt, line 66: why bother writing this file?

- the files will be deleted anyway
- `workflowSettingsDataformCoreVersion` should be a specific version anyway, so `package.json` is enough
- if `workflowSettingsDataformCoreVersion` isn't a specific version (i.e. semver), then this would be invalid

cli/api/commands/compile.ts
```diff
@@ -11,16 +13,26 @@ export class CompilationTimeoutError extends Error {}
 export async function compile(
   compileConfig: dataform.ICompileConfig = {}
 ): Promise<dataform.CompiledGraph> {
-  // Resolve the path in case it hasn't been resolved already.
-  path.resolve(compileConfig.projectDir);
+  let compiledGraph = dataform.CompiledGraph.create();
+  let workflowSettingsDataformCoreVersion = "";
+  try {
+    workflowSettingsDataformCoreVersion = await runInstallIfWorkflowSettingsDataformCoreVersion(
+      compileConfig.projectDir
```
BenBirt, line 20: I don't think you should be doing _any_ filesystem mutations inside the project directory. If the process dies, cleanup will not happen. Even if we could guarantee that cleanup would happen, it would still not be great from a filesystem-watching perspective.

I think this should all be happening in /tmp.

cli/api/commands/install.ts
```diff
@@ -6,37 +6,88 @@ import { promisify } from "util";
...
+    await promisify(childProcess.exec)("npm i --ignore-scripts", { cwd: resolvedProjectPath });
+  }
+
+  return workflowSettingsDataformCoreVersion;
+}
+
+export function cleanupNpmFiles(projectPath: string) {
```
BenBirt, line 73: ditto, I don't think this belongs here

BenBirt (changes requested):

cli/api/commands/install.ts
Ekrekr: To mimic internally, the `npm i` below should be `npm ci`, which requires a `package-lock.json` too - have updated

Ekrekr: From chat - we can remove `npm ci` from the server for this case too instead, which reduces overall complexity too

cli/api/commands/compile.ts
```diff
@@ -14,22 +17,53 @@ export async function compile(
...
+  "@dataform/core": "${workflowSettingsDataformCoreVersion}"
+  }
+}`
+    );
+
+    await promisify(exec)("npm i --ignore-scripts", {
+      cwd: temporaryInstallPath
+    });
+
+    fs.copySync(path.join(temporaryInstallPath, "node_modules"), resolvedProjectPath);
```
BenBirt, line 54: wait. you are still mutating the filesystem in the project directory. we should be doing the opposite: copy the project directory into /tmp, and then compile everything there.

cli/api/commands/compile.ts
```diff
@@ -14,22 +17,53 @@ export async function compile(
...
+  // project directories.
+  if (workflowSettingsDataformCoreVersion) {
+    if (fs.existsSync(projectNodeModulesPath)) {
+      fs.rmdirSync(projectNodeModulesPath);
     }
+
+    fs.writeFileSync(
+      path.join(temporaryInstallPath, "package.json"),
+      `{
+  "dependencies":{
```
BenBirt, line 44:
```suggestion
  "dependencies": {
```

cli/api/commands/compile.ts
```diff
@@ -14,22 +17,53 @@ export async function compile(
...
+
+    fs.copySync(path.join(temporaryInstallPath, "node_modules"), resolvedProjectPath);
+  }
+
+  const result = await CompileChildProcess.forkProcess().compile(compileConfig);
+
+  const decodedResult = decode64(dataform.CoreExecutionResponse, result);
+  compiledGraph = dataform.CompiledGraph.create(decodedResult.compile.compiledGraph);
+
+  compiledGraph.tables.forEach(setOrValidateTableEnumType);
```
BenBirt, line 62: why?

cli/api/commands/install.ts
```diff
@@ -1,10 +1,9 @@
 import * as fs from "fs";
-import { load as loadYaml, YAMLException } from "js-yaml";
 import * as path from "path";
 import { promisify } from "util";
 
 import * as childProcess from "child_process";
-import { dataform } from "df/protos/ts";
+import { readDataformCoreVersionFromWorkflowSettings } from "df/cli/api/commands/compile";
```
BenBirt, line 6: move this function into a shared utility file?

BenBirt (changes requested):

cli/api/commands/compile.ts
Ekrekr: No longer needed because v2 Core is no longer supported by the CLI! Removed.

cli/api/commands/compile.ts
```diff
@@ -20,8 +19,14 @@ export async function compile(
...
   const resolvedProjectPath = path.resolve(compileConfig.projectDir);
   const packageJsonPath = path.join(resolvedProjectPath, "package.json");
+  const packageLockJsonPath = path.join(resolvedProjectPath, "package-lock.json");
   const projectNodeModulesPath = path.join(resolvedProjectPath, "node_modules");
-  const temporaryInstallPath = tmp.dirSync().name;
+
+  const temporaryDirectoryPath = tmp.dirSync().name;
+  const temporaryProjectPath = path.join(
+    temporaryDirectoryPath,
+    resolvedProjectPath.split("/").slice(0, -1)[0]
```
BenBirt, line 28: Breaks on windows

BenBirt (approved):

PR #1676, opened 2024-02-16
Adding default-schema option to the init command

mohamed-saad:
The `init` CLI option always assume that the default-schema name is "dataform", so I have to go to the dataform.json file everytime to change it to a different schema.

In this PR, the `default-schema` option is added to the `init`

### DIFF
--- a/api/commands/init.ts
+++ b/api/commands/init.ts
@@ -49,8 +49,7 @@ export async function init(
     dataformJsonPath,
     prettyJsonStringify(
       dataform.ProjectConfig.create({
-        defaultSchema: "dataform",
-        assertionSchema: "dataform_assertions",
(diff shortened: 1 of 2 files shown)


Ekrekr:
> The `init` CLI option always assume that the default-schema name is "dataform", so I have to go to the dataform.json file everytime to change it to a different schema.
> 
> In this PR, the `default-schema` option is added to the `init`

I'm not sure I fully understand the pain point - having to go to the dataform.json to update the default schema isn't changed by not having a default.

Also FYI, these options have changed a lot in v3, due to `workflow_settings.yaml`. As we're not too far out from making that the default branch, I'd avoid working too much directly in main. I'll cover this in the OS onboarding next week!

PR #1683, opened 2024-03-07
Remove all analytics from the dataform CLI

GJMcGowan:
Per internal bug, segment is now deleted so this code does nothing.


package.json
```diff
@@ -161,4 +158,4 @@
     "**/node-fetch": "^2.6.7",
     "**/markdown-it": "^12.3.2"
   }
-}
+}
```
BenBirt, line 161: bring back trailing newline

BenBirt (approved):

PR #1686, opened 2024-03-12
Add notebook runtime options to workflow settings

Ekrekr:


GJMcGowan (approved):
Is it okay to merge a breaking change right into main? I guess we're not releasing anything

PR #1687, opened 2024-03-13
Add image annotate example

Ekrekr:
@G2H interested to see what you think about this cyclic project structure!


examples/image_annotate/definitions/unannotated_imagesets.sqlx
```diff
@@ -0,0 +1,19 @@
...
+js {
+    // Ref is avoided here to avoid the circular dependency chain.
+    const ANNOTATED_IMAGESETS = `${ctx.schema()}` + ".annotated_imagesets"
+}
+
+SELECT
+  *
+FROM
+  ${ref("all_imagesets")}
+LEFT JOIN
```
G2H, line 14: `WHERE uri NOT IN (SELECT uri FROM ${ANNOTATED_IMAGESETS})` is probably easier to read. Not sure if more or less efficient though

examples/image_annotate/definitions/annotated_imagesets.sqlx
```diff
@@ -0,0 +1,25 @@
...
+WHERE
+  ml_annotate_image_status NOT LIKE "${constants.RETRYABLE_ERROR}"
+  ${
+      when(incremental(), `AND
+      ${constants.MERGE_COLUMN} NOT IN (
+      SELECT
+        ${constants.MERGE_COLUMN}
+      FROM
+        ${self()})
+      OR updated > (
```
G2H, line 18: Where is field coming from? I don't see it listed in the output of ML.ANNOTATE_IMAGE

https://cloud.google.com/bigquery/docs/reference/standard-sql/bigqueryml-syntax-annotate-image

G2H: Also, why do you need that bit given the above?

G2H (approved):

Ekrekr:
This is a bit obsolete now, https://github.com/dataform-co/dataform-bqml supports equivalent use case in a better way

PR #1688, opened 2024-03-14
Add a configs reference page, and tidy up other readmes

Ekrekr:


scripts/regenerate_docs
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+docker pull pseudomuto/protoc-gen-doc
```
lewish, line 3: Obviously I would prefer that this was part of bazel but I won't block on that!

scripts/regenerate_docs
```diff
@@ -0,0 +1,8 @@
...
+
+docker run --rm \
+  -v $(pwd)/docs:/out \
+  -v $(pwd)/protos:/protos \
+  pseudomuto/protoc-gen-doc configs.proto --doc_opt=markdown,configs-reference.md
```
lewish, line 8: let's see if we can generate lowerCameCase names

lewish (approved):

scripts/regenerate_docs
Ekrekr: IIRC We need to update docker rules because they're still using Python 2.7, I'll look into updating this after that is done!

scripts/regenerate_docs
Ekrekr: There's no option for this currently - we would have to add these feature ourselves. I may do this!

PR #1689, opened 2024-03-14
Remove rules docker

Ekrekr:
* We're no longer meant to publish our own docker images.
* We can use the prebuilt bazel docker image


cloudbuild/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# TODO(ekrekr): update to image `gcr.io/cloud-builders/bazel` once we're at Bazel version 7+.
+FROM gcr.io/cloud-builders/bazel:5.4.0
```
lewish, line 2:
```
FROM index.docker.io/debian:bullseye
```

There's no point pulling from bazel as you install bazelisk below anyway

cloudbuild/cloudbuild.yaml
```diff
@@ -0,0 +1,6 @@
+steps:
```
lewish, line 1: Pretty sure you can't do this from inside cloud build easily. Docker in docker is a headache. Doing it manually is fine for now IMO!

cloudbuild-test.json
lewish, line 1: I'm confused, this doesn't seem to have a replacement? You will need to reconfigure cloud build if you move this file

cloudbuild/Dockerfile
Ekrekr: I realised we can get rid of the docker publish and run step entirely! see new script.

cloudbuild-test.json
Ekrekr: cloudbuild/cloudbuild.yaml is the new cloudbuild file. Having the scripts next to it seems to make structural sense

Ekrekr: Alternatively I could keep cloudbuild.yaml at the root, and put the scripts under scripts/ - I don't really mind either approach

cloudbuild/cloudbuild.yaml
```diff
@@ -0,0 +1,5 @@
...
+  - name:
+    args: ["./cloudbuild/install_requirements.sh"]
```
lewish, line 3: I don't see what this is for? This all happens in the docker image so none of it should be necessary (it will also be slow)

Ekrekr: Fair point, have removed that.

This command seems to work pretty well: `docker run -it -v /usr/local/google/home/eliaskassell/Documents/github/dataform:/dataform -v ~/.config/gcloud:/root/.config/gcloud -w /dataform --entrypoint bash gcr.io/cloud-builders/bazel:5.4.0 ./scripts/run_tests` - have updated the cloudbuild commands to do this instead.

cloudbuild-test.json
```diff
@@ -1,13 +1,9 @@
...
   "steps": [
     {
-      "name": "gcr.io/dataform-public/dataform-builder@sha256:413b7e076ae77f1bed9b254434919e40d4d4112c00e1a12ea076da112a7012e9",
+      "name": "gcr.io/cloud-builders/bazel:5.4.0",
       "entrypoint": "bash",
-      "args": ["./scripts/cloudbuild/bazel_test"]
+      "args": ["./scripts/run_tests"]
     }
-  ],
-  "timeout": "3600s",
```
lewish, line 9: Can you not remove these?
Beefy machine is definitely advantageous! Timeout though is probably a bit high based on my experience recently

lewish (approved):

Issue #1362, opened 2022-08-22
CLI Output Only shows up to the first "." in the filename


BenBirt:
Original public issue (with attached context): https://b.corp.google.com/issues/242250553

a2wd:
From the linked bug:

**Problem you have encountered:**

CLI output will not display the fill SQLX filename when a "." is present, instead the output is only showing up to the first "." in the filename.


**What you expected to happen:**

Display full name up to the file extension ".sqlx"


**Steps to reproduce:**

Create Operations Definition Files:

"LOAD_CSV.TABLE1.sqlx"
"LOAD_CSV.TABLE2.sqlx"

Execute `dataform compile`

Output would be:

```
Compiling...

Compiled 2 action(s).
2 operation(s):
  STAGING.LOAD_CSV 
  STAGING.LOAD_CSV
```

Ekrekr:
Playing around with this, I think it almost works as intended. If you allow dots in filenames to be added to action names, it makes the BigQuery table reference invalid.

E.g. `STAGING.LOAD_CSV` is valid, as it follows `dataset.name`, but `STAGING.LOAD_CSV.TABLE1` would then follow the BigQuery target reference of `project.dataset.name`, which is entirely different.

So to make it clear why this is invalid, I've filed https://github.com/dataform-co/dataform/pull/1699 to throw an error for it.

PR #1699, opened 2024-03-27
Throw errors when dots are in fields in targets

Ekrekr:
Fixes https://github.com/dataform-co/dataform/issues/1362#issuecomment-2022561105, which contains the explanation for this solution

Also update filename in utils to basename as it is more correct.

### DIFF
--- a/core/actions/assertion.ts
+++ b/core/actions/assertion.ts
@@ -92,7 +92,7 @@ export class Assertion extends ActionBuilder<dataform.Assertion> {
       config.name = Path.basename(config.filename);
     }
     const target = actionConfigToCompiledGraphTarget(config);
-    this.proto.target = this.applySessionToTarget(target);
+    this.proto.target = this.applySessionToTarget(target, config.filename);
     this.proto.canonicalTarget = this.applySessionCanonicallyToTarget(target);
 
     // Resolve the filename as its absolute path.
--- a/core/actions/assertion.ts
+++ b/core/actions/assertion.ts
@@ -89,7 +89,7 @@ export class Assertion extends ActionBuilder<dataform.Assertion> {
     }
 
     if (!config.name) {
-      config.name = Path.fileName(config.filename);
+      config.name = Path.basename(config.filename);
     }
     const target = actionConfigToCompiledGraphTarget(config);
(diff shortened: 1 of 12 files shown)


core/actions/assertion.ts
```diff
@@ -89,7 +89,7 @@ export class Assertion extends ActionBuilder<dataform.Assertion> {
     }
 
     if (!config.name) {
-      config.name = Path.fileName(config.filename);
+      config.name = Path.basename(config.filename);
```
BenBirt, line 92: i still think this capitalized import name is super weird (`Path` as opposed to `path` which would be more standard)

core/session.ts
```diff
@@ -392,6 +392,16 @@ export class Session {
       )
     );
 
+    this.checkNoDotsInTargetFields(
```
BenBirt, line 395: why do we need to do this at the end of compilation? IMO it should be happening well before then - at action declaration time.

Ekrekr: Where are you suggesting to put it?

Would it be both:
* Action functions in the session https://github.com/dataform-co/dataform/blob/b5368e04a047c6f2c958b6dd14d9bd5f872e4e0f/core/session.ts#L236C11-L236C21
* Action constructors https://github.com/dataform-co/dataform/blob/b5368e04a047c6f2c958b6dd14d9bd5f872e4e0f/core/actions/table.ts#L291

core/actions/assertion.ts
Ekrekr: It should probably be called `pathUtils` instead. It's capitalized for importing our path.ts file, but lowercase for our node path imports https://github.com/dataform-co/dataform/blob/b5368e04a047c6f2c958b6dd14d9bd5f872e4e0f/core/main_test.ts#L5.

Will update in a subsequent CL.

BenBirt: IMO it's fine as `path` everywhere, but up to you

core/session.ts
BenBirt: Everything passes through the action constructors, right? So we should only do it there.

BenBirt (approved):

PR #1705, opened 2024-04-02
Expose `defaultLocation` in `dataform.projectConfig`.

BenBirt:


Ekrekr:
Will this definitely fix the use case (I'm not sure what bug this is for). default location is already outputted in the compiled graph https://github.com/dataform-co/dataform/blob/8f288b8b82fe22d6b3e72ccb4da1802478e547da/core/main_test.ts#L378

Ekrekr (approved):

BenBirt:
> Will this definitely fix the use case (I'm not sure what bug this is for). default location is already outputted in the compiled graph
> 
> https://github.com/dataform-co/dataform/blob/8f288b8b82fe22d6b3e72ccb4da1802478e547da/core/main_test.ts#L378

Yes, it should fix https://issuetracker.google.com/issues/331919803.

Issue #1561, opened 2023-10-31
Generated incremental SQL when table does not exist should not do a `create or replace`


lewish:
In the rare situation that seems to have happened to a customer recently - that BigQuery tells us a incremental table does not exist even though it does - we should not generate a `create or replace` statement to rebuild the table in case it does actually exist, and instead generate a `create table if not exists` statement in case.

lewish:
Relevant code: https://github.com/dataform-co/dataform/blob/b5a561643e5020401639ef3a73aab6991083eb16/core/adapters/bigquery.ts#L34C46-L34C46

Should be: 

- Full refresh requested and NOT protected: `create or replace`
- Otherwise, if the table does not exist: `create table if not exists`

PR #1712, opened 2024-04-04
Update incremental table execution to avoid create or replace

Ekrekr:
Fixes https://github.com/dataform-co/dataform/issues/1561


cli/api/dbadapters/execution_sql.ts
```diff
@@ -196,17 +196,29 @@ from (${query}) as insertions`;
...
-        ? `partition by ${table.bigquery.partitionBy} `
-        : ""
-    }${
-      table.bigquery && table.bigquery.clusterBy && table.bigquery.clusterBy.length > 0
-        ? `cluster by ${table.bigquery.clusterBy.join(", ")} `
-        : ""
-    }${options.length > 0 ? `OPTIONS(${options.join(",")})` : ""}as ${table.query}`;
+    let statement = "create ";
+    if (table.enumType !== dataform.TableType.INCREMENTAL) {
+      statement += "or replace ";
```
lewish, line 201: From the bug - I think we still want to create or replace in the case that it is incremental AND it's not projected AND it's full refresh?

lewish: We should also add a test for this case

Ekrekr: What is projected? Have added the full refresh.

This is tested, but not obviously, across https://github.com/dataform-co/dataform/blob/c55ca6bcc79b1087545655605c7855447d091a45/tests/api/api.spec.ts#L463

lewish: *protected!

Ekrekr: Ah right! Done.

I had to update the protected logic, such as removing the error that we test for here https://github.com/dataform-co/dataform/blob/c55ca6bcc79b1087545655605c7855447d091a45/tests/api/api.spec.ts#L109. But not it should act as the incremental spec suggests.

lewish: Right, sorry I think this has got really confusing, a few thoughts.

1. leave this method as `createOrReplace` and don't pass in run options or do all that logic.
2. Add a `createIfNotExists` method next to it (they can share some private impl if you want, but I wouldn't bother)
3. Update the logic in publishTasks, only for the incremental tables, to do what you want to do and call `createIfNotExists` by default, and `createOrReplace` only when `!protected && fullRefresh`

The original method was clear what it did, but now you've moved this logic into it, I'd say it's much harder to follow what's going on in `publishTasks`.

Ekrekr:
CUJ is replaced in favor of the new onSchemaChange option

PR #1720, opened 2024-04-19
Feature: dependent assertions as dependencies

Tuseeq1:
### Summary:
This PR introduces changes that will allow users to add dependent assertions as dependencies. 
This required a new `ActionAssertionMap` that holds information regarding assertions that depend on each action. This map is updated while create assertions. 
New flags for Table and Operation to allow managing dependent assertions. These flags are updated for each dependency when its added in the Action either through config or ref(both happens in same place).
Lastly, when dependencies are qualified for compiledGraph, we add dependent assertions as dependencies using the ActionAssertionMap and the flags.

This PR also introduces new tests to test this feature and minor updates to some old test to adjust to new flags being introduces.

#core #feature #dependentAssertionsAsDependencies


core/actions/operation.ts
```diff
@@ -75,6 +77,10 @@ export class Operation extends ActionBuilder<dataform.Operation> {
   // We delay contextification until the final compile step, so hold these here for now.
   private contextableQueries: Contextable<ICommonContext, string | string[]>;
 
+  // This is Action level flag. If set to true, we will add assertions from all the depenedncies of
+  // current action as dependencies. 
+  public dependOnDependencyAssertions: boolean = false;
```
Ekrekr, line 82: I don't see this class variable being read from anywhere - can't we remove it, and just use the value from the config?

core/actions/table.ts
```diff
@@ -553,7 +562,19 @@ export class Table extends ActionBuilder<dataform.Table> {
   public dependencies(value: Resolvable | Resolvable[]) {
     const newDependencies = Array.isArray(value) ? value : [value];
     newDependencies.forEach(resolvable => {
-      this.proto.dependencyTargets.push(resolvableAsTarget(resolvable));
+      let dependencyTarget = resolvableAsTarget(resolvable);
+      dependencyTarget.includeDependentAssertions = dependencyTarget.includeDependentAssertions===undefined ? this.dependOnDependencyAssertions : dependencyTarget.includeDependentAssertions;
```
Ekrekr, line 566: This is logically identical to

```suggestion
      dependencyTarget.includeDependentAssertions = dependencyTarget.includeDependentAssertions || this.dependOnDependencyAssertions;
```

core/actions/table.ts
```diff
@@ -553,7 +562,19 @@ export class Table extends ActionBuilder<dataform.Table> {
...
+      const existingDependencies = this.proto.dependencyTargets.filter(dependency => isSameTarget(dependencyTarget, dependency) && dependency.includeDependentAssertions !== dependencyTarget.includeDependentAssertions)
```
Ekrekr, line 567: It's not particularly obvious to me why this is needed, could you explain?

IIUC:

A => B, where B has `includeDependentAssertions: true`, and A which generates inline assertions, has `includeDependentAssertions: false`.

Why does A need to have `includeDependentAssertions` to be true for this to be valid?

core/actions/operation.ts
```diff
@@ -156,7 +165,19 @@ export class Operation extends ActionBuilder<dataform.Operation> {
...
+      if (existingDependencies.length !== 0){
+          this.session.compileError(
+            `Conflicting "includeDependentAssertions" flag not allowed. Dependency ${dependencyTarget.name} have different value set for this flag.`,
+             this.proto.fileName,
+             this.proto.target
+          )
+          return this;
+      }
+
+      this.proto.dependencyTargets.push(dependencyTarget);
```
Ekrekr, line 180: These blocks are duplicated across action types, we should introduce a shared function for it in `core/actions/index.ts`

Ekrekr (commented):
There's quite a few outstanding comments in the design doc, and we haven't done a review session to close them, which we should do before this is merged!

Why is it just operations and tables that can have `dependOnDependencyAssertions`? Shouldn't all action types (excluding declarations) be able to require directly downstream assertions to have completed in order to run?

core/utils.ts
```diff
@@ -299,3 +299,21 @@ export function actionConfigToCompiledGraphTarget(
...
+          return false;
+      }
+  }
+  if (!!targetA.schema || !!targetB.schema){
+      if(targetA.schema !== targetB.schema){
+          return false
+      }
+  }
+  return true
+}
```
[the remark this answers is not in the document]
Tuseeq1: Stringify will also use the newly added "IncludeDependentAssertions" flag in target. We do not want to add that. The idea here is to find same target and compare this flag only.

core/actions/operation.ts
Tuseeq1: this variable is used inside `dependencies` on line 165. 
we do not attach config directly with the object. so when `dependencies` is called through ref() function, we need this information. thats why it is stored in this variable from config on an earlier stage.

core/actions/table.ts
Tuseeq1: this will not work. we want to prioritize the `dependencyTarget.includeDependentAssertions` over `this.dependOnDependencyAssertions` regardless of value itself(true or false). only in case this is not given we use the default setting `this.dependOnDependencyAssertions`.

core/actions/table.ts
Tuseeq1: This checks the case where one dependency A is added multiple times(in config.dependencies and in ref()). but with different value set to `includeDependentAssertions` flag. In this case we should throw an error. 

Consider following .sqlx file.
`config {
            type: "table",
            dependencies: [{name: "A", includeDependentAssertions: false}]
        }
select * from \${ref({name: "A", includeDependentAssertions: true})}
select 1 as test`

core/actions/operation.ts
Tuseeq1: this function is not used in all sub classes for ActionBuilder. that is why i kept it in operations and tables only.

core/common.ts
```diff
@@ -219,6 +224,8 @@ export interface ITarget {
   schema?: string;
 
   name?: string;
+
+  IncludeDependentAssertions?: boolean;
```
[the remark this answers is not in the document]
Tuseeq1: Code uses both proto and interface for target in different locations. so both required updates. 

I dont want to make code refactoring changes part of this PR. we can do that separately to replace interfaces with protos.

Tuseeq1:
> There's quite a few outstanding comments in the design doc, and we haven't done a review session to close them, which we should do before this is merged!
> 
> Why is it just operations and tables that can have `dependOnDependencyAssertions`? Shouldn't all action types (excluding declarations) be able to require directly downstream assertions to have completed in order to run?


We have 5 types of actions Assertions, Declaration, Notebook, Operations, and Tables. 

1. Assertions: 
We decided not to have this option for assertions to avoid circler graphs and avoid complexion as mentioned in DD. 
2. Declaration dont have dependencies
3. Notebooks can only use action.yaml to set config and as we discussed in chat that currently we dont have this option.

that leave only Tables and Operations.

core/actions/operation.ts
Ekrekr: Ah I see what you mean, we can't store and read it from the core proto, like we do with most other fields, because it's not a field there. This makes sense then.

Ekrekr: In the future once we've migrated to just using the config proto, we can store the config proto and read this field from there

core/session.ts
```diff
@@ -806,3 +816,71 @@ class ActionIndex {
     return this.byName.get(target.name) || [];
   }
 }
+
+
+class ActionAssertionMap {
```
Ekrekr, line 821: An actions index already exists
https://github.com/dataform-co/dataform/blob/754d582f4f588b3c812c40a38ec839607ec3d4f0/core/session.ts#L324 before we start compiling graph chunks https://github.com/dataform-co/dataform/blob/754d582f4f588b3c812c40a38ec839607ec3d4f0/core/session.ts#L344

Why can't that be used to retrieve the action as its fully qualified name, for comparing targets?

tests/core/core.spec.ts
```diff
@@ -901,7 +901,7 @@ suite("@dataform/core", () => {
       expect(
         cGraph.graphErrors.compilationErrors.filter(item =>
           item.message.match(
-            /Ambiguous Action name: {\"name\":\"a\"}. Did you mean one of: foo.a, bar.a./
+            /Ambiguous Action name: {\"name\":\"a\",\"includeDependentAssertions\":false}. Did you mean one of: foo.a, bar.a./
```
Ekrekr, line 904: Note to self: this is kind of gross, we should fix it by not using JSON stringify like this https://github.com/dataform-co/dataform/blob/754d582f4f588b3c812c40a38ec839607ec3d4f0/core/utils.ts#L129C42-L129C53

Ekrekr (changes requested):
This PR is pretty massive! If you want to speed up the review process, I'd suggest splitting it up a bit - up to you. For example:
1. add to the proto definition and ITarget
2. add support the option in the SQLX config
3. support the option in ref

core/actions/assertion.ts
```diff
@@ -56,7 +56,8 @@ export const IAssertionConfigProperties = strictKeysOf<IAssertionConfig>()([
   "name",
   "schema",
   "tags",
-  "type"
+  "type",
+  "dependOnDependencyAssertions"
```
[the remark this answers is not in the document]
Tuseeq1: I thought we skipped this for Weronika to decide.

core/common.ts
```diff
@@ -133,6 +133,11 @@ export interface IDependenciesConfig {
    * dependencies, then it should be set to `true`.
    */
   hermetic?: boolean;
+
+  /**
+   * If this flag is set to true, assertions depenedent upon any of the dependencies are added as depenedencies as well.
+   */
+  dependOnDependencyAssertions?: boolean;
```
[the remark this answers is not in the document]
Tuseeq1: Declarations does not use `IDependenciesConfig` so will not have this option either.

core/main_test.ts
```diff
@@ -1260,6 +1260,626 @@ actions:
...
+             assertions: {
+               rowConditions: ["test > 1"]
+            }
+          }
+          SELECT 1 as test
+        }
+        `);
+
+          const result = runMainInVm(coreExecutionRequestFromPath(projectDir));
+          expect(result.compile.compiledGraph.graphErrors.compilationErrors).deep.equals([]);
```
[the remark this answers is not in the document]
Tuseeq1: How? there should be no compilation error here.

core/session.ts
Tuseeq1: These might look quit similar because I used code from ActionMap for look up. However, these are two different things. 
ActionMap converts a simple list into a lookup list where each actions is mapped alongs its own target.
On the other hand ActionAssertionMap hold the information of assertions related to each action.
this information is updated in assertions and used later to find the assertions related to any actions. The implementation and usage for these two are very different.

core/common.ts
```diff
@@ -219,6 +224,8 @@ export interface ITarget {
   schema?: string;
 
   name?: string;
+
+  includeDependentAssertions?: boolean;
```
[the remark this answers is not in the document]
Tuseeq1: We discussed this, since we want to move to protos making another interface doesnt serve the purpose.

core/main_test.ts
```diff
@@ -1260,6 +1260,626 @@ actions:
...
+            type: "assertion",
+        }
+        select test from \${ref("A")} where test > 3`);
+        fs.writeFileSync(path.join(projectDir, "definitions/B.sqlx"),
+           `config {
+            type: "table",
+            dependOnDependencyAssertions: true,
+            dependencies: ["A"]
+        }
+        select * from \${ref({name: "C", includeDependentAssertions: false})}
```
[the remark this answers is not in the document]
Tuseeq1: This does not throw any error and is working as intended.

core/session.ts
kolina: As far I can see they only differ in the target type (`Action` vs `ITarget`), the actual code logic is the same. Can you just use `Map<string, dataform.ITarget[]>` and call `targetAsReadableString` to get a fully qualified name for an assertion target key? An alternative way is to create a generic class `ActionMap<ValueType>` and then instantiate it with `ActionMap<Action>` for `indexedActions` and `ActionMap<ITarget>` for `actionAssertionMap`.

core/utils.ts
```diff
@@ -293,9 +295,36 @@ export function actionConfigToCompiledGraphTarget(
...
+export function addDependenciesToActionDependencyTargets(action: actionsWithDependencies, resolvable: Resolvable) {
+  let dependencyTarget = resolvableAsTarget(resolvable);
+  if (!dependencyTarget.hasOwnProperty("includeDependentAssertions")) {
+    // dependency `includeDependentAssertions` takes precedence over the config's `dependOnDependencyAssertions`
+    dependencyTarget.includeDependentAssertions = action.dependOnDependencyAssertions;
+  }
+
+  // check if same dependency already exist in this action but with opposite value for includeDependentAssertions
+  const conflictingIncludeDependentAssertions =
+    action.proto.dependencyTargets.filter(existingDependency =>
```
kolina, line 317: @Ekrekr, is it ok to have a quadratic complexity from the number of action dependencies here? I assume with thousands of actions it may be acceptable, but if in the future we'll scale to tens of thousands potential dependencies it may become quite slow (though the workflows where one action has so many dependencies are weird).

kolina (changes requested):

core/utils.ts
Ekrekr: Where does the quadratic complexity come from? I don't really follow

core/main_test.ts
```diff
@@ -1261,88 +1261,67 @@ actions:
...
-          }
-          SELECT 1 as test
-        }
-        `);
-          fs.writeFileSync(path.join(projectDir, "definitions/A_assert.sqlx"), 
-          `config {
-            type: "assertion",
-        }
-        select test from \${ref("A")} where test > 3`);
+      test("Setting includeDependentAssertions to true in config.dependencies adds assertions from that dependeny to dependencyTargets", () => {
```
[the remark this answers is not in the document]
Tuseeq1: done

core/main_test.ts
```diff
@@ -1493,90 +1426,48 @@ actions:
...
-          SELECT 1 as test
-        }
-        `);
-          fs.writeFileSync(path.join(projectDir, "definitions/A_assert.sqlx"), 
-          `config {
-            type: "assertion",
-        }
-        select test from \${ref("A")} where test > 3`);
+
+      test("If assertions are added using includeDependentAssertions and also by add in dependencies, this duplications should be handle and added only once.", () => {
```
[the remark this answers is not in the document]
Tuseeq1: done

core/main_test.ts
```diff
@@ -1493,90 +1426,48 @@ actions:
...
-          }
-          SELECT 1 as test
-        }
-        `);
-          fs.writeFileSync(path.join(projectDir, "definitions/A_assert.sqlx"), 
-          `config {
-            type: "assertion",
-        }
-        select test from \${ref("A")} where test > 3`);
+      test("When includeDependentAssertions property in config and ref are set different for same dependency, compilationError is thrown.", () => {
```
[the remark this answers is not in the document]
Tuseeq1: done

core/utils.ts
```diff
@@ -6,7 +6,7 @@ import { Resolvable } from "df/core/common";
 import * as Path from "df/core/path";
 import { IActionProto, Session } from "df/core/session";
 import { dataform } from "df/protos/ts";
-import { Notebook } from "./actions/notebook";
+import { Notebook } from "df/core/actions/notebook";
```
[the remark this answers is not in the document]
Tuseeq1: done

core/utils.ts
Tuseeq1: replaced this loop with Map to make it O(n).

core/actions/assertion.ts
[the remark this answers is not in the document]
Ekrekr: `dependOnDependencyAssertions` confirmed it looks like!

core/main_test.ts
```diff
@@ -1260,6 +1260,453 @@ actions:
...
+            dependOnDependencyAssertions: true,
+            dependencies: ["A"]
+        }
+        select 1 as btest
+        `);
+
+        const result = runMainInVm(coreExecutionRequestFromPath(projectDir));
+
+        expect(result.compile.compiledGraph.graphErrors.compilationErrors).deep.equals([]);
+        expect(asPlainObject(result.compile.compiledGraph.tables[1].dependencyTargets.length)).equals(3)
```
Ekrekr, line 1308: Nit: to keep these tests flexible, we should probably do something like `result.compile.compiledGraph.tables.find(table => table.name === "A")` instead

core/main_test.ts
```diff
@@ -1260,6 +1260,453 @@ actions:
...
+              - operation:
+                  filename: C.sql
+                  dependOnDependencyAssertions: true
+                  dependencyTargets:
+                    - name: A
+              - notebook:
+                  filename: notebook.ipynb
+                  dependOnDependencyAssertions: true
+                  dependencyTargets:
+                    - name: A
```
Ekrekr, line 1531: I much prefer having file contents unindented, as the current tests in the file are - it prevents confusing diff errors when calling `expect`.

In any case, we should be consistent.

core/main_test.ts
```diff
@@ -1260,6 +1260,453 @@ actions:
...
+                  filename: A.sql
+              - assertion:
+                  filename: A_assert.sql
+                  dependencyTargets:
+                    - name: A
+              - view:
+                  filename: B.sql
+                  dependencyTargets:
+                    - 
+                      name: A
```
Ekrekr, line 1558: same line

core/main_test.ts
```diff
@@ -1260,6 +1260,453 @@ actions:
...
+        }
+        `);
+
+        const result = runMainInVm(coreExecutionRequestFromPath(projectDir));
+        expect(result.compile.compiledGraph.graphErrors.compilationErrors.length).deep.equals(2);
+        expect(result.compile.compiledGraph.graphErrors.compilationErrors[0].message).deep.equals(`Conflicting "includeDependentAssertions" properties are not allowed. Dependency A has different values set for this property.`);
+      })
+
+      suite("Action configs", ({ beforeEach }) => {
+        let projectDir: any;
```
Ekrekr, line 1487: You can technically have the same forEach for this suite and the one above, which would simplify the action configs below, by not having to include A and its assertion.

Ekrekr (changes requested):
The version should be increased here, for publishing the changes https://github.com/dataform-co/dataform/blob/main/version.bzl

core/main_test.ts
```diff
@@ -1260,6 +1260,626 @@ actions:
...
+        fs.writeFileSync(path.join(projectDir, "definitions/C.sqlx"), 
+          `config {
+            type: "table",
+             assertions: {
+               rowConditions: ["test > 1"]
+            }
+          }
+          SELECT 1 as test
+        }
+        `);
```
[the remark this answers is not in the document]
Ekrekr: I'm still confused why there are lots of C tables that don't depend on are are depended on other tables

core/main_test.ts
Tuseeq1: But I want to create the assertions using action yaml to make sure it is getting mapped correctly this way as well.

core/main_test.ts
Tuseeq1: Done

Ekrekr:
I've got most of the cloud build tests working in https://github.com/dataform-co/dataform, I think next time you push they should pop up!

If not, we can just run `./scripts/run_tests` locally before merging.

Ekrekr (approved):
Something weird is up with the PR - I shouldn't be seeing the diff from the new code in main. Something to do with the way the merge was done
image

But it doesn't matter too much anyway, because it will be squashed on merge.

PR #1733, opened 2024-05-09
Make proto object verification give more descriptive error messages

Ekrekr:
Given that we can't do proper type validation due to ProtobufJS, this is the next best thing that I can think of.

### DIFF
--- a/common/protos/index.ts
+++ b/common/protos/index.ts
@@ -18,6 +18,12 @@ export interface IProtoClass<IProto, Proto> {
   getTypeUrl(prefix: string): string;
 }
 
+export enum VerifyProtoErrorBehaviour {
+  DEFAULT,
+  SUGGEST_REPORTING_TO_DATAFORM_TEAM,
+  SHOW_DOCS_LINK
+}
+
 // This is a minimalist Typescript equivalent for the validation part of Profobuf's JsonFormat's
 // mergeMessage method:
 // https://github.com/protocolbuffers/protobuf/blob/670e0c2a0d0b64c994f743a73ee9b8926c47580d/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java#L1455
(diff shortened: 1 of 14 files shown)


common/protos/index.ts
```diff
@@ -24,7 +28,9 @@ export interface IProtoClass<IProto, Proto> {
 // meaning that the type of fields cannot be verified; an int can be confused with a string.
 export function verifyObjectMatchesProto<Proto>(
   protoType: IProtoClass<any, Proto>,
-  object: object
+  object: object,
+  suggestReportToDataformTeam: boolean = false,
+  showDocsLink: boolean = true
```
BenBirt, line 33: give me an `Options` type so that these can be named and therefore readable at the callsite

BenBirt (changes requested):

common/protos/index.ts
Ekrekr: Good shout

common/protos/index.ts
```diff
@@ -24,7 +28,8 @@ export interface IProtoClass<IProto, Proto> {
 // meaning that the type of fields cannot be verified; an int can be confused with a string.
 export function verifyObjectMatchesProto<Proto>(
   protoType: IProtoClass<any, Proto>,
-  object: object
+  object: object,
+  options?: { suggestReportToDataformTeam?: boolean; showDocsLink?: boolean }
```
BenBirt, line 32: sorry, now I see that it looks like these options are effectively an enumeration - it essentially never makes sense for them both to be true; because `suggestReportToDataformTeam` always takes precedence.

I'd suggest:

```
errorBehaviour?: enum ErrorBehaviour { SUGGEST_REPORTING_TO_DATAFORM_TEAM, SHOW_DOCS_LINK }
```

BenBirt (approved):

PR #1766, opened 2024-06-26
Operation config as protos

Ekrekr:
Validation of these options are locked to the same functionality as before this change via these tests: https://github.com/dataform-co/dataform/blob/b3638a8134f51afdb7afed13d3786b0e2e0e98c9/core/main_test.ts#L1849C11-L1849C25

### DIFF
--- a/core/actions/assertion.ts
+++ b/core/actions/assertion.ts
@@ -75,8 +75,6 @@ export class Assertion extends ActionBuilder<dataform.Assertion> {
       this.query(nativeRequire(config.filename).query);
     }
 
-    // TODO(ekrekr): load config proto column descriptors.
-
     if (config.dependencyTargets) {
       this.dependencies(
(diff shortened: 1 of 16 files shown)


Tuseeq1 (approved):

core/actions/assertion.ts
```diff
@@ -75,8 +75,6 @@ export class Assertion extends ActionBuilder<dataform.Assertion> {
       this.query(nativeRequire(config.filename).query);
     }
 
-    // TODO(ekrekr): load config proto column descriptors.
```
Tuseeq1, line 78: Why remove this  comment in assertion?

Ekrekr: The comment is a lie - columns aren't a field on assertions https://cloud.google.com/dataform/docs/reference/dataform-core-reference#iassertionconfig

PR #1770, opened 2024-06-27
Add workflow_settings.yaml validation/completion VSCode extension

moker-spaghetti:
Add workflow_settings.yaml validation/completion VSCode extension.
This feature depends on 3rd party extension "redhat.vscode-yaml", so please reject this PR if you mind this point.
(I implemented the feature as optional, so the extension should work fine even without the redhat extension.)


GJMcGowan (approved):
Good idea!

PR #1773, opened 2024-06-28
Incremental tables as protos, via a new class

Ekrekr:
Validation of these options are locked to the same functionality as before this change via these tests: https://github.com/dataform-co/dataform/blob/b3638a8134f51afdb7afed13d3786b0e2e0e98c9/core/main_test.ts#L1747C11-L1747C23


diasdauletov (approved):
LGTM!

PR #1775, opened 2024-07-02
feat: vscode extension, more intutive compile errors

ashish10alex:
Currently the default behaviour for when a compilation error occurs is to show a generic information message. The PR uses the graphErrors object and outputs the errors to the user similar to GCP Dataform UI 

**Current behaviour**

CleanShot 2024-07-02 at 20 21 58@2x

**New behaviour**

CleanShot 2024-07-02 at 20 14 57@2x

### DIFF
--- a/vscode/server.ts
+++ b/vscode/server.ts
@@ -60,7 +60,7 @@ async function applySettings() {
 }
 
 async function compileAndValidate() {
-  let hasCompilationError = false;
+  let compilationFailed = false;
   const spawnedProcess = spawn(
     (process.platform !== "win32") ? "dataform" : "dataform.cmd",
     ["compile", "--json", ...settings.compilerOptions]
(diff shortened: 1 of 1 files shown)


vscode/server.ts
```diff
@@ -59,6 +59,7 @@ async function applySettings() {
 }
 
 async function compileAndValidate() {
+  let hasCompilationError = false;
```
Ekrekr, line 62: This is a bit confusingly named, because the errors below are also compilation errors. A better name would probably be `let compilationFailed = false`

Ekrekr (approved):
Much nicer!

You need to merge from this repo's main I think, have just merged some other PRs.

vscode/server.ts
ashish10alex: Thanks @Ekrekr, I agree. I have made the suggested change and synced this branch with dataform repo's main branch.

Issue #1776, opened 2024-07-03
All assertions fail when dry run 


dali-rmw:
Hello, when I try to execute `datafrom run --dry-run`, all my assertions fail with the same error:
BigQuery error: Cannot convert undefined or null to object.
Example of an assertion 
create or replace view `project.dataform_assertions.table_name_assertions_uniqueKey_0` as 
  > SELECT
  >   *
  > FROM (
  >   SELECT
  >     row_id, col1,
  >     COUNT(1) AS index_row_count
  >   FROM TABLE_NAME
  >   GROUP BY row_id, col1
  >   ) AS data
  > WHERE index_row_count > 1

PR #1778, opened 2024-07-05
fix(1776): dry run of assertions failing

ashish10alex:
Fixes https://github.com/dataform-co/dataform/issues/1776

The dry of assertion currently fail due to the code expecting the dry run to return at least one row on [this line](https://github.com/dataform-co/dataform/blob/b3638a8134f51afdb7afed13d3786b0e2e0e98c9/cli/api/commands/run.ts#L427). 

**Tests**
1. Assertions passing by making the change in the MR. verified by running `./scripts/run run --dry-run <dataform_project_directory>`
2. Assertion when actually ran i.e. using `./scripts/run run ...` catches assertion errors 
3. `bazel test //core/...` passes 

**PS:**
This might not be the best solution, but happy to receive guidance

### DIFF
--- a/cli/api/commands/run.ts
+++ b/cli/api/commands/run.ts
@@ -424,7 +424,7 @@ export class Runner {
       if (task.type === "assertion") {
         // We expect that an assertion query returns 1 row, with 1 field that is the row count.
         // We don't really care what that field/column is called.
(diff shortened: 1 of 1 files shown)


Ekrekr (approved):
Thanks for the fix!

PR #1780, opened 2024-07-07
Table configs as protos

Ekrekr:


core/actions/table.ts
```diff
@@ -272,247 +83,109 @@ export class Table extends ActionBuilder<dataform.Table> {
...
-      });
-    }
-    this.query(nativeRequire(tableTypeConfig.filename).query);
-    if (tableTypeConfig.preOperations) {
-      this.preOps(tableTypeConfig.preOperations);
-    }
-    if (tableTypeConfig.postOperations) {
-      this.postOps(tableTypeConfig.postOperations);
+    if (configPath) {
+      config.filename = resolveActionsConfigFilename(config.filename, configPath);
```
Ekrekr, line 110: TODO: throw compilation error here when filename is empty.

core/actions/incremental_table.ts
```diff
@@ -196,12 +189,8 @@ export class IncrementalTable extends ActionBuilder<dataform.Table> {
     return this;
   }
 
-  public protected(defaultsToTrueProtected: boolean) {
-    // To prevent accidental data deletion, protected defaults to true if unspecified.
```
diasdauletov, line 200: this protection is not needed anymore?

diasdauletov: aah i see you have the check down below

core/actions/index.ts
```diff
@@ -119,4 +157,56 @@ export class LegacyConfigConverter {
     }
     return legacyConfig;
   }
+
+  public static insertLegacyBigQueryOptionsToConfigProto<
+    T extends ILegacyTableConfig | ILegacyIncrementalTableConfig
+  >(legacyConfig: T): T {
+    if (legacyConfig?.bigquery) {
```
diasdauletov, line 164: nit: you can consider early exit to avoid nested ifs
`if (!legacyConfig?.bigquery) {
        return legacyConfig;
    }`

core/actions/table.ts
```diff
@@ -1,247 +1,58 @@
...
-    "materialized",
-    "dependOnDependencyAssertions"
-  ]);
-
-/**
- * Context methods are available when evaluating contextable SQL code, such as
- * within SQLX files, or when using a [Contextable](#Contextable) argument with the JS API.
+ * @hidden
+ * This maintains backwards compatability with older versions.
+ * TODO(ekrekr): consider breaking backwards compatability of these in v4.
```
diasdauletov, line 44: for all the `TODO(erkerkr)` probably it is better to have `TODO(issue)`

core/actions/index.ts
```diff
@@ -96,15 +97,52 @@ export abstract class ActionBuilder<T> {
...
+  when: (cond: boolean, trueCase: string, falseCase?: string) => string;
+
+  /**
+   * Indicates whether the config indicates the file is dealing with an incremental table.
+   */
+  incremental: () => boolean;
+}
+
 export class LegacyConfigConverter {
+  // This is a workaround to make bigquery options output empty fields with the same behaviour as
```
diasdauletov, line 118: do we have a test that covers this?

core/actions/table.ts
```diff
@@ -1,247 +1,58 @@
...
-  incremental: () => boolean;
+export interface ILegacyTableConfig extends dataform.ActionConfig.TableConfig {
+  dependencies: Resolvable[];
+  database: string;
+  schema: string;
+  fileName: string;
+  type: string;
+  bigquery?: ILegacyTableBigqueryConfig;
+  // Legacy table config's table assertions cannot directly extend the protobuf table config
+  // definition because of legacy  table config's flexible types.
```
diasdauletov, line 54: nit: remove an extra space
```suggestion
  // definition because of legacy table config's flexible types.
```

core/session.ts
```diff
@@ -258,15 +261,26 @@ export class Session {
...
-    const newTable = new Table();
-    newTable.session = this;
-    utils.setNameAndTarget(this, newTable.proto, name);
+    queryOrConfig?:
+      | Contextable<ITableContext, string>
+      | ILegacyTableConfig
+      | ILegacyViewConfig
+      | ILegacyIncrementalTableConfig
+  ): Table | IncrementalTable | View {
+    let newTable: Table | IncrementalTable | View = new View(this, { type: "view", name });
```
diasdauletov, line 270: why do we init this with view?

core/actions/table.ts
Ekrekr: There are quite a lot of these, I'll make a triage plan and turn them into bugs in one go

core/actions/index.ts
Ekrekr: Yep! Most table tests actually touch this, for example the fact there's no bigquery field here: https://github.com/dataform-co/dataform/blob/0e98564706536c940111f7420694cd8a9f784c33/core/main_test.ts#L1453

PR #1788, opened 2024-07-17
Added proto definitions for Data Preparations

fernst:


protos/core.proto
```diff
@@ -247,6 +247,190 @@ message NotebookRuntimeOptions {
...
+message JoinKey {
+  string left_column = 1;
+  string right_column = 2;
+}
+
+message TableReference {
+  string project = 1;
+  string dataset = 2;
+  string table = 3;
+}
```
Ekrekr, line 356: It seems strange to duplicate the table target like this. Target is already defined https://github.com/dataform-co/dataform/blob/97123faf432781480bee6f888a4881d934aff349/protos/core.proto#L47

protos/core.proto
```diff
@@ -247,6 +247,190 @@ message NotebookRuntimeOptions {
...
+
+  bool disabled = 6;
+
+  DataPreparationDefinition data_preparation = 7;
+}
+
+message DataPreparationDefinition {
+  repeated DataPreparationNode nodes = 1;
+  DataPreparationGenerated generated = 2;
+}
```
Ekrekr, line 271: These values need to be populated from somewhere, I'd imagine that the config proto needs more options that roughly correspond to the data in here?

Ekrekr, line 271: I don't really understand why this is structured in this way - with nested nodes.

There are many downsides to having a node that contains other nested nodes:
* What prevents it from becoming a cyclic dag?
* How would one represent this in a UI?

It would be much better to keep each node as separate actions instead.

protos/core.proto
Ekrekr, line 1: IIUC this is essentially a copy of the public proto. Ideally we want to keep this in a healthy DAG structure, as described in my comment about avoiding nodes that contain nested nodes.

Ideally we would destructure it to preserve the compiled graph structure, then restructure it at execution time. For example, the execution proto could contain essentially what has been put here currently.

(note this is the execution proto for the CLI - internally we have a protos, but they look very similar).

https://github.com/dataform-co/dataform/blob/97123faf432781480bee6f888a4881d934aff349/protos/execution.proto#L32

I'm interested in your thoughts on this!

protos/configs.proto
```diff
@@ -515,6 +515,32 @@ message ActionConfig {
     // TODO(ekrekr): add a notebook runtime field definition.
   }
 
+  message DataPreparationConfig {
+    // The name of the data preparation.
+    string name = 1;
+
+    // Targets of actions that this action is dependent on.
+    repeated Target dependency_targets = 2;
```
chtyim, line 523: You can just call it `dependencies` as the type is already Target, no need to repeat it in the field name.

protos/core.proto
fernst: It looks similar, it just uses BigQuery specific terminology. Data Preparation is based on a BigQuery Source/Destination, thus we defined the protos with BQ specific terminology.

protos/core.proto
fernst: Yes, the Data Preparation YAML file is formatted according to this proto. During processing, we will parse the data preparation YAML and build this proto object from the parse result.

fernst: The DAG may contain branches or nodes that join 2 other nodes. The DAG is structured in a way that a node always points into a Source:

- A BigQuery Table
- 1 previous node
- 2 previous nodes, which are combined using a SQL JOIN operation

Ekrekr (commented):
In general I'm happy with this now, but one main change: I think `DataPreparationDefinition`, and its children definitions should be moved to its own proto file:
* It removes confusion about why we have multiple targets that look identical (agreed that it does make sense for them to be different protos), and why these nodes are different to Dataform's "action" concept.
* You can copybara the internal proto out (or in), so that keeping them in sync is easier.
* It seems to me that you'll always want the data prep config and what's sent in the compiled graph to be identical.

Ekrekr (approved):

PR #1792, opened 2024-07-30
Fix notebook dependency targets

Ekrekr:
Some fixes are:
* Notebooks can properly be dependency targets of other targets, by hacking from "location" to "dataset" fields.
* Notebooks, views and incremental tables now correctly convert dependency targets to compiled graph targets.
* Better type safeness by extracting configs proto target conversion and action configs proto conversion to separate functions.


chemerisuk (approved):
LGTM

PR #1796, opened 2024-08-05
Regenerate configs reference with camelCase fields

Ekrekr:
Also add a comment about camel casing.

### DIFF
--- a/scripts/regenerate_docs
+++ b/scripts/regenerate_docs
@@ -1,12 +1,12 @@
 #!/bin/bash
 
-docker pull pseudomuto/protoc-gen-doc
+# docker pull pseudomuto/protoc-gen-doc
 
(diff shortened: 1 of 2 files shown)


scripts/regenerate_docs
```diff
@@ -1,8 +1,12 @@
 #!/bin/bash
 
-docker pull pseudomuto/protoc-gen-doc
+# docker pull pseudomuto/protoc-gen-doc
```
chemerisuk, line 3: Usually it's better to remove commented code. You can find it in history.

chemerisuk (approved):

scripts/regenerate_docs
Ekrekr: This code is still needed as part of the release process, but the exit code below forces me to remember to update to the forked documentation generation beforehand.

Unusual but effective!

PR #1802, opened 2024-08-08
Fix catchall errors

Ekrekr:
`e instanceof Error` here caused issues in internal sandbox compilation, and was also just a bad idea - this makes it safer


lewish (approved):

Ekrekr:
We're having weird networking issues with cloud build, but I've run the tests locally and they're fine, so I'm going to merge despite the failing cloud build.

Issue #1805, opened 2024-08-12
Dry run fails when table/view/assertion to be created does not exists


ashish10alex:
There are two possible errors relating to dry run: 
#### Tables / views
1. If the table created by a .sqlx file is not already materialised on BigQuery the dry run will fail due to dry run trying to `setMetadata` on an view/table that does not exist yet. Relevant lines of code
 
https://github.com/dataform-co/dataform/blob/54dfa9ce51b141124608f0b5e15fa01f714e093a/cli/api/commands/run.ts#L371

https://github.com/dataform-co/dataform/blob/54dfa9ce51b141124608f0b5e15fa01f714e093a/cli/api/dbadapters/bigquery.ts#L235-L239

In these lines metadata for the table/view that already does not exist will be undefined and trying to access `.schema` attribute of undefined results in the error shown below
 
![CleanShot 2024-08-12 at 19 16 41@2x](https://github.com/user-attachments/assets/ec1d64aa-2b65-4909-8d30-e81072a107ae)

-----

#### Assertion 

Assertion dry run will face if the table it is trying to create does not already exists. Possibly due to trying to run the following query in the current code base which tries to select from a object that does not exist yet

https://github.com/dataform-co/dataform/blob/54dfa9ce51b141124608f0b5e15fa01f714e093a/cli/api/dbadapters/execution_sql.ts#L177

----- 

#### Minimal example to reproduce

1. Add the following as a new .sqlx file to a dataform project 
```sql
config {
  type: "table",
  schema: "dataset_name",
  description: "any non empty description",
  tags: ["TEST_DRY_RUN"]
}


SELECT 1 AS A
UNION ALL
SELECT 1 AS B
```

To recreate this error for view/assertion type, the type can be replaced by either of the two

2. Run  
   ```bash
   dataform run --dry-run --tags="TEST_DRY_RUN"
   ```

dataform version: `3.0.2` and also tested by running it from source

St-Hakky:
I’m also encountering the same issue, so I wanted to check if there have been any updates on this.

Ekrekr:
Agreed, we shouldn't be attempting to set metadata on a dry run!

Any chance you can submit a patch yourself? https://github.com/dataform-co/dataform/blob/main/contributing.md

mkamysz:
Hey @Ekrekr, I opened another PR to resolve. this issue, seeing that the original fix has been reverted

PR #1806, opened 2024-08-12
fix(1805): dry run fails when table/view/assertion already does not exists

ashish10alex:
Solves https://github.com/dataform-co/dataform/issues/1805

1. **Tables/views:** For a view / table the metadata need not be set when doing a dry run only query validation is sufficient as metadata cannot be set if the object is not yet being materialised in BigQuery 
2. **Assertions**: No need to do `select sum(1) ...` on the view that will be created by the assertion. The query validation is sufficient to establish if the query that will be created is valid.

**Tests**

1. Dry run on assertions pass if the query is correct 
2. Dry run on assertion correctly fails if the query is not valid
3. Dry run on table/view passes if the query is valid 
4. Dry run on table/view correctly fails if the query is not valid 
5. `bazel test //core/...` passes


Ekrekr (approved):
This is super helpful, thanks for digging into this and making the PR!

PR #1808, opened 2024-08-13
Update output for Data Preparation processing

fernst:
Updated data preparation proto to only carry binary encoded data preparation contents after processing the data preparation definition.

### DIFF
--- a/core/actions/data_preparation.ts
+++ b/core/actions/data_preparation.ts
@@ -34,8 +34,8 @@ export class DataPreparation extends ActionBuilder<dataform.DataPreparation> {
     }
 
     config.filename = resolveActionsConfigFilename(config.filename, configPath);
-    const dataPreparationContents = nativeRequire(config.filename).asJson;
-    const dataPreparationDefinition = parseDataPreparationDefinitionJson(dataPreparationContents);
+    const dataPreparationAsJson = nativeRequire(config.filename).asJson;
+    const dataPreparationDefinition = parseDataPreparationDefinitionJson(dataPreparationAsJson);
     this.proto.dataPreparationContents = dataform.DataPreparationDefinition.encode(dataPreparationDefinition).finish();
 
     // Find targets
--- a/core/actions/data_preparation.ts
+++ b/core/actions/data_preparation.ts
@@ -36,7 +36,7 @@ export class DataPreparation extends ActionBuilder<dataform.DataPreparation> {
     config.filename = resolveActionsConfigFilename(config.filename, configPath);
     const dataPreparationContents = nativeRequire(config.filename).asJson;
     const dataPreparationDefinition = parseDataPreparationDefinitionJson(dataPreparationContents);
-    this.proto.dataPreparation = dataPreparationDefinition;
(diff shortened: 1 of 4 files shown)


core/actions/data_preparation.ts
```diff
@@ -36,7 +36,7 @@ export class DataPreparation extends ActionBuilder<dataform.DataPreparation> {
     config.filename = resolveActionsConfigFilename(config.filename, configPath);
     const dataPreparationContents = nativeRequire(config.filename).asJson;
     const dataPreparationDefinition = parseDataPreparationDefinitionJson(dataPreparationContents);
```
Ekrekr, line 38: Nit: the term `dataPreparationContents` is overloaded now, so should be renamed - something like `dataPreparationAsJson = ...`

packages/@dataform/core/BUILD
```diff
@@ -99,7 +93,6 @@ pkg_npm_tar(
         ":bundle_with_license",
         ":configs_proto_with_license",
         ":core_proto_with_license",
-        ":data_preparation_proto_with_license",
```
Ekrekr, line 102: You're still going to need this exported, for the decode step

Ekrekr (changes requested):

core/actions/data_preparation.ts
fernst: Updated

packages/@dataform/core/BUILD
fernst: I added this back as discussed.

protos/core.proto
```diff
@@ -268,7 +267,10 @@ message DataPreparation {
 
   bool disabled = 6;
 
-  DataPreparationDefinition data_preparation = 7;
+  reserved 7;
```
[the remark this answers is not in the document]
fernst: Updated

core/main_test.ts
```diff
@@ -945,62 +945,7 @@ nodes:
...
-                            }
-                          ]
-                        }
-                      }
-                    }
-                  }
-                }
-              ]
-            }
+            dataPreparationContents: "Cn4KBW5vZGUxEhASDgoDcHJqEgJkcxoDc3JjIlAiFgoUCgFhEgVJTlQ2NBoITlVMTEFCTEUqGyIZEhcKFQoBYRIGU1RSSU5HGghOVUxMQUJMRTIZChcKFQoBYRIGU1RSSU5HGghOVUxMQUJMRSoRCg8KA3ByahICZHMaBGRlc3Q=",
```
chtyim, line 948: This is probably generated by some external script? Might be good to include that script / instruction on how to update this string.

Ekrekr: Good shout - @fernst if you bring back what you put here before, and just call `dataform.DataPreparationDefinition.encode(` on it in the test part, then its generation becomes clear

Ekrekr:
Our tests are being flakey atm because of the `Invalid JWT Signature` error - if you rerun them they should go away

image

core/main_test.ts
fernst: Good point. I'll add this logic so it's obvious.

PR #1812, opened 2024-08-16
Update unit test's result messages

### LINKED ISSUE #1791: About error messages during unit tests
Hello, I conducted the following test:

Test table:
```sql
config {
    type: "table",
    schema: "hoge",
    name: "fuga"
}

select
  user_id
from
  ${ref("users")}
```

Test code:
```sql
config {
  type: "test",
  dataset: "fuga",
}

select 1 as user_id union all
select cast(null as int64) as user_id

input "users" {
  select 1 as user_id union all
  select 2 as user_id
}
```

Even though null is cast to int, the expected message should indicate a value difference as shown here:
https://github.com/dataform-co/dataform/blob/main/cli/api/commands/test.ts#L98

However, the actual message indicates a type difference as shown here:
https://github.com/dataform-co/dataform/blob/main/cli/api/commands/test.ts#L88

I believe this should be corrected. What are your thoughts?

If fixing the type inference is difficult, I would like to propose the following branching as a fix, though it may be somewhat redundant. Is it possible to submit a PR for this?

```typescript
      // Null value check
      if (expectedValue === null && actualValue !== null) {
        rowMessages.push(
          `For row ${i} and column "${column}": expected null, but saw "${actualValue}".`
        );
        break;
      }
      if (expectedValue !== null && actualValue === null) {
        rowMessages.push(
          `For row ${i} and column "${column}": expected "${expectedValue}", but saw null.`
        );
        break;
      }
```

spider-man-tm:
This PR is linked to the following issue:

close #1791 

## Current
Prepare the following table and its corresponding unit test.

Test table:
```sql
config {
    type: "table",
    schema: "hoge",
    name: "fuga"
}

select
  user_id
from
  ${ref("users")}
```

Test code:
```sql
config {
  type: "test",
  dataset: "fuga",
}

select 1 as user_id union all
select cast(null as int64) as user_id

input "users" {
  select 1 as user_id union all
  select 2 as user_id
}
```

Given that the expected block includes cast(null as int64), it would be more user-friendly to have the output display as follows:

```shell
For row 1 and column "user_id": expected null, but saw "2"
```

However, the current output is as follows, which can cause confusion for the user:

```shell
For row 1 and column "user_id": expected type "object", but saw type "number".
```

## Before Committing
- Conducted unit testing
- Checked the behavior of dataform cli

### DIFF
--- a/cli/api/commands/test.ts
+++ b/cli/api/commands/test.ts
@@ -83,6 +83,15 @@ async function runTest(
       const normalizedColumn = normalizeColumnName(column);
       const expectedValue = expectedResultRow[normalizedColumn];
       const actualValue = actualResultRow[normalizedColumn];
+      // Check if exactly one of the expected or actual values is null (XOR condition)
+      if ((expectedValue === null) !== (actualValue === null)) {
+        const expectedText = expectedValue === null ? 'null' : `"${expectedValue}"`;
+        const actualText = actualValue === null ? 'null' : `"${actualValue}"`;
+        rowMessages.push(
+          `For row ${i} and column "${column}": expected ${expectedText}, but saw ${actualText}.`
+        );
+        break;
+      }
       if (typeof expectedValue !== typeof actualValue) {
         rowMessages.push(
           `For row ${i} and column "${column}": expected type "${typeof expectedValue}", but saw type "${typeof actualValue}".`
--- a/cli/api/commands/test.ts
+++ b/cli/api/commands/test.ts
@@ -83,12 +83,16 @@ async function runTest(
       const normalizedColumn = normalizeColumnName(column);
       const expectedValue = expectedResultRow[normalizedColumn];
       const actualValue = actualResultRow[normalizedColumn];
-      // Check if exactly one of the expected or actual values is null (XOR condition)
-      if ((expectedValue === null) !== (actualValue === null)) {
-        const expectedText = expectedValue === null ? 'null' : `"${expectedValue}"`;
-        const actualText = actualValue === null ? 'null' : `"${actualValue}"`;
+      // Null value check
+      if (expectedValue === null && actualValue !== null) {
         rowMessages.push(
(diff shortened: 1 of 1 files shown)


cli/api/commands/test.ts
```diff
@@ -83,6 +83,15 @@ async function runTest(
...
       const actualValue = actualResultRow[normalizedColumn];
+      // Check if exactly one of the expected or actual values is null (XOR condition)
+      if ((expectedValue === null) !== (actualValue === null)) {
+        const expectedText = expectedValue === null ? 'null' : `"${expectedValue}"`;
+        const actualText = actualValue === null ? 'null' : `"${actualValue}"`;
+        rowMessages.push(
+          `For row ${i} and column "${column}": expected ${expectedText}, but saw ${actualText}.`
+        );
+        break;
+      }
```
Ekrekr, line 94: I much preferred your suggestion of

```ts
      // Null value check
      if (expectedValue === null && actualValue !== null) {
        rowMessages.push(
          `For row ${i} and column "${column}": expected null, but saw "${actualValue}".`
        );
        break;
      }
      if (expectedValue !== null && actualValue === null) {
        rowMessages.push(
          `For row ${i} and column "${column}": expected "${expectedValue}", but saw null.`
        );
        break;
      }
```

This is much clearer to follow the logic of!

Ekrekr (changes requested):

cli/api/commands/test.ts
spider-man-tm: You're right, I also think that keeping it simple makes it easier to understand. I've committed.
https://github.com/dataform-co/dataform/pull/1812/commits/d40e265322ba45ec6236f52df0d3f9e7e8d2cd24

spider-man-tm: @Ekrekr 
Is there a plan to merge this PR?

Ekrekr (approved):

Ekrekr:
I've run the tests, and all pass. Will merge, thanks for doing this!

PR #1818, opened 2024-08-27
Added support for compilation overrides in Data Preparations

fernst:


common/protos/index.ts
```diff
@@ -35,7 +35,8 @@ export enum VerifyProtoErrorBehaviour {
 export function verifyObjectMatchesProto<Proto>(
   protoType: IProtoClass<any, Proto>,
   object: object,
-  errorBehaviour: VerifyProtoErrorBehaviour = VerifyProtoErrorBehaviour.DEFAULT
+  errorBehaviour: VerifyProtoErrorBehaviour = VerifyProtoErrorBehaviour.DEFAULT,
+  skipUnknownFields = false
```
Ekrekr, line 39: I don't think we should add this parameter - when it is set to true, there is no verification done, so it essentially just does `dataform.dataprep.DataPreparation.create(dataPreparationAsObject)`

core/main_test.ts
```diff
@@ -881,6 +881,11 @@ actions:
       const dataPreparationYaml = `
 nodes:
 - id: node1
+  unknown1: val1
+  unknown2: 1234
+  unknown3: 
+    nested1: 1234
+    nesteda: qwerty
```
Ekrekr, line 888: I don't think it's worth breaking proto safety to allow fields like this.

I think what you're looking for is adding a map instead:

```proto
message DataPreparation {
  ...
    map<string, SomeStructure> some_structure_map_name = 1;
}
```

Ekrekr (changes requested):

core/actions/data_preparation.ts
```diff
@@ -128,6 +131,64 @@ function parseDataPreparationDefinitionJson(dataPreparationAsJson: {
...
+              applySessionDefaultsToTableReference(actionBuilder, destinationTable);
+        }
+      }
+
+    );
+
+  return resolvedDataPreparation;
+}
+
+function applySessionDefaultsToTableReference(
```
Ekrekr, line 171: To mirror the naming of `applySessionToTarget`, this would better be `applySessionToTableReference` instead

core/actions/data_preparation.ts
```diff
@@ -128,6 +131,64 @@ function parseDataPreparationDefinitionJson(dataPreparationAsJson: {
   }
 }
 
+function resolveSourcesAndDestinations(
```
Ekrekr, line 134: Nit: rename to something like `applySessionToDataPreparationContents`

Ekrekr (changes requested):

core/actions/data_preparation.ts
fernst: Done.

core/actions/data_preparation.ts
fernst: Done.

core/actions/data_preparation.ts
```diff
@@ -128,6 +131,64 @@ function parseDataPreparationDefinitionJson(dataPreparationAsJson: {
...
+  }
+  const resolvedTarget =
+      actionBuilder.applySessionToTarget(
+          dataform.Target.create(target),
+          actionBuilder.session.projectConfig)
+  return dataform.dataprep.TableReference.create({
+    project: resolvedTarget.database,
+    dataset: resolvedTarget.schema,
+    table: resolvedTarget.name
+  })
```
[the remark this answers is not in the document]
fernst: Updated.

Ekrekr (approved):

PR #1826, opened 2024-09-04
Add support for Data Preparation Defaults during compilation.

fernst:
Data preparation defaults are applied AFTER compilation overrides and project defaults. This allows dataform defaults/overrides to take precedence over data preparaiton defaults.

Also made small improvements on data preparation tests.

### DIFF
--- a/core/actions/data_preparation.ts
+++ b/core/actions/data_preparation.ts
@@ -38,12 +38,23 @@ export class DataPreparation extends ActionBuilder<dataform.DataPreparation> {
 
     // Find targets
     const targets = getTargets(dataPreparationDefinition);
-    this.proto.targets = targets.map(target =>
-      this.applySessionToTarget(target, session.projectConfig, config.filename, true)
-    );
-    this.proto.canonicalTargets = targets.map(target =>
-      this.applySessionToTarget(target, session.canonicalProjectConfig)
-    );
+    this.proto.targets = targets
+        // Apply compilation overrides to target
+        .map(target =>
+            this.applySessionToTarget(target, session.projectConfig, config.filename, true)
+            // Apply data preparation defaults
+        ).map(target =>
+            applyDataPreparationDefaultsToTarget(dataPreparationDefinition, target)
+        );
+    this.proto.canonicalTargets = targets
+        // Apply canonical project config to target
+        .map(target =>
+            this.applySessionToTarget(target, session.canonicalProjectConfig)
+        )
+        // Apply data preparation defaults
+        .map(target =>
+            applyDataPreparationDefaultsToTarget(dataPreparationDefinition, target)
+        );
 
     // Resolve all table references with compilation overrides and encode resolved proto instance
     const resolvedDefinition = applySessionToDataPreparationContents(this, dataPreparationDefinition);
@@ -141,7 +152,11 @@ function applySessionToDataPreparationContents(
   const errorTable = definition.configuration?.errorTable;
   if (errorTable) {
     resolvedDataPreparation.configuration.errorTable =
-        applySessionToTableReference(actionBuilder, errorTable);
+        applySessionAndDefaultsToTableReference(
+            actionBuilder,
+            definition,
+            errorTable
+        );
   }
 
   // Loop through all nodes and resolve the compilation overrides for
@@ -152,20 +167,44 @@ function applySessionToDataPreparationContents(
         const sourceTable = node.source.table;
         if (sourceTable) {
           resolvedDataPreparation.nodes[index].source.table =
-              applySessionToTableReference(actionBuilder, sourceTable);
+              applySessionAndDefaultsToTableReference(
+                  actionBuilder,
+                  definition,
+                  sourceTable
+              );
         }
 
         // Resolve destination tables, if set.
         const destinationTable = node.destination?.table;
         if (destinationTable) {
           resolvedDataPreparation.nodes[index].destination.table =
-              applySessionToTableReference(actionBuilder, destinationTable);
+              applySessionAndDefaultsToTableReference(
+                  actionBuilder,
+                  definition,
+                  destinationTable
+              );
         }
       }
 
     );
 
   return resolvedDataPreparation;
+}
+
+function applySessionAndDefaultsToTableReference(
+    actionBuilder: ActionBuilder<dataform.DataPreparation>,
+    definition: dataform.dataprep.DataPreparation,
+    tableReference: dataform.dataprep.ITableReference
+): dataform.dataprep.ITableReference {
+  // Apply session compilation overrides and afterwards
+  // apply data preparation defaults if needed
+  return applyDataPreparationDefaultsToTableReference(
+      definition,
+      applySessionToTableReference(
+          actionBuilder,
+          tableReference
+      )
+  );
 }
 
 function applySessionToTableReference(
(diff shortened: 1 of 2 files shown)


core/actions/data_preparation.ts
```diff
@@ -168,6 +191,22 @@ function applySessionToDataPreparationContents(
   return resolvedDataPreparation;
 }
 
+function applySessionAndDefaultsToTableReference(
```
Ekrekr, line 194: Why isn't this a private method on the class, rather than passing the action builder as a parameter?

core/actions/data_preparation.ts
```diff
@@ -194,6 +233,46 @@ function applySessionToTableReference(
...
+      tableReference.project = defaults.project;
+    }
+    // Apply dataset default if needed
+    if (defaults.dataset && !tableReference.dataset) {
+      tableReference.dataset = defaults.dataset;
+    }
+  }
+
+  return tableReference
+}
```
Ekrekr, line 254: Product question: why are we allowing two places for defaults of the same properties? It seems overengineering at this stage not just to use the values from the workflow settings file

core/actions/data_preparation.ts
```diff
@@ -38,12 +38,23 @@ export class DataPreparation extends ActionBuilder<dataform.DataPreparation> {
 
     // Find targets
```
Ekrekr, line 39: Nit: remove unhelpful comment

core/actions/data_preparation.ts
```diff
@@ -38,12 +38,23 @@ export class DataPreparation extends ActionBuilder<dataform.DataPreparation> {
...
-      this.applySessionToTarget(target, session.projectConfig, config.filename, true)
-    );
-    this.proto.canonicalTargets = targets.map(target =>
-      this.applySessionToTarget(target, session.canonicalProjectConfig)
-    );
+    this.proto.targets = targets
+        // Apply compilation overrides to target
+        .map(target =>
+            this.applySessionToTarget(target, session.projectConfig, config.filename, true)
+            // Apply data preparation defaults
```
Ekrekr, line 45: Nit: comment is just stating the name of the function, remove

core/main_test.ts
```diff
@@ -21,6 +21,10 @@ defaultDataset: defaultDataset
 defaultLocation: US
 `;
 
+const VALID_WORKFLOW_SETTINGS_YAML_WITHOUT_PROJECT_DATASET_DEFAULTS = `
+defaultLocation: US
+`;
+
```
Ekrekr, line 27: Nit: inline this to the test, it's only used once and the name is more complex than the string

core/main_test.ts
```diff
@@ -1097,6 +1124,13 @@ nodes:
       const result = runMainInVm(coreExecutionRequestFromPath(projectDir));
```
Ekrekr, line 1125: This `resolvedYaml` would be better replaced with:

```ts
const resolvedDataPreparation = dataform.dataprep.DataPreparation.create({
  configuration: {
 ...
})

      const expectedEncodedDataPreparation = encode64(
          dataform.dataprep.DataPreparation,
          resolvedDataPreparation
      );

      expect(
          asPlainObject(resolvedDataPreparation)).deep.equals(
          asPlainObject(dataform.dataprep.DataPreparation.decode(
                  result.compile.compiledGraph.dataPreparations[0].dataPreparationContents
              )
          )
      )

      expect(asPlainObject(result.compile.compiledGraph.dataPreparations)).deep.equals(
          asPlainObject([
            {
...
              dataPreparationContents: expectedEncodedDataPreparation
            }
          ])
      );
```

core/main_test.ts
```diff
@@ -1131,6 +1165,182 @@ nodes:
...
+      field:
+      - name: a
+        type: INT64
+        mode: NULLABLE
+    destinationGenerated:
+      schema:
+        field:
+        - name: a
+          type: STRING
+          mode: NULLABLE
```
Ekrekr, line 1220: Nit: it looks like you only need one node here to cover the test surface, so this YAML file is longer than it needs to be. Suggestion: Remove node2, add a destinationGenerated field to node1.

core/main_test.ts
```diff
@@ -1131,6 +1165,182 @@ nodes:
...
+`
+
+      fs.writeFileSync(
+          path.join(projectDir, "definitions/data_preparation.yaml"),
+          dataPreparationYaml
+      );
+
+      const resolvedYaml = `
```
Ekrekr, line 1228: Same suggestion as above

Ekrekr (changes requested):
`applySessionToTarget`  is unused, and this only needs `applySessionAndDataPreparationDefaultsToTarget` and `applySessionAndDataPreparationDefaultsToTablereference` now.

Proposal:
* Rename `applySessionAndDataPreparationDefaultsToTarget` to `applyDefaultsToTarget`
* Rename `applySessionAndDataPreparationDefaultsToTablereference` to `applyDefaultsToTableReference`.
* Make them both private methods of the dataprep action builder.
* Only call those two methods, avoid the nested calling.

core/main_test.ts
fernst: Simplified into a single node.

core/main_test.ts
fernst: Done.

core/actions/data_preparation.ts
fernst: Done.

core/actions/data_preparation.ts
fernst: Done.

core/actions/data_preparation.ts
```diff
@@ -108,107 +114,168 @@ export class DataPreparation extends ActionBuilder<dataform.DataPreparation> {
...
-        dataPreparationAsJson as {
-          [key: string]: any;
-        },
-        VerifyProtoErrorBehaviour.SHOW_DOCS_LINK
-      )
-    );
-  } catch (e) {
-    if (e instanceof ReferenceError) {
-      throw ReferenceError(`Data Preparation parsing error: ${e.message}`);
+  private parseDataPreparationDefinitionJson(dataPreparationAsJson: {
```
Ekrekr, line 118: To match the naming convention of other action handlers, `verifyDataPreparationDefinition` would be better

core/actions/data_preparation.ts
```diff
@@ -108,107 +114,168 @@ export class DataPreparation extends ActionBuilder<dataform.DataPreparation> {
...
-          dataform.Target.create(target),
-          actionBuilder.session.projectConfig)
-  // Convert resolved target into a Data Preparation Table Reference
-  const resolvedTableReference = dataform.dataprep.TableReference.create({
-    table: resolvedTarget.name
-  });
-  if (resolvedTarget.database) {
-    resolvedTableReference.project = resolvedTarget.database;
+
+  private applySessionAndDefaultsToTableReference(
```
Ekrekr, line 184: See comment at top of previous review, about renaming and simplifying these

core/actions/data_preparation.ts
```diff
@@ -34,19 +34,25 @@ export class DataPreparation extends ActionBuilder<dataform.DataPreparation> {
...
-    this.proto.canonicalTargets = targets.map(target =>
-      this.applySessionToTarget(target, session.canonicalProjectConfig)
-    );
+    const targets = this.getTargets(dataPreparationDefinition);
+    this.proto.targets = targets
+        .map(target =>
+            this.applySessionToTarget(target, session.projectConfig, config.filename, true)
+        ).map(target =>
+            this.applyDataPreparationDefaultsToTarget(dataPreparationDefinition, target)
+        );
```
Ekrekr, line 45: I think this order should be reversed, against your PR description.

I think defaults should be applied at the level of granularity:

Workflow Setting defaults > Data Preparation defaults > explicit table targets

This way you can have two different data preparations in the same project have different default values within themselves, rather than always having the workflow settings override them

core/main_test.ts
```diff
@@ -1025,44 +1048,170 @@ nodes:
...
+                      name: "a",
+                      type: "STRING",
+                      mode: "NULLABLE"
+                    }
+                  ]
+                }
+              }
+            }
+          }
+        ]
```
Ekrekr, line 1142: Similarly, is this node actually needed for the test?

core/main_test.ts
```diff
@@ -1025,44 +1048,170 @@ nodes:
...
+              },
+              outputSchema: {
+                field: [
+                  {
+                    name: "a",
+                    type: "INT64",
+                    mode: "NULLABLE"
+                  }
+                ]
+              }
```
Ekrekr, line 1091: These parts of the proto don't contribute to the test, so let's remove them!

PR #1833, opened 2024-09-06
MOSS Cap1

Tuseeq1:
## Summary:

In order to use GCB for all building and publishing needs, new `cloudbuild-*.yaml` files are added that will be used by triggers in our GCP project. Moreover,  `yarn.lock` file has been updated to pull dependencies from google's verified sources.

## Changes:
* Updated `yarn.lock` to change dependencies location to google npm package manager.
* Removed `cloudbuild.yaml` file and replaced with following new build files.
* Added `cloudbuild-publish.yaml`. This build will get triggered when new code is pushed to main such that the `version.bzl` file has been updated i-e: a new version is ready for release. This will build and published to updated packages to NPM.
* Added `cloudbuild-test.yaml`. This will be trigger when a new PR is created or code is pushed to main such that the version.bzl is not changed. This will run tests.
* Added `cloudbuild-version.yaml`. This will be triggered on 1st of every month. This will update the version number in `version.bzl` in a new branch and create a PR to main.
* Added `scripts/update_version`. The previous step uses this script to update the patch version number.


cloudbuild-publish.yaml
```diff
@@ -0,0 +1,33 @@
+steps:
+  - name: 'gcr.io/cloud-builders/gcloud'
+    script: |
+      #!/usr/bin/env bash
+      set -e
+      REPO_TOKEN="$(gcloud auth print-access-token)"
```
Ekrekr, line 6: Why can't this secret be handled the same way as `NPM_TOKEN`, and be stored in secret manager?

cloudbuild-publish.yaml
```diff
@@ -0,0 +1,33 @@
...
+      cat > ~/.npmrc <<EOL
+      //registry.npmjs.org/:_authToken=$NPM_TOKEN"
+      registry=https://us-npm.pkg.dev/artifact-foundry-prod/ah-3p-staging-npm/
+      //us-npm.pkg.dev/artifact-foundry-prod/ah-3p-staging-npm/:always-auth=true
+      //us-npm.pkg.dev/artifact-foundry-prod/ah-3p-staging-npm/:_authToken=${REPO_TOKEN}
+      EOL
```
Ekrekr, line 12: The `npmrc` file can be a real file rather than doing this, and you can specify to use the one in the repo `--userconfig=./.npmrc`. This avoids having to cat contents of files as part of build scripts, which is a bit weird

cloudbuild-publish.yaml
```diff
@@ -0,0 +1,33 @@
...
+    secretEnv: ['NPM_TOKEN']
+  - name: gcr.io/cloud-builders/bazel:5.4.0
+    script: |
+      #!/usr/bin/env bash
+      set -e
+      bazel run @nodejs//:yarn config set registry https://us-npm.pkg.dev/artifact-foundry-prod/ah-3p-staging-npm/
+      ./scripts/publish
```
Ekrekr, line 19: Presumably we should only ever run the publish script when using the airlock registry from now on. So we should just update `./scripts/publish` to do all of this

cloudbuild-publish.yaml
```diff
@@ -0,0 +1,33 @@
...
+artifacts:
+  npmPackages:
+  - repository: 'https://us-central1-npm.pkg.dev/dataform-open-source/dataform-open-source'
+    packagePath: './bazel-bin/packages/@dataform/cli'
+  - repository: 'https://us-central1-npm.pkg.dev/dataform-open-source/dataform-open-source'
+    packagePath: './bazel-bin/packages/@dataform/core'
+options:
+  machineType: E2_HIGHCPU_8
+  requestedVerifyOption: VERIFIED
+timeout: 3600s
```
Ekrekr, line 33: nit: newline at end of file

cloudbuild-test.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - name: 'gcr.io/cloud-builders/gcloud'
+    script: |
+      #!/usr/bin/env bash
+      set -e
+      REPO_TOKEN="$(gcloud auth print-access-token)"
+      cat > ~/.npmrc <<EOL
+      registry=https://us-npm.pkg.dev/artifact-foundry-prod/ah-3p-staging-npm/
+      //us-npm.pkg.dev/artifact-foundry-prod/ah-3p-staging-npm/:always-auth=true
+      //us-npm.pkg.dev/artifact-foundry-prod/ah-3p-staging-npm/:_authToken=${REPO_TOKEN}
+      EOL
```
Ekrekr, line 11: Similarly with this, we should really be reading from the same npmrc file for both testing and publishing

cloudbuild-version.yaml
```diff
@@ -0,0 +1,35 @@
...
+    export git_branch_name=npm_veriosn_$(cat version.bzl | grep DF_VERSION | awk '{ print $3 }' | sed "s/\"//g")
+    echo "Create new branch $git_branch_name..."
+    git checkout -b $git_branch_name
+    echo "Update version..."
+    ./scripts/update_version
+    git add version.bzl
+    git commit -m "Update the npm package version"
+    echo "Push changes to remote..."
+    git push origin $git_branch_name
+    echo $git_branch_name > git_branch_name.txt
```
Ekrekr, line 24: Let's put this in an actual bash script, same as the others!

cloudbuild-version.yaml
```diff
@@ -0,0 +1,35 @@
+steps:
+- name: gcr.io/cloud-builders/gcloud
+  entrypoint: 'bash'
+  args: [ '-c', "gcloud secrets versions access latest --secret=github-token-access --format='get(payload.data)' | tr '_-' '/+' | base64 -d > token.txt" ]
```
Ekrekr, line 4: Add a comment about why we do this plz

Ekrekr (changes requested):
This fails tests currently, I think you can have multiple options for which cloud build file to use for testing. That would prevent merging this and having tests broken until it is updated

image

cloudbuild-publish.yaml
Tuseeq1: This generates fresh token every time with service account because a normal token expires after 24 hours. 
NPM_TOKEN is non expiry token we get from npm account.

cloudbuild-publish.yaml
Tuseeq1: I tried having a local one. somehow the `bazel yarn install` didn`t work with it. It was a bit weird because `yarn config ` showed the correct information.

Tuseeq1: Moreover, this is the method suggested by Airlock team.

cloudbuild-publish.yaml
Tuseeq1: 👍

cloudbuild-publish.yaml
Tuseeq1: 👍

cloudbuild-test.yaml
Tuseeq1: Each build have its own environment. token will be generated separately for each build. I am not sure what do you mean here by having same .npmrc file

cloudbuild-version.yaml
Tuseeq1: If we put this in a bash script GCB is unable to replace env variables such as  `$_GITHUB_USER`.

cloudbuild-version.yaml
Tuseeq1: 👍

cloudbuild-test.yaml
Ekrekr: If there are different files, then they're likely to diverge. Which means we would be a different setup for testing and releasing - which makes it more likely we're going to publish the wrong thing at some point

cloudbuild-test.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - name: gcr.io/cloud-builders/bazel:5.4.0
+    script: |
+      #!/usr/bin/env bash
+      set -e
+      bazel run @nodejs//:yarn config set registry https://us-npm.pkg.dev/artifact-foundry-prod/ah-3p-staging-npm/
+      ./scripts/run_tests
```
Ekrekr, line 17: Does the `./scripts/run` script still work when run locally? Or does it have to use this registry, now that the yarn lock has been updated?

cloudbuild-version.yaml
Ekrekr: Thanks - I'm also curious about the `tr '_-' '/+'` bit

cloudbuild-version.yaml
Ekrekr: Are you sure? If they're env varaibles, then they should be available in the whole bash context.

Alternatively, it can be passed in to the script as an arg.

It will make it much easier to modify and update the script in future, rather than having to run it in cloud build every time

yarn.lock
```diff
@@ -2,71 +2,55 @@
...
-  dependencies:
-    "@babel/highlight" "^7.0.0"
-
-"@babel/code-frame@^7.8.3":
+"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.8.3":
   version "7.8.3"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.8.3.tgz#33e25903d7481181534e12ec0a25f16b6fcf419e"
-  integrity "sha1-M+JZA9dIEYFTThLsCiXxa2/PQZ4= sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g=="
+  resolved "https://us-npm.pkg.dev/artifact-foundry-prod/ah-3p-staging-npm/@babel/code-frame/-/code-frame-7.8.3.tgz"
+  integrity "sha1-M+JZA9dIEYFTThLsCiXxa2/PQZ4=sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g== sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g=="
```
Ekrekr, line 8: What is the process for installing new packages now, or bumping versions? Does `yarn install x@1` still work?

cloudbuild-publish.yaml
Ekrekr: We're going to swap to `rules_js` soon, it will probably work then. Let's update it at that point.

scripts/update_version
```diff
@@ -0,0 +1,29 @@
...
+# Extract the current version from the file.
+current_version=$(grep -oP '(?<=DF_VERSION = ")[^"]+' "$file_path")
+
+# Split the version into its components.
+IFS='.-' read -r major minor patch label <<< "$current_version"
+
+# Increment the patch version.
+((patch++))
+
+# Toggle the label presence.
```
Ekrekr, line 16:
```suggestion
# If there's a label, add a hyphen before it.
```

cloudbuild-publish.yaml
Ekrekr: Fair enough!

cloudbuild-publish.yaml
```diff
@@ -0,0 +1,31 @@
...
+      //registry.npmjs.org/:_authToken=$NPM_TOKEN"
+      registry=https://us-npm.pkg.dev/artifact-foundry-prod/ah-3p-staging-npm/
+      //us-npm.pkg.dev/artifact-foundry-prod/ah-3p-staging-npm/:always-auth=true
+      //us-npm.pkg.dev/artifact-foundry-prod/ah-3p-staging-npm/:_authToken=${REPO_TOKEN}
+      EOL
+    secretEnv: ['NPM_TOKEN']
+  - name: gcr.io/cloud-builders/bazel:5.4.0
+    script: |
+      #!/usr/bin/env bash
+      ./scripts/publish
```
Ekrekr, line 17: Does this script still work, without having to run it in cloud build?

Ekrekr (changes requested):

cloudbuild-test.yaml
Tuseeq1: It have to use the new registry and local tests would not be possible. This is the discussion we had when we decided to keep single yarn.lock file. Ideal, user can use npm/yarn to update yarn.lock to point to yarnpkg.com registry and then run tests.

yarn.lock
Tuseeq1: no.

cloudbuild-publish.yaml
Tuseeq1: Yes. if you have collaboration access to npm.

scripts/update_version
Tuseeq1: 👍

cloudbuild-test.yaml
Tuseeq1: We can leave it for now. Also these are slightly different files. for publish we need NPM token for testing we do not. We can create a single script for creating both in future may be.

cloudbuild-test.yaml
Ekrekr: Wait, does this mean that local development isn't possible either? Is it no longer possible to run `yarn install` after cloning the repo? We need a workaround if this is the case!

scripts/create_gh_pr
```diff
@@ -0,0 +1,21 @@
...
+
+echo "Update version..."
+./scripts/update_version
+
+git add version.bzl
+git commit -m "Update the npm package version"
+
+echo "Push changes to remote..."
+git push origin $git_branch_name
+echo $git_branch_name > git_branch_name.txt
```
Ekrekr, line 21: newline

cloudbuild-test.yaml
Tuseeq1: Yes. Local development is no longer possible. To run local development, users will have to delete yarn.lock file and run install again. 

One workaround i can think about is to keep both versions of yarn.lock files. something like yarn.lock and yarn-local.lock. and users can change name before running yarn install. However, I do not support this solution as this will become problem very quickly trying to keep both files in sync. 

For authorization error you need o update the .npmrc file to same as the once created by `cloudbuild-test.yaml`

Ekrekr: So we definitely need to be able to keep developing... not quite sure why you think it would be OK to merge this.

Also ideally it needs to be done in a way that uses the same dependencies as when airlock is called

Some potential solutions:
* Have the tests run a replace of `registry.yarnpkg.com/` with `us-npm.pkg.dev/artifact-foundry-prod/ah-3p-staging-npm/` in the lockfile at test / publish time.
* Find a way for us to be able to access the airlock registry from our cloudtop machines.
* (We shouldn't do this due to non-determinism of installs) not write a yarn lockfile, and generate it at install

Ekrekr (changes requested):

yarn.lock
Ekrekr: Ok, what is the process?

cloudbuild-test.yaml
Ekrekr: This has now been resolved by the addition of a `./scripts/run_tests_local`

@Tuseeq1  could you rename `./scripts/run_tests_local` to `./scripts/run_tests`, and `./scripts/run_tests` to `./scripts/run_tests_on_cloudbuild`? This way we don't need to update the internal development guide.

yarn.lock
Tuseeq1: ideally.

bazel run @nodejs//:yarn config set registry https://us-npm.pkg.dev/artifact-foundry-prod/ah-3p-staging-npm/
bazel run @nodejs//:yarn add x@1.

Ekrekr: It would be good to document this, right? https://github.com/dataform-co/dataform/blob/main/contributing.md#add-new-npm-dependencies

Ekrekr (approved):

Issue #1835, opened 2024-09-10
BUG view + assertion uniqueKey on dataform v3.0.2


AntoineGiraud:
Hi,

Since dataform v3.0.2
With a view, and a `uniqueKey` assertion, it doesn't take my column name but puts `TableAssertionsConfig` instead in the SQL assertion

after some trial, it works fine :
- as is on v3.0.0
- on a table + uniqueKey & v3.0.2
- on a view + uniqueKey**s** & v3.0.2

At your service to provide more details on this issue :)
Antoine

recap img on the subject
![image](https://github.com/user-attachments/assets/4dc83869-d216-4a2b-95dc-632417c05d61)

alejandro-delgado-garcia:
it also does the same thing with incremental tables

AntoineGiraud:
Resolved with [3.0.7](https://github.com/dataform-co/dataform/releases/tag/3.0.7) 🥳 Thanks !

Issue #1804, opened 2024-08-12
uniqueKey assertions generate invalid SQL for incremental tables in v3.0.2


stevenleggdfe:
To reproduce, put the .sqlx file below in a Dataform repo with dataform-core v3.0.2 installed:

```
config {
  type: "incremental",
  assertions: {
    uniqueKey: ["foo"]
  }
}

SELECT 1 as foo, 2 AS bar
```

This generates the SQL below for the assertion (view):
```
SELECT
  *
FROM (
  SELECT
    TableAssertionsConfig,
    COUNT(1) AS index_row_count
  FROM `project-name.dataform.uniquekey_bug_test`
  GROUP BY TableAssertionsConfig
  ) AS data
WHERE index_row_count > 1
```

This in turn generates the compilation error "Unrecognized name: TableAssertionsConfig at [6:5]."
 
This bug does *not* occur when:
- An earlier version of dataform-core is used - function is as expected in v3.0.0 and not visible because of a different bug (now fixed) in v3.0.1.
- The uniqueKey assertion is applied to a non-incremental table
- ```uniqueKeys: [["foo"]]``` is used instead of ```uniqueKey: ["foo"]``` - this is a possible workaround if others are experiencing this, but not a fix I feel as ```uniqueKey``` is supported and documented here: https://cloud.google.com/dataform/docs/assertions#built-in .

mattimoller:
The same issue seems to be present for models of type `view`

blundin:
I'm seeing this on models of type `view` as well.

Ekrekr:
Thank you so much for this fix! It would have been approved earlier, but there's been a lot on.

It looks good - I'll merge main in with this and make sure the remote tests pass before merging.

Fixes https://github.com/dataform-co/dataform/issues/1835
Fixes https://github.com/dataform-co/dataform/issues/1804

Ceridan:
The issue was fixed by #1836

PR #1836, opened 2024-09-10
Fix `uniqueKey` assertions for views, incremental tables

bmagyarkuti:
# The Bug

This PR fixes a bug where `uniqueKey` assertions on views and incremental tables caused `dataform run` to exit with a query error. `uniqueKey` assertions on tables, as well as all `uniqueKeys` assertions still worked. Affected versions: `3.0.1`, `3.0.2`. I don't know whether executions on GCP were also affected.

 ## Steps to reproduce
1) Create an empty dataform project using version `3.0.1` or `3.0.2` of `@dataform/core`.
2) Add the following definition:
```
config {
  type: "view",
  name: "my table",
  assertions: {
    uniqueKey: ["id"]
  }
}

SELECT 1 as id
```
3) Execute `dataform run`. Observe the following error message:
```
Table created:  dataform.my table [view]
Assertion failed:  dataform_assertions.dataform_my table_assertions_uniqueKey_0
  >
  >       create or replace view `mbarna-mongo-to-bq-dataflow.dataform_assertions.dataform_my table_assertions_uniqueKey_0` as
  > SELECT
  >   *
  > FROM (
  >   SELECT
  >     TableAssertionsConfig,
  >     COUNT(1) AS index_row_count
  >   FROM `mbarna-mongo-to-bq-dataflow.dataform.my table`
  >   GROUP BY TableAssertionsConfig
  >   ) AS data
  > WHERE index_row_count > 1
  >
  bigquery error: Unrecognized name: TableAssertionsConfig at [7:5]
```

# The Fix

I don't fully understand what happened. The query complains about an undefined column `TableAssertionsConfig`, and the newly refactored implementations for `view` and `incremental_table` contain this hardcoded string for `uniqueKey` assertions. I'm not sure what the hardcoded string is meant to be referring to. Just replacing the hardcoded string with the user-supplied column names fixes the issue.

# The Tests
The suite in `core/main_test.ts` tested assertions for `views`, `tables` and `incremental_table`s, for the following assertions: `uniqueKeys`, `nonNull`, `rowConditions`. `uniqueKey` assertions were untested, I assume because they're incompatible with a `uniqueKeys` assertion. I extended the test suite to also include tests for `uniqueKey` assertions, for all three tested table kinds. I worked around the incompatibility by introducing a second sample file. Apart from the different `assertions` block, the second file's contents are identical to the contents of the first file.

This has confirmed the bug, as well as the fix.

I nominate @Ekrekr as reviewer.

### DIFF
--- a/core/actions/incremental_table.ts
+++ b/core/actions/incremental_table.ts
@@ -308,7 +308,7 @@ export class IncrementalTable extends ActionBuilder<dataform.Table> {
     if (!!assertions.uniqueKey?.length) {
       uniqueKeys = [
         dataform.ActionConfig.TableAssertionsConfig.UniqueKey.create({
-          uniqueKey: ["TableAssertionsConfig"]
+          uniqueKey: assertions.uniqueKey
         })
       ];
     }
--- a/core/actions/view.ts
+++ b/core/actions/view.ts
@@ -271,7 +271,7 @@ export class View extends ActionBuilder<dataform.Table> {
     if (!!assertions.uniqueKey?.length) {
       uniqueKeys = [
         dataform.ActionConfig.TableAssertionsConfig.UniqueKey.create({
-          uniqueKey: ["TableAssertionsConfig"]
+          uniqueKey: assertions.uniqueKey
         })
       ];
     }
--- a/core/main_test.ts
+++ b/core/main_test.ts
@@ -1690,11 +1690,15 @@ actions:
     };
 
     const exampleBuiltInAssertions = {
-      inputSqlxConfigBlock: `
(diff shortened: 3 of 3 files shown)


Ekrekr:
Thank you so much for this fix! It would have been approved earlier, but there's been a lot on.

It looks good - I'll merge main in with this and make sure the remote tests pass before merging.

Fixes https://github.com/dataform-co/dataform/issues/1835
Fixes https://github.com/dataform-co/dataform/issues/1804

Ekrekr (approved):

PR #1839, opened 2024-09-11
Update Data Preparation processor to remove proto dependency. 

fernst:
The data preparation in the compiled graph now contains the resolved YAML after compiling.

I will remove the proto reference in a follow up PR.


Ekrekr:
(Fix tests before merge plz)

PR #1845, opened 2024-09-13
Correct permission and branch name

Tuseeq1:
Correct permissions to run bash scripts for create_gh_pr and update_version.
Fixed name of the branch that is used for PR creation in update_version script to have the same branch name as the npm package version


kolina (approved):
Update PR title please because it may not be very clear what permissions and branch name are corrected.

PR #1846, opened 2024-09-14
Improve error messages when workflow settings have non-string values for default project or dataset

kolina:
Currently if `defautProject` or `defaultDataset` have an integer type in `workflow_settings.yaml`, a compilation will throw a not very obvious error like `t.target.schema.includes is not a function`. I've added a validation of these fields to be strings (if not, an explicit error is thrown).

### DIFF
--- a/core/main_test.ts
+++ b/core/main_test.ts
@@ -468,6 +468,38 @@ someKey: and an extra: colon
       );
     });
 
+    test(`main fails when workflow settings have non-string default dataset`, () => {
+      const projectDir = tmpDirFixture.createNewTmpDir();
+      fs.writeFileSync(
+        path.join(projectDir, "workflow_settings.yaml"),
+        `
+defaultProject: defaultProject
+defaultDataset: 12345
+defaultLocation: US
+`
+      );
+
+      expect(() => runMainInVm(coreExecutionRequestFromPath(projectDir))).to.throw(
+        "Default schema should be string."
+      );
+    });
+
+    test(`main fails when workflow settings have non-string default project`, () => {
+      const projectDir = tmpDirFixture.createNewTmpDir();
+      fs.writeFileSync(
+        path.join(projectDir, "workflow_settings.yaml"),
+        `
+defaultProject: 12345
+defaultDataset: defaultDataset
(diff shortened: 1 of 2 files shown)


Ekrekr (changes requested):
> Currently if `defautProject` or `defaultDataset` have an integer type in `workflow_settings.yaml`, a compilation will throw a not very obvious error like `t.target.schema.includes is not a function`. I've added a validation of these fields to be strings (if not, an explicit error is thrown).

I think there should be a better fix for this:
* workflow_settings.yaml is a version of the workflow settings proto https://github.com/dataform-co/dataform/blob/05341b7c2e8f6e55a7002ed29b4ff4a7e2cbf12b/protos/configs.proto#L13
* We should be catching this during the proto validation https://github.com/dataform-co/dataform/blob/05341b7c2e8f6e55a7002ed29b4ff4a7e2cbf12b/common/protos/index.ts#L53

I'm not sure why we're not catching it already - but fixing the proto validation would be the most extensible solution!

kolina:
> * We should be catching this during the proto validation https://github.com/dataform-co/dataform/blob/05341b7c2e8f6e55a7002ed29b4ff4a7e2cbf12b/common/protos/index.ts#L53

I also hoped that it should have been the case, but unfortunately this code https://github.com/dataform-co/dataform/blob/05341b7c2e8f6e55a7002ed29b4ff4a7e2cbf12b/common/protos/index.ts#L45-L46 doesn't validate types and actually allows to create `protoCastObject` where `defaultDataset` has an integer type and this check later doesn't work https://github.com/dataform-co/dataform/blob/05341b7c2e8f6e55a7002ed29b4ff4a7e2cbf12b/common/protos/index.ts#L53

The thing that actually verifies types is `protoType.verify()`. WDYT about using it?

Ekrekr:
`.verify()` is a lie, it doesn't actually do anything - there are no situations where the call will actually fail.

apilaskowski:
@Ekrekr @kolina is this item dead? can we close this PR?

kolina:
I'll check later this week if it's possible to resurrect this, let's not close it for now

PR #1852, opened 2024-09-30
Publish a new NPM versions

Tuseeq1:
Updating NPM package version to 3.0.3


diasdauletov (approved):
I think then you need to ask lewis|elias to publish the new package

PR #1853, opened 2024-09-30
Branch name from master to main

Tuseeq1:
In GCB the default branch name used is master while we check on main to publish. This also makes more sense since we use main everywhere else as default branch.

### DIFF
--- a/scripts/publish
+++ b/scripts/publish
@@ -4,6 +4,11 @@ set -e
 # TODO(ekrekr): re-enable once fork is merged to upstread.
 # ./scripts/regenerate_docs
 
+# By default GCB uses master branch name. This code change it to main.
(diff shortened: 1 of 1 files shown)


diasdauletov:
> In GCB the default branch name used is master while we check on main to publish. This also makes more sense since we use main everywhere else as default branch.

I don't have much knowledge in which cases we run these tests, but I assume these tests should run for any branches. 
Hence shouldn't this script run tests for master too? (even though we publish on main)

Tuseeq1:
No this change is only for publishing the npm package and does not affect any other tests or builds.

Tuseeq1:
yes

diasdauletov (approved):

PR #1857, opened 2024-10-02
Creating release notes as part of publishing process

Tuseeq1:
As part of publishing new NPM package every month we also need to add release notes for these published new versions on github. 
In this change. I am using Github Cli to create a new release when a new version is published to NPM.

### DIFF
--- a/cloudbuild-publish.yaml
+++ b/cloudbuild-publish.yaml
@@ -15,10 +15,10 @@ steps:
     - '-c'
     - |
       set -e
-      VERSION=$(cat version.bzl | grep DF_VERSION | awk '{ print $3 }' | sed "s/\"//g")
-      echo "Create release notes for $VERSION"
+      __VERSION=$(cat version.bzl | grep DF_VERSION | awk '{ print $3 }' | sed "s/\"//g")
+      echo "Create release notes for $__VERSION"
       gh auth login --with-token < token.txt
-      gh release create $VERSION --generate-notes
+      gh release create $__VERSION --generate-notes
 availableSecrets:
   secretManager:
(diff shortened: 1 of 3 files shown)


cloudbuild-publish.yaml
```diff
@@ -2,12 +2,23 @@ steps:
...
       #!/usr/bin/env bash
       ./scripts/publish
+  - name: 'gcr.io/$PROJECT_ID/github'
+    entrypoint: 'bash'
+    args:
+    - '-c'
+    - |
+      set -e
+      VERSION=$(cat version.bzl | grep DF_VERSION | awk '{ print $3 }' | sed "s/\"//g")
+      echo "Create release notes for $VERSION"
```
Ekrekr: Creating

cloudbuild-publish.yaml
```diff
@@ -2,12 +2,23 @@ steps:
   - name: 'gcr.io/cloud-builders/gcloud'
     script: |
       #!/usr/bin/env bash
+      gcloud secrets versions access latest --secret=github-token-access --format='get(payload.data)' | tr '_-' '/+' | base64 -d > token.txt
```
Ekrekr, line 5: It would be good to extract these clunky bits of bash to a shared script somewhere, and make it obvious why each piped stage is done.

Debugging issues with this in the future will be a huge pain without doing that, and we'd likely have drift between the commands!

Ekrekr (approved):

cloudbuild-publish.yaml
Tuseeq1: I will create a separate ticket for this.

PR #1866, opened 2024-10-31
Added logic to fully resolve targets (with prefixes/suffixes) when updating the Data Preparation YAML.

fernst:


core/main_test.ts
```diff
@@ -1072,9 +1079,9 @@ nodes:
           asPlainObject([
             {
               target: {
-                database: "defaultProject",
-                schema: "defaultDataset",
-                name: "dest"
+                database: "defaultProject_projectSuffix",
+                schema: "defaultDataset_datasetSuffix",
+                name: "tablePrefix_dest"
```
fernst, line 1084: In Line 1092 of this same file, the list of all possible targets is outlined, but this override isn't being applied there.

Do I need to add additional changes elsewhere so all targets are resolved correctly?

Ekrekr: Yes. Nice spot with this!

Ekrekr (approved):

Ekrekr (approved):

PR #1880, opened 2024-12-13
Fix string value for uniqueKey assertions, tidy tests

Ekrekr:
Restructuring test to be more DAMP now tests the issue


Tuseeq1 (approved):

Ekrekr:
Tested locally as CI/CD broken

Issue #1168, opened 2021-01-15
Power your code with TypeScript


moltar:
> [Power your code with ~Java~TypeScript](https://docs.dataform.co/guides/javascript)

Since Dataform is written in TypeScript, it would be great to get native support for TypeScript in scripts as well. 

That would then allow developers to get IntelliSense support and compiler can check for type errors.

Thanks!

dgillis91:
My team is moving to dataform and we will be relying pretty heavily on js. I'd much rather have ts support, though. 

I was considering just transpiling to js, but it would be better to have it in dataform. Any tips for tackling it if I make a fork?

dgillis91:
@BenBirt - wanted to get some advice on this one and didn't know who to bump. Any advice here? 

Probably a fairly heavy lift for me, but I'm willing to give it a shot. I also think I can get some people inside our company interested on working this, too.

Ekrekr:
Now that v3 is merged in as the main branch, this should be simpler to do - there's less code that this would have to be applied to.

This would probably need some hacking to figure out the best solution. I could see the solution to this being along the lines of:
* Add typescript as a dependency to the Core and CLI packages.
* Add an option to `workflow_settings.yaml` that swaps to expecting typescript rather than javascript.
* Transpile from TS to JS directly in what's currently the JS part of SQLX block extraction: https://github.com/dataform-co/dataform/blob/c0d1a7400f4aed74f90e032149561a3027df0ea4/core/compilers.ts#L77.
* Alternatively, set up `ts.transpile` for the vm2 compiler in the CLI, https://github.com/patriksimek/vm2/issues/323, https://github.com/dataform-co/dataform/blob/c0d1a7400f4aed74f90e032149561a3027df0ea4/cli/vm/compile.ts#L41.
* I don't know if these would provide nice type validation during compilation though.

If you want to chat further, hit me up! My email is eliaskassell@google.com.

moltar:
FYI with this landing #1550, I am also no longer interested in this project, thus do not need this feature. I will not close the issue, as it might be useful to others, but I will unsubscribe from further updates. Please don't hesitate to close this issue if you don't find it relevant.

Ekrekr:
No worries @moltar! From my last chat with them, https://www.datashell.dev are in the process of creating a CLI tool that can run Dataform compiled graphs (which are now agnostic to the SQL warehouse) against other warehouses - I would keep an eye on them.

toburger:
[...]

PR #1885, opened 2024-12-30
Move publish tests to main_test

Ekrekr:
Individual fields under `publish` are already tested by the config parameters test, e.g. https://github.com/dataform-co/dataform/blob/b951c3807873b27ab65833cef51a6c5aa0a5f03c/core/main_test.ts#L1883.

This avoids us type locking in session the publish method - which allows updating the interface, and is safe because it's not officially supported yet https://github.com/dataform-co/dataform/issues/1168. Directly hooking in with `session` isn't really something we document.

This unblocks https://github.com/dataform-co/dataform/pull/1780.

### DIFF
--- a/core/main_test.ts
+++ b/core/main_test.ts
@@ -5,12 +5,7 @@ import { dump as dumpYaml, load as loadYaml } from "js-yaml";
 import * as path from "path";
 import { CompilerFunction, NodeVM } from "vm2";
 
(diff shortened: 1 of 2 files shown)


diasdauletov (approved):

Ekrekr:
Tests passing manually, CI still broken

PR #1893, opened 2025-01-21
Fix #1805: failing dry-runs when table/view/assertion do not exist

### LINKED ISSUE #1805: Dry run fails when table/view/assertion to be created does not exists

mkamysz:
Fixes #1805, as former fix in #1806 has been reverted in #1834.

* Only attempt to set metadata for tables if BigQuery option dryRun is false
* Assertions consist of two separate queries, first creating a view and second selecting from the view, expecting an empty result set; the second part of this process needs to be skipped for dry runs, as the view does not actually get created


cli/api/dbadapters/bigquery.ts
```diff
@@ -330,72 +330,79 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
-            }
-            onCancel?.(() => {
-              resultStream.end();
-              reject(new Error("Query cancelled."));
+          if (dryRun && query.includes('-- @dry-run:skip-results-check')) {
+            // If the above evaluated to true, this means we would be running the row count query of an assertion (see execution_sql.ts).
+            // This must be stopped during a dry run, as the view is not actually created. Therefore we skip the creation of a query job
+            // and instead directly return.
+            return { rows: [], metadata: {} };
+          } else {
```
Ekrekr, line 338: This else can be skipped, because the function will already have been returned from the if. Also this will get rid of the indentation that is making reviewing a bit confusing.

cli/api/dbadapters/bigquery.ts
```diff
@@ -330,72 +330,79 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
-          return new Promise<IExecutionResult>((resolve, reject) => {
-            if (isCancelled) {
-              resultStream.end();
-              reject(new Error("Query cancelled."));
-              return;
-            }
-            onCancel?.(() => {
-              resultStream.end();
-              reject(new Error("Query cancelled."));
+          if (dryRun && query.includes('-- @dry-run:skip-results-check')) {
```
Ekrekr, line 333: I'm not a fan of making this depend on the SQL contents, because we have parameters that can be passed around instead - also this won't be backwards compatible with older core versions.

I think if you look upstream, we should just be able to skip the second statement of assertions here https://github.com/dataform-co/dataform/blob/6f13b4a283805928e0233004b50e8a074c90cd33/cli/api/commands/run.ts#L416 or one level higher up. This way inserting to the SQL during compilation won't be needed

Ceridan:
@mkamysz friendly ping, and thank you for the patch!

mkamysz:
Hey @Ekrekr @Ceridan, thanks for reviewing and giving pointers! Should be clearer now

Ekrekr (approved):
Nice! Very clean now

Ceridan:
@mkamysz, could you please merge the latest `main` branch to your branch? I believe it should fix presubmit test failures.

mkamysz:
@Ceridan done, don't think it fixed the test failure though, did it?

Ceridan (approved):

PR #1895, opened 2025-01-23
Add support for an empty data preparation. 

fernst:
This uses the file name as the target.

### DIFF
--- a/core/actions/data_preparation.ts
+++ b/core/actions/data_preparation.ts
@@ -143,7 +143,8 @@ export class DataPreparation extends ActionBuilder<dataform.DataPreparation> {
             session.canonicalProjectConfig
         );
     this.proto.canonicalTargets = [this.proto.canonicalTarget];
-    this.proto.dataPreparationYaml = dumpYaml(dataPreparationAsJson);
+    const resolvedDefinition = this.applySessionToDataPreparationContents(dataPreparationAsJson);
+    this.proto.dataPreparationYaml = dumpYaml(resolvedDefinition);
   }
 
   private resolveTargets(
@@ -177,7 +178,7 @@ export class DataPreparation extends ActionBuilder<dataform.DataPreparation> {
   }): { [key: string]: any } {
     // Handle empty definitions
     if (!definition) {
-      return {};
+      return definition;
     }
 
     // Resolve error table, if set
--- a/core/actions/data_preparation.ts
+++ b/core/actions/data_preparation.ts
@@ -12,6 +12,7 @@ import {
   resolveActionsConfigFilename
 } from "df/core/utils";
 import { dataform } from "df/protos/ts";
+import {filename} from 'df/core/path';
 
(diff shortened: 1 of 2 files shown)


core/actions/data_preparation.ts
```diff
@@ -116,9 +114,66 @@ export class DataPreparation extends ActionBuilder<dataform.DataPreparation> {
...
+        )
+    );
+    this.proto.targets = [this.proto.target];
+    this.proto.canonicalTarget =
+        this.applySessionToTarget(
+            defaultTarget,
+            session.canonicalProjectConfig
+        );
+    this.proto.canonicalTargets = [this.proto.canonicalTarget];
+    this.proto.dataPreparationYaml = dumpYaml(dataPreparationAsJson);
```
kolina, line 140: Should you call `applySessionToDataPreparationContents` similarly to how it's done in `resolveTargets`?

kolina (approved):

core/actions/data_preparation.ts
```diff
@@ -31,7 +31,13 @@ export class DataPreparation extends ActionBuilder<dataform.DataPreparation> {
     this.session = session;
 
     if (!config.name) {
-      config.name = Path.basename(config.filename);
+      // Ensure we extract only the file name.
+      // This handles both .yaml and .dp.yaml extensions
+      let name = Path.basename(config.filename);
+      if (name.indexOf(".") > 0) {
+        name = name.substring(0, name.indexOf("."));
```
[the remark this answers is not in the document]
fernst: I noticed targets cannot have a dot `.` in the name, which is why I wrote the logic like this. Essentially, we pick only the first segment before the first dot.

core/actions/data_preparation.ts
fernst: Since there are no targets in the data preparation, there is no entity to apply session to.

We apply the session (defaults, overrides, etc) to targets, and since there are none, there is no reason to do this.

core/actions/data_preparation.ts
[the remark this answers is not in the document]
fernst: But I see your point. Let me update to only drop `sqlx`, `dp.yaml` and `yaml`.

core/actions/data_preparation.ts
kolina: You also apply it here to `source` https://github.com/dataform-co/dataform/blob/026b7fb1a822a85d1fbe3e9d651c349f01e4fbf4/core/actions/data_preparation.ts#L191, that is not added to targets https://github.com/dataform-co/dataform/blob/026b7fb1a822a85d1fbe3e9d651c349f01e4fbf4/core/actions/data_preparation.ts#L247

fernst: I see your point now. This would give us a more "complete" data preparation, but it still won't execute.

Let me update.

Issue #1755, opened 2024-06-15
Integration test is not working for external contributors


moker-spaghetti:
Found in #1746 #1751 #1754 (Addressing the issue in #1754)
Originates in #1729 ?

External contirbutors cannot run the integration test (`scripts/run_tests`) since the decryption phase of `test_credentials/bigquery.json.enc` fails.

Ekrekr:
Some more context: the GCP project that the integration tests run in has an org policy which prevents non-Googlers from accessing it - and there's no current workaround.

Long term we may test against a BQ emulator instead, so that the GCP project connection isn't required for testing.

For now just disabling the integration tests when you run the test script is the only option, and having the PR approver run the tests before merging, is what has to be done.

bmagyarkuti:
A suggestion: if we just extracted the project name to an envvar, and set up the test environment to prefer application default credentials over the encrypted file, external contributors would be able to set up their own GCP project and run tests there, without having to make any modifications to the code. The project itself is trivial to set up, and luckily the tests don't depend on any assets there.

This would streamline my own workflow, which currently involves cherry-picking commits that change the GCP project and overwrite the encrypted credentials from my own local commit history.

It'd be great if there existed an officially supported BigQuery emulator, but I've yet to find one that is full-featured and is able to reliable replicate BQ's full feature set (I just mean its interface, I'm not expecting the emulator to replicate its performance.)

Ceridan:
Setting up own GCP project will make it harder for the new external contributors to set up the development environment. In #1900 we split unit and integration tests. Currently, if you will run `./scripts/run_tests` only unit tests will be executed and it is enough to validate the changes. The full test suite including integration tests will be executed on PR on the GCP project owned by Dataform team.

PR #1900, opened 2025-01-27
Split unit and integration tests

Ceridan:
At the moment external contributors can't run integration tests because they have no access to the GCP project used for tests. Split allows external contributors to run `./scripts/run_tests` without any issues (in fact only unit tests will run). Integration tests will run on PR.

Related issue: #1755

### DIFF
--- a/scripts/run_integration_tests
+++ b/scripts/run_integration_tests
@@ -7,5 +7,6 @@ bazel build @npm//tslint/bin:tslint && bazel-bin/external/npm/tslint/bin/tslint.
 
 # Run all the tests
 bazel test ... \
-  --test_tag_filters=-integration \
+  --test_tag_filters=integration \
   --build_tests_only \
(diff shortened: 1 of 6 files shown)


scripts/run_integration_tests
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -e
+
+# Run tslint.
+bazel run @nodejs//:yarn
+bazel build @npm//tslint/bin:tslint && bazel-bin/external/npm/tslint/bin/tslint.sh --project .
+
+# Run all the tests
```
Ekrekr, line 8: Comment isn't true now

Ekrekr (approved):
Nice, this should help people a lot

scripts/run_integration_tests
Ceridan: Fixed.

PR #1901, opened 2025-01-29
Remove unused import which broke unit tests.

Ceridan:
Currently unit tests are broken because #1895 introduced an unused import (I believe it was added by accident) in core/actions/data_preparations.ts


fernst:
Thank you. Yes, this was an accidental import.

fernst (approved):

PR #1902, opened 2025-01-29
Add support for OnSchemaChange field for incremental tables.

Ceridan:
The changes are part of a new feature: "Incremental Table Schema Update". The `onSchemaChange` in `IncrementalTableConfig` allows users to define the behavior of the action in case list of columns is changed in the select statement. E.g. new columns are added, or existing columns are removed.


core/main_test.ts
```diff
@@ -2609,6 +2611,79 @@ SELECT 2`
         });
       });
     });
+
+    suite("incremental table on schema change", () => {
```
Ekrekr, line 2615: It's a bit overkill to test all of these exhaustively, I would just add one of the (non-default) options to the action configs and SQLX tests. The other tests include the defaults.

Where I'm suggesting adding:
* https://github.com/dataform-co/dataform/blob/c799ec73800f23d0cbf03e1862a4a841e79b1b6d/core/main_test.ts#L2994, and
* https://github.com/dataform-co/dataform/blob/c799ec73800f23d0cbf03e1862a4a841e79b1b6d/core/main_test.ts#L2373

core/main_test.ts
```diff
@@ -3168,6 +3244,110 @@ actions:
         ])
       );
     });
+
+    suite("incremental table on schema change", () => {
```
Ekrekr, line 3248: Similarly with this - also the suite name is identical, even though this one is for action configs files rather than SQLX config blocks

protos/configs.proto
```diff
@@ -255,6 +255,17 @@ message ActionConfig {
     TableAssertionsConfig assertions = 17;
   }
 
+  enum OnSchemaChange {
+    // Ignore any schema changes (default).
+    IGNORE = 0;
+    // Fail if any schema changes introduced in the query.
```
Ekrekr, line 261:
```suggestion
    // Fails if the query would result in a new column(s) being added, deleted, or renamed.
```

protos/configs.proto
```diff
@@ -255,6 +255,17 @@ message ActionConfig {
...
+    FAIL = 1;
+    // Alter target table to add new columns.
+    EXTEND = 2;
+    // Alter target table to add new columns and remove missing columns.
```
Ekrekr, line 265:
```suggestion
    // Does not block any new column(s) from being added, deleted or renamed.
```

protos/configs.proto
```diff
@@ -255,6 +255,17 @@ message ActionConfig {
     TableAssertionsConfig assertions = 17;
   }
 
+  enum OnSchemaChange {
+    // Ignore any schema changes (default).
+    IGNORE = 0;
+    // Fail if any schema changes introduced in the query.
+    FAIL = 1;
+    // Alter target table to add new columns.
```
Ekrekr, line 263:
```suggestion
    // Does not block any new column(s) from being added.
```

protos/configs.proto
```diff
@@ -346,6 +357,9 @@ message ActionConfig {
     // the  action depends on data from a source which has not been declared as
     // a dependency.
     bool hermetic = 23;
+
+    // Defines the action behavior if the selected columns in query doesn't match columns in the target table.
```
Ekrekr, line 361:
```suggestion
    // Defines the action behavior if the selected columns in the query don't match the columns in the target table.
```

protos/core.proto
```diff
@@ -112,6 +112,17 @@ enum TableType {
   reserved 4;
 }
 
+enum OnSchemaChange {
```
Ekrekr, line 115: Ditto for my docs suggestions

core/actions/incremental_table.ts
```diff
@@ -465,6 +466,30 @@ export class IncrementalTable extends ActionBuilder<dataform.Table> {
       VerifyProtoErrorBehaviour.SHOW_DOCS_LINK
     );
   }
+
+  private mapOnSchemaChange(onSchemaChange?: string|number) : dataform.OnSchemaChange {
```
Ekrekr, line 470: The type `number` is supported here, but I'm not sure what that would look like if you actually wanted to reference the named enum in the code (writing integers isn't really a CUJ we need to support).

So from that, just expecting string would make sense - we're not writing tests for the number version, and that would add more complexity.

However, I'm slightly confused why string works here in the first place. This step https://github.com/dataform-co/dataform/blob/c799ec73800f23d0cbf03e1862a4a841e79b1b6d/core/actions/incremental_table.ts#L462 checks that `onSchemaChange` is of the type specified in the proto, which I guess for enums includes strings.

So overall I think this implementation correct, but it would be good to do some manual testing for weird edge case behavior later

Ekrekr (approved):

core/main_test.ts
Ceridan: Followed your suggestion remove additional tests and add some non-default values to the places you suggest. I also make sure that the whole mapper is properly covered.

core/main_test.ts
Ceridan: Similarly, fixed this.

protos/configs.proto
Ceridan: Thank you, fixed.

protos/configs.proto
Ceridan: Fixed.

protos/configs.proto
Ceridan: Fixed.

protos/configs.proto
Ceridan: Fixed.

protos/core.proto
Ceridan: Fixed.

core/actions/incremental_table.ts
Ceridan: I added comments to my mapping function which explains the logic:

The type of onSchemaChange depends on the source file:
- for sqlx it will have type "string"
- for action.yaml it will be converted to enum which is represented
in TypeScript as a "number".

It is a good idea to unify it during the verification case, will work on this in a separate PR.

zachrenwick:
Can the new onSchemaConfig options only be used with .js files?

I tried using it with a .sqlx file, added a new test column and tried an incremental load but got failures on all of the following config settings:
- synchronize
- ignore
- extend

image

image

kolina:
This is not yet ready for the actual usage: GCP Dataform backend doesn't process it yet. There will be a separate release not when this happens.

zachrenwick:
@kolina ah gotcha thanks!

Issue #1908, opened 2025-02-03
Bug: `publish` function stopped working in folder `includes/` which breaks udf unit-testing


mokshasoul:
Hello everyone,

We have an adapted [version of this](https://github.com/GoogleCloudPlatform/bigquery-utils/blob/master/udfs/tests/dataform_testing_framework/includes/unit_test_utils.js)

Behavior:

```text
TypeError: publish(...).type is not a function
```

This started happening with version dataform version `3.0.10`.

This functionality was essential and the library lives in `includes/`.

Ekrekr:
Hi Charis-Nicolas,

Thanks for the heads up. I expect a bug was introduced in https://github.com/dataform-co/dataform/pull/1780 that is causing this. We'll patch a fix for it in the next few days - for now just stick with 3.0.9.

Thanks,
Elias

mokshasoul:
Hi @Ekrekr thx for the quick update

Issue #1907, opened 2025-02-03
Incorrect expected types in config block


mattimoller:
Seems like `3.0.10` and specifically #1780 introduced some changes to the expected properties of the config block and making a bunch of the fields required vs optional. Seems like the only way to get typing to work now is to write out all the 20+ required fields in the config, which isn't a viable option. I'm no TS expert so could be something wrong with our setup here, open to any suggestions! 

Consider the below `my_table.js`

```
publish(`my_table`, {
    type: "table",
    schema: "my_schema",
    bigquery: {
        labels: {
            source: "my_source"
        }
    },
    tags: ["table"],
    assertions: {
        nonNull: ["id"]
    }
}).query(`select 1 as id`)
```

This table compiles and can be run both in Dataform `3.0.9` and `3.0.10` but in `3.0.10` our type checks complain about the below

**Error**
```
definitions/my_table.js:1:21 - error TS2345: Argument of type '{ type: string; schema: string; bigquery: { labels: { source: string; }; }; tags: string[]; assertions: { nonNull: string[]; }; }' is not assignable to parameter of type 'Contextable<ITableContext, string> | ILegacyTableConfig | ILegacyViewConfig | ILegacyIncrementalTableConfig | undefined'.
  Type '{ type: string; schema: string; bigquery: { labels: { source: string; }; }; tags: string[]; assertions: { nonNull: string[]; }; }' is missing the following properties from type 'ILegacyIncrementalTableConfig': dependencies, database, fileName, name, and 21 more.
```

**Steps to recreate**:
1. Install dataform version 3.0.10 or higher (see attached `package.json`)
2. Run npx tsc --project jsconfig.json --noEmit --allowJs --checkJs;

Package.json dependencies:
```
{
  "name": "any_name",
  "dependencies": {
    "@dataform/core": "3.0.10",
    "dataform-scd": "https://github.com/dataform-co/dataform-scd/archive/0.5.tar.gz",
    "date-fns": "4.1.0"
  },
  "devDependencies": {
    "@eslint/js": "^9.19.0",
    "@types/jest": "^29.5.14",
    "eslint": "^9.19.0",
    "eslint-plugin-jest": "^28.11.0",
    "globals": "^15.14.0",
    "jest": "^29.7.0",
    "prettier": "^3.4.2",
    "typescript": "^5.7.3"
  }
}
```

Jsconfig
[jsconfig.json](https://github.com/user-attachments/files/18640378/jsconfig.json)

[...]

PR #1910, opened 2025-02-05
Fix publish type backwards compatibility

Ekrekr:
in 3.0.10 to 3.0.12, the new test cases added to `core/main.test` in this PR would throw errors - but in 3.0.9 and prior they were supported. This re-adds backwards compatibility of this support.

It also fixes some typing used, though unfortunately disabling build type safety has to be done to facilitate it.

Fixes https://github.com/dataform-co/dataform/issues/1908
Fixes https://github.com/dataform-co/dataform/issues/1907

### DIFF
--- a/core/session.ts
+++ b/core/session.ts
@@ -2,14 +2,18 @@ import { default as TarjanGraphConstructor, Graph as TarjanGraph } from "tarjan-
 
 import { encode64, verifyObjectMatchesProto, VerifyProtoErrorBehaviour } from "df/common/protos";
 import { StringifiedMap, StringifiedSet } from "df/common/strings/stringifier";
-import { Action, ITableContext } from "df/core/actions";
+import { Action, ITableContext, TableType } from "df/core/actions";
 import { AContextable, Assertion, AssertionContext } from "df/core/actions/assertion";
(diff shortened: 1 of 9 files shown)


core/session.ts
```diff
@@ -264,23 +276,28 @@ export class Session {
...
+      // `any` is used here to facilitate the type merging of legacy table configs, which are very
+      // different to the new structures.
+      | any
   ): Table | IncrementalTable | View {
     let newTable: Table | IncrementalTable | View = new View(this, { type: "view", name });
     if (!!queryOrConfig) {
       if (typeof queryOrConfig === "object") {
-        if (queryOrConfig?.type === "table" || queryOrConfig.type === undefined) {
-          newTable = new Table(this, { type: "table", name, ...queryOrConfig });
+        if (queryOrConfig?.type === "view" || queryOrConfig.type === undefined) {
```
kolina, line 295: This seems like a change of behaviour to return `View` instead of `Table` when `queryOrConfig.type === undefined`

kolina (approved):

core/session.ts
Ekrekr: This is a fix - it was returning `View` when `queryOrConfig === undefined`, but table when `queryOrConfig.type === undefined`, which was an edge case I was missing

PR #1912, opened 2025-02-06
Remove custom generic stringifiers

Ekrekr:
This is because they are throwing issues with es6 shim generation, which is breaking the TSDoc generation that I'm trying to configure.

Also it's quite a lot of code that isn't used that much, so this reduces future maintenance costs too.


cli/console.ts
```diff
@@ -404,7 +404,7 @@ function printExecutedActionErrors(
   executionAction: dataform.IExecutionAction
 ) {
   const failingTasks = executedAction.tasks.filter(
-    task => task.status === dataform.ActionResult.ExecutionStatus.FAILED
+    task => task.status === dataform.TaskResult.ExecutionStatus.FAILED
```
GJMcGowan, line 407: This is a bugfix/type fix I assume? In theory it's breaking as the enums have different values.

GJMcGowan (approved):
I agree overall this is a good change. I'm sad that we're losing some of the typing here, and I worry it makes the code slightly less robust in that particular way.

I'm not going to block on this, but I would be happier if the `targetStringifier.parse` function explicitly checked that the object it returns IS a target (i.e. does it have any of the target fields, and only those fields) and then threw in tests or at runtime.

cli/console.ts
Ekrekr: Yeah, I'm surprised this wasn't throwing an error before - the new type is definitely right: https://github.com/dataform-co/dataform/blob/5995f4e169c2972db6207b44e4e45b55d5a20bcd/protos/execution.proto#L98

Ekrekr:
> `targetStringifier.parse` function explicitly checked that the object it returns IS a target

This is a good idea, but I want to do it properly with protobuf based verification - which doesn't currently work very well because of protobuf js. I'll try to find a decent workaround while I'm doing subsequent tidying

PR #1920, opened 2025-02-13
Add notebook js support

Ekrekr:
This tests and properly configures being able to create notebooks using the JS API.

Updating the parameter structure in a backwards breaking way is fine here because:
* It was broken and untested before.
* Therefore no one was using it.


Ceridan (approved):

Ceridan:
Please, fix linting lissue (see presubmits)

PR #1921, opened 2025-02-13
Replace deprecated `setNameAndTarget` use cases with preferred option

Ekrekr:
Also add some tests that cover the build API that can be called with these directly


core/actions/assertion.ts
```diff
@@ -243,12 +242,12 @@ export class Assertion extends ActionBuilder<dataform.Assertion> {
...
-      this.session,
-      this.proto,
-      this.proto.target.name,
-      this.proto.target.schema,
-      database
+    this.proto.target = this.applySessionToTarget(
+      dataform.Target.create({ ...this.proto.target, database }),
+      this.session.projectConfig,
+      this.proto.fileName,
+      true,
```
Ceridan, line 249: Having two booleans here is very confusing to the reader. Ideally I would change the method API to accept an object which contains all the fields, it will allow to pass named parameters.

But to make it simpler could we just add comments like:

```typescript
public database(database: string) {
    this.proto.target = this.applySessionToTarget(
      dataform.Target.create({ ...this.proto.target, database }),
      this.session.projectConfig,
      this.proto.fileName,
      /* validateTarget= */ true,
      /* useDefaultAssertionDataset= */ true
    );
```

I know it is annoying to write it but will be much easier to understand the code afterwards.

Ceridan (approved):

core/actions/assertion.ts
Ekrekr: Fairs, comments are the worst way to do this though - have made it an object so that naming is required

PR #1922, opened 2025-02-14
Move table config checks to actions

Ekrekr:
This makes config checking more consistent, as these options are now checked in the same places as the other config option checks.

Note that test coverage for all these options can be found here: https://github.com/dataform-co/dataform/blob/009252885087397ec32ce61745909446281307fc/core/main_test.ts#L4538


core/actions/index.ts
```diff
@@ -106,6 +106,27 @@ export abstract class ActionBuilder<T> {
   }
 }
 
+export function checkConfigAdditionalOptionsOverlap(
```
Ceridan, line 109: Just a question: it is not fully clear to me how we use the `index.ts`. Do I understand correctly that `index.ts` should be the "package" (in this case `core/actions`) entry point and should export public API?

I am a little bit confused about this method because it mostly used inside actions, but there is one use case in session. I have a feeling that we have messed up with abstractions somewhere.

No need to change anything, just wanted to discuss this.

Ceridan (approved):

core/actions/index.ts
Ekrekr: > mostly used inside actions

It's just to be used by actions - it was previously done in compilation as a post-compilation verification. But verifying that configs are correct from the output of the compiled graph doesn't really make sense now that configs are checked before the compiled graph is made now.

> index.ts should be the "package"

In general I'm using `core/actions/index.ts` for methods that are common across all actions, so yes!

PR #1924, opened 2025-02-17
Make action protos private

Ekrekr:
This prevents proto update logic from happening outside of actions. This makes action update logic more strict by having it happen inside the action files where they are configured.

### DIFF
--- a/core/actions/assertion.ts
+++ b/core/actions/assertion.ts
@@ -80,10 +80,8 @@ export type AContextable<T> = T | ((ctx: AssertionContext) => T);
 export class Assertion extends ActionBuilder<dataform.Assertion> {
   /**
    * @hidden Stores the generated proto for the compiled graph.
-   * <!-- TODO(ekrekr): make this field private, to enforce proto update logic to happen in this
-   * class. -->
    */
-  public proto = dataform.Assertion.create();
+  private proto = dataform.Assertion.create();
 
   /** @hidden Hold a reference to the Session instance. */
   public session: Session;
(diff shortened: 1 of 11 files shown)


core/actions/assertion.ts
```diff
@@ -274,6 +272,21 @@ export class Assertion extends ActionBuilder<dataform.Assertion> {
...
   }
 
+  /** @hidden */
+  public getParentAction() {
+    return dataform.Target.create(this.proto.parentAction);
+  }
+
+  /** @hidden */
+  public setParentAction(target: dataform.Target) {
+    return (this.proto.parentAction = target);
```
Ceridan, line 282: Why do we need returns on setters? My expectations that it just sets the value.

The same question about other setters

core/actions/operation.ts
```diff
@@ -323,6 +324,16 @@ export class Operation extends ActionBuilder<dataform.Operation> {
...
   }
 
+  /** @hidden */
+  public getHasOutput() {
+    return this.proto.hasOutput;
+  }
+
+  /** @hidden */
+  public setFilename() {
+    return this.proto.hasOutput;
```
Ceridan, line 334: This is very counterintuitive. Setter in fact sets nothing. What was the intention here?

core/actions/operation.ts
```diff
@@ -323,6 +324,16 @@ export class Operation extends ActionBuilder<dataform.Operation> {
     return dataform.Target.create(this.proto.target);
   }
 
+  /** @hidden */
+  public getHasOutput() {
```
Ceridan, line 328: Maybe specify a return type? I assume it is `boolean` but maybe it is better to make it clear?

Ekrekr: With most methods we just let the return value be implicit, but I'm also a fan of explicitly putting the returned types. Would be good to update everything else too!

core/actions/operation.ts
Ekrekr: This method is totally wrong, and it works because it's not being used. I've realized we don't actually need to expose a `setFilename` for most classes, so have removed and updated actions to pass is in via configs

core/actions/test.ts
```diff
@@ -137,6 +147,11 @@ export class Test extends ActionBuilder<dataform.Test> {
     return undefined;
   }
 
+  /** @hidden */
+  public setFilename(filename: string) {
+    return (this.proto.fileName = filename);
```
Ceridan, line 152: We missed this one.

Ceridan (approved):

core/actions/operation.ts
Ceridan: I think we can do it as we go, when touching some code.

PR #1929, opened 2025-02-20
Move actions tests to separate files

Ekrekr:
Bit hard to review sorry, but good to get over with! All the same tests remain, apart from a duplicate test for disabled options, but just moved. Plus some shared library changes.

I also want to:
* make these tests separate build targets, following Google standards, but I can't until resolving some circular dependencies.
* Have config options for YAML, SQLX, and the JS API merged.


core/main_test.ts
Ceridan, line 1: Could you please explain what is the current responsibility of `main_test.ts`?

Ekrekr: I think pretty much everything could be moved to separate test files!

So currently it's mainly a proxy for testing session, because session doesn't really work without the code in `main`. But this could all  moved under a `session_test.ts` file.

Things like SQLX testing in here are needed here because the literal JS parsing during SQLX compiling changes when passed through all the layers like this: https://github.com/dataform-co/dataform/blob/3035b54949b1fe39efc137c3d30e47347a822848/core/compilers.ts#L85. But they could be moved under `compilers_test.ts`, and call the methods in `run_core.ts`, like the actions tests do now.

Workflow settings could be moved to `workflow_settings_test.ts`, and general action configs tests could be moved to `actions/index_test.ts`.

image

PR #1942, opened 2025-03-25
Make  field on Incremental tables default to false as it was before.

### LINKED ISSUE #1914: Queries fail to run in full refresh mode
### How to reproduce:

1. Start a new execution in a Dataform (Cloud) workspace by specifying a tag.
2. Select run with full refresh and include dependendents
3. Start execution

### What happens:

The queries included in the workflow all run without errors. However, the incremental table queries are executed in incremental mode instead of full refresh mode.

In my case, the workflow included both regular tables and incremental tables.

### How to fix it:

The issue happens with Dataform core versions starting from 3.0.10.

Downgrading the Dataform core version back to 3.0.9 resolved the issue.

Ceridan:
Configs refactoring introduced by #1780 changes the default behavior of `protected` field on Incremental tables, see #1914

The current PR changes the default behavior to make it consistent with what we had before:

1. If config has `protected` field set, then the value of the field will be used.
2. If config omits `protected` field, it will be set to false.
3. JavaScript API will be slightly __changed__: by default, `protected` will be set to false.
4. If in JavaScript API `.protected()` will be called without arguments (like it was possible before) then the field will be set to through.
5. If in JavaScript API `.protected()` will be called with argument (true or false) then the field will be set to the value of the argument passed.

The logic is quite complicated here but we are trying to mimic the previous behavior to preserve backward compatibility.

Closes #1914

### DIFF
--- a/core/main_test.ts
+++ b/core/main_test.ts
@@ -1097,7 +1097,7 @@ publish("name", {
                           incrementalPostOps: ["post_op"],
                           incrementalPreOps: ["pre_op"],
                           incrementalQuery: "SELECT 1",
-                          protected: true,
+                          protected: false,
                           onSchemaChange: "IGNORE"
                         }
                       : {})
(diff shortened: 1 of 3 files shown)


core/main_test.ts
```diff
@@ -1162,13 +1162,63 @@ publish("name", {
...
+                fileName: "definitions/publish.js",
+                hermeticity: "NON_HERMETIC",
+                query: "SELECT 1",
+                target: {
+                  database: "defaultProject",
+                  name: "name",
+                  schema: "defaultDataset"
+                },
+                type: tableType,
+                ...(tableType === "incremental"
```
kolina, line 1211: Do we need this check here?

kolina (approved):

core/main_test.ts
Ceridan: Good catch, removed the check.

Issue #1944, opened 2025-03-26
[Feature Request] Print BQ job during CLI runs


HHammond:
When working with the dataform CLI it would be useful to print the BQ job ID so that detailed execution logs can be easily reviewed during development phases.

Ceridan:
JobId was added in #1949. Thanks @ashish10alex for contribution!

We plan to release the new CLI version this week which will include the changes.

ashish10alex:
thanks @HHammond , I really wanted to make the jobIds that will be shown clickable. But most terminal emulators will not support that so I was not sure if that will something that will be in scope of the project. But we can still do lot with the output of the job runs now, like getting cumulative billed cost of running a tag with dependencies using the jobId's.

PR #1948, opened 2025-03-27
feat: show bigquery jobId when executing the job

ashish10alex:
Solves: https://github.com/dataform-co/dataform/issues/1944

Shows BigQuery jobId against each ran model as follows 

image

**Tests**

- [x] JobId shows against the correct model 
- [ ] JobId shows against failed model / assertion run 

**P.S.**

It will be good to have have the jobIds as hyperlinks but most default terminal emulators do not support this. From the terminals I have only Kitty supported it. Default MacOS terminal and Ghostty terminal did not support it.


ashish10alex:
closing, committed with wrong git git creds that did not have class

PR #1949, opened 2025-03-27
feat: show bigquery jobId when executing the job

ashish10alex:
Solves: https://github.com/dataform-co/dataform/issues/1944

Shows BigQuery jobId against each ran model as follows. Happy to change the format the jobId is displayed 

image

**Tests**

- [x] JobId shows against the correct model 
- [x] JobId shows against failed model / assertion run 
- [x] Linting by running `./scripts/lint`

**P.S.**

It will be good to have have the jobIds as hyperlinks but most default terminal emulators do not support this. From the terminals I have only Kitty supported it. Default MacOS terminal and Ghostty terminal did not support it.

### DIFF
--- a/cli/api/commands/run.ts
+++ b/cli/api/commands/run.ts
@@ -441,6 +441,11 @@ export class Runner {
           ? dataform.TaskResult.ExecutionStatus.CANCELLED
           : dataform.TaskResult.ExecutionStatus.FAILED;
         taskResult.errorMessage = `${this.graph.projectConfig.warehouse} error: ${e.message}`;
+        taskResult.metadata = {
+          bigquery: {
+            jobId: e.metadata?.bigquery?.jobId
+          }
+        };
       }
     }
     taskResult.timing = timer.end();
--- a/cli/api/commands/run.ts
+++ b/cli/api/commands/run.ts
@@ -441,11 +441,13 @@ export class Runner {
           ? dataform.TaskResult.ExecutionStatus.CANCELLED
           : dataform.TaskResult.ExecutionStatus.FAILED;
         taskResult.errorMessage = `${this.graph.projectConfig.warehouse} error: ${e.message}`;
-        taskResult.metadata = {
-          bigquery: {
-            jobId: e.metadata?.bigquery?.jobId
-          }
-        };
+        if (e.metadata?.bigquery?.jobId) {
+          taskResult.metadata = {
+            bigquery: {
+              jobId: e.metadata.bigquery.jobId
+            }
+          };
+        }
       }
     }
     taskResult.timing = timer.end();
--- a/cli/api/dbadapters/bigquery.ts
+++ b/cli/api/dbadapters/bigquery.ts
@@ -389,11 +389,13 @@ export class BigQueryDbAdapter implements IDbAdapter {
                   const [jobMetadata] = await job[0].getMetadata();
                   if (!!jobMetadata.status?.errorResult) {
                     const error: any = new Error(jobMetadata.status.errorResult.message);
+                    if (jobMetadata.jobReference && jobMetadata.jobReference.jobId) {
                     error.metadata = {
                       bigquery: {
-                        jobId: jobMetadata.jobReference.jobId,
-                      }
-                    };
+                          jobId: jobMetadata.jobReference.jobId,
+                        }
+                      };
+                    }
                     reject(error);
                     return;
                   }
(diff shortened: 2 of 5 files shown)


cli/api/commands/run.ts
```diff
@@ -441,6 +441,11 @@ export class Runner {
           ? dataform.TaskResult.ExecutionStatus.CANCELLED
           : dataform.TaskResult.ExecutionStatus.FAILED;
         taskResult.errorMessage = `${this.graph.projectConfig.warehouse} error: ${e.message}`;
+        taskResult.metadata = {
```
Ceridan, line 444: Could we please set `metadata` it only when `e.metadata?.bigquery?.jobId` is defined? At the moment metadata is optional and it makes sense to set it only if we actually have some value.

It will also make integration tests pipeline green.

Ceridan (changes requested):
@ashish10alex, thanks a lot for the contribution!

I requested a minor change and with it in-place will happy to accept the PR.

cli/api/commands/run.ts
ashish10alex: thanks for the feedback  @Ceridan . I have made the requested change.

cli/api/dbadapters/bigquery.ts
```diff
@@ -356,12 +356,26 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
                 // Example: https://github.com/googleapis/python-bigquery/issues/118.
                 if (dryRun && e.message?.includes("Not found: Job")) {
                   resolve({ rows: [], metadata: {} });
+                  return;
+                }
+                
+                try {
+                  const [jobMetadata] = await job[0].getMetadata();
+                  if (jobMetadata && jobMetadata.jobReference && jobMetadata.jobReference.jobId) {
+                    e.metadata = {
```
Ceridan, line 371: Adding a `metadata` field is a little bit unexpected here and can be confusing. Because we use TypeScript it makes sense to define a new interface in `/cli/api/dbadapters/index.ts` like:

```typescript
interface BigQueryError extends Error {
  metadata?: dataform.IExecutionMetadata
}
```
and construct it here?

The same for the check below.

cli/api/dbadapters/bigquery.ts
```diff
@@ -374,7 +388,15 @@ export class BigQueryDbAdapter implements IDbAdapter {
                 try {
                   const [jobMetadata] = await job[0].getMetadata();
                   if (!!jobMetadata.status?.errorResult) {
-                    reject(new Error(jobMetadata.status.errorResult.message));
+                    const error: any = new Error(jobMetadata.status.errorResult.message);
```
Ceridan, line 391: Because we need to construct the error twice, I suggest to extract a helper method which accepts `jobMetadata` as argument, and returns `BigQueryError`.

Ceridan (changes requested):
Thank you for the quick response and fixes. I think we can make it a slightly better, could you please take a look at my comments on error construction?

cli/api/dbadapters/bigquery.ts
ashish10alex: created a new function `createBigQueryError`

cli/api/dbadapters/bigquery.ts
ashish10alex: added and exported `BigQueryError` as suggested

ashish10alex:
> Thank you for the quick response and fixes. I think we can make it a slightly better, could you please take a look at my comments on error construction?

thanks @Ceridan  for the suggestion to improve the code. I have made the requested changes now

Ceridan (changes requested):
The code looks great, thanks a lot!

The one last blocker: could you please fix the linter issues?

```
ERROR: /usr/local/google/home/mkumachev/repos/dataform-co/dataform/cli/api/dbadapters/bigquery.ts:6:51 - Named imports must be alphabetized.
ERROR: /usr/local/google/home/mkumachev/repos/dataform-co/dataform/cli/api/dbadapters/index.ts:11:18 - interface name must start with a capitalized I
```

You may reproduce them by running:

```
./scripts/lint
```

I know that at the moment our toolchain is not very contributor-friendly but I have some plans to improve it in the near future.

ashish10alex:
>  could you please fix the linter issues?

Thanks @Ceridan , I appreciate you guiding me through  the process :) 

I have fixed the lining error and also added the `./scripts/lint` to the `contributing.md` file. 

**Tests**

- [x] Linting

CleanShot 2025-03-31 at 09 42 27@2x

Ceridan (approved):
Thank you!

I will merge the changes, and plan to release the new CLI version this week which will include the changes.

Issue #1899, opened 2025-01-26
add new option for Built-in assertions name


hkak03key:
# Background
## we set defaultAssertionDataset same as defaultDataset
I am considering setting the destination for views generated by assertions to the same location as the models.  
(In other words, I plan to set `defaultDataset` and `defaultAssertionDataset` to the same value.)  
This is because, when `defaultDataset` is shared across multiple repositories, views generated by assertions may end up with duplicate or conflicting names between repositories.

## our model naming rule
For model naming, I use the following convention:  
- **Outputs models**: No prefix.  
- **Intermediate models**: Prefix `_`.  
- **Source models**: Prefix `__`.  
(This is somewhat similar to Python conventions.)  
I adopted this approach because prefixes like `stg_` or `intermediate_` might not be intuitive for non-engineer analysts who come from a marketing background.

# Problem
Built-in assertions generate views with the naming convention:  
`"${ACTION_DATASET_NAME}_${ACTION_NAME}_assertions_${BUILD_IN_ASSERTION}"`.  
This naming pattern conflicts with the convention used for outputs models.

# Solution
Introduce an additional option called `buildinAssertionPrefix`.

---

Of course, one alternative would be to create a separate `defaultAssertionDataset` for each repository.
However, this approach would significantly increase the number of resources, which I find less desirable.

I would appreciate it if you could kindly consider this approach.

Ceridan:
Hi @hkak03key, thank you for creating the issue!

Could you please provide examples of generated names and explain what exact conflict with naming we have here? I don't get why only output models have this problem?

Ceridan:
I added the new config parameter, see #1956 for details. It will be available in the next version (3.0.19).

PR #1956, opened 2025-04-03
Add builtinAssertionNamePrefix field to WorkflowSettings.

Ceridan:
Add a new field called `builtinAssertionNamePrefix` to `WorkflowSettings` to be able to modify the names which generated for built-in` assertions. In some cases the current name pattern may lead to name collisions. 

If, for example, the value of `builtinAssertionNamePrefix` will be set to some value in `workflow_settings.yaml` the assertion will be generated with the following pattern:

```javascript
${BUILT_IN_ASSERTION_NAME_PREFIX}${ACTION_DATASET_NAME}_${ACTION_NAME}_assertions_${BUILT_IN_ASSERTION_NAME}

```

See #1899 for more details.

### DIFF
--- a/core/actions/index.ts
+++ b/core/actions/index.ts
@@ -111,7 +111,7 @@ export abstract class ActionBuilder<T> {
       uniqueKeys.forEach(({ uniqueKey }, index) => {
         const uniqueKeyAssertion = this.session
           .assert(
-            `${proto.target.schema}_${proto.target.name}_assertions_${assertionPrefix}uniqueKey_${index}`,
+            `${assertionPrefix}${proto.target.schema}_${proto.target.name}_assertions_uniqueKey_${index}`,
             dataform.ActionConfig.AssertionConfig.create({ filename: proto.fileName })
           )
           .query(ctx =>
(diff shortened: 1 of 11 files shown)


core/actions/index_test.ts
```diff
@@ -128,70 +132,73 @@ export const exampleBuiltInAssertionsAsYaml = {
...
-      dependencyTargets: [
-        {
+  outputAssertions: (builtinAssertionPrefix?: string) => {
+    const assertionPrefix = !!builtinAssertionPrefix ? `${builtinAssertionPrefix}_` : "";
+    return [
+      {
+        target: {
+          database: "defaultProject",
+          schema: "defaultDataset",
+          name: `dataset_name_assertions_${assertionPrefix}uniqueKey_0`
```
kolina, line 142: nit: I'd have a separate test, because you're kinda repeating code logic here

kolina (changes requested):

core/actions/index.ts
```diff
@@ -111,7 +111,7 @@ export abstract class ActionBuilder<T> {
       uniqueKeys.forEach(({ uniqueKey }, index) => {
         const uniqueKeyAssertion = this.session
           .assert(
-            `${proto.target.schema}_${proto.target.name}_assertions_uniqueKey_${index}`,
+            `${proto.target.schema}_${proto.target.name}_assertions_${assertionPrefix}uniqueKey_${index}`,
```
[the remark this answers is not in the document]
Ceridan: Done.

core/actions/index_test.ts
Ceridan: Done.

kolina (approved):

Issue #1945, opened 2025-03-26
[Bug] Dataform CLI does not fail when running non-existent actions


HHammond:
Dataform CLI does not report an error when asked to run tasks which do not exist

E.g.

```
dataform run . --actions this_action_does_not_exist
Compiling...

Compiled successfully.

Running...
```

Ceridan:
Closed by #1965. Will be released in 3.0.21.

PR #1965, opened 2025-04-30
fix(actions): gracefully exit when no actions to run

ashish10alex:
Fixes: https://github.com/dataform-co/dataform/issues/1945

Show user a log message that there is no actions to run instead of prematurely showing running

Tests

- [x] `./scripts/lint` passes

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -532,11 +532,6 @@ export function runCli() {
             printSuccess("Unit tests completed successfully.\n");
           }
 
-          if (argv[dryRunOptionName]) {
-            print("Dry running (no changes to the warehouse will be applied)...");
-          } else {
-            print("Running...\n");
-          }
           let bigqueryOptions: {} = {
             actionRetryLimit: argv[actionRetryLimitName]
           };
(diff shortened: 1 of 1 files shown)


cli/index.ts
```diff
@@ -555,6 +550,18 @@ export function runCli() {
           executionGraph.actions.forEach(action => {
             actionsByName.set(targetAsReadableString(action.target), action);
           });
+
+          if (actionsByName.size === 0) {
+            print("No actions to run...\n");
+            return 0;
+          } 
+
+          if (argv[dryRunOptionName]) {
```
Ceridan, line 559: I think instead of moving this code, we can move the code on lines to line 540 and then return early if there is no actions to execute.

The idea is to do all this check and print something to the user before we actually create a runner which starts execution.

cli/index.ts
```diff
@@ -555,6 +550,18 @@ export function runCli() {
           executionGraph.actions.forEach(action => {
             actionsByName.set(targetAsReadableString(action.target), action);
           });
+
+          if (actionsByName.size === 0) {
+            print("No actions to run...\n");
```
Ceridan, line 555: nit: Let's put a single dot, it is the end of sentence and no process expected.

Ceridan (changes requested):

cli/index.ts
ashish10alex: certainly, changed to `.` .  Also, I was not sure if you wanted the text to be `No process expected` ? I have kept it as `No actions to run.` as it was in line with `action` keyword being used when the model that are ran are listed.

Ceridan: LGTM

cli/index.ts
Ceridan: @ashish10alex, it is pretty close to what I thinking about. Could we please move the block starting with `if (argv[dryRunOptionName]) { ... }` before the runner creation as well.

My idea is that we need to decide first do we want to run anything or not and output our intent to the user like "Running..." and then start the heavy task which is happen when runner is created.

ashish10alex: oh yes, thanks @Ceridan  for the further clarification. I have made the requested change now

Ceridan: Thank you! The code looks good to me. I will re-run the test pipeline and merge it.

Ceridan (approved):

Issue #1974, opened 2025-05-22
Feature request: Add --quiet flag to dataform compile


quassy:
Right now the output of `dataform compile` is overly verbose if you have a lot of models and/or errors.

```
$ dataform compile
Compiling...

Compiled 314 action(s).
82 dataset(s):
  dataform_inputs.customer_postcodes [table]
  dataform_inputs.high_revenue_postcodes [table]
  dataform_inputs.known_postcodes [table]
  ...
162 assertion(s):
  dataform_assertions.dataform_inputs_customer_postcodes_assertions_uniqueKey_0
  dataform_assertions.dataform_inputs_customer_postcodes_assertions_rowConditions
  dataform_assertions.dataform_inputs_high_revenue_postcodes_assertions_uniqueKey_0
  dataform_assertions.dataform_inputs_high_revenue_postcodes_assertions_rowConditions
  dataform_assertions.dataform_inputs_known_postcodes_assertions_uniqueKey_0
  dataform_assertions.dataform_inputs_known_postcodes_assertions_rowConditions
  ...
70 operation(s):
  udf.check_postcode
  ...

  Compilation errors:
  definitions/filtered_postcodes.sqlx: ReferenceError: ref is not defined
    at /var/folders/57/q_px7w_92ndg_jnpz0rts7sc0000gn/T/tmp-77098-HGfG1e0BXUBx/definitions/filtered_postcodes.sqlx:14:11
    at VM2 Wrapper.apply (/opt/homebrew/lib/node_modules/@dataform/cli/node_modules/vm2/lib/bridge.js:485:11)
    at NodeVM.run (/opt/homebrew/lib/node_modules/@dataform/cli/node_modules/vm2/lib/nodevm.js:497:23)
    at CustomResolver.loadJS (/opt/homebrew/lib/node_modules/@dataform/cli/node_modules/vm2/lib/resolver.js:210:6)
    at CustomResolver.loadJS (/opt/homebrew/lib/node_modules/@dataform/cli/node_modules/vm2/lib/resolver-compat.js:63:65)
    at /opt/homebrew/lib/node_modules/@dataform/cli/node_modules/vm2/lib/resolver.js:84:14
    at VM2 Wrapper.apply (/opt/homebrew/lib/node_modules/@dataform/cli/node_modules/vm2/lib/bridge.js:485:11)
[...]

PR #1976, opened 2025-05-23
feat(cli): add --quiet flag to compile argument

ashish10alex:
Solves: https://github.com/dataform-co/dataform/issues/1974

Adds `--quiet` option to `dataform compile` cli 

Tests:

- [x] Only shows high level error information when `--quiet` option is used, such that It is sufficient for the user to identify the issue. Please see screenshot 
- [x] Able to see stderr when using `--quiet` option 
- [x] Output as before when `--quiet` option is not used
- [x] `./scripts/lint` passes


cli/console.ts
```diff
@@ -129,29 +129,35 @@ export function printCompiledGraph(graph: dataform.ICompiledGraph, asJson: boole
...
-          `${datasetString(
-            compiledTable.target,
-            tableTypeEnumToString(compiledTable.enumType),
-            compiledTable.disabled
-          )}`,
-          1
-        );
-      });
+      writeStdOut(`${graph.tables.length} dataset(s)${quietCompilation ? "." : ":"}`);
+      if(quietCompilation === false){
```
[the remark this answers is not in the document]
ashish10alex: done

cli/console.ts
```diff
@@ -129,29 +129,35 @@ export function printCompiledGraph(graph: dataform.ICompiledGraph, asJson: boole
...
+          });
+      }
     }
     if (graph.assertions && graph.assertions.length) {
-      writeStdOut(`${graph.assertions.length} assertion(s):`);
-      graph.assertions.forEach(assertion => {
-        writeStdOut(assertionString(assertion.target, assertion.disabled), 1);
-      });
+      writeStdOut(`${graph.assertions.length} assertion(s)${quietCompilation ? "." : ":"}`);
+      if(quietCompilation === false){
```
Ceridan: `!quiteCompilation`?

cli/console.ts
```diff
@@ -129,29 +129,35 @@ export function printCompiledGraph(graph: dataform.ICompiledGraph, asJson: boole
...
+          });
+      }
     }
     if (graph.operations && graph.operations.length) {
-      writeStdOut(`${graph.operations.length} operation(s):`);
-      graph.operations.forEach(operation => {
-        writeStdOut(operationString(operation.target, operation.disabled), 1);
-      });
+      writeStdOut(`${graph.operations.length} operation(s)${quietCompilation ? "." : ":"}`);
+      if(quietCompilation === false){
```
Ceridan: `!quiteCompilation`?

cli/console.ts
```diff
@@ -162,7 +168,7 @@ export function printCompiledGraphErrors(graphErrors: dataform.IGraphErrors) {
     graphErrors.compilationErrors.forEach(compileError => {
       writeStdErr(
         `${calloutOutput(compileError.fileName)}: ${errorOutput(
-          compileError.stack || compileError.message
+          compileError.message || compileError.stack
```
Ceridan: I do not think that we could change the current behavior here. I propose to change it only for the `quite` flag. Having a stack trace could be very handy to debug things, and I propose to keep it.

cli/console.ts
```diff
@@ -117,7 +117,7 @@ export function printInitCredsResult(writtenFilePath: string) {
   writeStdOut("To change connection settings, edit this file directly.");
 }
 
-export function printCompiledGraph(graph: dataform.ICompiledGraph, asJson: boolean) {
+export function printCompiledGraph(graph: dataform.ICompiledGraph, asJson: boolean, quietCompilation: boolean) {
```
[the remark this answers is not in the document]
Ceridan: Sorry for the late reply and thanks for the change! Looking at the screenshot I think the biggest problem we have with this approach - it is not actionable. For example, the first error: "Unexpected identifier". It is very hard to say where it is happen. Could you please check if there is an easy way (without manually parsing stacktrace) to get a line number of error?

cli/console.ts
ashish10alex: done

cli/console.ts
ashish10alex: done

cli/console.ts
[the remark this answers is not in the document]
ashish10alex: Hi @Ceridan , sorry was on annual leave and just getting back to this. I agree with you assessment. I have now modified the code to only change the behaviour of what happens when we use `--quiet` flag. 

---

**More informative error:**  Not sure, if what I have done is completely without parsing the stack trace. However, based on my experience, I have identified that we can use certain keywords to identify some generic errors and based on that we can show user specific portion of the stderr. Please see the screenshot in which I have covered verity of standard compilation errors we get. 

--


**Line number of the error:**  Even in the stack trace the line number we get as where the error, some how does not make sense. I think there is some compilation etc happening in the background which writes a version of the sqlx file to a staging area for compilation , in the example in the screenshot that is `/var/folders/vz/mwhrymmx5xs3j45cv039kv_w0000gn/T/tmp-64750-gWnt7gmyeQAY/definitions/0100_CLUBS.sqlx:9`. However, I can not seem to get to this file in my file system post running of the compilation.

cli/console.ts
ashish10alex: Yes, agree. I have now retained the original behaviour when `--quiet` flag is not used

cli/console.ts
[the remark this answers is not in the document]
Ceridan: I liked the idea, let's keep it like you proposed. Seems like we could provide the bare minimum of details for the quiet mode, and if needed the user can run the compilation of the action in normal mode for more details.

cli/console.ts
```diff
@@ -129,40 +129,51 @@ export function printCompiledGraph(graph: dataform.ICompiledGraph, asJson: boole
...
     }
   }
 }
 
-export function printCompiledGraphErrors(graphErrors: dataform.IGraphErrors) {
+export function printCompiledGraphErrors(graphErrors: dataform.IGraphErrors, quietCompilation: boolean) {
   if (graphErrors.compilationErrors && graphErrors.compilationErrors.length > 0) {
     printError("Compilation errors:", 1);
-    graphErrors.compilationErrors.forEach(compileError => {
+    graphErrors.compilationErrors.forEach((compileError, index) => {
```
Ceridan: We do not use `index` currently, right? Maybe we do not need it here then.

cli/console.ts
```diff
@@ -129,40 +129,51 @@ export function printCompiledGraph(graph: dataform.ICompiledGraph, asJson: boole
...
+      let errorMeta = ""
+
+      if (compileError.message.includes("Unexpected identifier")) {
+        errorMeta = compileError.stack.split("\n").slice(0, 3).join("\n");
+      }
```
Ceridan: I propose to extract it to a helper private method and call it directly on line 176.

Additionally, we need to add a safety check in the method: make sure that compileError.stack is actually defined and not empty.

Also I propose to left a small comment on what we are doing here.

Ceridan (changes requested):
It looks nice, I left a few small comments to make it a little bit nicer and more robust. Thanks for all your effort you put on the CL so far.

cli/console.ts
ashish10alex: thanks @Ceridan  for the feedback  !
 
1. Created a method `formatStackTraceForQuietCompilation` which takes `compileError` object as input.
2. Additionally checks have been added so that the method does not fail
3. Added comment to describe the function in (1). Let me know if that is too verbose. I thought adding a descriptive example would make the intension clear.

cli/console.ts
ashish10alex: removed `index` . I was planning to use it but forgot to remove later :)

cli/console.ts
```diff
@@ -129,40 +145,47 @@ export function printCompiledGraph(graph: dataform.ICompiledGraph, asJson: boole
...
     }
   }
 }
 
-export function printCompiledGraphErrors(graphErrors: dataform.IGraphErrors) {
+export function printCompiledGraphErrors(graphErrors: dataform.IGraphErrors, quietCompilation: boolean) {
   if (graphErrors.compilationErrors && graphErrors.compilationErrors.length > 0) {
     printError("Compilation errors:", 1);
     graphErrors.compilationErrors.forEach(compileError => {
+      const errorMeta = formatStackTraceForQuietCompilation(compileError);
```
Ceridan: I propose to inline it to line 188. The idea is to call it only if we are in `quiet` mode.

cli/console.ts
```diff
@@ -92,6 +92,22 @@ export function print(text: string) {
   writeStdOut(text);
 }
 
+export function formatStackTraceForQuietCompilation(compileError: dataform.ICompilationError): string {
```
Ceridan: I have a few suggestions about this one:

1. Maybe let's move the function closer to `printCompiledGraphErrors` because it is used there?
2. We do not need `export` keyword here, it's internal function.
3. The formatting inside function seems broken.

Ceridan (changes requested):
I left 2 small comments and will merge after it. You did a great job so fat. Thank you!

cli/console.ts
ashish10alex: done

cli/console.ts
ashish10alex: Addressed (1), (2). For (3), from what I could see, the issue was tabs vs 2 spaces ? I have attempted to change that. Please let me know if that is not what it was

Ceridan: We use 2 spaces at the moment. LGTM, thanks!

ashish10alex:
> LGTM! Maybe it will require merge the main after the other PR will be submitted. Will see after I merge the other one. Thanks for the great job!

Thanks for taking your time for the review and providing great feedback. I have merged with the main branch

PR #1991, opened 2025-06-27
Fail compilation if onSchemaChange is set on incremental table.

### LINKED ISSUE #1906: Bug - OnSchemaChange doesnt work as expected in Bigquery Dataform
Steps Taken to Address the Issue:
1)Core Version Update: Upgraded the core version to 3.0.12 in workflow_settings.yaml.

2)Package Installation: Installed necessary dependencies.

3)Initial SQLX Configuration: Created a SQLX file with the following configuration:
```
config {
    type: "incremental",
    schema: "test",
    onSchemaChange: "EXTEND",
}
select 1 as a, 2 as b, "asd" as c;
```
Column Addition: Introduced a new column d to the SQLX configuration:

```
config {
    type: "incremental",
    schema: "test",
    onSchemaChange: "EXTEND",
}
select 1 as a, 2 as b, "asd" as c, "test" as d;
```

Expected vs. Actual Outcome:

- Expected Behavior: The schema should recognize the newly added column d and automatically update the table structure.

- Actual Behavior:  the table schema was not updated accordingly.

- Additional Issue: When adding uniqueKey, and dropped a column ,an error occurred `Query error: Name c not found inside S `

Root Cause Analysis & Suggested Fixes:

- The current logic retrieves column names from INFORMATION_SCHEMA.COLUMNS, which does not immediately reflect newly added columns / removed columns.

Ceridan:
While Incremental Table Schema Update is not officially released on Dataform GCP, we need to fail the compilation if `onSchemaChange` field is set on incremental table config.

Resolves #1906.

### DIFF
--- a/core/actions/incremental_table.ts
+++ b/core/actions/incremental_table.ts
@@ -607,6 +607,12 @@ export class IncrementalTable extends ActionBuilder<dataform.Table> {
       checkConfigAdditionalOptionsOverlap(config, this.session);
     }
 
+    // Temporary fail the compilation if non-default `onSchemaChange` value is provided until the official release of the feature on Dataform GCP.
+    const onSchemaChange = this.mapOnSchemaChange(config.onSchemaChange)
+    if (onSchemaChange != dataform.OnSchemaChange.IGNORE) {
+      throw new Error(`OnSchemaChange is not supported yet. It will be available as soon as Dataform GCP will officially release this feature.`);
+    }
+
     return config;
   }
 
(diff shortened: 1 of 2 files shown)


kolina:
> @kolina, this is a good question. The proper way of doing this is to introduce the feature flag system on OSS code but unfortunately we do not have a time budget to do this. There are following options we have:
> 
> 1. I think we could test it by using the older version of Dataform Core package (up to 3.0.24) before we introduce this error.
> 2. Maybe, we can do nothing. I mean, yes it is not very convenient at the moment that we have a config which is technically available to the users, but it wasn't announced in any public documentation including https://cloud.google.com/dataform/docs/reference/dataform-core-reference . So, yes, we need to apologize in the issue, and close the issue as soon as we will release the feature.
> 
> WDYT?

WDYT about throwing this error in the API instead of modifying `@dataform/core`?

Tuseeq1:
I agree a failure makes most sense as this is a failed state. This way user will be aware at the earliest stage of development that this feature will not work.

kolina:
> My point is that it shouldn't be a warning. It should fail the compilation completely. If I use `onSchemaChange` configuration in one action, I may rely on this changes in further actions. I think it is better to fail for now, and then remove this check after we release it.

Advantage of the API is that you can disable it for all `@dataform/core` why it's not supported (including versions in the past).

Issue #1993, opened 2025-07-01
feat: improve the formatting of dataform run stdout


ashish10alex:
With the introduction of [jobId](https://github.com/dataform-co/dataform/pull/1971) & [Bytes billed](https://github.com/dataform-co/dataform/pull/1971) in the ouput of _dataform run_ command. The stdout looks cluttered and hard to read as everything is put on a single line. 

I would be nice if the output is formatted better. One simple solution would be to move the metadata to next line and indent it with tab like yaml format.

Ceridan:
PR was merged, I think we can close this one.

PR #1994, opened 2025-07-01
feat: improve formatting for `dataform run` stdout

ashish10alex:
Solves: https://github.com/dataform-co/dataform/issues/1993

Moves the metadata (jobId, Bytes billed) of dataform run for table/view/operation to a new line and indented to the right making the output more readable. 

**Tests**

- [x] bazel test //core/... passes
- [x] Linting by running ./scripts/lint


Ceridan (approved):
Thank you for the yet another contribution! :)

Issue #2000, opened 2025-08-06
Dataform CLI ignores impersonation


wintermi:
Rather than use the `--credentials` and provide JSON keys, I was looking to use service account impersonation.  Executing as follows:

```
gcloud config set auth/impersonate_service_account <sa>
dataform run
gcloud config unset auth/impersonate_service_account
```

However the Dataform CLI ignores the impersonation and executes using your core account.

Could we get an option to specify an impersonation account as part of the `run` command?

PR #2001, opened 2025-08-06
feat: add support for service account impersonation

ashish10alex:
Fixes: https://github.com/dataform-co/dataform/issues/2000

**Solution:** 

Update [@google-cloud/bigquery](https://www.npmjs.com/package/@google-cloud/bigquery?activeTab=readme) to a newer version which supports service account impersonation. The version number was set from finding the maximum version of @google-cloud/bigquery such that current minimum Node JS version (16.6.0) does not need to be changed. 

**Tests**

1. Able to create .df-credentials.json after authenticating to service account to impersonate by running the following. Earlier to would throw an error: `The incoming JSON object does not contain a client_email field`

```
gcloud auth application-default login --impersonate-service-account=<service-account-here>
```

CleanShot 2025-08-06 at 16 15 31@2x

2. Able to do dataform run with the impersonated service account 

CleanShot 2025-08-06 at 16 17 02@2x

CleanShot 2025-08-06 at 16 18 09@2x

3. `bazel test //core/...` & `./scripts/lint` passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "@bazel/labs": "^0.42.3",
     "@bazel/rollup": "^3.0.0",
     "@bazel/typescript": "^3.0.0",
-    "@google-cloud/bigquery": "^5.6.0",
+    "@google-cloud/bigquery": "7.1.1",
     "@google-cloud/storage": "^5.8.2",
     "@rollup/plugin-node-resolve": "^7.1.3",
     "@types/chai": "^4.1.7",
--- a/yarn.lock
+++ b/yarn.lock
@@ -79,37 +79,22 @@
   resolved "https://registry.npmjs.org/@discoveryjs/json-ext/-/json-ext-0.5.7.tgz#1d572bfbbe14b7704e0ba0f39b74815b84870d70"
   integrity sha512-dBVuXR082gk3jsFp7Rd/JI4kytwGHecnCoTtXFb7DB6CNHp4rg5k1bhg0nWdLGLnOV71lmDzGQaLMy8iPLY0pw==
 
-"@google-cloud/bigquery@^5.6.0":
-  version "5.6.0"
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
     "@bazel/labs": "^0.42.3",
     "@bazel/rollup": "^3.0.0",
     "@bazel/typescript": "^3.0.0",
-    "@google-cloud/bigquery": "^5.6.0",
+    "@google-cloud/bigquery": "7.1.1",
```
kolina: I think it shouldn't be pinned to a specific version similar to other dependencies

kolina (changes requested):

package.json
ashish10alex: Thanks @kolina for the review. Using `^7.1.1` results in the following error when doing ` bazel run @nodejs//:yarn add @google-cloud/bigquery@^7.1.1` 

```bash
error @google-cloud/promisify@4.1.0: The engine "node" is incompatible with this module. Expected version ">=18". Got "16.16.0"
```

However, we can use `~` to allow patch updates only (minor and major fixed). This installs without any issues. I have made this change in the lates commit. 

Happy to change based on feedback. Trying a lower major version is another option.

kolina: @Ceridan, can you please take a look here?

kolina (approved):

Ceridan (commented):
Unfortunately, we need to put it on hold for a while. We cannot easily upgrade to @google-cloud/bigquery@7.1.1 because it is not available on our npm mirror at the moment.

Additionally, there is a an ongoing work to update the whole toolchain for Dataform which will make it possible to upgrade NodeJS version and other dependencies to up to date versions.

wintermi:
Any further updates on when this will be merged?

ashish10alex:
> @ashish10alex, can you please sync this PR to the latest `main`? After this I'll re-run tests and if they're successful, I'll merge this PR.

@kolina done

mganeshphani:
@kolina thankyou for getting this in.. when can we expect a new release with this fix?

PR #2007, opened 2025-08-25
Prototype runtime parameters.

Ceridan:


core/actions/table.ts
kvlaskarolina, line 1: Should we also add 'inputParams: [],' in proto in core/actions/incremental_table.ts?

Ceridan: Yes, it would be better to define a default as empty array on every action.

apilaskowski:
@Ceridan do you know if this PR is going to be finished? should we close it?

PR #2009, opened 2025-08-29
Add Iceberg fields to compilation logic

leonarda-pribanic:
This change adds Iceberg fields to config.proto and core.proto and ensures they are correctly parsed and compiled.

### DIFF
--- a/core/actions/table.ts
+++ b/core/actions/table.ts
@@ -184,11 +184,18 @@ export class Table extends ActionBuilder<dataform.Table> {
       labels: config.labels,
       partitionExpirationDays: config.partitionExpirationDays,
       requirePartitionFilter: config.requirePartitionFilter,
-      additionalOptions: config.additionalOptions
+      additionalOptions: config.additionalOptions,
+      ...(config.iceberg ? {
+        connection: getConnectionForIcebergTable(config.iceberg.connection),
+        fileFormat: getFileFormatValueForIcebergTable(config.iceberg.fileFormat?.toString()),
+        tableFormat: dataform.TableFormat.ICEBERG,
+        storageUri: getStorageUriForIcebergTable(
+          config.iceberg.bucketName,
+          getEffectiveTableFolderSubpath(this.proto.target.schema, this.proto.target.name, config.iceberg.tableFolderSubpath),
+          config.iceberg.tableFolderRoot,
+        ),
+      } : {}),
     });
-    if (config.iceberg) {
-      this.iceberg(config.iceberg);
-    }
 
     return this;
   }
@@ -279,27 +286,6 @@ export class Table extends ActionBuilder<dataform.Table> {
   public postOps(posts: Contextable<ITableContext, string | string[]>) {
     this.contextablePostOps.push(posts);
     return this;
-  }
-
-
-  /**
-   * Sets the configuration options for the creation of Apache Iceberg tables.
-   * @param icebergOptions Iceberg options provided in the iceberg {} subblock of the config file.
-   */
-  public iceberg(
-    icebergOptions: dataform.ActionConfig.IIcebergTableConfig,
-  ) {
-    if (!this.proto.bigquery) {
-      this.proto.bigquery = dataform.BigQueryOptions.create();
-    }
-    this.proto.bigquery.tableFormat = dataform.TableFormat.ICEBERG;
-    this.proto.bigquery.fileFormat = getFileFormatValueForIcebergTable(icebergOptions?.fileFormat?.toString());
-    this.proto.bigquery.connection = getConnectionForIcebergTable(icebergOptions.connection);
-    this.proto.bigquery.storageUri = getStorageUriForIcebergTable(
-      icebergOptions.bucketName,
-      getEffectiveTableFolderSubpath(this.proto.target.schema, this.proto.target.name, icebergOptions.tableFolderSubpath),
-      icebergOptions.tableFolderRoot,
-    );
   }
 
   /**
--- a/core/actions/incremental_table.ts
+++ b/core/actions/incremental_table.ts
@@ -191,11 +191,18 @@ export class IncrementalTable extends ActionBuilder<dataform.Table> {
       labels: config.labels,
       partitionExpirationDays: config.partitionExpirationDays,
       requirePartitionFilter: config.requirePartitionFilter,
-      additionalOptions: config.additionalOptions
+      additionalOptions: config.additionalOptions,
+      ...(config.iceberg ? {
+        connection: getConnectionForIcebergTable(config.iceberg.connection),
+        fileFormat: getFileFormatValueForIcebergTable(config.iceberg.fileFormat?.toString()),
+        tableFormat: dataform.TableFormat.ICEBERG,
+        storageUri: getStorageUriForIcebergTable(
+          config.iceberg.bucketName,
+          getEffectiveTableFolderSubpath(this.proto.target.schema, this.proto.target.name, config.iceberg.tableFolderSubpath),
+          config.iceberg.tableFolderRoot,
+        ),
+      } : {}),
     });
-    if (config.iceberg) {
-      this.iceberg(config.iceberg);
-    }
     this.proto.onSchemaChange = this.mapOnSchemaChange(config.onSchemaChange);
 
     return this;
@@ -287,26 +294,6 @@ export class IncrementalTable extends ActionBuilder<dataform.Table> {
   public postOps(posts: Contextable<ITableContext, string | string[]>) {
     this.contextablePostOps.push(posts);
     return this;
-  }
-
-  /**
-   * Sets the configuration options for the creation of Apache Iceberg tables.
-   * @param icebergOptions Iceberg options provided in the iceberg {} subblock of the config file.
-   */
-  public iceberg(
-    icebergOptions: dataform.ActionConfig.IIcebergTableConfig,
-  ) {
-    if (!this.proto.bigquery) {
-      this.proto.bigquery = dataform.BigQueryOptions.create();
-    }
-    this.proto.bigquery.tableFormat = dataform.TableFormat.ICEBERG;
-    this.proto.bigquery.fileFormat = getFileFormatValueForIcebergTable(icebergOptions?.fileFormat?.toString());
-    this.proto.bigquery.connection = getConnectionForIcebergTable(icebergOptions.connection);
-    this.proto.bigquery.storageUri = getStorageUriForIcebergTable(
-      icebergOptions.bucketName,
-      getEffectiveTableFolderSubpath(this.proto.target.schema, this.proto.target.name, icebergOptions.tableFolderSubpath),
-      icebergOptions.tableFolderRoot,
-    );
   }
 
   /**


core/actions/table_test.ts
```diff
@@ -266,4 +267,236 @@ ${exampleBuiltInAssertionsAsYaml.inputActionConfigBlock}
...
+            tableFolderRoot: "my-root",
+            tableFolderSubpath: "my-subpath",
+        }`,
+        expected: {},
+        expectError: "The connection must be in the format `{project}.{location}.{connection_id}` or `projects/{project}/locations/{location}/connections/{connection_id}`, or be set to `DEFAULT`.",
+      },
+      {
+        testName: "invalid file format",
+        configBlock: `
+        fileFormat: "AVRO",
```
fafsdafsafs: could you please add a test that checks when fileformat is empty

protos/configs.proto
```diff
@@ -189,6 +218,20 @@ message ActionConfig {
...
+    // Supported file formats for BigQuery tables.
+    enum FileFormat {
+      // Default value.
+      FILE_FORMAT_UNSPECIFIED = 0;
+      // Apache Parquet format.
+      PARQUET = 1;
+    }
+
+    // The file format for the BigQuery table.
+    FileFormat file_format = 21;
```
fafsdafsafs: let's decide if we want to put file_format under IcebergTableConfig or not

leonarda-pribanic: We decided to not have it under IcebergTableConfig because this will make it easier to reuse for other table formats in the future.

core/actions/table_test.ts
leonarda-pribanic: Done.

protos/configs.proto
```diff
@@ -189,6 +221,17 @@ message ActionConfig {
     // the  action depends on data from a source which has not been declared as
     // a dependency.
     bool hermetic = 20;
+
+    // Supported file formats for BigQuery tables.
+    enum FileFormat {
```
Ceridan: `FileFormat` logically belongs to `IcebergTableConfig`. I propose to move the enum definition there.

protos/configs.proto
```diff
@@ -112,6 +112,38 @@ message ActionConfig {
     repeated string row_conditions = 4;
   }
 
+  message IcebergTableConfig {
```
Ceridan, line 115: Based on the agreement in design document we will not have a separate config for IcebergTable but it will go to bigquery section of existing table, right? I propose to make it a little bit more flexible:
Could we call it `ExternalTableConfig` and add a field `TableFormat` here as well? It will allow us to reuse this config later when we decide to support Apache Hudi or Delta Lake. WDYT?

core/utils.ts
```diff
@@ -245,6 +245,110 @@ export function validateQueryString(session: Session, query: string, filename: s
   }
 }
 
+/**
+ * Checks if the Cloud Resource connection has a valid format.
+ * @param session Used to throw a compile error if connection is not valid.
+ * @param connection String to be validated.
+ * @param filename Used to throw a compile error if connection is not valid.
+ */
+export function validateConnectionFormat(
```
Ceridan, line 252: New functions here require some unit testing especially because we are relying on magic regexes. Eventually I plan to cover all utils with unit tests.

core/utils.ts
```diff
@@ -245,6 +245,110 @@ export function validateQueryString(session: Session, query: string, filename: s
...
+ * @returns Storage URI used when creating an Iceberg table.
+ */
+export function getStorageUriForIcebergTable(
+  datasetName: string,
+  tableName: string,
+  bucketName?: string,
+  tableFolderRoot?: string,
+  tableFolderSubpath?: string,
+): string | undefined {
+  const effectiveTableFolderRoot = tableFolderRoot || '_dataform';
```
Ceridan: How the default value was chosen here?

core/utils.ts
```diff
@@ -245,6 +245,110 @@ export function validateQueryString(session: Session, query: string, filename: s
...
+    );
+  }
+}
+
+/**
+ * Returns a file format for an Iceberg table, as specified in the user's config file.
+ * @param configFileFormat User-provided file format, if it exists.
+ * @return File format used when creating an Iceberg table.
+ */
+export function getFileFormatValueForIcebergTable(
```
Ceridan, line 298: I propose to have a logic here like: if FileFormat is defined and and not unknown, then use it, otherwise fallback to parquet.

protos/core.proto
```diff
@@ -52,6 +52,16 @@ message Target {
   bool include_dependent_assertions = 4;
 }
 
+enum FileFormat {
+  UNKNOWN = 0;
```
Ceridan: Let's follow the same pattern on all enums. Should be `FILE_FORMAT_UNSPECIFIED`.

protos/core.proto
```diff
@@ -52,6 +52,16 @@ message Target {
...
+  PARQUET = 1;
+}
+
+enum TableFormat {
+  UNKNOWN = 0;
```
Ceridan: Let's follow the same pattern on all enums. Should be `TABLE_FORMAT_UNSPECIFIED`.

Ceridan (changes requested):

protos/configs.proto
fafsdafsafs: Based on updated design we will have separate config block for IcebergTable that will contain all new fields.

version.bzl
```diff
@@ -1 +1 @@
-DF_VERSION = "3.0.27"
+DF_VERSION = "3.0.28"
```
Ceridan: version.bzl and yarn.lock changes shouldn't be in this PR. I believe it is some issue with merging with main branch. Could you, please, try to merge the main branch to this one once again?

core/actions/table.ts
```diff
@@ -268,6 +285,47 @@ export class Table extends ActionBuilder<dataform.Table> {
...
   }
 
+
+  /**
+   * Sets the configuration options for the creation of Apache Iceberg tables.
+   * @param session Used to throw a compile error.
+   * @param filename Used to throw a compile error.
+   * @param icebergOptions Iceberg options provided in the iceberg {} subblock of the config file.
+   */
+  public iceberg(
```
Ceridan: Why do we need this temporary state - contextableIcbergOpts? For `preOps` and `postOps` it is needed because they are not part of the config. I propose to set necessary fields on BigQueryOptions directly after we initialize the BigQueryOptions on line 188

Because we also have a validation for icberg config I am fine to extract it to a helper method but we still do not need intermediate state, and it should be a private method which accepts `bigquery` and `config` objects and enriches `bigquery` with additional fields. The overall solution will become simpler.

core/utils.ts
```diff
@@ -245,6 +245,129 @@ export function validateQueryString(session: Session, query: string, filename: s
...
+ * tableFolderRoot, and tableFolderSubpath provided in the config file. Returns
+ * undefined if a complete URI cannot be formed.
+ * @param datasetName Might be used to construct the storage URI if no alternative value is provided.
+ * @param tableName Might be used to construct the storage URI if no alternative value is provided.
+ * @param bucketName User-provided Cloud Storage bucket name, if it exists.
+ * @param tableFolderRoot User-provided table folder root, if it exists.
+ * @param tableFolderSubpath User-provided table folder subpath, if it exists.
+ * @returns Storage URI used when creating an Iceberg table.
+ */
+export function getStorageUriForIcebergTable(
```
Ceridan, line 333: I am concerned about the API we expose here. We need to provide `datasetName` and `tableName` as mandatory arguments just to use them as a fallback in case `tableFolderSubpath` is not provided. Additionally, it is not clear what behavior do we expect if `bucketName` is null.

I propose to refactor it like this:

```typescript
export function getStorageUriForIcebergTable(
  bucketName: string,
  tableFolderSubpath: string,
  tableFolderRoot: string = "_dataform",
) {...}
```

And force the caller to make a decision on what should be provided as `tableFolderSubpath`.

core/actions/table.ts
```diff
@@ -557,15 +651,18 @@ export class Table extends ActionBuilder<dataform.Table> {
...
+            ...strictKeysOf<ILegacyBigQueryOptions>()([
+              "partitionBy",
+              "clusterBy",
+              "updatePartitionFilter",
+              "labels",
+              "partitionExpirationDays",
+              "requirePartitionFilter",
+              "additionalOptions",
           ]),
+            "iceberg"
```
Ceridan: It should be inside the array of `strictKeysOf`

core/utils.ts
```diff
@@ -245,6 +245,129 @@ export function validateQueryString(session: Session, query: string, filename: s
...
+  // Example: projects/my-project/locations/us-central1/connections/my-connection
+  // Substrings cannot contain '/'.
+  const resourcePattern =
+    /^projects\/[^/]+\/locations\/[^/]+\/connections\/[^/]+$/;
+
+  const isValidFormat =
+    dotPattern.test(connection) || resourcePattern.test(connection);
+
+  if (connection !== 'DEFAULT' && !isValidFormat) {
+    session.compileError(
```
Ceridan: It is fine to just throw a normal error, I believe it will be wrapped by session.compileError on a few layers above. The same for other places.

core/utils.ts
```diff
@@ -245,6 +245,129 @@ export function validateQueryString(session: Session, query: string, filename: s
...
+ * @param filename Used to throw a compile error if fileFormat is not valid.
+ * @param configFileFormat User-provided file format, if it exists.
+ * @return File format used when creating an Iceberg table.
+ */
+export function getFileFormatValueForIcebergTable(
+  session: Session,
+  filename: string,
+  configFileFormat?: string,
+): dataform.FileFormat {
+  if (configFileFormat === undefined || configFileFormat === "") {
```
Ceridan: Here could be a `null` value either, let make it simpler and check like: 

```typescript
if (!configFileFormat) { ... }
```

core/utils_test.ts
```diff
@@ -0,0 +1,169 @@
+import { validateConnectionFormat, validateStorageUriFormat } from './utils';
```
Ceridan: I would at at least 1 "sunny day" check for `getSomething` functions which were introduced in this PR.

core/actions/table.ts
leonarda-pribanic: This throws an error because iceberg is not part of the LegacyBigQueryOptions. 

I changed this part of the code to validate if fields are inside BigQueryOptions instead of LegacyBigQueryOptions, I hope that's fine.

core/actions/table.ts
```diff
@@ -268,6 +281,27 @@ export class Table extends ActionBuilder<dataform.Table> {
     return this;
   }
 
+
+  /**
+   * Sets the configuration options for the creation of Apache Iceberg tables.
+   * @param icebergOptions Iceberg options provided in the iceberg {} subblock of the config file.
+   */
+  public iceberg(
```
Ceridan: I think this method shouldn't be public. Additionally, seems like now all what it is doing is just set fields on proto.bigquery without any additional work. I propose, to inline it to the bigquery definition on line 181. I mean, extend the code there like:

```typescript
    this.bigquery({
      partitionBy: config.partitionBy,
      clusterBy: config.clusterBy,
      labels: config.labels,
      partitionExpirationDays: config.partitionExpirationDays,
      requirePartitionFilter: config.requirePartitionFilter,
      additionalOptions: config.additionalOptions,
      tableFormat: dataform.TableFormat.ICEBERG,
      fileFormat: getFileFormatValueForIcebergTable(...),
      connection: getConnectionForIcebergTable(...),
      storageUri: getStorageUriForIcebergTable(...),
    });
```

core/actions/table_test.ts
```diff
@@ -266,4 +267,256 @@ ${exampleBuiltInAssertionsAsYaml.inputActionConfigBlock}
...
+            const error = result.compile.compiledGraph.graphErrors.compilationErrors[0];
+            expect(error.message).contains(testCase.expectError);
+            expect(error.fileName).equals(`definitions/${params.filename}`);
+          } else {
+            expect(result.compile.compiledGraph.graphErrors.compilationErrors).deep.equals([]);
+            const compiledTable = result.compile.compiledGraph.tables[0];
+            expect(compiledTable.target.name).equals(testCase.expected.target.name);
+            expect(compiledTable.target.schema).equals(testCase.expected.target.schema);
+            expect(compiledTable.target.database).equals(testCase.expected.target.database);
+            expect(compiledTable.bigquery.tableFormat).equals(dataform.TableFormat.ICEBERG);
```
Ceridan: I would compare the whole bigquery object. Additionally, I would add a test which sets something else in the test setup which also lands in bigquery to make sure that it is not overridden.

core/actions/table.ts
```diff
@@ -557,15 +615,9 @@ export class Table extends ActionBuilder<dataform.Table> {
...
-            "partitionBy",
-            "clusterBy",
-            "updatePartitionFilter",
-            "labels",
-            "partitionExpirationDays",
-            "requirePartitionFilter",
-            "additionalOptions"
-          ]),
+          [
+            ...strictKeysOf<dataform.IBigQueryOptions>()([]),
```
Ceridan: I slightly concerned about this change. I believe we need to have it to preserve backward compatibility with a legacy format of the config.

core/actions/table.ts
leonarda-pribanic: I thought about including the Iceberg fields in this method but hesitated because it's deprecated and the Javadoc says to use the fields available on TableConfig directly.

Let me know if you think they should still be included!

core/actions/table.ts
leonarda-pribanic: Makes sense, I dropped the commit.

I think the only way iceberg can be included with other bq options here is if I add it to LegacyBigQueryOptions, and I don't know if that's a good idea.

Let me know what you think!

core/actions/table.ts
Ceridan: Ah, OK, I remembered this one. We do not need to validation of `iceberg` here in this case. In fact the new `iceberg` field is a separate field in the config. it is not part of bigquery. It will be added to bigquery section in the bigquery section after the compilation but on config level it is not in bigquery.

core/actions/table.ts
Ceridan: I think it is fine to set it like I propose. It is deprecated because it is public. It shouldn't be public, it should be private method. But we need some kind of logic there. The idea is that we need to have some method which accepts the config, and assigns all the necessary fields to `this.proto.bigquery`. Ideally all the parsing logic should be inside it, but for now it is fine to just reuse an existing method on line 181 and set a bunch of fields there.

Ceridan: My goal is to avoid of creating new public methods which shouldn't exist on the API level.

PR #2011, opened 2025-09-01
Update test credentials.

Ceridan:
Test credentials rotation.


a2wd:
Unclear what the credential change is (binary file) but since the checks passed, will approve

a2wd (approved):

PR #2013, opened 2025-09-08
Enable setting default values for storageUri substrings when constructing an Iceberg table.

leonarda-pribanic:
This change allows users to set defaultBucketName, defaultTableFolderRoot and defaultTableFolderSubpath in workflow_settings.yaml. These fields are parsed and stored as part of projectConfig. If no alternative values are provided inside the user's config block, we use the defaults defined inside workflow_settings.yaml. If the default values inside workflow_settings.yaml are also not provided, we fall back to preexisting defaults.

### DIFF
--- a/core/actions/incremental_table.ts
+++ b/core/actions/incremental_table.ts
@@ -199,9 +199,9 @@ export class IncrementalTable extends ActionBuilder<dataform.Table> {
         fileFormat: getFileFormatValueForIcebergTable(config.iceberg.fileFormat?.toString()),
         tableFormat: dataform.TableFormat.ICEBERG,
         storageUri: getStorageUriForIcebergTable(
-          getEffectiveBucketName(session.projectConfig.defaultBucketName, config.iceberg.bucketName),
-          getEffectiveTableFolderRoot(session.projectConfig.defaultTableFolderRoot, config.iceberg.tableFolderRoot),
-          getEffectiveTableFolderSubpath(this.proto.target.schema, this.proto.target.name, session.projectConfig.defaultTableFolderSubpath, config.iceberg.tableFolderSubpath),
+          getEffectiveBucketName(session.projectConfig.defaultIcebergConfigs?.defaultBucketName, config.iceberg.bucketName),
+          getEffectiveTableFolderRoot(session.projectConfig.defaultIcebergConfigs?.defaultTableFolderRoot, config.iceberg.tableFolderRoot),
+          getEffectiveTableFolderSubpath(this.proto.target.schema, this.proto.target.name, session.projectConfig.defaultIcebergConfigs?.defaultTableFolderSubpath, config.iceberg.tableFolderSubpath),
         ),
       } : {}),
     });
--- a/core/actions/incremental_table.ts
+++ b/core/actions/incremental_table.ts
@@ -199,9 +199,9 @@ export class IncrementalTable extends ActionBuilder<dataform.Table> {
         fileFormat: getFileFormatValueForIcebergTable(config.iceberg.fileFormat?.toString()),
         tableFormat: dataform.TableFormat.ICEBERG,
         storageUri: getStorageUriForIcebergTable(
-          getEffectiveBucketName(session.projectConfig.defaultIcebergConfigs?.defaultBucketName, config.iceberg.bucketName),
-          getEffectiveTableFolderRoot(session.projectConfig.defaultIcebergConfigs?.defaultTableFolderRoot, config.iceberg.tableFolderRoot),
-          getEffectiveTableFolderSubpath(this.proto.target.schema, this.proto.target.name, session.projectConfig.defaultIcebergConfigs?.defaultTableFolderSubpath, config.iceberg.tableFolderSubpath),
+          getEffectiveBucketName(session.projectConfig.defaultIcebergConfig?.defaultBucketName, config.iceberg.bucketName),
+          getEffectiveTableFolderRoot(session.projectConfig.defaultIcebergConfig?.defaultTableFolderRoot, config.iceberg.tableFolderRoot),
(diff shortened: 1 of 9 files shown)


protos/configs.proto
```diff
@@ -45,6 +45,13 @@ message WorkflowSettings {
 
   // Optional. The prefix to append to built-in assertion names.
   string builtin_assertion_name_prefix = 11;
+
+  // Optional.
```
fafsdafsafs: you can add:
1.  'Optional' to all fields
2. Short description for each field

protos/configs.proto
```diff
@@ -45,6 +45,18 @@ message WorkflowSettings {
 
   // Optional. The prefix to append to built-in assertion names.
   string builtin_assertion_name_prefix = 11;
+
+  // Optional. Bucket name used to construct a storage URI when creating an
```
Ceridan, line 54: I propose to extract these default properties here and in core.proto to a separate message similar to `NotebookRuntimeOptionsConfig`, The justification for this: default fields like `default_location` or `default_dataset` are general defaults for all the actions, however `default_bucket_name` is very specific for a particular type of tables. I propose to denote it by specifying for example `IcebergOptions default_icberg_options`.

core/actions/incremental_table.ts
```diff
@@ -197,9 +199,9 @@ export class IncrementalTable extends ActionBuilder<dataform.Table> {
         fileFormat: getFileFormatValueForIcebergTable(config.iceberg.fileFormat?.toString()),
         tableFormat: dataform.TableFormat.ICEBERG,
         storageUri: getStorageUriForIcebergTable(
-          config.iceberg.bucketName,
-          getEffectiveTableFolderSubpath(this.proto.target.schema, this.proto.target.name, config.iceberg.tableFolderSubpath),
-          config.iceberg.tableFolderRoot,
+          getEffectiveBucketName(session.projectConfig.defaultBucketName, config.iceberg.bucketName),
```
Ceridan: Do we actually need to have validation on this level? My understanding was that `getStorageUriForIcebergTable` will do all the validation of input parameters. Instead of introducing `getEffectiveBucketName, getEffectiveTableFolderRoot, getEffectiveTableFolderSubpath` I propose to simplify it to constructions like:

```typescript
config.iceberg?.bucketName ?? session.projectConfig.defaultBucketName
```

In case both are undefined, `getStorageUriForIcebergTable` will throw an error.

core/actions/incremental_table_test.ts
```diff
@@ -352,6 +352,25 @@ actions:
       fs.writeFileSync(path.join(projectDir, `definitions/${filename}`), fileContents);
     };
 
+    // Helper to set up files with custom workflow settings
+    const setupFilesWithCustomWs = (projectDir: string, filename: string, fileContents: string, wsContent: string) => {
```
Ceridan: I suggest to use an optional parameter instead of defining a completely new function, we may update the existing one in a way:

```typescript
const setupFiles = (
  projectDir: string,
  filename: string,
  fileContents: string,
  wsContent?: string) => {
  const ws = wsContent ?? VALID_WORKFLOW_SETTINGS_YAML
  ...
  }
```

core/actions/table_test.ts
```diff
@@ -547,7 +692,12 @@ ${exampleBuiltInAssertionsAsYaml.inputActionConfigBlock}
       paramsToTest.forEach(params => {
         test(`${testCase.testName} in ${params.filename}`, () => {
           const projectDir = tmpDirFixture.createNewTmpDir();
-          setupFiles(projectDir, params.filename, params.fileContents);
+          // Use the custom setup function if wsContent is specified, otherwise default to VALID_WORKFLOW_SETTINGS_YAML
+          if (testCase.wsContent) {
```
Ceridan: You also may avoid of using `if-else` here if you have only one function for `setupFiles`.

core/actions/incremental_table.ts
leonarda-pribanic: The construction logic is a bit more complicated and it's different for bucketName vs tableFolderRoot and tableFolderSubpath.
For bucketName it goes like this: 
    If the value is provided in the config file, use it,
    Otherwise, if the value is provided in workflow_settings.yaml, use it.
    Otherwise, throw an error.
For tableFolderRoot and tableFolderSubpath it goes like this:
    If the value is provided in the config file, use it,
    Otherwise, if the value is provided in workflow_settings.yaml, use it.
    Otherwise, use '_dataform' for tableFodlerRoot and {dataset name}/{table name} for tableFolderSubpath.

I'm not sure the logic for tableFolderRoot and tableFolderSubpath can be nicely inlined, so I added the getEffective... methods for those two. And then it also seemed right to have getEffectiveBucketName to have defaulting logic for each of the fields in their own methods, instead of getStorageUri handling erroring for bucketName only.

Let me know what you think about this!

protos/configs.proto
```diff
@@ -45,6 +45,23 @@ message WorkflowSettings {
 
   // Optional. The prefix to append to built-in assertion names.
   string builtin_assertion_name_prefix = 11;
+
+  // Optional. Default config options for Iceberg tables.
+  DefaultIcebergConfigs default_iceberg_configs = 12;
+}
+
+message DefaultIcebergConfigs {
```
Ceridan: nit comment on naming: I propose to have either `DefaultIcbergConfig` (singular) or `DefaultIcbergOptions` (plural). It is aligned with naming we use across Dataform (the same comment applicable to core.proto).

core/actions/incremental_table.ts
Ceridan: My idea was to push this validation logic inside `getStorageUriForIcebergTable` to make it responsible to validate all the arguments if it set, or not set, and decide what to do because currently it is split across 3 different methods.

Actually, I do not want to force you to change this, I am fine with the current solution. My idea was just to keep tightly coupled logic in one place.

PR #2026, opened 2025-09-26
Enable setting Iceberg config in init CLI

leonarda-pribanic:
This change enables users to invoke `dataform init` with an `--iceberg` option. The users will then be prompted to input defaultBucketName, defaultTableFolderRoot and defaultTableFolderSubpath. The fields users input will be set inside workflow_settings.yaml.

### DIFF
--- a/cli/console.ts
+++ b/cli/console.ts
@@ -56,12 +56,8 @@ export function interactiveQuestion(questionText: string): string {
   // If running in a non-TTY environment and test inputs are provided
   if (!process.stdin.isTTY && process.env.DATAFORM_CLI_TEST_INPUTS !== undefined) {
     const testInput = getTestInput(questionText);
-    if (testInput !== undefined) {
-      print(`${questionText} ${testInput}\n`); // Echo the test input for clarity in logs
-      return testInput;
-    }
-    // If we are in a non-TTY  (i.e. test) environment and not enough test inputs were provided
-    throw new Error("Not enough test inputs provided in DATAFORM_CLI_TEST_INPUTS.");
+    print(`${questionText} ${testInput}\n`); // Echo the test input for clarity in logs
+    return testInput;
   }
 
   // If running in TTY environment or if testInputs are not available in non-TTY
@@ -73,12 +69,12 @@ export function interactiveQuestion(questionText: string): string {
  * Helper function to enable testing interactive CLI. Retrieves testInput from
  * the DATAFORM_CLI_TEST_INPUTS environment variable.
  * @param questionText The exact question text displayed to the user.
- * @returns Test input, or undefined if no test inputs are available for this question.
+ * @returns Test input for the provided question.
  */
-function getTestInput(questionText: string): string | undefined {
+function getTestInput(questionText: string): string {
   const envVar = process.env.DATAFORM_CLI_TEST_INPUTS;
   if(!envVar) {
-    return undefined; // Environment variable not set
+    throw new Error("Environment variable DATAFORM_CLI_TEST_INPUTS not set.");
   }
 
   try {
@@ -89,18 +85,16 @@ function getTestInput(questionText: string): string | undefined {
       const answer = inputs.get(trimmedQuestion);
 
       if (answer !== undefined) {
-        printError(`[TEST_INPUT for "${trimmedQuestion}"]: "${answer}"`);
+        print(`[TEST_INPUT for "${trimmedQuestion}"]: "${answer}"`);
+        return answer;
       } else {
-        printError(`[MISSING TEST_INPUT for "${trimmedQuestion}"]`);
+        throw new Error(`[MISSING TEST_INPUT for "${trimmedQuestion}"]`);
       }
-      return answer;
     } else {
-      printError(`Failed to parse DATAFORM_CLI_TEST_INPUTS: Expected a JSON object, but got ${typeof parsed}.`);
-      return undefined;
+      throw new Error(`Failed to parse DATAFORM_CLI_TEST_INPUTS: Expected a JSON object, but got ${typeof parsed}.`);
     }
   } catch (e) {
-    printError(`Failed to parse DATAFORM_CLI_TEST_INPUTS: ${e.message || e}`);
-    return undefined;
+    throw new Error(`Failed to parse DATAFORM_CLI_TEST_INPUTS: ${e.message || e}`);
   }
 }
 


cli/console.ts
```diff
@@ -45,6 +45,100 @@ export function question(questionText: string) {
...
 }
 
+/**
+ * Test-friendly function that prompts the user for input. If running in a
+ * non-TTY (i.e. test) environment and DATAFORM_CLI_TEST_INPUTS is set,
+ * it consumes inputs from the environment variable.
+ * @param questionText The question text to display.
+ * @returns Users's input or test input.
+ */
+export function interactiveQuestion(questionText: string): string {
```
leonarda-pribanic, line 55: So, it turns out testing interactive CLI is kind of hard. 
We use readline-sync to gather user input. That function can only be used in a TTY environment, and blaze test environments are not TTY environments.
I added this function which does the same thing as question() above, but enables simulating user input for testing. This is not the most elegant function, but I wasn’t able to find any other way to test interactive CLI.

It might be good to replace usages of question() with interactiveQuestion() and add tests for other interactive CLI we have (there are currently no tests for them). I’d be up for doing this in a separate PR. Let me know what you think!

cli/util.ts
```diff
@@ -45,3 +50,37 @@ export function formatBytesInHumanReadableFormat(bytes: number): string {
...
+  // Only return the config object if at least one field was set.
+  if (Object.keys(tempIcebergConfig).length > 0) {
+    return tempIcebergConfig;
+  }
+  return undefined;
+}
+
+export const ICEBERG_CONFIG_PROMPT_TEXT = "\nSet repository-level configuration for Iceberg bucket name, table folder root and table folder subpath. If you do not want to set a field, enter an empty string in response to the prompt.\n";
+export const ICEBERG_CONFIG_COLLECTED_TEXT = "Default Iceberg configuration collected.\n";
+const ICEBERG_BUCKET_NAME_PROMPT_TEXT = "Enter the default Iceberg bucket name:";
```
Ceridan: Could we please provide an example in parenthesis to make sure that users have a hint. For example for folder ROOT and SUBPATH it is critical because there are multiple formats who path can be provided.

Additionally, it would be nice to have some validation for the path, for example make sure that there is no things like "./" or "../". WDYT?

cli/util.ts
```diff
@@ -45,3 +50,37 @@ export function formatBytesInHumanReadableFormat(bytes: number): string {
...
+  printSuccess(ICEBERG_CONFIG_COLLECTED_TEXT);
+
+  // Only return the config object if at least one field was set.
+  if (Object.keys(tempIcebergConfig).length > 0) {
+    return tempIcebergConfig;
+  }
+  return undefined;
+}
+
+export const ICEBERG_CONFIG_PROMPT_TEXT = "\nSet repository-level configuration for Iceberg bucket name, table folder root and table folder subpath. If you do not want to set a field, enter an empty string in response to the prompt.\n";
```
Ceridan: I suggest to inline all these constants. They needed only in a single method and will not be reused anywhere. It will be easier to read and understand the logic if it will be in a single place.

cli/console.ts
Ceridan: I like the idea but I am worrying about shared state we use for the tests to be able to track the question index. I propose to resolve it for now in a simple way (and unblock the PR): use map instead of array. I mean, that for each particular question we will look in the DATAFORM_CLI_TEST_INPUTS map the answer for the question. Yes, it will limit us in a way we cannot repeat the questions but maybe it is fine for now. WDYT?

Ceridan (changes requested):
The PR looks good, I actually learned a few things from it. Thank you! 

I am worried about shared state for testing.
I see two options here:
1. Use map instead of array (see more detailed comment below).
2. Give up and drop testing for now to unblock PR faster :(

cli/util.ts
Ceridan: If we will switch to map for testing, the exported variables will make sense.

cli/console.ts
leonarda-pribanic: Sounds good, I replaced an array with a map.

cli/util.ts
leonarda-pribanic: I added validation for user inputs. If the inputs are invalid, we keep prompting until we get a valid string.

cli/util.ts
leonarda-pribanic: I kept them exported to be able to easily use them in tests. Let me know if you think some should be removed!

cli/util.ts
```diff
@@ -45,3 +51,150 @@ export function formatBytesInHumanReadableFormat(bytes: number): string {
...
+export function validateIcebergConfigBucketName(bucketName: string): void {
+  if (bucketName.length < 3 || bucketName.length > 63) {
+    throw new Error("Bucket name must be between 3 and 63 characters long.");
+  }
+  if (!/^[a-z0-9][a-z0-9-._]{1,220}[a-z0-9]$/.test(bucketName) || bucketName.includes("..")) {
+    throw new Error(
+      "Invalid bucket name. Must start and end with a letter or number, contain only lowercase letters, numbers, hyphens (-), underscores (_), and periods (.)."
+    );
+  }
+  if (bucketName.startsWith("goog") || bucketName.includes("--")) {
```
Ceridan, line 148: Let also check: "Bucket names cannot contain "google""

cli/console.ts
```diff
@@ -45,6 +45,100 @@ export function question(questionText: string) {
...
+ * Represents the state for managing test inputs provided via environment
+ * variables. This state uses a map to associate question texts with answers.
+ */
+interface ITestInputsState {
+  /**
+   * A map where keys are question strings and values are the pre-defined answers.
+   */
+  inputs: Map<string, string>;
+}
+let testInputsState: ITestInputsState | undefined;
```
Ceridan: We do not need this state anymore. Let's remove it.

cli/console.ts
```diff
@@ -45,6 +45,100 @@ export function question(questionText: string) {
...
+
+/**
+ * Helper function to enable testing interactive CLI. Initializes
+ * testInputsState by parsing the DATAFORM_CLI_TEST_INPUTS environment variable.
+ * This function ensures the parsing happens only once.
+ */
+function ensureTestInputsInitialized(): void {
```
Ceridan: I do not think we need this method anymore.

cli/console.ts
```diff
@@ -45,6 +45,100 @@ export function question(questionText: string) {
...
+  }
+}
+
+/**
+ * Helper function to enable testing interactive CLI. Retrieves testInput from
+ * the testInputsState map using the question text as the key.
+ * @param questionText The exact question text displayed to the user.
+ * @returns Test input, or undefined if no test inputs are available for this question.
+ */
+function getTestInput(questionText: string): string | undefined {
```
Ceridan: Could we make it simple and here check if there is a process.env.DATAFORM_CLI_TEST_INPUTS provided, than try to parse it. If it fails - throw, if parsed correctly - read from map by key.

Ceridan (changes requested):

cli/util.ts
leonarda-pribanic: Done.

Just to clarify - we expect users to type in already existing buckets (we don't handle creating a bucket and if the user tries to use a nonexistent bucket, execution of Iceberg SQL will error). So we don't need to be too thorough in enforcing naming requirements here since they are already enforced when creating a bucket.

cli/console.ts
```diff
@@ -45,6 +45,66 @@ export function question(questionText: string) {
...
+ * non-TTY (i.e. test) environment and DATAFORM_CLI_TEST_INPUTS is set,
+ * it consumes inputs from the environment variable.
+ * @param questionText The question text to display.
+ * @returns Users's input or test input.
+ */
+export function interactiveQuestion(questionText: string): string {
+  // If running in a non-TTY environment and test inputs are provided
+  if (!process.stdin.isTTY && process.env.DATAFORM_CLI_TEST_INPUTS !== undefined) {
+    const testInput = getTestInput(questionText);
+    if (testInput !== undefined) {
```
Ceridan: I propose to simplify it, and throw from `getTestInput` if there is any error instead of returning undefined.

cli/console.ts
```diff
@@ -45,6 +45,66 @@ export function question(questionText: string) {
...
+
+  try {
+    const parsed = JSON.parse(envVar);
+    if (typeof parsed === 'object' && parsed !== null && !Array.isArray(parsed)) {
+      const inputs = new Map<string, string>(Object.entries(parsed).map(([key, value]) => [key, String(value)]));
+      const trimmedQuestion = questionText.trim();
+      const answer = inputs.get(trimmedQuestion);
+
+      if (answer !== undefined) {
+        printError(`[TEST_INPUT for "${trimmedQuestion}"]: "${answer}"`);
```
Ceridan: Why do we use `printError` here? Seems like it is expected situation,m isn't it?

Ceridan (approved):
Thank you for the contribution!

PR #2030, opened 2025-10-06
Enable setting default connection for Iceberg table

leonarda-pribanic:
This change enables a workflow-level default connection to be set inside workflow_settings.yaml.

### DIFF
--- a/core/utils.ts
+++ b/core/utils.ts
@@ -317,13 +317,13 @@ export function getFileFormatValueForIcebergTable(
 /**
  * Returns the connection for an Iceberg table, as specified in the user's config file.
  * Defaults to "DEFAULT" if no connection is provided.
- * @param defaultConnection defined in workflow_settings.yaml.
  * @param configConnection defined in the config block
+ * @param defaultConnection defined in workflow_settings.yaml.
  * @returns Connection used when creating an Iceberg table.
  */
 export function getConnectionForIcebergTable(
+  configConnection?: string,
   defaultConnection?: string,
-  configConnection?: string
 ): string {
   if(configConnection) {
     return configConnection;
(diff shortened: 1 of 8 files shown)


core/utils.ts
```diff
@@ -317,11 +317,21 @@ export function getFileFormatValueForIcebergTable(
...
+ * @param configConnection defined in the config block
  * @returns Connection used when creating an Iceberg table.
  */
-export function getConnectionForIcebergTable(connection?: string): string {
-  return connection || "DEFAULT";
+export function getConnectionForIcebergTable(
+  defaultConnection?: string,
+  configConnection?: string
+): string {
+  if(configConnection) {
```
Ceridan, line 328: Maybe make it a little bit simpler:

```typescript
configConnection ?? defaultConnection ?? "DEFAULT";
```

References:
- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

core/utils.ts
```diff
@@ -317,11 +317,21 @@ export function getFileFormatValueForIcebergTable(
...
  * Defaults to "DEFAULT" if no connection is provided.
- * @param connection User-provided connection, if it exists.
+ * @param defaultConnection defined in workflow_settings.yaml.
+ * @param configConnection defined in the config block
  * @returns Connection used when creating an Iceberg table.
  */
-export function getConnectionForIcebergTable(connection?: string): string {
-  return connection || "DEFAULT";
+export function getConnectionForIcebergTable(
+  defaultConnection?: string,
```
Ceridan, line 326: I would change the order of parameters, `configConnection` should go first as it is the first choice. Additionally, it is totally fine to call the method with only `configConnection` provided and omit the other parameter.

Ceridan (approved):

core/utils.ts
leonarda-pribanic: This wouldn't work - if configConnection or defaultConnection is an empty string "", the above command will return the empty string, which is not what we expect.

PR #2037, opened 2025-10-29
Compile extension interface

ikholopov-omni:
Introducing a compilation extension interface - the way to plug ad-hoc compilation stages as NPM package without a need to create an explicit JS definition, unblocking the transpilation of additional syntax and formats.

The interface at the moment includes a compilation callback, which receives the original request and Session to operate on.

Two modes are supported:
 1. "prologue" - extension performs compilation in session before the regular Dataform compilation of sources is done.
 2. "application code" - extension replaces the normal Dataform sources compilation.

### DIFF
--- a/core/main_test.ts
+++ b/core/main_test.ts
@@ -1850,7 +1850,7 @@ publish("name", {type: "${fromType}", schema: "schemaOverride"}).type("${toType}
       expect(result.compile.compiledGraph.targets?.map(t => t.name)).deep.equals(["sample-action"]);
     });
 
-    test("catches extension exceptions", () => {
+    test("catches extension import exceptions", () => {
       const projectDir = setUpProjectWithExtension();
       const request = coreExecutionRequestFromPath(projectDir);
       request.compile.compileConfig.extension = {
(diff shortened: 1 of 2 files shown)


core/main_test.ts
```diff
@@ -1783,4 +1783,86 @@ publish("name", {type: "${fromType}", schema: "schemaOverride"}).type("${toType}
...
+        compilationMode: dataform.CompileConfig.Extension.CompilationMode.APPLICATION_CODE,
+      };
+
+      const result = runMainInVm(request);
+
+      expect(result.compile.compiledGraph.graphErrors.compilationErrors).deep.equals([]);
+      expect(result.compile.compiledGraph.targets?.map(t => t.name)).deep.equals(["sample-action"]);
+    });
+
+    test("catches extension exceptions", () => {
```
kolina: I'd add a test where an extension was imported, but then its call throws an error

protos/core.proto
```diff
@@ -48,6 +48,25 @@ message CompileConfig {
   // Override compilation timeout settings.
   int32 timeout_millis = 6;
 
+  message Extension {
```
kolina: I'd prefer to not define this message as inner to avoid breaking changes if we decide to reuse it in the future

ikholopov-omni: Done

core/main_test.ts
ikholopov-omni: Added two more cases around compilation errors - the one with exception thrown and explicit compilation error set in session.

PR #2041, opened 2025-11-05
Add overview metadata to sqlxconfig

rahulmadanahalli:


core/actions/incremental_table.ts
```diff
@@ -160,6 +160,12 @@ export class IncrementalTable extends ActionBuilder<dataform.Table> {
     if (config.description) {
       this.description(config.description);
     }
+    if (config.metadata) {
```
fernst, line 163: Can we also add this for Dataprep?

fernst (approved):

core/actions/incremental_table.ts
chtyim: Let's not to expand on data prep as we want to move everything SQL into regular table action.

protos/configs.proto
```diff
@@ -258,6 +258,14 @@ message ActionConfig {
...
     // Configuration options for an Iceberg table.
     IcebergTableConfig iceberg = 22;
+
+    // Metadata for this table.
+    Metadata metadata = 23;
+  }
+
+  message Metadata {
+    // A detailed description of the data object.
+    string overview = 1;
```
chtyim, line 268: Isn't that we want to allow Metadata to be a map of any key value pairs?

PR #2043, opened 2025-11-13
Support bq job labeling

de-lan:
## What

Add support for bq job labelling when in `dataform run` command. This provides user more flexibility of correlating created jobs with custom global level labels

### DIFF
--- a/tests/api/api.spec.ts
+++ b/tests/api/api.spec.ts
@@ -1382,12 +1382,12 @@ suite("@dataform/api", () => {
         });
         
         // For the second action (assertion), verify only global labels (no action-level labels)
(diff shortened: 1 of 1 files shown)


cli/index.ts
```diff
@@ -171,6 +171,26 @@ const jobPrefixOption: INamedOption<yargs.Options> = {
   }
 };
 
+const bigqueryLabelsOption: INamedOption<yargs.Options> = {
+  name: "bigquery-labels",
```
fernst: Should this be `job-labels`? In BQ, jobs can be both in the job and in the table itself.

de-lan: Good point, updated to use `job-labels`

PR #2052, opened 2025-12-16
Added new properties to Test proto

fernst:
Adds new proto properties to Test entity.

Includes test target + fully resolved reference names for inputs.

### DIFF
--- a/protos/core.proto
+++ b/protos/core.proto
@@ -264,6 +264,16 @@ message Test {
 
   // Generated.
   string file_name = 4;
+
+  Target test_target = 5;
+  repeated TestInput inputs = 6; 
+  string query = 7;
+  bool resolve_schema = 8;
+
+  message TestInput {
+    Target target = 1;
+    string query = 2;
+  }
 }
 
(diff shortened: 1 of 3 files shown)


protos/core.proto
```diff
@@ -264,6 +264,16 @@ message Test {
...
+
+  Target test_target = 5;
+  repeated TestInput inputs = 6; 
+  string query = 7;
+  bool resolve_schema = 8;
+
+  message TestInput {
+    Target target = 1;
+    string query = 2;
+  }
```
kolina, line 276: Sorry, I'm not ready to accept this proposal:
* You're adding a lot of new fields and it'll be super confusing for users to understand the relationship between new `query` and `inputs` and old `test_query` and `expected_output_query` (I myself spend a lot of time to understand it)
* I'm not sure if all of these fields are really needed, I'll need some explanation on how all of these fields are intended to be used and discuss if it's feasible to have more compact compilation output

kolina (changes requested):

PR #2055, opened 2025-12-17
Add extra_properties struct to metadata

rahulmadanahalli:


protos/core.proto
```diff
@@ -110,6 +111,7 @@ message ActionDescriptor {
 
 message Metadata {
   string overview = 1;
+  google.protobuf.Struct custom_attributes = 2;
 }
```
kolina, line 115: Are attributes is relevant name here?

kolina (changes requested):
Integration tests are failing

kolina (approved):

PR #2057, opened 2025-12-23
Fix incremental table recreation when defaultProject not set

sibyabin:
Add changes to 
- fix the incremental table getting recreated when the defaultProject is not available as part of the workflow_settings.yaml
- Minimal unit test for the same

### DIFF
--- a/core/actions/incremental_table_test.ts
+++ b/core/actions/incremental_table_test.ts
@@ -882,4 +882,39 @@ defaultIcebergConfig:
       });
     });
   });
+
+  test("incremental table without defaultProject in workflow_settings", () => {
+    const projectDir = tmpDirFixture.createNewTmpDir();
+    // Create workflow_settings without defaultProject (only defaultDataset and defaultLocation)
+    fs.writeFileSync(
+      path.join(projectDir, "workflow_settings.yaml"),
+      `defaultDataset: dataform
+defaultLocation: europe-west2
+`
+    );
+    fs.mkdirSync(path.join(projectDir, "definitions"));
+    fs.writeFileSync(
+      path.join(projectDir, "definitions/incremental_table_without_default_project.sqlx"),
+      `config {
+    type: "incremental",
+    name: "incremental_table_without_default_project"
+}
+
+select \${incremental()} as is_incremental`
+    );
+
+    const result = runMainInVm(coreExecutionRequestFromPath(projectDir));
+
+    expect(result.compile.compiledGraph.graphErrors.compilationErrors).deep.equals([]);
+    const compiledTable = result.compile.compiledGraph.tables[0];
+    // Verify the table compiles without database field since defaultProject is not set.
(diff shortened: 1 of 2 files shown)


cli/api/commands/build.ts
```diff
@@ -49,8 +49,13 @@ export class Builder {
...
     const tableMetadataByTarget = new Map<string, dataform.ITableMetadata>();
 
+    // Normalize warehouse state targets to match compiled targets for consistent lookups.
+    // This is necessary because targets from the warehouse may have a database field set
+    // (from BigQuery credentials) while compiled targets may not have a database field
+    // if defaultDatabase is not specified in workflow_settings.yaml.
     this.warehouseState.tables.forEach(tableState => {
-      tableMetadataByTarget.set(targetStringifier.stringify(tableState.target), tableState);
+      const normalizedTarget = this.normalizeWarehouseTarget(tableState.target);
+      tableMetadataByTarget.set(targetStringifier.stringify(normalizedTarget), tableState);
```
kolina: I'd suggest to fix a bit differently: [here](https://github.com/dataform-co/dataform/blob/e16820781911670184452db8bed65dabc3885941/cli/api/dbadapters/bigquery.ts#L209) we can set `target` instead of `metadataTarget`  in the metadata structure.

It seems to me a more robust and concise solution.

core/actions/incremental_table_test.ts
```diff
@@ -882,4 +882,39 @@ defaultIcebergConfig:
...
+      path.join(projectDir, "definitions/incremental_table_without_default_project.sqlx"),
+      `config {
+    type: "incremental",
+    name: "incremental_table_without_default_project"
+}
+
+select \${incremental()} as is_incremental`
+    );
+
+    const result = runMainInVm(coreExecutionRequestFromPath(projectDir));
```
kolina, line 906: I don't think that this test will validate the fix because the problem is not on the compilation stage. But anyway the more tests the better :)

You'd need to add an integration test here https://github.com/dataform-co/dataform/blob/main/cli/index_test.ts#L951 to validate fix correctness, but I'm ok to accept PR without it.

kolina (commented):
Can you please rebase the PR to the latest `main` after I fixed integration tests in https://github.com/dataform-co/dataform/commit/ee5b5d1be7c1d9f2fcfc71829a247082998836df.

I want to re-run them after my fix in your PR

cli/api/commands/build.ts
sibyabin: branch rebased and changes made as suggested

core/actions/incremental_table_test.ts
sibyabin: I have not added the integration tests as I was not really sure how to validate this specific scenario for the change is made . 

I see that the current integration tests are running with the --dry-run flag . In this case, we need to validate the --dry-run of an existing **incremental** table is not firing a **create or replace** when there is no defaultProject set in the workflow settings yaml

Let me know if there is a better way to check this.

PR #2063, opened 2026-01-09
Populate new unit test properties and add flag to include unit tests in compiled graph

fernst:
Populated new properties in unit test proto during compilation.

Also added a new workflow config setting to include unit tests in compiled DAG.

### DIFF
--- a/core/actions/index.ts
+++ b/core/actions/index.ts
@@ -5,6 +5,7 @@ import { IncrementalTable } from "df/core/actions/incremental_table";
 import { Notebook } from "df/core/actions/notebook";
 import { Operation } from "df/core/actions/operation";
 import { Table } from "df/core/actions/table";
+import { Test } from "df/core/actions/test";
 import { View } from "df/core/actions/view";
 import { IColumnsDescriptor } from "df/core/column_descriptors";
 import { Resolvable } from "df/core/contextables";
@@ -19,7 +20,8 @@ export type Action =
   | Assertion
   | Declaration
   | Notebook
-  | DataPreparation;
+  | DataPreparation
+  | Test;
 
 export type ActionProto =
   | dataform.Table // core.proto's Table represents the Table, View or IncrementalTable action type.
@@ -27,7 +29,8 @@ export type ActionProto =
   | dataform.Assertion
   | dataform.Declaration
   | dataform.Notebook
-  | dataform.DataPreparation;
+  | dataform.DataPreparation
+  | dataform.Test;
 
 // In v4, consider making methods on inheritors of this private, forcing users to use constructors
 // in order to populate actions.
--- a/core/actions/test.ts
+++ b/core/actions/test.ts
@@ -77,10 +77,8 @@ export class Test extends ActionBuilder<dataform.Test> {
 
   /** @hidden We delay contextification until the final compile step, so hold these here for now. */
   public contextableInputs = new Map<string, Contextable<IActionContext, string>>();
-  private inputToTargets = new Map<string, dataform.Target>();
   private contextableQuery: Contextable<IActionContext, string>; 
   private testTarget: dataform.ITarget;
-  private canonicalTestTarget:  dataform.ITarget;
 
   /**
    * @hidden Stores the generated proto for the compiled graph.
@@ -149,13 +147,10 @@ export class Test extends ActionBuilder<dataform.Test> {
    * Sets the input query to unit test against.
    */
   public input(refName: string | string[], contextableQuery: Contextable<IActionContext, string>) {
-    const target = resolvableAsTarget(toResolvable(refName));
-    const inputName = targetStringifier.stringify(target);
     this.contextableInputs.set(
-      inputName,
+      targetStringifier.stringify(resolvableAsTarget(toResolvable(refName))),
       contextableQuery
     );
-    this.inputToTargets.set(inputName, target); 
     return this;
   }
 
--- a/core/actions/test.ts
+++ b/core/actions/test.ts
@@ -32,10 +32,13 @@ export interface ITestConfig extends INamedConfig {
 
   /** Path to the source file that the contents of the action is loaded from. */
   filename?: string;
+
+  /** Tags for this test action. */
+  tags?: string[];
 }
 
 /** @hidden */
-const ITestConfigProperties = strictKeysOf<ITestConfig>()(["type", "dataset", "name", "filename"]);
+const ITestConfigProperties = strictKeysOf<ITestConfig>()(["type", "dataset", "name", "filename", "tags"]);
 
 /**
  * Dataform test actions can be used to write unit tests for your generated SQL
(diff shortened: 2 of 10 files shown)


core/actions/test.ts
```diff
@@ -188,6 +202,28 @@ export class Test extends ActionBuilder<dataform.Test> {
       } else {
         const refReplacingContext = new RefReplacingContext(testContext);
         this.proto.testQuery = refReplacingContext.apply(dataset.contextableQuery);
+        this.testTarget = dataform.Target.create(dataset.getTarget());
+
+        // Set the test query with the fully qualified table references.
+        if (dataset instanceof Table) {
```
ikholopov-omni: These if/else branches are equivalent

ikholopov-omni (changes requested):

core/actions/test.ts
fernst: True. I've consolidated this code.

core/actions/test.ts
```diff
@@ -188,6 +202,19 @@ export class Test extends ActionBuilder<dataform.Test> {
...
+
+        // Set the test query with the fully qualified table references.
+        this.proto.target = this.replaceTestNameInTarget(
+          testTarget.getTarget(),
+          this.proto.name
+        );
+        this.proto.canonicalTarget = this.replaceTestNameInTarget(
+          testTarget.getCanonicalTarget(),
+          this.proto.name
+        );
```
kolina: Do I understand correctly that this can change the result target of the test?

I imagine this can break in some corner cases (the first thing coming to my mind: if tests become regular DAG nodes, someone can explicitly add dependency on the test node and this dependency can break if dataset / database part changed).

Maybe we can have behaviour similar to assertion nodes here instead (AFAIK they don't have this logic)

core/actions/test_test.ts
kolina, line 1: I think it's worth adding a test where logic of calling `replaceTestNameInTarget` actually matters

kolina (changes requested):
Can you please against the latest `main`? So integration tests will run with non-expired credentials

core/actions/test.ts
fernst: The function naming is a bit confusing now that I reread it.

This simply takes the parent target/canonical target and sets the test name as the new action name.

This is so we can generate a whole new target for the current test, which ensures we can use this as a proper DAG element (where the target is the identifier).

core/actions/test_test.ts
fernst: I renamed the function so it's more clear that this is only impacting the current action.

I'll add a test where I have 2 targets with similar name, and test names, but different schema

`defaultProject.schema1.tbl` and `defaultProject.schema2.tbl`, both with a test named `tbl_test` for each one.

fernst: Actually, there is an issue with 2 actions with the same name (meaning, you can't write tests for an action if 2 actions share the same name). I will fix this in a follow-up PR.

core/actions/test.ts
fernst: I've renamed this function, PTAL.

core/actions/test_test.ts
fernst: I added a test case with 2 targets using the same table name but under different schemas. This tests the resolution logic for the targets and that actions are labeled under the same database/schema as the action being tested.

core/actions/test.ts
```diff
@@ -198,6 +226,14 @@ export class Test extends ActionBuilder<dataform.Test> {
       VerifyProtoErrorBehaviour.SUGGEST_REPORTING_TO_DATAFORM_TEAM
     );
   }
+
+  private overrideTargetWithNewName(target: dataform.ITarget, testName: string): dataform.Target {
```
ikholopov-omni: Nit: Can be a local function, as it doesn't depend on `this`.

core/session.ts
```diff
@@ -549,11 +558,24 @@ export class Session {
...
+          .forEach(tableOrViewAction => tableOrViewAction.dependencies(utils.resolvableAsTarget(test.getTarget())));
+      });
+  }
+
   private fullyQualifyDependencies(actions: ActionProto[]) {
     actions.forEach(action => {
       const fullyQualifiedDependencies: { [name: string]: dataform.ITarget } = {};
       if (action instanceof dataform.Declaration || !action.dependencyTargets) {
-        // Declarations cannot have dependencies.
+        // Declarations cannot have dependencies.
```
ikholopov-omni: Nit: dangling whitespace?

core/actions/index.ts
```diff
@@ -19,15 +20,17 @@ export type Action =
   | Assertion
   | Declaration
   | Notebook
-  | DataPreparation;
+  | DataPreparation
+  | Test;
```
ikholopov-omni, line 24: The addition of tests to the list of targets in the compiled graph and a dependency, changes in session.test backing structure is technically a breaking change.

At the very least we should bump a minor version, once we merger theses changes.

ikholopov-omni (dismissed):

core/actions/index.ts
fernst: I've since removed the dependency changes (meaning tests are in the DAG, but not connected to any other action). 

Should we still do the minor version bump? Or can we keep it under patch versions until we add the new tests into the middle of the DAG?

core/actions/test.ts
fernst: Moved.

core/session.ts
fernst: Removed

core/actions/index.ts
ikholopov-omni: At the moment there are still breaking changes in Dataform Core API, specifically in Session object. Bottom line is these changes shouldn't be released under a patch, which means that we should either:
1) Exercise control and block all @dataform/core releases until all breaking changes are submitted from the feature branch
2) Prepare all breaking changes in a separate branch and them merge them in a single commit with a minor version bump.

At the moment we don't have any automation to enforce (1) repo-wide. I think executing (2) is easier.

kolina: Current changes LGTM from my side, will wait for resolution in this thread to approve the PR

kolina (commented):
Waiting on resolution of how to update Dataform core version with these changes

core/actions/index.ts
fernst: I've added a new configuration property to only add the unit tests to the compiled DAG when this setting is set. This ensures backwards compatibility and ensures the graph remains consistent for users who want to keep this behavior.

core/session.ts
```diff
@@ -732,6 +743,19 @@ export class Session {
...
   }
 
+  private addTestsToCompiledGraph(actions: Action[]) {
+    actions
+      .filter(action => action instanceof Test)
+      .map(test => test as Test)
+      .forEach(test => {
+        this.indexedActions
+          .find(test.getTestTarget())
+          .filter(action => action instanceof Table || action instanceof View)
```
kolina, line 753: should we throw an error otherwise?

kolina (approved):

core/session.ts
fernst: This would have failed already in https://github.com/dataform-co/dataform/blob/main/core/actions/test.ts#L180-L195 , as that already checks the type of the action that the test is testing.

PR #2079, opened 2026-02-03
JiT context RPC service and protos

ikholopov-omni:
Protos for JiT DB adapter service (initial set of methods) and JiT compilation messages. 

Relying on existing Core and Execution messages for Tables and Job representations. Refactoring those into a common proto, attaching it to @dataform/core NPM package.

### DIFF
--- a/protos/execution.proto
+++ b/protos/execution.proto
@@ -1,11 +1,11 @@
-syntax="proto3";
+syntax = "proto3";
 
 package dataform;
 
 import "protos/core.proto";
+import "protos/db_adapter.proto";
 
 option go_package = "github.com/dataform-co/dataform/protos/dataform";
-
 
 message RunConfig {
   repeated string actions = 1;
@@ -14,7 +14,7 @@ message RunConfig {
   bool include_dependents = 11;
   bool full_refresh = 2;
   int32 timeout_millis = 7;
-  
+
   // For internal use only, will be removed at a later date.
   bool disable_set_metadata = 9;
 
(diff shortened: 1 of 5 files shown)


protos/jit.proto
```diff
@@ -0,0 +1,108 @@
...
+  bool fetch_results = 4;
+
+  BigQueryExecuteOptions big_query_options = 5;
+}
+
+message BigQueryExecuteOptions {
+  // Priority - interactive if true or batch if false (default).
+  bool interactive = 1;
+  // See https://docs.cloud.google.com/bigquery/docs/reference/legacy-sql.
+  bool use_legacy_sql = 2;
```
kolina, line 43: Is it needed?

protos/execution.proto
```diff
@@ -133,66 +121,3 @@ message TestResult {
   bool successful = 2;
   repeated string messages = 3;
 }
-
-message Field {
```
kolina, line 137: Just want to confirm that it's not a breaking change for TypeScript proto generation.

I hope people don't try depending on these in Python :)

protos/jit.proto
ikholopov-omni: Yes, it is still supported at the moment, even if not widely used across target libs, so better to expose this flag for compatibility.

protos/execution.proto
ikholopov-omni: Typescript is fine, it concats all protos into a single package. The concern about usages of protos that we are not aware of in other languates is valid, but at least in case of this proto we don't ship it as part of the node package. So, all breaking changes will be tied to an original code checkout :)

PR #2080, opened 2026-02-03
JiT data in Session

ikholopov-omni:
Stacked on https://github.com/dataform-co/dataform/pull/2079

Adding a JiT data to the session. Renaming the proto field as well,
it was submitted before the update of the design, but since it is not used yet,
it should be fine to rename it directly.


core/session.ts
```diff
@@ -408,6 +412,49 @@ export class Session {
     return notebook;
   }
 
+  public jitData(key: string, data: unknown): void {
```
kolina, line 415: Add a test with JS objects we fail to add to a context?

ikholopov-omni: Done, also added a handling of undefined.

PR #2082, opened 2026-02-05
Added logic to fail test compilation if test SQL or expected SQL are empty

fernst:


core/actions/test.ts
```diff
@@ -192,6 +192,20 @@ export class Test extends ActionBuilder<dataform.Test> {
     }
     this.proto.expectedOutputQuery = testContext.apply(this.contextableQuery);
 
+    // Check if the test query and expected output query are non-empty.
+    if (this.proto.testQuery === null || this.proto.testQuery.trim() === "") {
```
ikholopov-omni: You can `!this.proto.testQuery` to cover both null and undefined.

Same below

fernst: Thanks! Will change

core/actions/test.ts
```diff
@@ -192,6 +192,20 @@ export class Test extends ActionBuilder<dataform.Test> {
     }
     this.proto.expectedOutputQuery = testContext.apply(this.contextableQuery);
 
+    // Check if the test query and expected output query are non-empty.
+    if (!this.proto.testQuery || this.proto.testQuery.trim() === "") {
```
kolina: `!this.proto.testQuery.trim()`, same below

PR #2086, opened 2026-02-10
JiT compilation support in JS API

ikholopov-omni:
JS API for JiT compilation actions definition (mutually exclusive with AoT properties) and JiT context/result typings.

### DIFF
--- a/core/actions/incremental_table.ts
+++ b/core/actions/incremental_table.ts
@@ -8,10 +8,10 @@ import {
   TableType
 } from "df/core/actions";
 import { Assertion } from "df/core/actions/assertion";
-import { Table } from "df/core/actions/table";
+import { JitTableResult, Table } from "df/core/actions/table";
 import { View } from "df/core/actions/view";
 import { ColumnDescriptors } from "df/core/column_descriptors";
-import { Contextable, ITableContext, Resolvable } from "df/core/contextables";
+import { Contextable, ITableContext, JitContextable, Resolvable } from "df/core/contextables";
 import * as Path from "df/core/path";
 import { Session } from "df/core/session";
 import {
@@ -32,6 +32,7 @@ import {
   strictKeysOf,
   toResolvable,
   validateConnectionFormat,
+  validateNoMixedCompilationMode,
   validateQueryString,
   validateStorageUriFormat,
 } from "df/core/utils";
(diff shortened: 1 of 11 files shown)


core/actions/incremental_table.ts
```diff
@@ -517,6 +527,49 @@ export class IncrementalTable extends ActionBuilder<dataform.Table> {
...
+    let flattenPostOps: unknown[] = [];
+    this.contextablePostOps.forEach(op => {
+      flattenPostOps = flattenPostOps.concat(typeof op === "object" ? op : [op]);
+    })
+    let flattenPreOps: unknown[] = [];
+    this.contextablePreOps.forEach(op => {
+      flattenPreOps = flattenPreOps.concat(typeof op === "object" ? op : [op]);
+    })
+    if (!!this.contextableQuery || !!this.contextableWhere || !!flattenPostOps.length || !!flattenPreOps.length) {
+      const err = new Error(`Cannot mix AoT and JiT compilation in action: ${this.contextableQuery}, ${this.contextableWhere}, ${this.contextablePostOps}, ${this.contextablePreOps}`);
```
kolina: I'd report specific violation reason here (i.e. what specific config field triggered it)

core/actions/table.ts
```diff
@@ -652,7 +694,7 @@ export class Table extends ActionBuilder<dataform.Table> {
  * @hidden
  */
 export class TableContext implements ITableContext {
-  constructor(private table: Table, private isIncremental = false) {}
+  constructor(private table: Table, private isIncremental = false) { }
```
kolina, line 694: what is the cause of these diffs?

core/actions/table.ts
```diff
@@ -485,6 +499,48 @@ export class Table extends ActionBuilder<dataform.Table> {
...
+    })
+    let flattenPreOps: unknown[] = [];
+    this.contextablePreOps.forEach(op => {
+      flattenPreOps = flattenPreOps.concat(typeof op === "object" ? op : [op]);
+    })
+    if (!!this.contextableQuery || !!this.contextableWhere || !!flattenPostOps.length || !!flattenPreOps.length) {
+      const err = new Error(`Cannot mix AoT and JiT compilation in action: ${this.contextableQuery}, ${this.contextableWhere}, ${this.contextablePostOps}, ${this.contextablePreOps}`);
+      this.session.compileError(err, this.getFileName());
+      throw err;
+    }
```
kolina: This validation can be reused for different action types?

kolina (approved):

core/actions/table.ts
ikholopov-omni: My IDE (vscode) is extremely opinionated about TS formatiing, apparently (like golang-level of opinionation). 

I could clean those up by splitting it out of the commit, but this will continue being an overhead long-term :/

Probably worth setting up formatter next to the linter check.

core/actions/table.ts
ikholopov-omni: Moved to util

core/actions/incremental_table.ts
ikholopov-omni: Done

PR #2088, opened 2026-02-10
JiT contexts implementation

ikholopov-omni:
Implementation of context objects accessible at JiT compilation stage for supported action types.

They expose access to adapter/data and context methods, equivalent to AoT context, but with a more restrictive selection of resolvable targets.

### DIFF
--- a/core/jit_context.ts
+++ b/core/jit_context.ts
@@ -1,0 +1,108 @@
+import { IActionContext, ITableContext, JitContext, Resolvable } from "df/core/contextables";
+import { ambiguousActionNameMsg, resolvableAsTarget, ResolvableMap, stringifyResolvable, toResolvable } from "df/core/utils";
+import { dataform } from "df/protos/ts";
+
+function canonicalTargetValue(target: dataform.ITarget): string {
+    return `${target.database}.${target.schema}.${target.name}`;
+}
+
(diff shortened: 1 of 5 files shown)


core/jit_context.ts
```diff
@@ -0,0 +1,115 @@
...
+    constructor(
+        public readonly adapter: dataform.DbAdapter,
+        public readonly data: { [k: string]: any } | undefined,
+        private readonly target: dataform.ITarget,
+        dependencies: dataform.ITarget[],
+    ) {
+        const resolvables = [target, ...dependencies];
+        this.resolvablesSet = new Set(resolvables.map(canonicalTargetValue));
+        this.resolvablesPerName = new Map(resolvables.map(dep => [dep.name, canonicalTargetValue(dep)]));
+        this.resolvablesPerSchema = new Map(resolvables.map(dep => [schemaTargetValue(dep), canonicalTargetValue(dep)]));
```
kolina: nit: refactor `ActionMap` so it could support it as well https://github.com/dataform-co/dataform/blob/4246ebb2aa6189170128d7fc869b426e5d4183e2/core/session.ts#L825

ikholopov-omni: Moved into a generic "ResolvableMap" (in core/utils.ts).

PR #2090, opened 2026-02-11
JiT compilation stage

ikholopov-omni:
Builds on top of https://github.com/dataform-co/dataform/pull/2090

Implement JiT compilation stage in Core:
 - a new entry point jitCompiler(rpcCallback), which instantiates a
single compile method, performing a compilation against a supplied
RPC implementation.
 - table/operations/incremental table context(s) construction and
invocation of the JiT code.

### DIFF
--- a/core/jit_compiler.ts
+++ b/core/jit_compiler.ts
@@ -1,0 +1,128 @@
+import * as $protobuf from "protobufjs";
+
+import { JitOperationResult } from "df/core/actions/operation";
+import { JitTableResult } from "df/core/actions/table";
+import { IActionContext, ITableContext, JitContext } from "df/core/contextables";
+import { IncrementalTableJitContext, SqlActionJitContext, TableJitContext } from "df/core/jit_context";
+import { dataform } from "df/protos/ts";
+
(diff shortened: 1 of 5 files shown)


core/jit_compiler.ts
```diff
@@ -0,0 +1,131 @@
...
+import { JitOperationResult } from "df/core/actions/operation";
+import { JitTableResult } from "df/core/actions/table";
+import { IActionContext, ITableContext, JitContext } from "df/core/contextables";
+import { IncrementalTableJitContext, SqlActionJitContext, TableJitContext } from "df/core/jit_context";
+import { dataform } from "df/protos/ts";
+
+function makeMainBody<Context, T>(code: string): (jctx: JitContext<Context>) => Promise<T> {
+  return (
+    jctx => {
+      const mainAsync = async () => {
```
kolina: Do you need this outer `mainAsync` wrapper? If yes, please write a comment for people without context :)

kolina (approved):

core/jit_compiler.ts
ikholopov-omni: It is no longer needed - it was the remnant of original hard-coded implementation. Removed (or rather renamed f() into it, which is used to handle different function definition format).

PR #2092, opened 2026-02-11
Support both 'yaml' and 'yml' file extensions

apilaskowski:
We can support both formats in Dataform. 

Native configurations are stored in YAML format, but other configurations eg. from extensions might use YML too.


apilaskowski:
Added some unit-testing for compilers.

Issue #1981, opened 2025-06-04
Feature Request: support BigQuery reservations as Dataform config


rsenseman:
### Background
As far as I can tell, BigQuery [reservations](https://cloud.google.com/bigquery/docs/reservations-workload-management) can only be applied by supplying the reservation syntax inline with the query, so reservations must be coded into every query, like this:

```
config {
  type: "view"
}
pre_operations { SET @@reservation='projects/<project_name>/locations/<region>/reservations/<reservation_name>'; }


SELECT
  *
FROM ${ ref("table_name") }
```

Ideally, this could be applied at a higher level so that reservations could be applied at the project level or at the very least supplied at the model config level

### Desired behavior
At the model level:
```
config {
  type: "view",
  bigquery: {
    reservation: 'projects/<project_name>/locations/<region>/reservations/<reservation_name>'
  }
}

SELECT
  *
FROM ${ ref("table_name") }
```

At the workflow_settings.yaml level:
```
defaultReservation: 'projects/<project_name>/locations/<region>/reservations/<reservation_name>'
```

PR #2096, opened 2026-02-24
Add BigQuery reservation configuration

max-ostapenko:
This pull request introduces support for specifying BigQuery reservations at both the project and action level, allowing more granular control over query resource allocation. The changes span CLI flags, project configuration, action builders, and test coverage to ensure correct application and precedence of reservation settings.

**BigQuery reservation support:**

* Added a new `--default-reservation` CLI flag and corresponding `defaultReservation` project config option, enabling users to set a default BigQuery reservation for all actions in a project.
* Updated `IBigQueryExecutionOptions` and `BigQueryDbAdapter` to accept and propagate a `reservation` parameter, ensuring reservation information is passed through to BigQuery job execution.

**Action-level reservation configuration:**

* Extended action builders (`Assertion`, `Table`, `IncrementalTable`, `Operation`, `View`) to support an optional `reservation` config property, storing it in the action descriptor for runtime precedence over project-level reservations.

**Test coverage and validation:**

* Added and updated tests to verify correct application of project-level and action-level reservations, including precedence rules and propagation to compiled outputs.

**Project configuration precedence:**

* Ensured that action-level `reservation` overrides the project-level `defaultReservation` at runtime, with tests confirming the correct behavior.

These changes provide users with flexible reservation management for Dataform workload.

Related to #1981

### DIFF
--- a/protos/configs.proto
+++ b/protos/configs.proto
@@ -51,13 +51,14 @@ message WorkflowSettings {
   // Optional. Default config options for Iceberg tables.
   DefaultIcebergConfig default_iceberg_config = 12;
 
-  // Optional. Disables all assertions including built-in assertions 
+  // Optional. Disables all assertions including built-in assertions
   // (uniqueKey, nonNull, rowConditions) and manual assertions (type: assertion).
   // When true, assertions will still be compiled but marked as disabled.
   bool disable_assertions = 13;
 
   // Optional. The default BigQuery reservation to use for execution.
   // If unset, default BigQuery behavior applies.
+  // Dataform CLI only (GCP Dataform support pending).
   string default_reservation = 14;
 }
 
@@ -270,6 +271,7 @@ message ActionConfig {
 
     // Optional. The BigQuery reservation to use for execution.
     // If unset, the value from workflow_settings.yaml is used. If neither is set, default BigQuery behavior applies.
+    // Dataform CLI only (GCP Dataform support pending).
     string reservation = 24;
   }
 
@@ -357,6 +359,7 @@ message ActionConfig {
 
     // Optional. The BigQuery reservation to use for execution.
     // If unset, the value from workflow_settings.yaml is used. If neither is set, default BigQuery behavior applies.
+    // Dataform CLI only (GCP Dataform support pending).
     string reservation = 21;
   }
 
@@ -473,6 +476,7 @@ message ActionConfig {
 
     // Optional. The BigQuery reservation to use for execution.
     // If unset, the value from workflow_settings.yaml is used. If neither is set, default BigQuery behavior applies.
+    // Dataform CLI only (GCP Dataform support pending).
     string reservation = 27;
   }
 
@@ -515,6 +519,7 @@ message ActionConfig {
 
     // Optional. The BigQuery reservation to use for execution.
     // If unset, the value from workflow_settings.yaml is used. If neither is set, default BigQuery behavior applies.
+    // Dataform CLI only (GCP Dataform support pending).
     string reservation = 11;
   }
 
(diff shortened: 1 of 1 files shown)


cli/index.ts
```diff
@@ -863,6 +863,16 @@ class ProjectConfigOptions {
     }
   };
 
+  public static defaultReservation: INamedOption<yargs.Options> = {
+    name: "default-reservation",
+    option: {
+      describe:
+        "The default BigQuery reservation to use for execution. If unset, the value from " +
+        "workflow_settings.yaml is used.",
```
kolina: Let's also describe behaviour when it's not set in `workflow_settings.yaml`

cli/index_test.ts
```diff
@@ -1062,4 +1062,90 @@ SELECT 1 as id
...
+          CREDENTIALS_PATH,
+          "--dry-run",
+          "--json",
+          `--default-reservation=${RESERVATION}`,
+          "--actions=example_table"
+        ])
+      );
+
+      expect(runResult.exitCode).equals(0);
+      const executionGraph = JSON.parse(runResult.stdout);
```
kolina, line 1155: can we validate the whole object here?

protos/configs.proto
```diff
@@ -263,6 +266,9 @@ message ActionConfig {
 
     // Metadata for this table.
     Metadata metadata = 23;
+
+    // Optional. The BigQuery reservation to use for execution.
+    string bigquery_reservation = 24;
```
kolina: I'd name these fields (and in other action configs) just `reservation`

protos/core.proto
```diff
@@ -119,6 +121,7 @@ message ActionDescriptor {
   map<string, string> bigquery_labels = 3;
   Metadata metadata = 4;
   ActionCompilationMode compilation_mode = 5;
+  string bigquery_reservation = 6;
```
kolina: same: name it just `reservation`?

cli/index.ts
max-ostapenko: Done. Also added similar description to the `protos/config.proto` for each type of config.

cli/index_test.ts
```diff
@@ -25,6 +25,7 @@ import { TmpDirFixture } from "df/testing/fixtures";
 
 const DEFAULT_DATABASE = "dataform-open-source";
 const DEFAULT_LOCATION = "US";
+const DEFAULT_RESERVATION = "projects/dataform-open-source/locations/us-central1/reservations/dataform-test";
```
max-ostapenko: @kolina Moved the default reservation to the other defaults, and mentioned this config in `contributing.md`

max-ostapenko: Removed mention from contributing.md, as not sure it is required.

protos/configs.proto
```diff
@@ -57,6 +57,7 @@ message WorkflowSettings {
   bool disable_assertions = 13;
 
   // Optional. The default BigQuery reservation to use for execution.
+  // If unset, default BigQuery behavior applies.
   string default_reservation = 14;
```
kolina, line 62: I'd add a comment that it'll be supported by the Dataform CLI only for now (support in GCP Dataform will require about to the GCP Dataform executor to use this value), same for action fields below

cli/index_test.ts
kolina: Thanks :)

kolina (approved):

protos/configs.proto
max-ostapenko: 😔
So cloud Dataform executor has its own version of `cli/api/dbadapters/bigquery.ts`?
Then all the action `reservation` attributes will not impact anything in GCP Dataform.

How often does it catch up?

kolina: Yeah, the GCP Dataform executor is different.

We were planning to implement it later this year (Q2-Q3 timeframe), we may do it sooner taking your contribution.

PR #2101, opened 2026-03-03
Support onSchemaChange for incremental tables

SuchodolskiEdvin:
This change introduces support for the onSchemaChange option in incremental tables for the BigQuery adapter. It adds the incrementalSchemaChangeBody() strategy to handle schema changes.

End-to-end tests have been created to verify the new functionality.


kolina (changes requested):
I haven't taken a deep look, but in the current logic we add granular SQL statements as separate tasks.

I'd be great to preserve this consistency (as incremental schema change generates quite a lot of statements).

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +231,232 @@ from (${query}) as insertions`;
...
+  private runFinalDmlSql(
+    table: dataform.ITable,
+    qualifiedTargetTableName: string,
+    dataTempTableName: string
+  ): string {
+    let finalDmlSql = "\n-- Run final MERGE/INSERT.";
+
+    // Create temp table for incremental data.
+    finalDmlSql += `
+CREATE OR REPLACE TEMP TABLE ${dataTempTableName} AS (
```
spatel11: Not a maintainer of this repo but:

I've noticed this in the GCP Dataform compiled version of an EXTEND schema change incremental table, but for the life of me I cannot discern why this intermediate table is needed. It doesn't seem to be used for anything other than the MERGE statement, and the INSERT on line 420. Could the query be used directly in the MERGE, and in the INSERT? Materializing into an intermediate table increases the cost of the query, sometimes by quite a lot.

SuchodolskiEdvin: The dataTempTableName is used to ensure the columns selected for the MERGE or INSERT operation exactly match the schema determined after potential ALTER TABLE operations, as reflected in the temp_table_columns variable. The final DML statement is constructed dynamically within EXECUTE IMMEDIATE using column lists derived from temp_table_columns (e.g., dataform_columns_list). To use these dynamic column lists in the SELECT part of the MERGE's USING clause or the INSERT's SELECT statement, the source data must be in a table format. Using the raw table.incrementalQuery directly would not allow dynamic column selection based on dataform_columns_list within the EXECUTE IMMEDIATE context.

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +231,232 @@ from (${query}) as insertions`;
...
+    private createProcedureName(target: dataform.ITarget, uniqueId: string): string {
+    // Procedure names cannot contain hyphens.
+    const sanitizedUniqueId = uniqueId.replace(/-/g, "_");
+    return this.resolveTarget({
+      ...target,
+      name: `df_osc_${sanitizedUniqueId}`
+    });
+  }
+
+  private safeCallProcedure(
```
ikholopov-omni: The name should also describe the fact that procedure is being removed in all cases.

cli/api/dbadapters/execution_sql.ts
```diff
@@ -141,23 +142,67 @@ from (${query}) as insertions`;
...
-                  table.bigquery && table.bigquery.updatePartitionFilter
-                )
-              : this.insertInto(
-                  table.target,
-                  tableMetadata?.fields.map(f => f.name).map(column => `\`${column}\``),
-                  this.where(table.incrementalQuery || table.query, table.where)
-                )
-          )
-        );
+        const onSchemaChange = table.onSchemaChange || dataform.OnSchemaChange.IGNORE;
```
ikholopov-omni: `??` operator is more appropriate here.

ikholopov-omni (changes requested):

cli/api/dbadapters/execution_sql.ts
Tuseeq1: Hey,  we actually don't have a specific reason for doing it. It is clean up leftover from the when we were using store procedure inside the execute immediate. @SuchodolskiEdvin it might be worth running some tests on the engine directly with the generated queries with table.incrementalQuery instead of temp table. 

@spatel11 Thanks for pointing this out. Please feel free to open a feature request in our public tracker and we will take a look and plan this change. 
https://issuetracker.google.com/issues?q=status:open%20componentid:1193995&s=created_time:desc

spatel11: Done, thank you: https://issuetracker.google.com/issues/493440483

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +198,285 @@ from (${query}) as insertions`;
     return `drop ${this.tableTypeAsSql(type)} if exists ${this.resolveTarget(target)}`;
   }
 
+  private buildIncrementalSchemaChangeTasks(tasks: Tasks, table: dataform.ITable) {
+    const uniqueId = crypto.randomUUID().replace(/-/g, "_");
```
kolina: I don't think we need this UUID to be cryptographically secure

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +198,285 @@ from (${query}) as insertions`;
...
+
+    const shortEmptyTableName = `${table.target.name}_df_temp_${uniqueId}_empty`;
+    const emptyTempTableName = this.resolveTarget({
+      ...table.target,
+      name: shortEmptyTableName
+    });
+
+    const shortDataTableName = shortEmptyTableName.replace("_empty", "_data");
```
kolina: Let's try to not use a temporary table for results of incremental query following this [comment](https://github.com/dataform-co/dataform/pull/2101#discussion_r2906187361).

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +198,285 @@ from (${query}) as insertions`;
...
+      name: shortDataTableName
+    });
+
+    const procedureName = this.createProcedureName(table.target, uniqueId);
+    const procedureBody = this.incrementalSchemaChangeBody(
+      table,
+      this.resolveTarget(table.target),
+      emptyTempTableName,
+      dataTempTableName,
+      shortEmptyTableName
```
kolina: Do we need to pass both `shortEmptyTableName` and `emptyTempTableName` here?

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +198,285 @@ from (${query}) as insertions`;
...
+      dataTempTableName
+    );
+    tasks.add(Task.statement(createProcedureSql));
+    tasks.add(Task.statement(callProcedureSql));
+    tasks.add(Task.statement(`DROP PROCEDURE IF EXISTS ${procedureName};`));
+  }
+
+  private createProcedureName(target: dataform.ITarget, uniqueId: string): string {
+    // Procedure names cannot contain hyphens.
+    const sanitizedUniqueId = uniqueId.replace(/-/g, "_");
```
kolina: `uniqueId` already seems to be sanitized in the code above?

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +198,285 @@ from (${query}) as insertions`;
...
+END;`;
+
+    const callProcedureSql = this.safeCallAndDropProcedure(
+      procedureName,
+      emptyTempTableName,
+      dataTempTableName
+    );
+    tasks.add(Task.statement(createProcedureSql));
+    tasks.add(Task.statement(callProcedureSql));
+    tasks.add(Task.statement(`DROP PROCEDURE IF EXISTS ${procedureName};`));
```
kolina: It seems to be called in `safeCallAndDropProcedure`?

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +198,285 @@ from (${query}) as insertions`;
...
+EXCEPTION WHEN ERROR THEN
+  DROP TABLE IF EXISTS ${emptyTempTableName};
+  DROP TABLE IF EXISTS ${dataTempTableName};
+  DROP PROCEDURE IF EXISTS ${procedureName};
+  RAISE;
+END;
+DROP PROCEDURE IF EXISTS ${procedureName};`;
+  }
+
+  private inferSchemaSql(emptyTempTableName: string, query: string): string {
```
kolina: This code doesn't actually infers schema, so I'd rename this function

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +198,285 @@ from (${query}) as insertions`;
...
+    const onSchemaChange = table.onSchemaChange || dataform.OnSchemaChange.IGNORE;
+    let sql = `
+-- Apply schema change strategy (${dataform.OnSchemaChange[onSchemaChange]}).`;
+
+    switch (onSchemaChange) {
+      case dataform.OnSchemaChange.FAIL:
+        sql += `
+IF ARRAY_LENGTH(columns_added) > 0 OR ARRAY_LENGTH(columns_removed) > 0 THEN
+  RAISE USING MESSAGE = FORMAT(
+    "Schema mismatch defined by on_schema_change = 'FAIL'. Added columns: %t, removed columns: %t",
```
kolina: Is `%t` correct formatting here?

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +198,285 @@ from (${query}) as insertions`;
...
+IF ARRAY_LENGTH(columns_removed) > 0 THEN
+  RAISE USING MESSAGE = FORMAT(
+    "Column removals are not allowed when on_schema_change = 'EXTEND'. Removed columns: %t",
+    columns_removed
+  );
+END IF;
+
+FOR column_info IN (SELECT * FROM UNNEST(columns_added)) DO
+  EXECUTE IMMEDIATE FORMAT(
+    "ALTER TABLE ${qualifiedTargetTableName} ADD COLUMN IF NOT EXISTS %s %s",
```
kolina: Let's optimize following one of reported issues (multiple `ALTER TABLE` calls can eat quota): it should be possible to add all columns in one `ALTER TABLE` statement

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +198,285 @@ from (${query}) as insertions`;
...
+END FOR;
+`;
+        break;
+      case dataform.OnSchemaChange.SYNCHRONIZE:
+        const uniqueKeys = table.uniqueKey || [];
+        sql += `
+FOR removed_column_name IN (SELECT * FROM UNNEST(columns_removed)) DO
+  IF removed_column_name IN UNNEST(${JSON.stringify(uniqueKeys)}) THEN
+    RAISE USING MESSAGE = FORMAT(
+      "Cannot drop column %s as it is part of the unique key for table ${qualifiedTargetTableName}",
```
kolina: Let's not mix different string formatting ways like `%s` and `${qualifiedTargetTableName}`

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +198,285 @@ from (${query}) as insertions`;
...
+  END IF;
+END FOR;
+
+FOR column_info IN (SELECT * FROM UNNEST(columns_added)) DO
+  EXECUTE IMMEDIATE FORMAT(
+    "ALTER TABLE ${qualifiedTargetTableName} ADD COLUMN IF NOT EXISTS %s %s",
+    column_info.column_name,
+    column_info.data_type
+  );
+END FOR;
```
kolina: Let's not duplicate add statements and put it into separate function

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +198,285 @@ from (${query}) as insertions`;
...
+      this.createIncrementalDataTempTableSql(table, dataTempTableName),
+      this.declareDataformColumnsListSql(),
+      this.executeMergeOrInsertSql(table, qualifiedTargetTableName, dataTempTableName)
+    ].join("\n");
+  }
+
+  private createIncrementalDataTempTableSql(table: dataform.ITable, dataTempTableName: string): string {
+    return `
+CREATE OR REPLACE TEMP TABLE ${dataTempTableName} AS (
+  ${this.where(table.incrementalQuery || table.query, table.where)}
```
kolina: It's better to have a separate subfunction to generate `SELECT` part of an incremental query so it won't be repeated in multiple places (or just initialize it in one place)

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +198,285 @@ from (${query}) as insertions`;
...
+  private declareDataformColumnsListSql(): string {
+    return `
+DECLARE dataform_columns_list STRING;
+SET dataform_columns_list = (
+  SELECT IFNULL(STRING_AGG(CONCAT('\`', column_name, '\`'), ', '), '')
+  FROM UNNEST(temp_table_columns)
+);`;
+  }
+
+  private executeMergeOrInsertSql(
```
kolina: So it seems that you added a new logic of generating `MERGE` or `INSERT` statements from what we had before in `mergeInto` / `insertInto` functions.

I think we should avoid duplicating this logic and different behaviour when incremental schema change is enabled / disabled. So let's consolidate them

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +198,285 @@ from (${query}) as insertions`;
...
+        : mergeOnClause;
+
+      return `
+DECLARE dataform_columns_merge STRING;
+SET dataform_columns_merge = (
+  SELECT IFNULL(STRING_AGG(CONCAT('\`', column_name, '\` = S.\`', column_name, '\`'), ', '), '')
+  FROM UNNEST(temp_table_columns)
+);
+
+IF ARRAY_LENGTH(temp_table_columns) > 0 THEN
```
kolina: why do we need these checks?

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +198,285 @@ from (${query}) as insertions`;
...
+    "ON ${mergeOnClauseWithFilter} " ||
+    "WHEN MATCHED THEN " ||
+    "  UPDATE SET " || dataform_columns_merge || " " ||
+    "WHEN NOT MATCHED THEN " ||
+    "  INSERT (" || dataform_columns_list || ") VALUES (" || dataform_columns_list || ")"
+  );
+END IF;`;
+    } else {
+      return `
+IF ARRAY_LENGTH(temp_table_columns) > 0 THEN
```
kolina: Let's separate generations of merge / insert statements into separate functions

cli/api/execution_sql_test.ts
kolina, line 1: These tests don't seem to check a lot, for example there is no checks about how logic of merge / insert or actual altering schema is generated

kolina (changes requested):
Ideally we should add an e2e test checking a generated query like [here](https://github.com/dataform-co/dataform/blob/412d523108dbe411461c7d3c2e1b51484ede7e72/cli/index_run_e2e_test.ts#L269)

SuchodolskiEdvin:
> Ideally we should add an e2e test checking a generated query like [here](https://github.com/dataform-co/dataform/blob/412d523108dbe411461c7d3c2e1b51484ede7e72/cli/index_run_e2e_test.ts#L269)

I was wondering if we can actually create an e2e test for the onSchemaChange parameter? During a dataform run --dry-run, Dataform connects to BigQuery to fetch the current warehouseState. Because our example_incremental table doesn't actually exist in the test BigQuery project yet, tableMetadata comes back as undefined. When that happens, shouldWriteIncrementally() correctly evaluates to false, and Dataform natively falls back to generating a standard CREATE OR REPLACE TABLE.

Because the CLI does not expose a flag to inject a fake warehouseState JSON, I am not sure how we can force the E2E test to simulate a pre-existing table. I still can create the test to verify the parser and the dry-run fallback.

Technically we could run a real dataform run first to create the table in BigQuery, followed by a --dry-run to check the generated MERGE query. However, this would mutate the test BigQuery project, potentially introducing flakiness if tests run concurrently, and it would slow down the suite. The SQL generation is already mostly covered in the execution_sql_test.ts unit tests by mocking tableMetadata.

WDYT?

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: We could remove emptyTempTableName from the parameters and just compute it inside incrementalSchemaChangeBody() using shortEmptyTableName. However, the parent function buildIncrementalSchemaChangeTasks() still needs emptyTempTableName anyway to pass it to safeCallAndDropProcedure(). Since the parent has to compute it, passing it down to the child prevents us from duplicating the resolveTarget logic in two different places. So I would stay with this implementation.

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Done

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Done

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Done

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Done

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Done

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Fixed

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Fixed

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: I kept ${qualifiedTargetTableName} as a JS template literal and used %T for the BigQuery variables because they evaluate at different stages. This felt more readable than injecting the static table name as a %s argument into BigQuery's FORMAT() function.

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Done

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Done

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Consolidating them entirely is tricky because they operate differently: mergeInto/insertInto generate static SQL, while executeMergeOrInsertSql generates dynamic SQL via EXECUTE IMMEDIATE. Forcing both into one function hurts readability.                                                                                                     
                                                                                                                                           
To reduce duplication I suggest we extract the shared ON clause logic (handling uniqueKey and updatePartitionFilter) into a single helper method, and rename the original methods to clarify they are for static SQL.

WDYT?

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: originally I added it to prevent syntax errors if the query returned zero columns. However, a zero-column query is fundamentally invalid anyway. I removed the check to let BigQuery fail natively, so the user sees the real error instead of the operation silently skipping.

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Done

cli/api/execution_sql_test.ts
SuchodolskiEdvin: I enhanced the tests. Additionally they check the STRING_AGG formatting for ADD COLUMN and DROP COLUMN command, and they verify the structural generation of the procedural MERGE (when a uniqueKey is present) and INSERT (when a uniqueKey is absent) queries.

ikholopov-omni:
> > Ideally we should add an e2e test checking a generated query like [here](https://github.com/dataform-co/dataform/blob/412d523108dbe411461c7d3c2e1b51484ede7e72/cli/index_run_e2e_test.ts#L269)
> 
> I was wondering if we can actually create an e2e test for the onSchemaChange parameter? During a dataform run --dry-run, Dataform connects to BigQuery to fetch the current warehouseState. Because our example_incremental table doesn't actually exist in the test BigQuery project yet, tableMetadata comes back as undefined. When that happens, shouldWriteIncrementally() correctly evaluates to false, and Dataform natively falls back to generating a standard CREATE OR REPLACE TABLE.
> 
> Because the CLI does not expose a flag to inject a fake warehouseState JSON, I am not sure how we can force the E2E test to simulate a pre-existing table. I still can create the test to verify the parser and the dry-run fallback.
> 
> Technically we could run a real dataform run first to create the table in BigQuery, followed by a --dry-run to check the generated MERGE query. However, this would mutate the test BigQuery project, potentially introducing flakiness if tests run concurrently, and it would slow down the suite. The SQL generation is already mostly covered in the execution_sql_test.ts unit tests by mocking tableMetadata.
> 
> WDYT?

you can create a separate dataset per-run and prepare the table with the desired state in operation action before running the one you actually want to test. Just make sure to have a tear-down mechanism.

cli/api/dbadapters/execution_sql.ts
kolina: Can you pass `ITarget` instances as function arguments and call `resolveTarget` or `name` where you actually need a specific string?

It'll be more readable than passing these strings that have quite similar names

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +201,279 @@ from (${query}) as insertions`;
...
+-- Create empty table to extract schema of new query.
+CREATE OR REPLACE TABLE ${emptyTempTableName} AS (
+  SELECT * FROM (${query}) AS insertions LIMIT 0
+);`;
+  }
+
+  private compareSchemasSql(
+    database: string,
+    schema: string,
+    targetName: string,
```
kolina: You can also just pass `ITarget` here

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +201,279 @@ from (${query}) as insertions`;
...
+);`;
+  }
+
+  private applySchemaChangeStrategySql(
+    table: dataform.ITable,
+    qualifiedTargetTableName: string
+  ): string {
+    const onSchemaChange = table.onSchemaChange || dataform.OnSchemaChange.IGNORE;
+    let sql = `
+-- Apply schema change strategy (${dataform.OnSchemaChange[onSchemaChange]}).`;
```
kolina, line 305: what is the purpose of this?

cli/api/dbadapters/execution_sql.ts
```diff
@@ -186,6 +201,279 @@ from (${query}) as insertions`;
...
+    )
+  );
+END IF;`;
+  }
+
+  private runFinalDmlSql(
+    table: dataform.ITable,
+    qualifiedTargetTableName: string
+  ): string {
+    const query = this.getIncrementalQuery(table);
```
kolina: Do I understand correctly that this query will be inserted twice into generated SQL? Could we avoid it?

cli/api/dbadapters/execution_sql.ts
kolina: > Consolidating them entirely is tricky because they operate differently: mergeInto/insertInto generate static SQL, while executeMergeOrInsertSql generates dynamic SQL via EXECUTE IMMEDIATE. Forcing both into one function hurts readability.

I assume this logic with `EXECUTE IMMEDIATE` was copied from GCP code? If static SQL in CLI didn't use it before, I don't think you should introduce it as part of your changes: merge / insert logic should stay consistent for all modes of schema updates including `IGNORE`.

We can bring `EXECUTE IMMEDIATE` to synchronize GCP and CLI, but it can be done later.

cli/api/execution_sql_test.ts
```diff
@@ -0,0 +1,147 @@
...
+    const procedureSql = tasks.build()[0].statement;
+    expect(procedureSql).to.match(
+      /create or replace procedure `project-id.dataset-id.df_osc_.*`\(\)\s+options\(strict_mode=false\)/i
+    );
+    expect(procedureSql).to.include(
+      `"Schema mismatch defined by on_schema_change = 'FAIL'. Added columns: %T, removed columns: %T"`
+    );
+    expect(procedureSql).to.match(/call `project-id.dataset-id.df_osc_.*`\(\)/i);
+    expect(procedureSql).to.include("EXCEPTION WHEN ERROR THEN");
+    expect(procedureSql).to.match(/drop procedure if exists `project-id.dataset-id.df_osc_.*`/i);
```
kolina: Can we just match the whole query?

If there are non-deterministic places in logic, you can inject an interface for UUID generation to have a determinitic output in tests.

For readability you can put expected "golden" SQL into separate file and read it in tests for comparison

kolina:
> you can create a separate dataset per-run and prepare the table with the desired state in operation action before running the one you actually want to test. Just make sure to have a tear-down mechanism.

Igor's proposal is good

SuchodolskiEdvin:
> > > Ideally we should add an e2e test checking a generated query like [here](https://github.com/dataform-co/dataform/blob/412d523108dbe411461c7d3c2e1b51484ede7e72/cli/index_run_e2e_test.ts#L269)
> > 
> > 
> > I was wondering if we can actually create an e2e test for the onSchemaChange parameter? During a dataform run --dry-run, Dataform connects to BigQuery to fetch the current warehouseState. Because our example_incremental table doesn't actually exist in the test BigQuery project yet, tableMetadata comes back as undefined. When that happens, shouldWriteIncrementally() correctly evaluates to false, and Dataform natively falls back to generating a standard CREATE OR REPLACE TABLE.
> > Because the CLI does not expose a flag to inject a fake warehouseState JSON, I am not sure how we can force the E2E test to simulate a pre-existing table. I still can create the test to verify the parser and the dry-run fallback.
> > Technically we could run a real dataform run first to create the table in BigQuery, followed by a --dry-run to check the generated MERGE query. However, this would mutate the test BigQuery project, potentially introducing flakiness if tests run concurrently, and it would slow down the suite. The SQL generation is already mostly covered in the execution_sql_test.ts unit tests by mocking tableMetadata.
> > WDYT?
> 
> you can create a separate dataset per-run and prepare the table with the desired state in operation action before running the one you actually want to test. Just make sure to have a tear-down mechanism.

Done

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Fixed

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: It is just an informative comment added into the compiled SQL. In case users would want to inspect the query logs they can easily find which strategy was triggered. I can remove it if you find it unnecessary.

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Yes, the query is inserted twice into the final SQL script. It goes into the CREATE OR REPLACE TABLE ... statement to extract the schema metadata, and then again into the final MERGE or INSERT statement. AFAIK, we cannot avoid injecting it twice, because BigQuery requires it to materialize the INFORMATION_SCHEMA metadata before running the validation rules. 

Right now I can avoid recalculating it twice. Locally I've updated the code so getIncrementalQuery() is only called once in the incrementalSchemaChangeBody() and after the result of it passed down as a constant. The GCP code works the same way.

SuchodolskiEdvin:
@kolina 

Coming back to https://github.com/dataform-co/dataform/pull/2101#discussion_r3081875580

Yes, dynamic EXECUTE IMMEDIATE logic was adapted from the GCP. Ok, I will keep the merge / insert logic consistent with the existing static SQL in the CLI for now and avoid expanding the scope of this PR too much.                                                                
                                                                                                                                           
I will revert the dynamic DML generation and reuse the existing mergeInto and insertInto functions for the final data loading across all schema change strategies. I'll push the updated implementation shortly!

cli/api/execution_sql_test.ts
SuchodolskiEdvin: Done

cli/index_run_e2e_test.ts
```diff
@@ -646,4 +646,167 @@ select 2
...
+                }
+              ],
+              type: "table"
+            }
+          ],
+          warehouseState: executionGraph.warehouseState
+        };
+        
+        expect(executionGraph).deep.equals(expectedRunResult);
+        expect(statement).to.include("CREATE OR REPLACE PROCEDURE");
```
kolina, line 793: ideally you can use NodeJS BigQuery client (see how we use it in our code [here](https://github.com/dataform-co/dataform/blob/dc2ec7b3955e414f0ebab7f51b7bdf0d8ea3e176/cli/api/dbadapters/bigquery.ts#L37)) and just check the schema after running this action in non-dry mode.

kolina (approved):

ikholopov-omni (approved):

PR #2107, opened 2026-03-09
JiT: protos update to enable execution in CLI

rafal-hawrylak:
Updates the proto files to support execution in CLI.

### DIFF
--- a/common/errors/errors.ts
+++ b/common/errors/errors.ts
@@ -15,10 +15,10 @@ export function coerceAsError<T extends Error | any>(errorLike: T): T extends Er
   if ((errorLike as any) instanceof Error) {
     return errorLike as any;
   }
-  const error = (errorLike as any) as Error;
+  const error = errorLike as any;
   // Otherwise, attempt to reconstruct an error class from the object.
-  const message = error.message ? String(error.message) : String(errorLike);
-  const coercedError = new Error(message);
+  const message = error.message || error.error || (typeof errorLike === "string" ? errorLike : JSON.stringify(errorLike));
+  const coercedError = new Error(String(message));
   if (error.stack) {
     coercedError.stack = error.stack;
   }
--- a/protos/execution.proto
+++ b/protos/execution.proto
@@ -4,6 +4,7 @@ package dataform;
 
 import "protos/core.proto";
 import "protos/db_adapter.proto";
+import "google/protobuf/struct.proto";
 
 option go_package = "github.com/dataform-co/dataform/protos/dataform";
 
@@ -43,6 +44,8 @@ message ExecutionAction {
 
   ActionDescriptor action_descriptor = 9;
 
+  string jit_code = 12;
+
   reserved 1, 3, 7;
 }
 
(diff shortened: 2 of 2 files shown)


core/jit_compiler.ts
```diff
@@ -17,15 +17,61 @@ function makeMainBody<Context, T>(code: string): (jctx: JitContext<Context>) =>
...
-    jitResult.query = result;
-  } else {
-    jitResult = result;
+    return dataform.JitTableResult.create({ query: result });
   }
 
-  return dataform.JitTableResult.create(jitResult);
+  return dataform.JitTableResult.create({
+    query: result.query,
+    actionDescriptor: makeActionDescriptor(result),
```
ikholopov-omni: action descriptor must not be jit-compilable.

common/errors/errors.ts
```diff
@@ -15,10 +15,10 @@ export function coerceAsError<T extends Error | any>(errorLike: T): T extends Er
   if ((errorLike as any) instanceof Error) {
     return errorLike as any;
   }
-  const error = (errorLike as any) as Error;
+  const error = (errorLike as any) as any;
```
kolina: Is this second cast needed here?

protos/jit.proto
```diff
@@ -108,6 +108,8 @@ message JitCompilationRequest {
   JitCompilationTargetType compilation_target_type = 5;
   // List of additional file paths accessible at compilation.
   repeated string file_paths = 6;
+  // File name where the target is defined.
+  string file_name = 7;
```
ikholopov-omni: What's this for? For error handling or something else?

protos/execution.proto
```diff
@@ -43,6 +43,10 @@ message ExecutionAction {
 
   ActionDescriptor action_descriptor = 9;
 
+  string jit_code = 12;
+  google.protobuf.Struct jit_data = 13;
```
ikholopov-omni: Given that jit data is workflow-execution specific, not action-specific, why do we need it here?

protos/jit.proto
rafal-hawrylak: This is to resolve relative path in jit_worker if require is used.
Example:

```
publish("jit_require", { type: "table" }).jitCode(async (jctx) => {
  // This relative import requires the VM to know it's currently in the "definitions/" directory.
  const utils = require("../helpers/sql_utils.js"); 
  return `SELECT ${utils.getStaticValue()} as value`;
});
```

When the CLI sends this code to the jit_worker (NodeVM), it also sends file_name: "definitions/jit_require.js". If file_name were not provided, the VM would default to the project root, and the require call would fail because it would be looking for helpers/sql_utils.js one level above the project root.

https://github.com/dataform-co/dataform/pull/2109/changes/01fa6bab7ad76f8e502b0d78d74369cf5a5fec15#diff-c88d00fcc6836e8a1bbecf72502bf3cd2f54cf0a585c58b1ba234bcf8b00bb4cR50

ikholopov-omni: Ok. We explicitly DO NOT support requiring anything that is not a system Dataform library. E.g. in Cloud we only allow `require("@dataform/<pkg>")`.

We should actively block non-system import in JiT compile stage here as well.

And as a result, we don't need this field either.

Issue #1947, opened 2025-03-26
npm WARN deprecated vm2@3.9.19


Nkurra059:
Dear Git Team,

I am currently using the Dataform CLI in my build and release pipelines (Azure DevOps CI/CD) as part of our deployment process from lower to higher environments. However, my organization has raised concerns regarding the following warning:

npm WARN deprecated vm2@3.9.19: The library contains critical security issues and should not be used for production! The maintenance of the project has been discontinued. Consider migrating your code to isolated-vm.

Given that vm2 is marked as deprecated and has security issues, we would like to understand the following:

Any updates on this warning?
Is there a planned update or migration path for the Dataform CLI to replace vm2 with a safer alternative like isolated-vm?

Security Considerations:
If vm2 is still required for Dataform’s functionality, could you provide guidance on the risks involved and whether it's safe to use in production environments? Specifically, we are seeking advice on how to present this issue to our security team, outlining any steps we can take to mitigate risks while continuing to use the tool.

Suggested Alternatives:
Are there any official recommendations for alternatives to vm2 that would integrate with the Dataform CLI without compromising functionality or security?

We are looking to understand the situation better and align with best practices to ensure the security of our production deployments.

Thank you for your assistance!

mitchea5:
This has been brought up before.
https://github.com/dataform-co/dataform/issues/1615

Ceridan:
Hi @Nkurra059, thank you for opening the issue. You raised a fair concern. I want to confirm that we have migration from vm2 in our plans. Unfortunately, I can't provide an exact timeline yet but I will get back to you when we have it.

At the moment, the workaround is to run your workloads on GCP using UI or [Dataform API](https://cloud.google.com/dataform/reference/rest). In this case, the compilation will be done on GCP side and will have additional protection layers.

rafal-hawrylak:
Thank you for raising this important security concern. We have investigated the dependency on `vm2`. While `vm2@3.9.19` did have critical vulnerabilities, we've confirmed that the maintainers have released patches. Specifically, the fix for the reported vulnerability was included in `vm2@3.10.2` (see details in https://security.snyk.io/package/npm/vm2). What is more this version is already used in the Dataform CLI (https://github.com/dataform-co/dataform/blob/main/package.json#L75) starting **3.0.46** (https://github.com/dataform-co/dataform/releases/tag/3.0.46). Users should upgrade their `@dataform/cli` installation to at least version to ensure they are using a patched version of `vm2`.

PR #2109, opened 2026-03-09
JiT Worker in CLI and RPC infrastructure

rafal-hawrylak:
This change establishes the foundation for executing JiT compilation in an isolated environment. It introduces the worker process management, the RPC bridge for database access during JiT, and the necessary Bazel targets.


cli/api/commands/base_worker.ts
```diff
@@ -0,0 +1,70 @@
...
+    });
+  }
+
+  private resolveScript() {
+    try {
+      return require.resolve(this.loaderPath);
+    } catch (e) {
+      try {
+        return require.resolve("./worker_bundle.js");
+      } catch (e2) {
```
kolina: No need to obfuscate code yet :)

cli/api/commands/jit_compile_child_process.ts
```diff
@@ -0,0 +1,100 @@
...
+    dbclient: IDbClient,
+    timeoutMillis: number,
+    dryRun?: boolean
+  ): Promise<dataform.IJitCompilationResponse> {
+    return await this.runWorker(
+      timeoutMillis,
+      child => {
+        child.send({
+          type: "jit_compile",
+          request: request instanceof dataform.JitCompilationRequest ? request.toJSON() : request,
```
kolina: Such casts defies the purpose of interfaces TBH

cli/api/commands/jit/compiler.ts
```diff
@@ -0,0 +1,100 @@
...
+      timeoutMillis,
+      child => {
+        child.send({
+          type: "jit_compile",
+          request: request instanceof dataform.JitCompilationRequest ? request.toJSON() : request,
+          projectDir
+        });
+      },
+      async (message, child, resolve, reject) => {
+        if (message.type === "rpc_request") {
```
kolina, line 64: rename it to `rpc_request`?

cli/api/commands/jit_compile_child_process.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  private async handleRpcRequest(
+    message: IJitWorkerMessage,
+    child: ChildProcess,
+    dbadapter: IDbAdapter,
+    dbclient: IDbClient,
+    dryRun?: boolean
+  ) {
+    try {
+      const response = await handleRpc(
```
kolina: `handleDbRequest`?

cli/api/commands/base_worker.ts
```diff
@@ -0,0 +1,70 @@
+import { ChildProcess, fork } from "child_process";
```
kolina, line 1: It makes sense to put JiT-related logic into subdirectory

cli/api/commands/jit/rpc.ts
```diff
@@ -0,0 +1,101 @@
...
+  request: Uint8Array,
+  dryRun?: boolean
+): Promise<Uint8Array> {
+  const executeRequest = dataform.ExecuteRequest.decode(request);
+  const results = await dbclient.execute(executeRequest.statement, {
+    bigquery: {
+      labels: executeRequest.bigQueryOptions?.labels,
+      location: executeRequest.bigQueryOptions?.location,
+      jobPrefix: executeRequest.bigQueryOptions?.jobPrefix,
+      dryRun: dryRun || executeRequest.bigQueryOptions?.dryRun
```
kolina: Is this set by a compiled code or do we set it automatically? it'd be great to have consistency with running regular BQ actions

cli/api/commands/jit/rpc.ts
```diff
@@ -0,0 +1,101 @@
...
+    case "GetTable":
+      return await handleGetTable(dbadapter, request);
+    case "DeleteTable":
+      return await handleDeleteTable(dbadapter, request, dryRun);
+    default:
+      throw new Error(`Unrecognized RPC method: ${method}`);
+  }
+}
+
+async function handleExecute(
```
kolina, line 29: Should we enforce timeouts here?

cli/api/commands/jit/rpc.ts
```diff
@@ -0,0 +1,101 @@
...
+  });
+  return dataform.ExecuteResponse.encode(executeResponse).finish();
+}
+
+async function handleListTables(dbadapter: IDbAdapter, request: Uint8Array): Promise<Uint8Array> {
+  const listTablesRequest = dataform.ListTablesRequest.decode(request);
+  const targets = await dbadapter.tables();
+  const tablesMetadata = await Promise.all(
+    targets
+      .filter(target => !listTablesRequest.schema || target.schema === listTablesRequest.schema)
```
kolina: Is this filtering needed? Because `ListTables` already requires to provide a dataset

cli/api/commands/jit/rpc.ts
```diff
@@ -0,0 +1,101 @@
...
+    tables: tablesMetadata
+  });
+  return dataform.ListTablesResponse.encode(listTablesResponse).finish();
+}
+
+async function handleGetTable(dbadapter: IDbAdapter, request: Uint8Array): Promise<Uint8Array> {
+  const getTableRequest = dataform.GetTableRequest.decode(request);
+  const tableMetadata = await dbadapter.table(getTableRequest.target);
+  if (!tableMetadata) {
+    return dataform.TableMetadata.encode(dataform.TableMetadata.create({})).finish();
```
kolina: when this can happen?

cli/api/commands/jit/rpc.ts
```diff
@@ -0,0 +1,101 @@
...
+): Promise<Uint8Array> {
+  const deleteTableRequest = dataform.DeleteTableRequest.decode(request);
+  if (dryRun) {
+    return new Uint8Array();
+  }
+  await dbadapter.deleteTable(deleteTableRequest.target);
+  return new Uint8Array();
+}
+
+function mapRowToProto(row: { [key: string]: any }): google.protobuf.IStruct {
```
kolina: Makes sense to put into common util library

cli/api/commands/jit/rpc.ts
```diff
@@ -0,0 +1,101 @@
...
+    } else if (typeof val === "string") {
+      fields[key] = { stringValue: val };
+    } else if (typeof val === "boolean") {
+      fields[key] = { boolValue: val };
+    } else if (val === null || val === undefined) {
+      fields[key] = { nullValue: 0 };
+    } else {
+      // For more complex types (Date, Buffer, nested Objects), stringify or structify.
+      // This is a simplified mapper for IPC data transfer.
+      fields[key] = { structValue: val };
```
kolina: Will it be consistent with GCP behaviour?

cli/api/utils/constants.ts
```diff
@@ -0,0 +1 @@
+export const DEFAULT_COMPILATION_TIMEOUT_MILLIS = 30000;
```
kolina, line 1: I think we should already have a default compilation timeout for the `compile` command? Is it different?

cli/vm/jit_worker.ts
```diff
@@ -0,0 +1,118 @@
...
+    const requestMessage = dataform.JitCompilationRequest.fromObject(request);
+    const requestBytes = dataform.JitCompilationRequest.encode(requestMessage).finish();
+    const requestBase64 = Buffer.from(requestBytes).toString("base64");
+
+    // Use the action's file name as the VM filename for correct relative requires.
+    const vmFileName = requestMessage.fileName
+       ? path.resolve(projectDir, requestMessage.fileName)
+       : path.resolve(projectDir, "index.js");
+
+    const vm = new NodeVM({
```
kolina, line 57: TBH I'm not sure if we should continue using `vm2` here. It has known vulnerabilities, and though we don't need it for sandboxing, I was hoping that we'll get rid of it, otherwise users will continue asking questions if it's safe.

cli/vm/jit_worker.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+}
+
+if (require.main === module) {
+  if (process.send) {
+    process.send({ type: "worker_booted" });
+  }
+  process.on("message", async (message: any) => {
+    if (message.request) {
+      await handleJitRequest(message);
```
kolina, line 114: technically it'll be possible to send more than 1 JiT request to this process? if it's not intended, I'd reject subsequent requests

cli/api/commands/jit/rpc_test.ts
```diff
@@ -0,0 +1,210 @@
...
+    const decoded = dataform.ExecuteResponse.decode(response);
+
+    expect(decoded.rows.length).equals(1);
+    const fields = decoded.rows[0].fields;
+    expect(fields.num.numberValue).equals(42);
+    expect(fields.str.stringValue).equals("val");
+    expect(fields.bool.boolValue).equals(true);
+    expect(fields.n.nullValue).equals(0);
+
+    verify(mockClient.execute(statement, anything())).once();
```
kolina: should you check that request fields are passed to BQ correctly? applies to other tests as well

cli/api/commands/jit/rpc.ts
```diff
@@ -0,0 +1,101 @@
+import { IDbAdapter, IDbClient } from "df/cli/api/dbadapters";
```
kolina, line 3: To clarify: is this protobuf bridge protocol the same for GCP and CLI implementations?

kolina (changes requested):

cli/vm/jit_worker.ts
rafal-hawrylak: We can get back to this topic after JiT is finished. However vm2, at least at the moment has all the vulns solved: https://github.com/dataform-co/dataform/issues/1947#issuecomment-3932989010

kolina: I'll be fine as long as `npm` doesn't show a scary warning during the Dataform CLI installation :)

cli/api/commands/jit_compile_child_process.ts
```diff
@@ -0,0 +1,100 @@
...
+      projectDir,
+      dbadapter,
+      dbclient,
+      timeoutMillis,
+      dryRun
+    );
+  }
+
+  constructor() {
+    super("../../vm/jit_loader");
```
apilaskowski: I wonder if it is possible to avoid such hard-coded links. I can imagine that if we create jit directory for jit related actions it will fail.

cli/api/utils/constants.ts
apilaskowski: I think we wanted to have specific timeout for JiT, and it was smaller than default. It indeed is 300 seconds.

apilaskowski: Given that we might want to change name of this constant to reflect that.

protos/jit.proto
```diff
@@ -108,6 +108,8 @@ message JitCompilationRequest {
   JitCompilationTargetType compilation_target_type = 5;
   // List of additional file paths accessible at compilation.
   repeated string file_paths = 6;
+  // File name where the target is defined.
+  string file_name = 7;
```
ikholopov-omni: What do we need it for? For error handling?

cli/api/commands/jit/rpc.ts
ikholopov-omni: jit.proto declares that struct contains:
```
// Rows. For BigQuery, see
  // https://docs.cloud.google.com/bigquery/docs/reference/rest/v2/jobs/getQueryResults.
```

In other words, right now we expect a raw API result of "f,v" JSON struct, not bespoke conversion.

The BigQuery client for Node is strange in respect for this - it forcefully decodes those and removes rows from the original request. We could either:
a) Use `googleapis` package client instead for JiT
b) Come up with another protocol for encoding rows and implement it both here and in GCP. Let me know on chat if you need code pointers for GCP part.

ikholopov-omni (changes requested):

cli/api/commands/jit/rpc.ts
rafal-hawrylak: Both the CLI and GCP implementations follow the dataform.DbAdapter service defined in jit.proto.
- They both use the same binary "payloads" for the actual data (e.g., ExecuteRequest, ExecuteResponse).
- In CLI all 4 methods are implemented (Execute, ListTables, GetTable, and DeleteTable) and in GCP only Execute as of now (https://source.corp.google.com/piper///depot/google3/cloud/dataform/compilation/jitrunner/worker/engine.cc;rcl=881961593;l=163)

ikholopov-omni: See my comment below on the issue with "Execute" compatibility

cli/api/utils/constants.ts
rafal-hawrylak: @kolina Yes, but not as a constant. Now both use this constant.
@apilaskowski that's 30s, not sure we need different for JiT

cli/api/commands/jit/rpc.ts
rafal-hawrylak: Ok, I will. I will also extract timeout mechanism to bigquery.ts adapter so that it'd be reused by both rpc.ts calling the adapter and run.ts calling the adapter. In run.ts the time is cumulative across all the actions. In rpc.ts I think I'll keep timeout per rpc call as these should be rather shorter metadata checks from js. But still the mechanism can be shared, in rpc.ts I'll pass fixed timeout and in run.ts I will pass the remaining time to spend from budget.
Nevertheless, to keep this PR relatively small I'll add such a refactor once the JiT in CLI is fully merged.

cli/api/commands/jit/rpc.ts
rafal-hawrylak: The interface in CLI has only parameterless method:
`tables(): Promise<dataform.ITarget[]>;`

bigquery implementation as well.

ListTables - do you refer to proto method? This one is used to call this rpc code, not db.
`rpc ListTables(ListTablesRequest) returns (ListTablesResponse) {}`

Eventually `IDbAdapter` can have method with optional database and schema prams. Even though BigQuery service allows to filter by database (GCP project id) and schema filtering needs to happen on js side. It makes sense as we'd not get all the schemas in all the projects on js side to do late filtering. I'll add this as a follow up task it that makes sense to you.

cli/api/commands/jit/rpc.ts
rafal-hawrylak: Any time the table with given name (`getTableRequest.target`) is not found.
bigquery table method implementation is:
```
  public async table(target: dataform.ITarget): Promise<dataform.ITableMetadata> {
    const metadata = await this.getMetadata(target);

    if (!metadata) {
      return null;
    }
```

cli/api/commands/jit/rpc.ts
rafal-hawrylak: Done. Moved also the mirror version of this function added recently in Core. All moved to common/protos/struct (https://github.com/dataform-co/dataform/pull/2111/changes#diff-f05c6f4ca1b1cb84283c40f2a96541c3980d76d0b54690c5d461161a2b2cecfcR106)

cli/api/commands/jit/rpc.ts
rafal-hawrylak: Good point, I enhanced structValue handling and added listValue handling - all in the moved logic to common/protos/struct

cli/api/commands/jit/rpc.ts
rafal-hawrylak: PTAL on the approach with executeRaw

cli/api/commands/jit/rpc.ts
rafal-hawrylak: PTAL on the approach with executeRaw

ikholopov-omni: It has exactly the same problem. query() inside rawExecute already removes rows from response and only returns decoded ones as first component in the tuple `[rows, _, response]`

cli/api/commands/jit/rpc.ts
```diff
@@ -0,0 +1,87 @@
...
+    rowLimit: executeRequest.rowLimit ? executeRequest.rowLimit.toNumber() : undefined,
+    bigquery: {
+      labels: {
+        ...(options?.labels || {}),
+        ...(requestOptions?.labels || {})
+      },
+      location: requestOptions?.location || options?.location,
+      jobPrefix: [options?.jobPrefix, requestOptions?.jobPrefix].filter(Boolean).join("-") || undefined,
+      dryRun: !!(options?.dryRun || requestOptions?.dryRun)
+    }
```
kolina, line 51: I think it should be a consistent order of what takes priority here.

Also we'd need to update this logic every time a new option is added to `IBigQueryExecutionOptions`? I'd suggest using the full value of `options` by default (and overwrite with `requestOptions` where it makes sense)

cli/api/commands/jit/rpc.ts
kolina: > The interface in CLI has only parameterless method:
> `tables(): Promise<dataform.ITarget[]>;`

Will it be working correctly then? JiT request of `ListTablesRequest` can have different values of `database`, it may not be the same as whatever project is used by `dbadapter.tables()` to list tables in.

cli/api/commands/jit/rpc.ts
kolina: I'd rather rethrow `NOT_FOUND` back to the compiler in such cases

cli/api/dbadapters/bigquery.ts
```diff
@@ -92,6 +92,34 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+      .addSingleTask({
+        generator: async () => {
+          const [rows, , apiResponse] = (await this.getClient().query({
+            query: statement,
+            location: options.bigquery?.location,
+            maxResults: options.rowLimit,
+            useLegacySql: false,
+            labels: options.bigquery?.labels,
+            jobPrefix: options.bigquery?.jobPrefix,
+            dryRun: options.bigquery?.dryRun
```
kolina: Have a common logic how we set options for BQ job calls?

cli/api/dbadapters/bigquery.ts
```diff
@@ -392,8 +427,13 @@ export class BigQueryDbAdapter implements IDbAdapter {
                     reject(error);
                     return;
                   }
+                  const [, , apiResponse] = (await job[0].getQueryResults({
+                    maxResults: rowLimit,
+                    location
+                  })) as any;
```
kolina: Not sure that it's correct way to generate one more API request here

cli/vm/jit_worker.ts
```diff
@@ -0,0 +1,120 @@
...
+        correlationId
+      });
+    };
+
+    const projectLocalCorePath = path.join(projectDir, "node_modules", "@dataform", "core", "bundle.js");
+    const hasProjectLocalCore = fs.existsSync(projectLocalCorePath);
+
+    const requestMessage = dataform.JitCompilationRequest.fromObject(request);
+    const requestBytes = dataform.JitCompilationRequest.encode(requestMessage).finish();
+    const requestBase64 = Buffer.from(requestBytes).toString("base64");
```
kolina: Do we need to do base64 encoding here?

cli/vm/jit_worker.ts
```diff
@@ -0,0 +1,120 @@
...
+      env: process.env,
+      require: {
+        builtin: [],
+        context: "sandbox",
+        external: {
+          modules: ["@dataform/*"]
+        },
+        root: projectDir,
+        mock: hasProjectLocalCore ? {} : {
+          "@dataform/core": require("@dataform/core")
```
kolina: I don't think we have such a fallback for regular compilation?

cli/vm/jit_worker.ts
```diff
@@ -0,0 +1,120 @@
...
+    const jitCompileInVm = vm.run(`
+      const { jitCompiler } = require("@dataform/core");
+
+      global.require = require;
+
+      module.exports = async (requestBase64, armoredRpcCallback) => {
+        const requestBytes = new Uint8Array(Buffer.from(requestBase64, "base64"));
+
+        const internalRpcCallback = (method, reqBytes, callback) => {
+           const reqBase64 = Buffer.from(reqBytes).toString("base64");
```
kolina: same: do we need base64 encoding here?

kolina (changes requested):

cli/vm/jit_worker.ts
rafal-hawrylak: We use base64 also in regular compile:
function createCoreExecutionRequest(compileConfig: dataform.ICompileConfig): string {
  const filePaths = Array.from(
    new Set<string>(glob.sync("!(node_modules)/**/*.*", { cwd: compileConfig.projectDir }))
  );

  return encode64(dataform.CoreExecutionRequest, {
    // Add the list of file paths to the compile config if not already set.
    compile: { compileConfig: { filePaths, ...compileConfig } }
  });

cli/vm/jit_worker.ts
rafal-hawrylak: As well here

cli/vm/jit_worker.ts
ikholopov-omni: I think having base64 here is a bit of a legacy (and it brings additional overhead). I think we should be good to just patch serialized ArrayBuffer. See "else" branch here: https://github.com/dataform-co/dataform/blob/967a0020272da37d576a8fe052799a22eadbc7f5/core/main.ts#L37 and here https://github.com/dataform-co/dataform/blob/967a0020272da37d576a8fe052799a22eadbc7f5/core/main.ts#L75

cli/vm/jit_worker.ts
ikholopov-omni: See other "base64" comment thread

cli/api/commands/jit/rpc.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  return dataform.ExecuteResponse.encode({
+    rows: (results.rows || []).map(row => Structs.fromObject(row)),
+    schemaFields: results.schema || []
+  } as any).finish();
+}
+
+async function handleListTables(dbadapter: IDbAdapter, request: Uint8Array): Promise<Uint8Array> {
+  const listTablesRequest = dataform.ListTablesRequest.decode(request);
+  const targets = await dbadapter.tables(listTablesRequest.database, listTablesRequest.schema);
```
kolina: I think the `ListTables` API  could potentially return all needed metadata without the need to request it separately for each table?

cli/api/commands/jit/rpc_test.ts
```diff
@@ -0,0 +1,479 @@
...
+  });
+
+  test("Execute RPC merges global BigQuery options with request-specific options", async () => {
+    const mockAdapter = mock<IDbAdapter>();
+    const mockClient = mock<IDbClient>();
+
+    const statement = "SELECT 1";
+    const executeRequest = dataform.ExecuteRequest.create({
+      statement,
+      bigQueryOptions: {
```
kolina, line 273: Add `location` here to validate overrides?

cli/api/commands/jit/rpc_test.ts
```diff
@@ -0,0 +1,479 @@
...
+    const statement = "SELECT 1";
+    const encodedRequest = dataform.ExecuteRequest.encode({
+      statement,
+      bigQueryOptions: { labels: { request_label: "request_val" } }
+    }).finish();
+
+    when(mockClient.executeRaw(statement, anything())).thenResolve({ rows: [], metadata: {} });
+
+    await handleDbRequest(instance(mockAdapter), instance(mockClient), "Execute", encodedRequest, {
+      labels: { global_label: "global_val" }
```
kolina: I'd add `request_label` here to validate priority

cli/api/dbadapters/bigquery.ts
```diff
@@ -36,8 +36,14 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
   private readonly clients = new Map<string, BigQuery>();
 
-  constructor(credentials: dataform.IBigQuery, options?: { concurrencyLimit: number }) {
+  constructor(
+    credentials: dataform.IBigQuery,
+    options?: { concurrencyLimit?: number; bigqueryClient?: BigQuery }
+  ) {
     this.bigQueryCredentials = credentials;
+    if (options?.bigqueryClient) {
+      this.clients.set(credentials.projectId, options.bigqueryClient);
```
kolina: Do we need to update `getClient` function below as well?

cli/api/dbadapters/bigquery.ts
```diff
@@ -92,6 +98,31 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
+    if (!statement) {
+      throw new Error("Query string cannot be empty");
+    }
+    return this.pool
+      .addSingleTask({
+        generator: async () => {
+          const [rows, , apiResponse] = (await this.getClient().query({
+            ...this.prepareQueryOptions(statement, options.rowLimit, options.bigquery, options.params),
+            skipParsing: true
+          } as any)) as any;
```
kolina: This double conversion to `any` is not clear TBH

cli/api/dbadapters/bigquery.ts
kolina: So this is needed to return a schema? Could we use dry run here instead? Also it'd be great if we didn't make this call for AoT queries.

cli/vm/jit_worker.ts
kolina: Yep, we do exactly in GCP compiler (and it was a big performance improvement IIRC). I think the Dataform CLI was just missed from applying the same optimization.

cli/api/dbadapters/bigquery.ts
rafal-hawrylak: This is to inject client from test. bigquery didn't have unit tests so far.

kolina: I'd then use this client below for consistency

cli/api/dbadapters/bigquery.ts
```diff
@@ -35,9 +42,14 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
   private readonly clients = new Map<string, BigQuery>();
+  private readonly bigqueryClient?: BigQuery;
 
-  constructor(credentials: dataform.IBigQuery, options?: { concurrencyLimit: number }) {
+  constructor(
+    credentials: dataform.IBigQuery,
+    options?: { concurrencyLimit?: number; bigqueryClient?: BigQuery }
+  ) {
     this.bigQueryCredentials = credentials;
+    this.bigqueryClient = options?.bigqueryClient;
```
rafal-hawrylak, line 52: @kolina like this? Honestly I don't like having the field in class just for tests. Maybe extract logic of creating BigQuery clients outside BigQueryDbAdapter to BigQueryClientProvider - it would be easily mockable and we'd use the benefits of dependency injection?

cli/api/commands/jit/rpc.ts
rafal-hawrylak: Yes, we can hide it in bigquery adapter. Eventually, we can't avoid api requests per table to bigquery and that's because to fetch metadata there is no suitable api in bigquery client. We may get most of the metadata directly from INFORMATION_SCHEMA.TABLE and INFORMATION_SCHEMA.COLUMNS but we'd miss lastModifiedTime and labels. Labels though can be obtained from INFORMATION_SCHEMA.TABLE_OPTIONS but this is not stightforward, requires parsing and what the most important this is low level information for which we'd couple bigquery implementation knowledge into our code.

cli/api/commands/jit/rpc.ts
rafal-hawrylak: This one is waiting for:
- this to be released: https://github.com/googleapis/google-cloud-node/pull/7848
- this to be reviewed, merged and released: https://github.com/googleapis/google-cloud-node/pull/7905

cli/api/commands/jit/rpc.ts
rafal-hawrylak: This one is waiting for:

- this to be released: https://github.com/googleapis/google-cloud-node/pull/7848
- this to be reviewed, merged and released: https://github.com/googleapis/google-cloud-node/pull/7905

cli/api/dbadapters/bigquery.ts
kolina: Yeah, your suggestion SGTM

cli/api/dbadapters/bigquery.ts
```diff
@@ -129,22 +166,31 @@ export class BigQueryDbAdapter implements IDbAdapter {
...
-    const allTables: dataform.ITarget[] = [];
-    tables.forEach((tablesResult: GetTablesResponse) =>
-      tablesResult[0].forEach(table =>
-        allTables.push({
-          database: table.bigQuery.projectId,
-          schema: table.dataset.id,
-          name: table.id
-        })
-      )
+  public async tables(database: string, schema?: string): Promise<dataform.ITableMetadata[]> {
```
kolina, line 169: Why did you remove `autoPaginate` and `maxResults` options in your new version?

cli/api/dbadapters/bigquery.ts
```diff
@@ -226,7 +279,7 @@ export class BigQueryDbAdapter implements IDbAdapter {
   public async createSchema(database: string, schema: string): Promise<void> {
     await this.execute(
       `create schema if not exists \`${database || this.bigQueryCredentials.projectId}.${schema}\``,
-      { bigquery: { location: this.bigQueryCredentials.location || "US" } }
+      { bigquery: { location: this.bigQueryCredentials.location } }
```
kolina, line 282: This diff is also not clear

cli/api/commands/jit/rpc.ts
kolina: If `ListTable` can't return all required data, then I'm ok

kolina (commented):
The current PR LGTM, there are a couple of questions about diffs in BigQuery adapters.

After they're responded, I can approve from my side

apilaskowski:
@kolina can you take a look here?

cli/api/commands/jit/rpc.ts
rafal-hawrylak: bigquery lib is available and set as dependency

cli/api/dbadapters/bigquery.ts
kolina: Can you explain it here please?

cli/api/dbadapters/bigquery.ts
kolina: Same, let's explain it here

cli/api/dbadapters/bigquery.ts
rafal-hawrylak: Right, overlooked

cli/api/dbadapters/bigquery.ts
rafal-hawrylak: The hardcoded "US" fallback was removed to ensure consistency with how the BigQuery client is instantiated elsewhere in the file. Hardcoding it here overrides intentional default behaviors from GCP or configurations passed from higher up, which could accidentally create datasets in the US region when the user intended otherwise.

kolina: It's true, but from the other side there can always be someone relying on existing behaviour even if it's not documented.

Anyway, up to you

kolina (approved):

cli/api/dbadapters/bigquery.ts
rafal-hawrylak: I'd treat it as fixing bug. Not documented behaviour should not be relied on.

apilaskowski (approved):

common/protos/structs.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  public static fromObject(obj: { [key: string]: any }): google.protobuf.IStruct {
+    const fields: { [key: string]: google.protobuf.IValue } = {};
+    for (const [key, val] of Object.entries(obj)) {
+      fields[key] = this.toValue(val);
+    }
+    return { fields };
+  }
+
+  private static fromValue(value: google.protobuf.IValue): any {
```
ikholopov-omni: Do we need `any` here? Can we be more specific and use 'null|number|string|bool|{[key: string]: {}}|undefined instead?

common/protos/structs.ts
```diff
@@ -0,0 +1,70 @@
+import { google } from "df/protos/ts";
+
+export class Structs {
+  public static toObject(struct?: google.protobuf.IStruct): { [key: string]: any } | undefined {
```
ikholopov-omni: Let's not use any here, we can put { [key: string]: {} }

ikholopov-omni (changes requested):
Overall LGTM, feel free to submit once the comments are addressed.

common/protos/structs.ts
rafal-hawrylak: It needs to accept also multi-nested arrays of those, so:
export type AnyValue = null | number | string | boolean | undefined | AnyValue[] | { [key: string]: AnyValue };

rafal-hawrylak:
> Overall LGTM, feel free to submit once the comments are addressed.

Addressed. However the merged is blocked due to request for a change.

ikholopov-omni (approved):

PR #2110, opened 2026-03-09
Integrate JiT compilation into CLI

rafal-hawrylak:


cli/api/commands/compile.ts
```diff
@@ -96,68 +98,24 @@ export async function compile(
...
-          resolve(messageOrError);
-          return;
-        }
-        reject(coerceAsError(messageOrError));
-      });
-
-      this.childProcess.on("close", exitCode => {
-        if (exitCode !== 0) {
-          reject(new Error(`Compilation child process exited with exit code ${exitCode}.`));
+    const timeoutValue = compileConfig.timeoutMillis || DEFAULT_COMPILATION_TIMEOUT_MILLIS;
```
kolina, line 109: It seems that the default timeout is smaller currently?

cli/api/commands/compile.ts
```diff
@@ -96,68 +98,24 @@ export async function compile(
...
-
-      this.childProcess.on("message", (messageOrError: string | Error) => {
-        if (typeof messageOrError === "string") {
-          resolve(messageOrError);
-          return;
-        }
-        reject(coerceAsError(messageOrError));
-      });
-
-      this.childProcess.on("close", exitCode => {
```
kolina, line 136: Just to clarify: in the base worker you have a hook for "exit", they'll be equivalent?

cli/api/commands/compile.ts
```diff
@@ -96,68 +98,24 @@ export async function compile(
...
-          () => reject(new CompilationTimeoutError("Compilation timed out")),
-          compileConfig.timeoutMillis || 5000
-        ))
-    );
-    try {
-      await Promise.race([timeout, compileInChildProcess]);
-      return await compileInChildProcess;
-    } finally {
-      if (!this.childProcess.killed) {
-        this.childProcess.kill("SIGKILL");
```
kolina, line 158: In the base worker you don't send SIGKILL, I'd preserve this logic to avoid compiler ignoring SIGTERM for some reason

cli/api/commands/run.ts
```diff
@@ -68,10 +84,15 @@ export class Runner {
   constructor(
     private readonly dbadapter: dbadapters.IDbAdapter,
     private readonly graph: dataform.IExecutionGraph,
-    private readonly executionOptions: IExecutionOptions = {},
+    private readonly projectDir: string,
```
kolina: if it's part of `IExecutionOptions`, do you need to pass it separately?

cli/api/commands/run.ts
```diff
@@ -329,43 +350,58 @@ export class Runner {
     actionResult.timing = timer.current();
     this.notifyListeners();
 
-    await this.dbadapter.withClientLock(async client => {
-      // Start running tasks from the last executed task (if any), onwards.
-      for (const task of action.tasks.slice(actionResult.tasks.length)) {
-        if (this.stopped) {
-          return actionResult;
+    try {
+      await this.dbadapter.withClientLock(async client => {
```
kolina: This `withClientLock` function doesn't do anything meaningful today (maybe it was needed before when we supported other warehouses), I'd remove it.

cli/api/commands/run.ts
```diff
@@ -329,43 +350,58 @@ export class Runner {
...
+                  ...(this.executionOptions?.bigquery?.labels || {}),
+                  ...(action.actionDescriptor?.bigqueryLabels || {})
+                },
+                actionRetryLimit: this.executionOptions?.bigquery?.actionRetryLimit,
+                jobPrefix: this.executionOptions?.bigquery?.jobPrefix,
+                dryRun: this.executionOptions?.bigquery?.dryRun,
+                reservation:
+                  action.actionDescriptor?.reservation ||
+                  this.graph.projectConfig?.defaultReservation
+              }
```
kolina: Do we preserve this logic for BQ queries initiated from JiT?

cli/api/commands/run.ts
```diff
@@ -329,43 +350,58 @@ export class Runner {
...
           }
-        } else {
-          actionResult.tasks.push({
-            status: dataform.TaskResult.ExecutionStatus.SKIPPED
-          });
         }
-      }
-    });
+      });
+    } catch (e) {
```
kolina, line 452: what new exceptions are we catching here?

cli/api/commands/run.ts
```diff
@@ -423,6 +459,9 @@ export class Runner {
     };
     parentAction.tasks.push(taskResult);
     this.notifyListeners();
+    if (options.bigquery?.dryRun) {
+      (taskResult as any).statement = task.statement;
+    }
```
kolina, line 521: What is the purpose of this logic?

cli/api/commands/run.ts
```diff
@@ -466,6 +505,72 @@ export class Runner {
...
+      compilationTargetType,
+      jitData: this.graph.jitData
+    });
+
+    const timeoutMillis = this.graph.runConfig?.timeoutMillis || DEFAULT_COMPILATION_TIMEOUT_MILLIS;
+
+    try {
+      const jitResponse = await this.executionOptions.jitCompiler(
+        jitRequest,
+        this.projectDir,
```
kolina: Is it consistent with behaviour in GCP that JiT compilation needs to access original source code?

cli/api/commands/run.ts
```diff
@@ -466,6 +505,72 @@ export class Runner {
...
+        this.executionOptions.bigquery?.dryRun
+      );
+
+      if (jitResponse.table) {
+        const table = dataform.Table.create({
+          ...action,
+          ...jitResponse.table,
+          enumType: action.tableType === "view" ? dataform.TableType.VIEW : (action.tableType === "incremental" ? dataform.TableType.INCREMENTAL : dataform.TableType.TABLE)
+        });
+        action.tasks = this.executionSql.publishTasks(table, this.graph.runConfig, this.warehouseStateByTarget.get(targetStringifier.stringify(action.target))).build();
```
kolina: Can we create a helper function to create tasks from compiled action? So it can be reused here and in `cli/api/commands/build.ts`

cli/vm/compile.ts
```diff
@@ -82,6 +82,9 @@ export function listenForCompileRequest() {
 }
 
 if (require.main === module) {
+  if (process.send) {
+    process.send({ type: "worker_booted" });
+  }
```
kolina, line 91: Just to clarify: this code is bundled with the Dataform CLI (not with Dataform core) and so it shouldn't break compatibility with different version of the Dataform core?

cli/bigquery_test.ts
```diff
@@ -0,0 +1,64 @@
+import { expect } from "chai";
```
kolina: Should we put this test in the same directory as `BigQueryDbAdapter` code?

cli/console.ts
```diff
@@ -511,9 +511,14 @@ function printExecutedActionErrors(
     task => task.status === dataform.TaskResult.ExecutionStatus.FAILED
   );
   failingTasks.forEach((task, i) => {
-    executionAction.tasks[i].statement.split("\n").forEach(line => {
-      writeStdErr(`${DEFAULT_PROMPT}${line}`, 1);
-    });
+    // For JiT actions, the original executionAction.tasks might be empty
+    // since they are generated during re-compilation.
+    const statement = (executionAction.tasks[i] || (task as any)).statement;
```
kolina: This is a bit hacky :) TBH it makes sense to extend `TaskResult` to store something like `jit_statement` for JiT-compiled code

cli/index.ts
```diff
@@ -583,25 +583,34 @@ export function runCli() {
...
+            isJsonOutput &&
+            !executionGraph.actions.some(action => !!action.jitCode)
+          ) {
+            printExecutionGraph(executionGraph, isJsonOutput);
             return;
           }
 
           if (argv[runTestsOptionName]) {
-            print(`Running ${compiledGraph.tests.length} unit tests...\n`);
+            if (!isJsonOutput) {
```
kolina: Put logging into a helper? To have check for json output in one place

cli/index.ts
```diff
@@ -583,25 +583,34 @@ export function runCli() {
...
             },
             dbadapter
           );
 
-          if (argv[dryRunOptionName] && argv[jsonOutputOption.name]) {
-            printExecutionGraph(executionGraph, argv[jsonOutputOption.name]);
+          if (
+            argv[dryRunOptionName] &&
+            isJsonOutput &&
+            !executionGraph.actions.some(action => !!action.jitCode)
```
kolina, line 654: Do we have a test validating correct handling of dry run in graph execution with JiT? I saw that it should be respected, but it's not immediately clear from code in this file.

cli/tests/jit/index_jit_advanced_test.ts
```diff
@@ -0,0 +1,169 @@
...
+    const projectDir = tmpDirFixture.createNewTmpDir();
+    await setupJitProject(tmpDirFixture, projectDir);
+    const incPrePostPath = path.join(projectDir, "definitions", "inc_pre_post_jit.js");
+    fs.writeFileSync(
+      incPrePostPath,
+      `publish("inc_pre_post_jit", { type: "incremental" }).jitCode(async (jctx) => {
+        if (jctx.incremental()) {
+          return {
+            query: "SELECT 'inc_path_query' as q",
+            preOps: ["SELECT 'inc_path_pre' as p"]
```
kolina, line 64: Add validation of incremental mode as well below?

cli/index_jit_dependency_test.ts
```diff
@@ -0,0 +1,91 @@
...
+        "--include-deps"
+      ])
+    );
+
+    expect(runResult.exitCode).equals(0);
+    const executedGraph = JSON.parse(runResult.stdout);
+    // Should have BOTH tables because of --include-deps
+    expect(executedGraph.actions.length).to.equal(2);
+    expect(executedGraph.actions.some((a: any) => a.target.name === "table_a")).to.equal(true);
+    expect(executedGraph.actions.some((a: any) => a.target.name === "table_b")).to.equal(true);
```
kolina: check that generated query references `table_a`?

cli/tests/jit/index_jit_advanced_test.ts
```diff
@@ -0,0 +1,169 @@
+import { expect } from "chai";
```
kolina, line 1: I'd put all JiT tests into subdirectory

cli/index_jit_main_test.ts
```diff
@@ -0,0 +1,273 @@
...
+        "--json",
+        "--actions=disabled_jit"
+      ])
+    );
+
+    // It should exit successfully.
+    expect(runResult.exitCode).equals(0);
+    // If no actions are run, stdout might be empty or empty JSON.
+    const stdout = runResult.stdout.trim();
+    if (stdout.length > 0) {
```
kolina: If this test is deterministic, this `if` is not needed?

cli/tests/jit/index_jit_main_test.ts
```diff
@@ -0,0 +1,273 @@
...
+    );
+
+    expect(runResult.exitCode).equals(1);
+
+    const executedGraph = JSON.parse(runResult.stdout);
+    const rpcAction = executedGraph.actions.find((a: any) => a.target.name === "rpc_jit");
+
+    expect(!!rpcAction).to.equal(true);
+    expect(rpcAction.status).to.equal(3); // FAILED
+    expect(rpcAction.tasks[0].errorMessage).to.include("JiT compilation error");
```
kolina, line 211: can you add a comment in the code about why it fails?

cli/jit_run_test.ts
```diff
@@ -0,0 +1,318 @@
...
+    const actionResult = result.actions[0];
+    expect(actionResult.target.name).equals("jit_table");
+    expect(actionResult.status).equals(dataform.ActionResult.ExecutionStatus.SUCCESSFUL);
+
+    // Verify task results
+    expect(actionResult.tasks.length).equals(1);
+    expect(actionResult.tasks[0].status).equals(dataform.TaskResult.ExecutionStatus.SUCCESSFUL);
+
+    // Verifying that it used the database to generate the task (via createRpcCallback in jitCompile)
+    verify(mockClient.execute(anything(), anything())).atLeast(1);
```
kolina: JiT code in this test doesn't seem to do any BQ calls? So it seems that this request is made for the compiled query?

Applies to other tests as well

cli/tests/jit/jit_run_test.ts
```diff
@@ -0,0 +1,318 @@
...
+    const aotResult = result.actions.find((a: dataform.IActionResult) => a.target.name === "aot_table");
+    const jitResult = result.actions.find((a: dataform.IActionResult) => a.target.name === "jit_table");
+
+    expect(aotResult.status).equals(dataform.ActionResult.ExecutionStatus.SUCCESSFUL);
+    expect(aotResult.tasks.length).equals(1);
+    expect(aotResult.tasks[0].status).equals(dataform.TaskResult.ExecutionStatus.SUCCESSFUL);
+
+    expect(jitResult.status).equals(dataform.ActionResult.ExecutionStatus.SUCCESSFUL);
+    expect(jitResult.tasks.length).equals(1);
+    expect(jitResult.tasks[0].status).equals(dataform.TaskResult.ExecutionStatus.SUCCESSFUL);
```
kolina, line 130: Should we validate BQ calls in these tests as well?

cli/tests/jit/jit_run_test.ts
```diff
@@ -0,0 +1,318 @@
...
+  test("Handles JiT incremental table compilation", async () => {
+    const { mockClient, adapterInstance } = createMocks();
+
+    const executionGraph = createGraph([
+      {
+        target: { database: "db", schema: "sch", name: "incremental_jit" },
+        type: "table",
+        tableType: "incremental",
+        jitCode: `async (jctx) => {
+          return jctx.incremental() ? "SELECT 'inc' as t" : "SELECT 'full' as t";
```
kolina, line 215: add a test for incremental mode?

cli/api/commands/compile.ts
```diff
@@ -96,68 +98,24 @@ export async function compile(
...
-      fork(require.resolve(forkScript), [], { stdio: [0, 1, 2, "ipc", "pipe"] })
-    );
-  }
-  private readonly childProcess: ChildProcess;
-
-  constructor(childProcess: ChildProcess) {
-    this.childProcess = childProcess;
+export class CompileChildProcess extends BaseWorker<string, string | Error> {
+  constructor() {
+    super("../../vm/compile_loader");
```
kolina: Do I understand correctly that this will change the order of fork script resolution? I'd prefer to keep the original order

packages/@dataform/cli/worker.ts
```diff
@@ -1,2 +1,36 @@
...
+    if (message.request) {
+      // It's a JiT request
+      await handleJitRequest(message);
+    } else {
+      // It's an AoT compile request
+      const compiledGraphJson = compile(message);
+
+      // AoT compile() in cli/vm/compile.ts returns a JSON string of CompiledGraph.
+      // We need to wrap it in a CoreExecutionResponse and Base64 encode it
+      // to match what cli/api/commands/compile.ts expects.
```
kolina: Is it true? From my reading of code:
* `compile` in `cli/vm/compile.ts` calls `main` function of Dataform core and returns its result [here](https://github.com/dataform-co/dataform/blob/df8cb18c249ef29d44cf383c337fd9270058c0a2/cli/vm/compile.ts#L63), in the CLI we always pass base-64 core execution request [here](https://github.com/dataform-co/dataform/blob/df8cb18c249ef29d44cf383c337fd9270058c0a2/cli/vm/compile.ts#L96)
* The `main` function in the Dataform core in this case returns base64-encoded serialized proto response [here](https://github.com/dataform-co/dataform/blob/df8cb18c249ef29d44cf383c337fd9270058c0a2/core/main.ts#L72) (comment about older version of the Dataform CLI is a bit confusing)

protos/execution.proto
```diff
@@ -46,7 +46,9 @@ message ExecutionAction {
 
   string jit_code = 12;
 
-  reserved 1, 3, 7, 13;
+  bool disabled = 15;
+
+  reserved 1, 3, 7;
```
kolina, line 59: why do we have less reserved field numbers here?

tests/api/api.spec.ts
```diff
@@ -976,7 +976,7 @@ suite("@dataform/api", () => {
       mockDbAdapterInstance.withClientLock = async callback =>
         await callback(mockDbAdapterInstance);
 
-      const runner = new Runner(mockDbAdapterInstance, RUN_TEST_GRAPH);
+      const runner = new Runner(mockDbAdapterInstance, RUN_TEST_GRAPH, ".");
```
kolina: If someone uses the Dataform CLI as a library, will they need to update their code in the same manner?

kolina (changes requested):
Let's test compatibility with different Dataform core versions.

Also this change seems worth it to bump a minor version of Dataform core

ikholopov-omni (changes requested):
Blocked on review of https://github.com/dataform-co/dataform/pull/2109/changes

cli/api/commands/compile.ts
rafal-hawrylak: I changed the original timeout from 5s to 30s. Now increased to 5 minutes if 30s might not be enough.

cli/api/commands/compile.ts
rafal-hawrylak: The previous compile.ts used close to detect if the process crashed before sending a message. The new base worker uses exit for the same purpose.

cli/api/commands/run.ts
rafal-hawrylak: Right, added merging of labels and fixed reservation

cli/api/commands/run.ts
rafal-hawrylak: Removing withClientLock simplified the execution flow. JiT and SQL tasks now handle specific errors internally. This outer block serves as a safety net for unexpected failures. I've updated the logic to avoid reporting duplicate errors.

cli/api/commands/run.ts
rafal-hawrylak: this allows cli to display dynamically generated SQL for JiT actions during dry runs and error reporting. It gives users a chance to audit or debug JiT logic

cli/api/commands/run.ts
rafal-hawrylak: In the CLI, projectDir allows the JiT worker to resolve local require() calls and find @dataform/core. While this is convenient for local development, it does differ from the cloud environment where source code isn't available during execution. We may want to consider if JiT code should be strictly self-contained for better parity. This would make the project to be unportable from local to GCP and we don't want it, right?

cli/vm/compile.ts
rafal-hawrylak: Yes, this is CLI-only. It dynamically loads the project's Core version at runtime, so it doesn't affect compatibility

cli/jit_run_test.ts
rafal-hawrylak: Right, compilation is mocked - the run is tested. Updated the comment message.

packages/@dataform/cli/worker.ts
rafal-hawrylak: That's right - cleared this out

cli/tests/jit/index_jit_main_test.ts
rafal-hawrylak: Fixed so that errors are propagated properly now

cli/api/commands/run.ts
```diff
@@ -24,18 +28,27 @@ export interface IExecutedAction {
...
+    timeoutMillis?: number,
+    options?: IBigQueryExecutionOptions
+  ) => Promise<dataform.IJitCompilationResponse>;
 }
 
 export function run(
   dbadapter: dbadapters.IDbAdapter,
   graph: dataform.IExecutionGraph,
-  executionOptions?: IExecutionOptions,
+  executionOptions: string | IExecutionOptions | dataform.IRunResult = ".",
```
kolina: This change in types looks quite unclear

cli/api/commands/run.ts
```diff
@@ -329,7 +386,14 @@ export class Runner {
     actionResult.timing = timer.current();
     this.notifyListeners();
 
-    await this.dbadapter.withClientLock(async client => {
+    try {
+      if (action.jitCode) {
+        await this.compileJitAction(action, actionResult, this.dbadapter);
+        if ((actionResult.status as dataform.ActionResult.ExecutionStatus) === dataform.ActionResult.ExecutionStatus.FAILED) {
```
kolina: Do we need casts here? Here and below

cli/api/commands/run.ts
```diff
@@ -423,6 +495,10 @@ export class Runner {
     };
     parentAction.tasks.push(taskResult);
     this.notifyListeners();
+    if (options.bigquery?.dryRun) {
+      taskResult.compiledSql = task.statement;
+      (taskResult as any).statement = task.statement;
```
kolina: Can you do it without converting to `any`?

cli/api/commands/run.ts
```diff
@@ -466,6 +542,87 @@ export class Runner {
...
+      compilationTargetType,
+      jitData: this.graph.jitData
+    });
+
+    const timeoutMillis = this.graph.runConfig?.timeoutMillis || DEFAULT_COMPILATION_TIMEOUT_MILLIS;
+
+    try {
+      const jitResponse = await this.executionOptions.jitCompiler(
+        jitRequest,
+        this.executionOptions.projectDir,
```
kolina: Reponding to the older comment

> In the CLI, projectDir allows the JiT worker to resolve local require() calls and find @dataform/core. While this is convenient for local development, it does differ from the cloud environment where source code isn't available during execution. We may want to consider if JiT code should be strictly self-contained for better parity. This would make the project to be unportable from local to GCP and we don't want it, right?

I think it should be consistent. Allowing JiT compilation to access other files in a project seems ok to me, but if we don't support it in JiT, it'd make sense to not allow it in CLI as well

cli/api/commands/run.ts
```diff
@@ -466,6 +542,87 @@ export class Runner {
...
+          jobPrefix: this.executionOptions.bigquery?.jobPrefix,
+          labels: {
+            ...(this.executionOptions?.bigquery?.labels || {}),
+            ...(action.actionDescriptor?.bigqueryLabels || {})
+          },
+          actionRetryLimit: this.executionOptions.bigquery?.actionRetryLimit,
+          reservation:
+            action.actionDescriptor?.reservation ||
+            this.graph.projectConfig?.defaultReservation
+        }
```
kolina: Can you deduplicate this with filling very similar options in `executeAction`?

cli/api/commands/run.ts
```diff
@@ -49,7 +62,44 @@ export function run(
...
+  private static handleParamsOverloads(
+    executionOptions: string | IExecutionOptions | dataform.IRunResult,
+    partiallyExecutedRunResult: dataform.IRunResult
+  ): { options: IExecutionOptions; runResult: dataform.IRunResult } {
+    if (
+      typeof executionOptions !== "string" &&
+      (executionOptions as dataform.IRunResult).actions &&
+      !(executionOptions as IExecutionOptions).projectDir &&
+      !(executionOptions as IExecutionOptions).bigquery &&
+      !(executionOptions as IExecutionOptions).jitCompiler
```
kolina: Should it be so complicated?

cli/api/dbadapters/bigquery.ts
```diff
@@ -30,14 +37,52 @@ export interface IBigQueryExecutionOptions {
...
+export class BigQueryDbAdapter implements IDbAdapter {
+  private bigQueryCredentials: dataform.IBigQuery;
+  private pool: PromisePoolExecutor;
+  private clientProvider: IBigQueryClientProvider;
+
+  constructor(
+    credentials: dataform.IBigQuery,
+    options?: {
+      concurrencyLimit?: number;
+      bigqueryClient?: BigQuery;
```
kolina: I suppose we can just have a separate `IBigQueryClientProvider` returning a static client

cli/vm/jit_worker.ts
```diff
@@ -0,0 +1,132 @@
...
+          "@dataform/core": require("@dataform/core")
+        },
+        resolve: (moduleName, parentDirName) => {
+          if (moduleName.startsWith(".")) {
+            return path.resolve(parentDirName, moduleName);
+          }
+          return moduleName;
+        }
+      },
+      sourceExtensions: ["js", "json"]
```
kolina: add `yaml` here?

cli/console.ts
```diff
@@ -568,9 +528,14 @@ function printExecutedActionErrors(
     task => task.status === dataform.TaskResult.ExecutionStatus.FAILED
   );
   failingTasks.forEach((task, i) => {
-    executionAction.tasks[i].statement.split("\n").forEach(line => {
-      writeStdErr(`${DEFAULT_PROMPT}${line}`, 1);
-    });
+    // For JiT actions, the original executionAction.tasks might be empty
+    // since they are generated during re-compilation.
+    const statement = task.compiledSql || (executionAction.tasks[i] || (task as any)).statement;
```
kolina: Do we need `task as any` here?

cli/index.ts
```diff
@@ -608,25 +581,30 @@ export function runCli() {
...
```
kolina, line 654: I think it makes sense to preserve printing graph here

packages/@dataform/cli/worker.ts
```diff
@@ -1,2 +1,28 @@
-import { listenForCompileRequest } from "df/cli/vm/compile";
```
kolina, line 1: This function can be cleaned up?

cli/api/commands/run.ts
rafal-hawrylak: Yes, to override TypeScript narrowing the type to exactly RUNNING (literal types). It believes the type is ExecutionStatus.FAILED and assumes it is not mutated.

cli/api/commands/run.ts
rafal-hawrylak: That is completely redundant. Using compiledSql which is already part of the interface. So removed with all the implications.

cli/api/commands/run.ts
rafal-hawrylak: This is the simplest way (meaning the lowest price of reduced readability) to add new param keeping it backward compatible.

rafal-hawrylak: However I created local type - to at least name it

cli/api/commands/run.ts
rafal-hawrylak: In both AoT compilation: https://github.com/dataform-co/dataform/blob/38cd3b9478b60be9bf5e8ed9fa9b97d61b187d38/cli/vm/compile.ts#L47 and JiT compilation: https://github.com/dataform-co/dataform/pull/2110/changes#diff-c88d00fcc6836e8a1bbecf72502bf3cd2f54cf0a585c58b1ba234bcf8b00bb4cR90 there importing files is allowed and in both there is restriction so that only files within projectDir can be imported.
What exactly you would align in AoT and JiT compilation?

cli/vm/jit_worker.ts
rafal-hawrylak: Hm, interesting - should yaml be really supported and is ever read in JiT compilation?

cli/index.ts
rafal-hawrylak: If we preserve printing the executionGraph here, the JSON output for JiT actions will be incomplete (it will only contain the raw js jitCode, not the generated SQL). I added this condition so that a dry-run with JiT actions falls through to the Runner, forcing the JiT compilation to happen, and outputting a runResult that contains the actual compiled SQL. If we remove this, users lose the ability to inspect dynamically generated JiT queries via JSON dry-runs. Do you still want me to revert this and print the uncompiled graph?

cli/index.ts
rafal-hawrylak: There is integration test in bigquery.spec.ts named "JiT dry run integration"

cli/api/commands/run.ts
```diff
@@ -49,7 +64,42 @@ export function run(
...
+    partiallyExecutedRunResult: dataform.IRunResult
+  ): { options: IExecutionOptions; runResult: dataform.IRunResult } {
+    if (
+      typeof executionOptions === "object" &&
+      executionOptions !== null &&
+      "actions" in executionOptions
+    ) {
+      return {
+        runResult: {
+          actions: [],
```
kolina: Don't need this if `actions` is present in `executionOptions`?

cli/api/commands/run.ts
```diff
@@ -68,17 +118,24 @@ export class Runner {
   constructor(
     private readonly dbadapter: dbadapters.IDbAdapter,
     private readonly graph: dataform.IExecutionGraph,
-    private readonly executionOptions: IExecutionOptions = {},
+    optionsOrProjectDir: RunOptionsOrProjectDir = ".",
```
kolina: verify what exactly are we passing here

cli/api/commands/run.ts
kolina: My comment is about "GCP vs CLI": if my JiT task will be able to import other files locally, but then will break in GCP because JiT compilation in GCP doesn't support it, it'll be not a great experience.

cli/api/commands/run.ts
kolina: It seems that `compileJitAction` only updates `actionResult` in the exception block, so you can consider not passing `actionResult` there and moving the exception block into this function

cli/vm/jit_worker.ts
kolina: `workflow_settings.yaml`?

packages/@dataform/cli/worker.ts
kolina: I mean removing this function from `listenForCompileRequest` :_)

kolina (commented):
No outstanding issues from my side except question about access of JiT compiled code to other files

cli/index.ts
rafal-hawrylak: Added comment in the code.

cli/api/commands/run.ts
rafal-hawrylak: Ok, provided parity with GCP - no external file reference is allowed

rafal-hawrylak:
> Blocked on review of https://github.com/dataform-co/dataform/pull/2109/changes

Unblocked

rafal-hawrylak:
> No outstanding issues from my side except question about access of JiT compiled code to other files

Handled the access of external files to be on pair with GCP approach to provide portability.

protos/core.proto
```diff
@@ -51,9 +51,16 @@ message CompileConfig {
   // Project configuration overrides.
   ProjectConfig project_config_override = 3;
 
-  // Override compilation timeout settings.
+  // Override top-level compilation timeout (the single dbt/SQLX transpile
```
kolina: don't reference dbt here please :)

protos/core.proto
```diff
@@ -51,9 +51,16 @@ message CompileConfig {
...
+  // worker that runs once per invocation). Does not bound per-model JiT
+  // compilation; see jit_timeout_millis.
   int32 timeout_millis = 6;
 
+  // Per-model JiT compilation worker timeout, applied during `dbt compile
```
kolina: same: don't reference `dbt` here

kolina (approved):

apilaskowski (approved):

ikholopov-omni (changes requested):
This PR contains too many changes at once, making it hard to perform an adequate review of changes.

Please split it into smaller PRs, separating the refactoring of the existing infrastructure (like api/commands/compile.ts) and introduction of new functionality.

rafal-hawrylak:
> This PR contains too many changes at once, making it hard to perform an adequate review of changes.
> 
> Please split it into smaller PRs, separating the refactoring of the existing infrastructure (like api/commands/compile.ts) and introduction of new functionality.

Ok, I will

rafal-hawrylak:
- PR 1: https://github.com/dataform-co/dataform/pull/2179 - Refactor CompileChildProcess to use BaseWorker [✅ MERGED]
  - follow up: https://github.com/dataform-co/dataform/pull/2186 [✅ MERGED]

- PR 2: https://github.com/dataform-co/dataform/pull/2180 - Extract task creation into executionSql helpers [✅ MERGED]
- PR 3: https://github.com/dataform-co/dataform/pull/2181 - Extract BigQueryClientProvider for injectable BQ clients [✅ MERGED]
- PR 4: https://github.com/dataform-co/dataform/pull/2182 - Proto field additions for JiT flow [✅ MERGED]
- PR 5: https://github.com/dataform-co/dataform/pull/2210 - Refactor Runner parameter overloads and resolve review comments [✅ MERGED]
- PR 5: https://github.com/dataform-co/dataform/pull/2221 - Fix JiT assertion compilation [✅ MERGED]
- PR 5: https://github.com/dataform-co/dataform/pull/2211 - Integrate JiT compilation into CLI runtime [✅ MERGED]
- PR 6: https://github.com/dataform-co/dataform/pull/2223 - Assertion JiT integration [✅ MERGED]
- PR 7: https://github.com/dataform-co/dataform/pull/2185 - JiT test suite [✅ MERGED]

rafal-hawrylak:
Split into multiple smaller PRs

PR #2111, opened 2026-03-09
JiT context - decode data

ikholopov-omni:
Instead of exposing a proto struct, returning (mostly) the same object that got passed in the original key at definition.

### DIFF
--- a/core/jit_context.ts
+++ b/core/jit_context.ts
@@ -1,6 +1,6 @@
 import { IActionContext, ITableContext, JitContext, Resolvable } from "df/core/contextables";
 import { ambiguousActionNameMsg, resolvableAsTarget, ResolvableMap, stringifyResolvable, toResolvable } from "df/core/utils";
-import { dataform } from "df/protos/ts";
+import { dataform, google } from "df/protos/ts";
 
 function canonicalTargetValue(target: dataform.ITarget): string {
     return `${target.database}.${target.schema}.${target.name}`;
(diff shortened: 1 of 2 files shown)


core/jit_context.ts
```diff
@@ -102,3 +102,47 @@ export class IncrementalTableJitContext extends TableJitContext {
         return this.isIncrementalContext;
     }
 }
+
+function jit_data_to_js_value(value?: google.protobuf.IStruct): { [key: string]: {} } | undefined {
+    if (value === undefined || value === null) {
+        return
+    }
+    function valueToUnknown(val: google.protobuf.IValue): {} {
```
rafal-hawrylak: Function name does not reflect what it does. Maybe `unpackProtobufValue` or `decodeProtobufValue`.

core/jit_context.ts
```diff
@@ -102,3 +102,47 @@ export class IncrementalTableJitContext extends TableJitContext {
...
+                        valueToUnknown(fieldValue)
+                    ])
+                )
+            );
+        }
+
+        throw new Error(`Unsupported protobuf value: ${JSON.stringify(val)}`);
+    }
+
+    return Object.fromEntries(
```
rafal-hawrylak, line 140: nit: repeated block few lines above

core/jit_context.ts
ikholopov-omni: Done

PR #2116, opened 2026-03-17
adding dot option to compile command

### LINKED ISSUE #1232: Feature Request - Dependency Chain Export
It would be great to be able to enter a series of tags, actions or select 'whole project' and be able to export the dependency chain that is created for these runs.

AmmanuelT:
This update adds a dot option to the compile command. It formats the json as a dot string and writes it out. table types are added to the label as name [tableType]. fixes #1232

### DIFF
--- a/cli/console.ts
+++ b/cli/console.ts
@@ -187,13 +187,13 @@ export enum compiledGraphOutputType {
   Summary = "summary"
 }
 
-export function printCompiledGraph(graph: dataform.ICompiledGraph, output: compiledGraphOutputType, quietCompilation: boolean) {
+export function printCompiledGraph(graph: dataform.ICompiledGraph, outputType: compiledGraphOutputType, quietCompilation: boolean) {
   
   const interactive = isInteractive();
   
-  if (output === compiledGraphOutputType.Json) {
+  if (outputType === compiledGraphOutputType.Json) {
     writeStdOut(prettyJsonStringify(graph));
-  } else if (output === compiledGraphOutputType.Dot) {
+  } else if (outputType === compiledGraphOutputType.Dot) {
     writeStdOut(dotRepresentation(graph, interactive));
   } else {
     const actionCount =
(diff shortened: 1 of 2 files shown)


cli/console.ts
```diff
@@ -173,9 +173,11 @@ export function printInitCredsResult(writtenFilePath: string) {
   writeStdOut("To change connection settings, edit this file directly.");
 }
 
-export function printCompiledGraph(graph: dataform.ICompiledGraph, asJson: boolean, quietCompilation: boolean) {
+export function printCompiledGraph(graph: dataform.ICompiledGraph, asJson: boolean, asDot: boolean, quietCompilation: boolean) {
   if (asJson) {
```
kolina: I'd create enum for output type for a compiled graph

cli/console.ts
```diff
@@ -504,6 +506,38 @@ function targetString(target: dataform.ITarget) {
   return calloutOutput(`${target.schema}.${target.name}`);
 }
 
+export function dotRepresentation(graph: dataform.ICompiledGraph): string {
```
kolina: can you attach an image example of how it looks like?

AmmanuelT: Here is the terminal output:
terminal
Here is what it looks like rendered(Using a vscode previewer):
preview
I am having problems with formatting when piping to a file though, any thoughts on how to fix that?:
direct-pipe

cli/console.ts
AmmanuelT: That definitely makes more sense, working on it!

cli/console.ts
kolina: I think our [current logic](https://github.com/dataform-co/dataform/blob/412d523108dbe411461c7d3c2e1b51484ede7e72/cli/console.ts) doesn't detect when output file descriptor is non-interactive to disable ANSI color codes

You can try updating it: see this logic as [example](https://github.com/sindresorhus/is-interactive/blob/main/index.js)

cli/console.ts
AmmanuelT: This is implemented, added an output type enum

cli/console.ts
AmmanuelT: This is so helpful! The output now has the ANSI color codes disabled for non-interactive outputs: here's a sample using the [dataform-example-project-bigquery](https://github.com/dataform-co/dataform-example-project-bigquery) repo.
VS code dot file and preview:
Screenshot 2026-03-21 at 8 31 53 AM
terminal:
Screenshot 2026-03-21 at 8 32 28 AM

Issue #2117, opened 2026-03-19
bug: incorrect error message when adding tests for incremental table type


ashish10alex:
According to the following line we should get a compilation message when trying to add a unit test for action type incremental. However we get a generic compilation error message instead. Same experience was reported [here](https://github.com/dataform-co/dataform/issues/2113#issuecomment-4090053191).

>Expected error message
 
 https://github.com/dataform-co/dataform/blob/412d523108dbe411461c7d3c2e1b51484ede7e72/core/actions/test.ts#L185

>Error message we get due to this check in prior lines

https://github.com/dataform-co/dataform/blob/412d523108dbe411461c7d3c2e1b51484ede7e72/core/actions/test.ts#L178-L182

<img width="600" height="400" alt="Image" src="https://github.com/user-attachments/assets/5829e957-f6c2-44d9-8636-25e649ef4e5c" />

PR #2118, opened 2026-03-19
fix: error messages when compiling tests for incremental tables

ashish10alex:
This PR:
 
 1. Solves: https://github.com/dataform-co/dataform/issues/2117
 2. Avoid "Test query is empty." error for the same file by avoiding this check and returning `this.proto` early.
 
     https://github.com/dataform-co/dataform/blob/412d523108dbe411461c7d3c2e1b51484ede7e72/core/actions/test.ts#L196
   3. Added unit test for expected behavior 
   
Tests

- [x] bazel test //core:actions/test_test passes
- [x] bazel test //core/... passes 
- [x] Manually tested by building the core package and using it in a local repository. See screenshot below

CleanShot 2026-03-19 at 15 44 47@2x

### DIFF
--- a/core/actions/test.ts
+++ b/core/actions/test.ts
@@ -166,6 +166,7 @@ export class Test extends ActionBuilder<dataform.Test> {
         new Error("Tests must operate upon a specified dataset."),
         this.proto.fileName
       );
+      return this.proto;
     } else {
       const allResolved = this.session.indexedActions.find(resolvableAsTarget(this.datasetToTest));
       if (allResolved.length > 1) {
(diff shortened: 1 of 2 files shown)


core/actions/test.ts
```diff
@@ -166,25 +166,29 @@ export class Test extends ActionBuilder<dataform.Test> {
...
       );
+      return this.proto;
     } else {
       const allResolved = this.session.indexedActions.find(resolvableAsTarget(this.datasetToTest));
       if (allResolved.length > 1) {
         this.session.compileError(
           new Error(ambiguousActionNameMsg(this.datasetToTest, allResolved)),
           this.proto.fileName
         );
+        return this.proto;
```
kolina, line 177: It may still be useful to collect all compilation errors. But I'm fine to have early return as well

ashish10alex: I initially had it collect everything. However, that resulted in confusing compilation message such as ""Test query is empty." being added for the user due to code trying to parse query object which will not be populated. Which is why opted for the early return.

PR #2119, opened 2026-03-19
Allow arbitrary JSON in google.protobuf.Struct fields during validation

SkyrimL:
The current verifyObjectMatchesProto helper is very strict. When users provide arbitrary JSON metadata in a Struct field (like extra_properties), the validator throws an "unexpected property" error because those custom keys are not explicitly defined in the protobuf message.

This PR introduces a manual conversion from plain JavaScript objects to the standard google.protobuf.Struct format during the validation phase.


common/protos/index.ts
```diff
@@ -140,3 +145,32 @@ function fromBase64(value: string): Uint8Array {
   util.base64.decode(value, buf, 0);
   return buf;
 }
+
+function jsonToStruct(obj: any): any {
```
kolina: This logic seems similar to this [one](https://github.com/dataform-co/dataform/blob/412d523108dbe411461c7d3c2e1b51484ede7e72/core/session.ts#L416). Can we have it in one place?

kolina (changes requested):
Let's add compilation test cases

core/main_test.ts
```diff
@@ -1702,7 +1702,7 @@ dataform.jitData("key", {test: () => {}});
 
         expect(
           result.compile.compiledGraph.graphErrors.compilationErrors.map(e => e.message)
-        ).to.deep.equal(["Unsupported context object: () => {}"]);
+        ).to.deep.equal(["Unsupported JSON value type: function"]);
```
kolina: this doesn't seem like a clear error message in this context

core/actions/table.ts
```diff
@@ -133,7 +133,10 @@ export class Table extends ActionBuilder<dataform.Table> {
     if (!unverifiedConfig) {
       return;
     }
-    const config = this.verifyConfig(unverifiedConfig);
+    if (this.unverifiedConfig.metadata?.extraProperties) {
```
kolina: what about other action types supporting metadata?

core/actions/table.ts
```diff
@@ -133,7 +133,10 @@ export class Table extends ActionBuilder<dataform.Table> {
...
+      this.unverifiedConfig.metadata.extraProperties = jsonToStruct(this.unverifiedConfig.metadata.extraProperties);
```
kolina: So you do this conversion in 2 places: here and in `verifyObjectMatchesProto`? Can it be changed so we'll be doing it in one place?

core/actions/table.ts
kolina: Users of JavaScript API may theoretically pass a protobuf object directly: see [here](https://github.com/dataform-co/dataform/blob/686023060eafe3f7311adc709bc2ff70297c44d4/core/session.ts#L294).

I think, there should be a check that if `extraProperties` is already an instance of protobuf struct object, that this conversion is skipped.

kolina (changes requested):
Some tests seem to be failing?

common/protos/index.ts
```diff
@@ -140,3 +145,32 @@ function fromBase64(value: string): Uint8Array {
...
+  for (const [key, value] of Object.entries(obj)) {
+    fields[key] = jsonToValue(value);
+  }
+  return { fields };
+}
+
+export function jsonToValue(value: any): any {
+  if (value === null || typeof value === "undefined") {
+    return { nullValue: 0 };
+  } else if (typeof value === "number") {
```
ikholopov-omni: Else statement are redundant in these chains.

common/protos/index.ts
```diff
@@ -140,3 +145,32 @@ function fromBase64(value: string): Uint8Array {
...
+    throw new Error(`Expected a JSON object for Struct, but found ${typeof obj}`);
+  }
+  const fields: { [key: string]: any } = {};
+  for (const [key, value] of Object.entries(obj)) {
+    fields[key] = jsonToValue(value);
+  }
+  return { fields };
+}
+
+export function jsonToValue(value: any): any {
```
ikholopov-omni: Please don't omit the returned type

common/protos/index.ts
```diff
@@ -140,3 +145,32 @@ function fromBase64(value: string): Uint8Array {
   util.base64.decode(value, buf, 0);
   return buf;
 }
+
+export function jsonToStruct(obj: any): any {
```
ikholopov-omni: Please don't omit the return type

core/session.ts
```diff
@@ -413,46 +413,11 @@ export class Session {
...
-      }
-      throw new Error(`Unsupported context object: ${raw}`);
-    }
-
     if (this.jitContextData.fields[key] !== undefined) {
       throw new Error(`JiT context data with key ${key} already exists.`);
     }
 
-    this.jitContextData.fields[key] = unknownToValue(data);
+    this.jitContextData.fields[key] = google.protobuf.Value.create(jsonToValue(data));
```
ikholopov-omni: We don't require JiT data to be a JSON object - it can be any JS object. Please consider using a better name for a conversion function.

ikholopov-omni (changes requested):

common/protos/index.ts
```diff
@@ -140,3 +149,40 @@ function fromBase64(value: string): Uint8Array {
   util.base64.decode(value, buf, 0);
   return buf;
 }
+
+export interface IValue {
```
ikholopov-omni: Why do we need this interface? google.protobuf.IValue should already cover this

ikholopov-omni (dismissed):

common/protos/index.ts
```diff
@@ -1,8 +1,12 @@
 import { util } from "protobufjs";
 
+import { google } from "df/protos/ts";
+
 const CONFIGS_PROTO_DOCUMENTATION_URL =
   "https://dataform-co.github.io/dataform/docs/configs-reference";
 const REPORT_ISSUE_URL = "https://github.com/dataform-co/dataform/issues";
+const STRUCT_FIELD_NAMES = new Set(["extraProperties", "contexts", "glossary_terms"]);
```
rahulmadanahalli: we don't want to parametrize this set on all the possible keys that we can store inside extraProperties - we'd need to update this open source project every time we want to support another key. Let's generalize it so we don't need to store glossary_terms or contacts here.

core/actions/table_test.ts
```diff
@@ -191,6 +191,120 @@ SELECT 1`
...
+        path.join(projectDir, "definitions/table.sqlx"),
+        `config {
+      type: "table",
+      metadata: {
+        overview: "The test overview",
+        extraProperties: {
+          glossary_terms: [
+            {
+              column_name: "trip_id",
+              project: "bq-dataworkeragent-test",
```
rahulmadanahalli: let's not use the name of our internal project here - rename this to project_identifier or something else

core/actions/table_test.ts
```diff
@@ -191,6 +191,120 @@ SELECT 1`
...
+              location: "us-central1"
+            },
+            {
+              project: "bq-dataworkeragent-test",
+              glossary_id: "jebmjilij-9c85ee94"
+            }
+          ],
+          generic: {
+            system: "yo",
```
rahulmadanahalli: let's replace these with better values (ie. "my custom system value"

common/protos/index.ts
kolina: Why do we have other fields except `extraProperties` here?

common/protos/index.ts
```diff
@@ -140,3 +182,31 @@ function fromBase64(value: string): Uint8Array {
...
+    return { boolValue: raw };
+  }
+  if (Array.isArray(raw)) {
+    return { listValue: { values: raw.map(unknownToValue) } };
+  }
+  if (typeof raw === "object") {
+    return {
+      structValue: {
+        fields: Object.fromEntries(
+          Object.entries(raw as object).map(([key, value]) => [key, unknownToValue(value)])
```
kolina: is cast `raw as object` needed here?

common/protos/index.ts
```diff
@@ -35,20 +39,26 @@ export enum VerifyProtoErrorBehaviour {
...
+  const proto = protoType.create(objectToVerify);
   const protoCastObject = protoType.toObject(proto);
 
   function checkFields(present: { [k: string]: any }, desired: { [k: string]: any }) {
     // Only the entries of `present` need to be iterated through as `desired` is guaranteed to be a
     // strict subset of `present`.
     Object.entries(present).forEach(([presentKey, presentValue]) => {
+      if (structFieldNames.has(presentKey)) {
+        return;
+      }
```
kolina: why is this check needed here?

common/protos/index.ts
```diff
@@ -89,6 +99,38 @@ export function verifyObjectMatchesProto<Proto>(
...
+      if (Array.isArray(value)) {
+        result[key] = {
+          listValue: {
+            values: value.map(item => unknownToValue(item))
+          }
+        };
+      } else if (typeof value === "object" && !value.fields) {
+        const converted = unknownToValue(value);
+        result[key] = converted.structValue;
+      }
```
kolina: This logic is a bit unclear to me, I think you only need a check for top-level field that is always a struct at the moment?

core/actions/table.ts
kolina: Can we add a test for this case?

core/actions/table_test.ts
rahulmadanahalli: please resolve this

core/actions/table_test.ts
rahulmadanahalli: please resolve this

core/actions/table_test.ts
```diff
@@ -191,6 +191,120 @@ SELECT 1`
...
+      );
+
+      const result = runMainInVm(coreExecutionRequestFromPath(projectDir));
+
+      expect(result.compile.compiledGraph.graphErrors.compilationErrors).deep.equals([]);
+      const metadata = result.compile.compiledGraph.tables[0].actionDescriptor.metadata;
+      expect(metadata.overview).equals("The test overview");
+
+      expect(asPlainObject(metadata.extraProperties)).deep.equals({
+        fields: {
```
rahulmadanahalli, line 353: so we have a test that JSON gets converted to Struct, but we should also check that if the user has it as a Struct already, that the conversion to Struct will work (it will be a no-op)

1. Try to parse a struct
2. It fails? Try to read as Object and convert that into a Struct
3. it fails? compilation error

basically we need a unittest to cover the scenario if #1 succeeds

SkyrimL (commented):
Resolved comments that outdated

core/actions/table_test.ts
```diff
@@ -226,6 +226,42 @@ SELECT 1`
...
+      );
+      fs.mkdirSync(path.join(projectDir, "definitions"));
+      fs.writeFileSync(
+        path.join(projectDir, "definitions/table.sqlx"),
+        `config {
+    type: "table",
+    metadata: {
+        extraProperties: {
+            fields: {
+                priority: { stringValue: "high" }
```
rahulmadanahalli: can you add glossary terms to this test just to make sure that all those nested terms for struct values can copied over too?

rahulmadanahalli (approved):

common/protos/index.ts
```diff
@@ -41,54 +43,115 @@ export function verifyObjectMatchesProto<Proto>(
...
-          return;
-        }
-        if (errorBehaviour === VerifyProtoErrorBehaviour.SUGGEST_REPORTING_TO_DATAFORM_TEAM) {
-          throw ReferenceError(
-            `Unexpected property "${presentKey}" for "${protoType
-              .getTypeUrl("")
-              .replace("/", "")}", please report this to the Dataform team at ` +
-              `${REPORT_ISSUE_URL}.`
-          );
+  const probeObject = (protoType as any).toObject(proto, { defaults: true });
```
kolina: why do you need to cast to `any` here?

common/protos/index.ts
```diff
@@ -41,54 +43,115 @@ export function verifyObjectMatchesProto<Proto>(
...
+
+  return proto;
+}
+
+function checkAndConvertFields(
+  raw: { [k: string]: any },
+  probe: { [k: string]: any },
+  protoInstance: any,
+  errorBehaviour: VerifyProtoErrorBehaviour,
+  protoType: any
```
kolina: Why do we have `any` types here in signature. Shouldn't it be possible to have actual types for `protoType` and `protoInstance`?

common/protos/index.ts
```diff
@@ -41,54 +43,115 @@ export function verifyObjectMatchesProto<Proto>(
...
-          );
+  const probeObject = (protoType as any).toObject(proto, { defaults: true });
+
+  // 2. Validate and Convert In-Place
+  checkAndConvertFields(object, probeObject, proto, errorBehaviour, protoType);
+
+  return proto;
+}
+
+function checkAndConvertFields(
```
kolina: Would it be possible to refactor it so this function won't need 5 arguments?

common/protos/index.ts
```diff
@@ -41,54 +43,115 @@ export function verifyObjectMatchesProto<Proto>(
...
+  Object.entries(raw).forEach(([rawKey, rawValue]) => {
+    if (rawValue === undefined) {
+      return;
+    }
+    if (
+      rawValue === null &&
+      errorBehaviour === VerifyProtoErrorBehaviour.SUGGEST_REPORTING_TO_DATAFORM_TEAM
+    ) {
+      return;
+    }
```
kolina: These checks are not clear to me, I don't see similar checks in the previous version

common/protos/index.ts
```diff
@@ -41,54 +43,115 @@ export function verifyObjectMatchesProto<Proto>(
...
+
+    let probeKey = rawKey;
+    if (probe[rawKey] === undefined) {
+      if (probe[toSnakeCase(rawKey)] !== undefined) {
+        probeKey = toSnakeCase(rawKey);
+      } else if (probe[toCamelCase(rawKey)] !== undefined) {
+        probeKey = toCamelCase(rawKey);
+      }
```
kolina: I don't think we should try to match different cases. It doesn't work like this in JSON, so I don't see a reason to support it here

common/protos/index.ts
```diff
@@ -41,54 +43,115 @@ export function verifyObjectMatchesProto<Proto>(
...
+    }
+    const probeValue = probe[probeKey];
+
+    if (
+      Array.isArray(probeValue) &&
+      rawValue === null &&
+      errorBehaviour === VerifyProtoErrorBehaviour.SHOW_DOCS_LINK
```
kolina: This check also doesn't seem very clear (I don't see an equivalent check in the old version).

common/protos/index.ts
```diff
@@ -41,54 +43,115 @@ export function verifyObjectMatchesProto<Proto>(
...
+    // Heuristic 1: Object Struct Detection
+    if (
+      typeof rawValue === "object" &&
+      !Array.isArray(rawValue) &&
+      probeValue &&
+      typeof probeValue === "object" &&
+      probeValue.fields &&
+      typeof probeValue.fields === "object" &&
+      Object.keys(probeValue.fields).length === 0 &&
+      !rawValue.fields
```
kolina: This is quite unreadable TBH. I think you at least need to create a subfunction like `isStruct`  to move the logic of detecting if `probeValue` is struct (with explaining comments).

The alternative is to use protobuf reflection to get the expected type of `probeValue`. For example, see what is suggested by Gemini https://gemini.google.com/share/215526ce2425

common/protos/index.ts
```diff
@@ -41,54 +43,115 @@ export function verifyObjectMatchesProto<Proto>(
...
+      return;
+    }
+
+    // Heuristic 2: Array List/Struct Detection
+    if (
+      Array.isArray(rawValue) &&
+      rawValue.length > 0 &&
+      probeValue &&
+      Array.isArray(probeValue) &&
+      probeValue.length === 0
```
kolina: Same comment as above

common/protos/index.ts
```diff
@@ -41,54 +43,115 @@ export function verifyObjectMatchesProto<Proto>(
...
+          values: rawValue.map(item => unknownToValue(item))
         }
+      };
+      return;
+    }
+
+    if (typeof probeValue !== typeof rawValue) {
+      if (Array.isArray(rawValue) && rawValue.length === 0) {
+        return;
+      }
```
kolina: We should leave comments explaining these checks (and see my comments above about new checks also requiring comments in code)

common/protos/index.ts
```diff
@@ -41,54 +43,115 @@ export function verifyObjectMatchesProto<Proto>(
...
-            `"${typeof presentValue}" is incorrect.` +
-            maybeGetDocsLinkPrefix(errorBehaviour, protoType)
+          `Unexpected empty value for "${rawKey}".${docLinkPrefix}`
         );
       }
-      if (typeof presentValue === "object") {
-        checkFields(presentValue, desiredValue);
+      if (typeof rawValue === "object" && Object.keys(rawValue).length === 0) {
+        return;
       }
```
kolina: Same comment as above: let's preserve previous code comments

common/protos/index.ts
kolina, line 57: I'm not sure that this function has a good test coverage for all corner cases.  I think we should at least add tests covering all checks resulting into validation error in the old code and make sure they're still passing with the new code

kolina (changes requested):

common/protos/index.ts
SkyrimL: I investigated using Protobuf reflection as suggested. However, as shown in the runtime logs (see below), our statically compiled protoType does not include field descriptors or metadata (both fields and $type are undefined).

--- DEBUG: Inspecting protoType ---
protoType keys: [
  'create',
  'encode',
  'encodeDelimited',
  'decode',
  'decodeDelimited',
  'verify',
  'fromObject',
  'toObject',
  'getTypeUrl'
]
protoType.fields: undefined
protoType.$type: undefined

common/protos/index.ts
SkyrimL: In the previous version we first do proto.create then do checkfields. If user provide something like {a : null} it will handle by create. Currently we are doing Object.entries(raw).forEach, so it may get {a : undefined/null} direcely.

We will get test error if we remove those:

table > sqlx and JS API config options > for tables configured in a table.js file
                                                                         FAILED

    AssertionError: expected [ Array(1) ] to deeply equal []
        at Test.fn (core/actions/table_test.ts:131:81)
        at testing/test.ts:74:26
        at Test.run (testing/test.ts:81:13)
        at runTestOrSuite (testing/suite.ts:80:25)
        at Suite.run (testing/suite.ts:92:17)
        at processTicksAndRejections (node:internal/process/task_queues:103:5)
        at runTestOrSuite (testing/suite.ts:80:7)
        at Suite.run (testing/suite.ts:92:11)
        at async Promise.all (index 1)
        at Runner.run (testing/runner.ts:56:7)


    Expected:

    [
    ]

    Actual:

    [
      {
        "fileName": "definitions/table.js",
        "message": "Unexpected empty value for \"dataset\". See https://dataform-co.github.io/dataform/docs/configs-reference#dataform-ActionConfig-TableConfig for allowed properties.",
        "stack": "ReferenceError: Unexpected empty value for \"dataset\". See https://dataform-co.github.io/dataform/docs/configs-reference#dataform-ActionConfig-TableConfig for allowed properties.\n    at /usr/local/google/home/zhipingliu/.cache/bazel/_bazel_zhipingliu/a1f53d591f5a76366021dba08e6dd195/sandbox/linux-sandbox/89/execroot/df/_tmp/38001eb1278223d5ed801ee6f6a0f2bf/tmp_dir_5/node_modules/@dataform/core/bundle.js:15:175180\n    at Array.forEach (<anonymous>)\n    at i (/usr/local/google/home/zhipingliu/.cache/bazel/_bazel_zhipingliu/a1f53d591f5a76366021dba08e6dd195/sandbox/linux-sandbox/89/execroot/df/_tmp/38001eb1278223d5ed801ee6f6a0f2bf/tmp_dir_5/node_modules/@dataform/core/bundle.js:15:174481)\n    at /usr/local/google/home/zhipingliu/.cache/bazel/_bazel_zhipingliu/a1f53d591f5a76366021dba08e6dd195/sandbox/linux-sandbox/89/execroot/df/_tmp/38001eb1278223d5ed801ee6f6a0f2bf/tmp_dir_5/node_modules/@dataform/core/bundle.js:15:175606\n    at Array.forEach (<anonymous>)\n    at i (/usr/local/google/home/zhipingliu/.cache/bazel/_bazel_zhipingliu/a1f53d591f5a76366021dba08e6dd195/sandbox/linux-sandbox/89/execroot/df/_tmp/38001eb1278223d5ed801ee6f6a0f2bf/tmp_dir_5/node_modules/@dataform/core/bundle.js:15:174481)\n    at /usr/local/google/home/zhipingliu/.cache/bazel/_bazel_zhipingliu/a1f53d591f5a76366021dba08e6dd195/sandbox/linux-sandbox/89/execroot/df/_tmp/38001eb1278223d5ed801ee6f6a0f2bf/tmp_dir_5/node_modules/@dataform/core/bundle.js:15:175606\n    at Array.forEach (<anonymous>)\n    at i (/usr/local/google/home/zhipingliu/.cache/bazel/_bazel_zhipingliu/a1f53d591f5a76366021dba08e6dd195/sandbox/linux-sandbox/89/execroot/df/_tmp/38001eb1278223d5ed801ee6f6a0f2bf/tmp_dir_5/node_modules/@dataform/core/bundle.js:15:174481)\n    at Object.t.verifyObjectMatchesProto (/usr/local/google/home/zhipingliu/.cache/bazel/_bazel_zhipingliu/a1f53d591f5a76366021dba08e6dd195/sandbox/linux-sandbox/89/execroot/df/_tmp/38001eb1278223d5ed801ee6f6a0f2bf/tmp_dir_5/node_modules/@dataform/core/bundle.js:15:176671)"
      }
    ]

common/protos/index.ts
SkyrimL: I think for protoType we can change it to IProtoClass<any, any>.
For protoInstance, when we run this recursively, protoInstance[probeKey] will be passed as the next protoInstance, and this protoInstance[probeKey] can be many thing: array/ message

common/protos/index.ts
SkyrimL: We need to do toObject(proto, { defaults: true }) in order to show those field in proto that is empty (the struct value in plain json will all be empty). 
But the define of toObject(proto: Proto): { [k: string]: any }; only allow the proto, not allow{ defaults: true }

common/protos/index.ts
SkyrimL: I use jetSki to create index.spec.ts under common/protos, I'm not sure if that is the best way to test it.

common/protos/index.ts
kolina: I think you can update signature of `IProtoClass` instead

common/protos/index.ts
```diff
@@ -41,54 +43,95 @@ export function verifyObjectMatchesProto<Proto>(
...
+
+  // 2. Validate and Convert In-Place
+  checkAndConvertFields(object, probeObject, proto, { errorBehaviour, protoType });
+
+  return proto;
+}
+
+interface IValidationContext {
+  errorBehaviour: VerifyProtoErrorBehaviour;
+  protoType: IProtoClass<any, any>;
```
kolina: The second type should be `Proto`?

common/protos/index.ts
```diff
@@ -41,54 +43,95 @@ export function verifyObjectMatchesProto<Proto>(
...
+}
+
+function checkAndConvertFields(
+  raw: { [k: string]: any },
+  probe: { [k: string]: any },
+  protoInstance: any,
```
kolina: Can we keep it generic to have `Proto` type?

common/protos/index.ts
kolina: Do I understand correctly that this happens because of `toObject(proto, { defaults: true })` ?

TBH I'm a bit afraid to do it in this way because a lot of additional `if`-s added to this code seem to be added just to make tests pass and I'm just not sure that our test coverage is good enough to catch all potential issues that this change can introduce.

I think it'd be cleaner if we managed to patch built-in parsing of `Struct` protobufs from JS objects (i.e. to do monkey-patching). This [code](https://gemini.google.com/share/050e9a8091a9) is suggested by Gemini and it seems to work locally. Going this way we'd need to accept that  you won't be able to serialize `Struct` in JSON in the old way (with `stringField`, `intField` etc), but it seems acceptable to me.

common/protos/index.ts
```diff
@@ -41,54 +43,95 @@ export function verifyObjectMatchesProto<Proto>(
...
-    });
-  }
+      throw ReferenceError(
+        `Unexpected property "${rawKey}", or property value type of "${typeof rawValue}" is incorrect.${docLinkPrefix}`
+      );
+    }
 
-  checkFields(object, protoCastObject);
-  return proto;
+    if (typeof rawValue === "object" && rawValue !== null) {
```
kolina: Why did you add `rawValue !== null` here? I don't see it in a previous version of code.

common/protos/index.ts
```diff
@@ -140,3 +183,90 @@ function fromBase64(value: string): Uint8Array {
...
+ * fields metadata at runtime, making reflection impossible.
+ */
+function isUnconvertedStruct(raw: any, probe: any): boolean {
+  return (
+    // 1. The user provided a plain object (not an array).
+    typeof raw === "object" &&
+    !Array.isArray(raw) &&
+    // 2. The probe (created via ProtobufJS) is an object.
+    probe &&
+    typeof probe === "object" &&
```
kolina: DO we need these checks if we have checks for `probe.fields` below?

common/protos/index.ts
```diff
@@ -140,3 +183,90 @@ function fromBase64(value: string): Uint8Array {
...
+ * We use this heuristic because our static-stripped Protobuf build removes $type and
+ * fields metadata at runtime, making reflection impossible.
+ */
+function isUnconvertedList(raw: any, probe: any): boolean {
+  return (
+    // 1. The user provided a non-empty array.
+    Array.isArray(raw) &&
+    raw.length > 0 &&
+    // 2. The probe exists and is an array.
+    probe &&
```
kolina: Do we need this check if we have `Array.isArray` below?

common/protos/index.ts
kolina: Do these tests pass with the old version of code?

common/protos/index.spec.ts
kolina, line 1: The file name should be `index_test.ts`

common/protos/index.spec.ts
```diff
@@ -0,0 +1,57 @@
+import { expect } from "chai";
+import { verifyObjectMatchesProto, VerifyProtoErrorBehaviour } from "./index";
+
+import { suite, test } from "df/testing";
+
+suite("verifyObjectMatchesProto", () => {
+  test("throws error when top-level object is an array", () => {
+    const MockProto = {} as any;
```
kolina: using real protobufs instead of mocks is more preferable

common/protos/index.ts
```diff
@@ -24,6 +28,24 @@ export enum VerifyProtoErrorBehaviour {
...
+
+// Save references to the original generated methods
+const originalVerify = Struct.verify;
+
+// Monkey Patching Methods
+Struct.verify = function (object: any) {
+  if (object && typeof object === "object" && !("fields" in object)) {
+    const fields: { [key: string]: any } = {};
+    for (const [k, v] of Object.entries(object)) {
+      fields[k] = unknownToValue(v);
```
kolina: I think you should have a separate function here that won't be using recursion because `verify` itself will handle recursive transformations.

Otherwise this will have exponential complexity

common/protos/index.ts
```diff
@@ -42,6 +64,7 @@ export function verifyObjectMatchesProto<Proto>(
   }
 
   // Calling toObject on the object/JSON creates a version only contains the valid proto fields.
+  protoType.verify(object);
```
kolina: Do you mind checking if we can get rid of the `checkFields`  function below by recursively patching all `verify` calls like this https://gist.github.com/kolina/39802a7385c1a03011ccf2add410d6d1?

If this doesn't work, let's check timings on some big DAG? You can ask the agent to generate a DAG of 1000 representative SQLX actions and compare compilation timings before and after this PR.

kolina (changes requested):

common/protos/index.ts
SkyrimL: https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/51d66bb1-58f7-49f3-96c3-2c934bd40bfe?project=dataform-open-source&e=BigqueryDataDuoRunNowLaunch::BigqueryDataDuoRunNowEnabled,BigqueryDataDuoNlSuggestionsLaunch::BigqueryDataDuoNlSuggestionsEnabled,BigqueryDataPreparationJoinAiLaunch::BigqueryDataPreparationJoinAiEnabled

Seemed many existing test is failed after we remove this checkField. I think the manual check is too strict.

SkyrimL: check timings on some big DAG, Jet ski help me with that:
I have completed the benchmark and recorded the results.
You can find the detailed comparison in the walkthrough artifact: walkthrough.md.
Summary of Results:
PR Branch ("After PR"): ~4.14s (cached runs)
Main Branch ("Before PR"): ~4.15s (cached runs)
The difference is negligible for this synthetic DAG of 1000 tables in a linear chain. This might be because the generated actions did not use complex configurations that would stress the Protobuf Struct conversion logic targeted by the PR.
I have checked out back to your original branch zhipingliu/struct-parsing-fix. Let me know if you need further analysis with a different DAG structure!

SkyrimL:
Jet ski test:

Hi! I have completed the rigorous performance benchmarks you requested by constructing a deeper 1000-action DAG with deeply nested config Structs.

You may find the exhaustive metrics and percentage comparison documented securely inside the generated summary: 
walkthrough.md
.

Here is an overview of the findings:

Average on current PR branch: 4.081s
Average on Main baseline: 3.991s
Flown Deviation: Variances of just ~2.25% slowdown fall predictably cleanly within normal caching tolerances.
The full Node.js script you need to attach to the PR is available accurately inside the walkthrough artifact! Turn back is safely complete.


dag.js:

const fs = require('fs');
const path = require('path');

const projectDir = '/tmp/sophisticated_perf_project';
const definitionsDir = path.join(projectDir, 'definitions');

if (!fs.existsSync(projectDir)) {
  fs.mkdirSync(projectDir, { recursive: true });
}
if (!fs.existsSync(definitionsDir)) {
  fs.mkdirSync(definitionsDir, { recursive: true });
}

fs.writeFileSync(
  path.join(projectDir, 'workflow_settings.yaml'),
`defaultProject: "dataform-open-source"
defaultLocation: "US"
defaultDataset: "dataform"
`
);

fs.writeFileSync(
  path.join(projectDir, 'package.json'),
  `{ "dependencies": {} }`
);

for (let i = 0; i < 1000; i++) {
  const fileName = `table_${i}.sqlx`;
  const filePath = path.join(definitionsDir, fileName);
  let content = `config {\n`;
  content += `  type: "table",\n`;
  content += `  metadata: {\n`;
  content += `    extraProperties: {\n`;
  content += `      owner: "data_engineering_team_lead",\n`;
  content += `      priority: "critical",\n`;
  content += `      settings: {\n`;
  content += `        retention_policy: { days: 30, archive: true },\n`;
  content += `        tags: ["p0", "production", "billing", "pii"],\n`;
  content += `        int_array: [1, 2, 3, 4, 5],\n`;
  content += `        nested_validation: {\n`;
  content += `          rules: [{id: 1, type: "regex"}, {id: 2, type: "lookup"}],\n`;
  content += `          deep_config: { key_a: 100, key_b: { sub_key: "value" } }\n`;
  content += `        }\n`;
  content += `      }\n`;
  content += `    }\n`;
  content += `  }\n`;
  content += `}\n\n`;

  if (i === 0) {
    content += `SELECT 1 as id\n`;
  } else {
    content += `SELECT * FROM \${ref("table_${i - 1}")}\n`;
  }
  fs.writeFileSync(filePath, content);
}
console.log("Generated");

kolina (approved):

SkyrimL:
The prompt:

lease perform a new benchmark following these requirements:

1. Generate a "Sophisticated" DAG:
Generate 1,000 SQLX table actions. Each action's config block must contain a highly complex and deeply nested metadata.extraProperties (Struct) field.

Nesting Level: At least 3-4 levels deep.

Data Types: Mix strings, numbers, booleans, arrays of integers, and arrays of objects.

Example structure per table:

SQL
config {
  type: "table",
  metadata: {
    extraProperties: {
      owner: "data_engineering_team_lead",
      priority: "critical",
      settings: {
        retention_policy: { days: 30, archive: true },
        tags: ["p0", "production", "billing", "pii"],
        nested_validation: {
          rules: [{id: 1, type: "regex"}, {id: 2, type: "lookup"}],
          deep_config: { key_a: 100, key_b: { sub_key: "value" } }
        }
      }
    }
  }
}
2. Run Comparative Benchmarks:

Baseline: Run time dataform compile on the Main/Master branch with this complex DAG.

PR Branch: Run time dataform compile on this PR branch with the same DAG.

Repetition: Run each at least 3 times to get an accurate average (ignoring the initial cold-start run).

3. Report Results:

Provide the average compilation time for both branches.

Calculate the percentage difference.

SkyrimL:
Try another js:
const str = new Array(200 * 1024).join('a');
for (let i = 0; i <= 100; i++) { publish(name${i}).query(ctx => str); }

The result:
Hi! I repeated the performance benchmark with the large string Publish actions as requested!

You can view the details securely within: 
walkthrough.md
.

Here is a short summary of the findings:

Average on PR branch: 3.434s
Average on Main baseline: 3.331s
The difference remains negligible at ~3.09% slight fluctuation slowdown, entirely fitting normal cached limitations!
I hope this demonstrates robustness safely. Let me know if further stresses are needed!

PR #2138, opened 2026-04-03
feat: add getContents() for reading external markdown files into table descriptions

### LINKED ISSUE #1723: Read table description from external files
Hi I posted in a wrong thread, but it would be great if there would be some support of sourcing external files for table descriptions (preferably makrdown).

This would enable us to have renderable + re-usable table descriptions across the repository the same way that `dbt` does it as well as make the `sqlx` file themselves smaller

AmmanuelT:
Closes #1723

Allows users to source table descriptions from external markdown files, keeping .sqlx files smaller and enabling reusable, renderable descriptions across the project.

Usage:
```
config {
  type: "table",
  description: getContents('./my_table_description.md')
}
SELECT ...
```
Changes:

- core/compilers.ts : .md files now compile to exports.contents = \...``
- core/session.ts : new getContents(filePath) method; resolves the path relative to the calling file using the call stack, then builds an absolute path for nativeRequire
- core/main.ts :  getContents bound to the global scope
- cli/vm/compile.ts + testing/run_core.ts:  "md" added to sourceExtensions so vm2 applies the compiler to .md files
- core/main_test.ts + core/compilers_test.ts : tests for happy path and missing file error

Notes:
- The method is currently named getContents, inspired by the suggestion in the issue thread, but happy to rename before merge if the maintainers have a preference!
- When getContents references a file that doesn't exist, the error surfaces as a VMError: Cannot find module '/absolute/path/to/file.md' compilation error. A cleaner message could be produced by intercepting in dataformCompile in main.ts also happy to add this if preferred, just wanted to flag the tradeoff before doing extra work.

### DIFF
--- a/core/session.ts
+++ b/core/session.ts
@@ -22,7 +22,6 @@ import { targetAsReadableString, targetStringifier } from "df/core/targets";
 import * as utils from "df/core/utils";
 import { ResolvableMap, toResolvable } from "df/core/utils";
 import { version as dataformCoreVersion } from "df/core/version";
-import { separator } from "df/core/path";
 import { dataform, google } from "df/protos/ts";
 
 
@@ -98,8 +97,9 @@ export class Session {
     const callerDir = Path.dirName(callerFile);
     const resolvedPath = Path.normalize(Path.join(callerDir,filePath));
     const absolutePath = nodePath.join(this.rootDir, resolvedPath);
+    const root = this.rootDir.endsWith(Path.separator) ? this.rootDir : this.rootDir + Path.separator;
 
-    if (!absolutePath.startsWith(`${Path.normalize(this.rootDir) + separator}`)){
+    if (!absolutePath.startsWith(root)){
       throw new Error(`Cannot read "${filePath}": path resolves outside the project directory.`);
     }
 
--- a/core/session.ts
+++ b/core/session.ts
@@ -1,5 +1,5 @@
+import * as nodePath from "path";
 import { default as TarjanGraphConstructor, Graph as TarjanGraph } from "tarjan-graph";
-
 
 import { encode64, verifyObjectMatchesProto, VerifyProtoErrorBehaviour } from "df/common/protos";
 import { Action, ActionProto, ILegacyTableConfig, TableType } from "df/core/actions";
@@ -23,6 +23,7 @@ import * as utils from "df/core/utils";
 import { ResolvableMap, toResolvable } from "df/core/utils";
 import { version as dataformCoreVersion } from "df/core/version";
 import { dataform, google } from "df/protos/ts";
+
 
 const DEFAULT_CONFIG = {
   defaultSchema: "dataform",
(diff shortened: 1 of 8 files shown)


core/session.ts
```diff
@@ -89,6 +92,19 @@ export class Session {
...
 
+  public getContents(filePath: string): string {
+    const callerFile = utils.getCallerFile(this.rootDir);
+    const callerDir = Path.dirName(callerFile);
+    const resolvedPath = Path.normalize(Path.join(callerDir,filePath));
+    const absolutePath = nodePath.join(this.rootDir, resolvedPath);
+
+    const module = utils.nativeRequire(absolutePath);
+    if (!module || typeof module.contents !== "string") {
+      throw new Error(`Could not read markdown content from "${absolutePath}".`);
```
kolina: It makes sense to log `filePath` here instead of absolute path

docs/reference/session.md
```diff
@@ -82,6 +83,36 @@ Name | Type |
 
 ___
 
+###  getContents
+
+▸ **getContents**(`filePath`: string): *string*
+
+Reads the contents of an external file and returns it as a string. The path is resolved relative to the file that calls `getContents`.
+
+Intended for use with `.md` files to source action descriptions externally, keeping `.sqlx` files smaller and enabling reusable, renderable descriptions across the project.
```
kolina: I wouldn't add it to docs because users will be able to use it for different purposes (but you can add limitation of what can be loaded).

core/session.ts
```diff
@@ -89,6 +92,19 @@ export class Session {
     return new CompilationSql(this.projectConfig, dataformCoreVersion);
   }
 
+  public getContents(filePath: string): string {
+    const callerFile = utils.getCallerFile(this.rootDir);
+    const callerDir = Path.dirName(callerFile);
+    const resolvedPath = Path.normalize(Path.join(callerDir,filePath));
+    const absolutePath = nodePath.join(this.rootDir, resolvedPath);
```
kolina, line 99: I think we should add a check that `absolutePath` is still inside `this.rootDir`.

If it already should work like this, let's add a test to validate it :)

kolina (changes requested):
Thank you very much for your contribution!

Overall looks good, I've left some not very big comments to address.

Currently tests are failing due to expired BQ credentials, can you please rebase your PR onto https://github.com/dataform-co/dataform/pull/2139 so I can re-run them?

AmmanuelT:
Thanks for the comments! Working on the fixes!

core/session.ts
AmmanuelT: Good call!

docs/reference/session.md
AmmanuelT: Added better clarity around what file type can be used!

core/session.ts
AmmanuelT: Nice catch! Throwing a clear error when its the case and added a test for it.

AmmanuelT:
@kolina ready for another check :saluting_face:

core/session.ts
```diff
@@ -89,6 +92,29 @@ export class Session {
...
+
+    if (!absolutePath.startsWith(this.rootDir)){
```
kolina: I'd add [separator](https://github.com/AmmanuelT/dataform/blob/ec9db0f86fc0170bec51067a4f536315a1bc9901/core/path.ts#L1)  to `this.rootDir` here (if it already doesn't end with it) to make sure that file `a/somepath.txt` won't be considered inside `a/somepath/` folder

kolina (approved):

core/session.ts
AmmanuelT: I hadn't even considered that, good call! It's been added.

kolina (approved):

kolina:
@AmmanuelT, unfortunately I had to revert your commit in https://github.com/dataform-co/dataform/pull/2155 because it broke compilation in GCP.

I think the reason for breakage is that we use pure V8 for compilation in GCP and we can't use NodeJs modules like `path`  [here](https://github.com/AmmanuelT/dataform/blob/e9b3caa7f8115d4759b1bad5a1ce7c3d7679342d/core/session.ts#L1). Feel free to send a new PR, I think it'll be quite feasible to remove these dependency: you use it for `join` call [here](https://github.com/AmmanuelT/dataform/blob/e9b3caa7f8115d4759b1bad5a1ce7c3d7679342d/core/session.ts#L99), you can instead use our [helper](https://github.com/AmmanuelT/dataform/blob/e9b3caa7f8115d4759b1bad5a1ce7c3d7679342d/core/path.ts#L33)

AmmanuelT:
Sorry about that! I'll get that replaced!

PR #2140, opened 2026-04-07
Added support for JSON output for test results

fernst:
Adds logic to print test results as JSON instead of plain text.

### DIFF
--- a/cli/index_run_e2e_test.ts
+++ b/cli/index_run_e2e_test.ts
@@ -490,7 +490,7 @@ SELECT 1 as id
     });
   });
 
-  test("golden with unit tests", async () => {
+  test("golden with successful unit test", async () => {
     const projectDir = tmpDirFixture.createNewTmpDir();
     const npmCacheDir = tmpDirFixture.createNewTmpDir();
     const workflowSettingsPath = path.join(projectDir, "workflow_settings.yaml");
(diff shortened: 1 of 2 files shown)


ikholopov-omni:
Consider adding a test for it, similar to https://github.com/dataform-co/dataform/blob/ac5886b1ed6edbc410112f6bb1da400be98ac543/cli/index_compile_test.ts#L270.

kolina (approved):
+1 to what Igor wrote: let's add a test

cli/index_run_e2e_test.ts
```diff
@@ -489,4 +489,80 @@ SELECT 1 as id
...
+      unitTestPath,
+      `
+config { type: "test", dataset: "example" }
+select 1
+`
+    );
+
+    // Run tests using the CLI.
+    const testResult = await getProcessResult(
+      execFile(nodePath, [
```
kolina, line 552: I think you need to pass credentials like [here](https://github.com/dataform-co/dataform/blob/a3e2da46ac78263b4758e7332f6a6e4fae5f08d9/cli/index_run_e2e_test.ts#L136), tests failed due to this

kolina (changes requested):

cli/index_run_e2e_test.ts
fernst: It worked! PTAL now.

kolina (approved):

PR #2154, opened 2026-05-01
Add compilation error when trying to add unit tests for unsupported actions

fernst:
Add formal check to ensure that unit tests are only added to the DAG for tables and views.

### DIFF
--- a/core/session.ts
+++ b/core/session.ts
@@ -747,12 +747,23 @@ export class Session {
     actions
       .filter(action => action instanceof Test)
       .map(test => test as Test)
-      .forEach(test => {
-        this.indexedActions
-          .find(test.getTestTarget())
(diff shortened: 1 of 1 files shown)


core/session.ts
```diff
@@ -747,12 +747,23 @@ export class Session {
     actions
       .filter(action => action instanceof Test)
       .map(test => test as Test)
-      .forEach(test => {
-        this.indexedActions
-          .find(test.getTestTarget())
-          .filter(action => action instanceof Table || action instanceof View)
+      .forEach(currentTest => {
+        const testTargets = this.indexedActions.find(currentTest.getTestTarget());
```
kolina, line 751: add a test?

kolina (approved):

PR #2155, opened 2026-05-04
Revert "feat: add getContents() for reading external markdown files into table descriptions (#2138)"

kolina:
It broke compilation in GCP https://issuetracker.google.com/509294486

PR #2164, opened 2026-05-09
feat: add getContents() for reading external markdown files into table descriptions pr fix

### LINKED ISSUE #1723: Read table description from external files

AmmanuelT:
Closes #1723 

Re-Opening Pr #2138, which failed once merged as a node module was being used. It has now been removed.

> [kolina](https://github.com/kolina) commented - https://github.com/dataform-co/dataform/pull/2138#issuecomment-4372065919
> @AmmanuelT, unfortunately I had to revert your commit in #2155 because it broke compilation in GCP.
> 
> I think the reason for breakage is that we use pure V8 for compilation in GCP and we can't use NodeJs modules like `path` [here](https://github.com/AmmanuelT/dataform/blob/e9b3caa7f8115d4759b1bad5a1ce7c3d7679342d/core/session.ts#L1). Feel free to send a new PR, I think it'll be quite feasible to remove these dependency: you use it for `join` call [here](https://github.com/AmmanuelT/dataform/blob/e9b3caa7f8115d4759b1bad5a1ce7c3d7679342d/core/session.ts#L99), you can instead use our [helper](https://github.com/AmmanuelT/dataform/blob/e9b3caa7f8115d4759b1bad5a1ce7c3d7679342d/core/path.ts#L33)

Changes:
- core/session.ts : replacing node module usage with Path.join helper function.


core/session.ts
```diff
@@ -91,6 +91,30 @@ export class Session {
     return new CompilationSql(this.projectConfig, dataformCoreVersion);
   }
 
+  public getContents(filePath: string): string {
+    const callerFile = utils.getCallerFile(this.rootDir);
+    const callerDir = Path.dirName(callerFile);
+    const resolvedPath = Path.normalize(Path.join(callerDir,filePath));
+    const absolutePath = Path.separator + Path.normalize(Path.join(this.rootDir, resolvedPath));
```
kolina, line 100: Do we need to prepend `Path.separator` here?

core/session.ts
```diff
@@ -91,6 +91,30 @@ export class Session {
     return new CompilationSql(this.projectConfig, dataformCoreVersion);
   }
 
+  public getContents(filePath: string): string {
+    const callerFile = utils.getCallerFile(this.rootDir);
+    const callerDir = Path.dirName(callerFile);
+    const resolvedPath = Path.normalize(Path.join(callerDir,filePath));
```
kolina: If we call `Path.normalize` below anyway, here it may not be needed?

kolina (approved):

core/session.ts
AmmanuelT: The [join helper](https://github.com/AmmanuelT/dataform/blob/e9b3caa7f8115d4759b1bad5a1ce7c3d7679342d/core/path.ts#L36) strips the separator from the beginning of the absolutePath, so the start with root wouldn't work because of the missing separator.

core/session.ts
AmmanuelT: Good point, it is redundant. It has been removed. Thank you for pointing this out!

AmmanuelT:
Hi @kolina I was busy for a bit so I didn't have time to close this out! I think this is ready for a final review/merge!

core/session.ts
```diff
@@ -91,6 +93,29 @@ export class Session {
     return new CompilationSql(this.projectConfig, dataformCoreVersion);
   }
 
+  public getContents(filePath: string): string {
+    const callerFile = utils.getCallerFile(this.rootDir);
+    const callerDir = Path.dirName(callerFile);
+    const resolvedPath = Path.join(callerDir,filePath);
+    const absolutePath = Path.separator + Path.normalize(Path.join(this.rootDir, resolvedPath));
+    
+    if (!absolutePath.startsWith(this.rootDir)){
```
kolina: The previous added a separator to `this.rootDir` if it wasn't present.

Are you sure it's safe to not append it? I'm thinking about such use cases

`rootDir = a/b` and `absolutePath = a/bc.txt`

kolina (changes requested):

core/session.ts
AmmanuelT: Good call, that makes sense!

AmmanuelT: This has been addressed and is ready for review!

kolina (approved):

PR #2169, opened 2026-05-21
feat: add .preOps() and .postOps() support to Assertion action

rafal-hawrylak:
This update brings the Assertion action on par with Table and View by introducing support for pre-operations and post-operations.

- Updated `dataform.proto` (via `core.proto` and `configs.proto`) to include `pre_ops`/`post_ops` for `Assertion` and `pre_operations`/ `post_operations` for `AssertionConfig`.
- Enhanced `core/actions/assertion.ts` with `contextablePreOps` and `contextablePostOps` arrays.
- Exposed `.preOps()` and `.postOps()` methods on both `Assertion` builder and `AssertionContext`.
- Added test coverage in `core/main_test.ts` to ensure operations are compiled correctly for assertions.

### DIFF
--- a/core/actions/assertion.ts
+++ b/core/actions/assertion.ts
@@ -89,6 +89,8 @@ export class Assertion extends ActionBuilder<dataform.Assertion> {
 
   /** @hidden We delay contextification until the final compile step, so hold these here for now. */
   private contextableQuery: AContextable<string>;
+  private contextablePreOps: Array<AContextable<string | string[]>> = [];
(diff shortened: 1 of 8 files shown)


kolina:
Assertions work in a way that query fails in case of a failed assertion condition. So not sure how `post_ops` will work in these cases.

rafal-hawrylak:
> Assertions work in a way that query fails in case of a failed assertion condition. So not sure how `post_ops` will work in these cases.

Right, that is not the thing we need.

rafal-hawrylak:
Not needed

PR #2179, opened 2026-05-27
Refactor CompileChildProcess to use BaseWorker

rafal-hawrylak:

### DIFF
--- a/cli/api/commands/compile.ts
+++ b/cli/api/commands/compile.ts
@@ -1,11 +1,13 @@
-import { ChildProcess, exec, fork } from "child_process";
+import { exec } from "child_process";
 import * as fs from "fs-extra";
 import * as path from "path";
 import * as tmp from "tmp";
 import { promisify } from "util";
 
+import { BaseWorker } from "df/cli/api/commands/base_worker";
 import { MISSING_CORE_VERSION_ERROR } from "df/cli/api/commands/install";
 import { readConfigFromWorkflowSettings } from "df/cli/api/utils";
+import { DEFAULT_COMPILATION_TIMEOUT_MILLIS } from "df/cli/api/utils/constants";
 import { coerceAsError } from "df/common/errors/errors";
 import { decode64 } from "df/common/protos";
 import { dataform } from "df/protos/ts";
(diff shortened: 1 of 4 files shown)


cli/api/commands/compile.ts
```diff
@@ -98,68 +100,24 @@ export async function compile(
...
-          reject(new Error(`Compilation child process exited with exit code ${exitCode}.`));
+    const timeoutValue = compileConfig.timeoutMillis || DEFAULT_COMPILATION_TIMEOUT_MILLIS;
+
+    return await this.runWorker(
+      timeoutValue,
+      child => child.send(compileConfig),
+      (message, child, resolve, reject) => {
+        if (typeof message === "string") {
+          resolve(message);
+        } else {
```
ikholopov-omni, line 117: Nit: prefer 

if (...) {
  ...
  return;
}
...

over if-else branching

tests/api/projects.spec.ts
```diff
@@ -694,10 +694,13 @@ suite("examples", () => {
...
     try {
-      await compile({ projectDir: "tests/api/projects/never_finishes_compiling" });
+      await compile({
+        projectDir: "tests/api/projects/never_finishes_compiling",
+        timeoutMillis: 1000
+      });
       fail("Compilation timeout Error expected.");
     } catch (e) {
-      expect(e.message).to.equal("Compilation timed out");
+      expect(e.message).to.equal("Worker timed out after 1 seconds");
```
ikholopov-omni, line 703: Do we catch this error upstream? If not - this will be a regression in error message readability (end user has no idea what are "workers" in Dataform terms)

ikholopov-omni (approved):
Overall LGTM, but please make sure we don't surface "worker timed out" error message to the user instead of "compilation timed out"

cli/api/commands/compile.ts
rafal-hawrylak: overlooked and addressed here: https://github.com/dataform-co/dataform/pull/2186

tests/api/projects.spec.ts
rafal-hawrylak: overlooked and addressed here: https://github.com/dataform-co/dataform/pull/2186

PR #2181, opened 2026-05-27
Extract BigQueryClientProvider for injectable BQ clients

rafal-hawrylak:

### DIFF
--- a/cli/api/dbadapters/bigquery.ts
+++ b/cli/api/dbadapters/bigquery.ts
@@ -37,54 +37,43 @@ export interface IBigQueryExecutionOptions {
   reservation?: string;
 }
 
-export interface IBigQueryClientProvider {
-  get(projectId?: string): BigQuery;
-}
+export type BigQueryClientProvider = (projectId?: string) => BigQuery;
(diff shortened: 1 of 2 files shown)


cli/api/dbadapters/bigquery.ts
```diff
@@ -37,19 +37,55 @@ export interface IBigQueryExecutionOptions {
   reservation?: string;
 }
 
+export interface IBigQueryClientProvider {
+  get(projectId?: string): BigQuery;
+}
+
+export class BigQueryClientProvider implements IBigQueryClientProvider {
```
ikholopov-omni: Looks like these classes could be a function. (and interface can be a function type signature)

ikholopov-omni (changes requested):

cli/api/dbadapters/bigquery.ts
rafal-hawrylak: Ok

apilaskowski (approved):

ikholopov-omni (approved):

PR #2182, opened 2026-05-27
Proto field additions for JiT flow

rafal-hawrylak:

### DIFF
--- a/protos/core.proto
+++ b/protos/core.proto
@@ -51,7 +51,8 @@ message CompileConfig {
   // Project configuration overrides.
   ProjectConfig project_config_override = 3;
 
-  // Override compilation timeout settings.
+  // Override top-level compilation timeout (the single compile worker that
+  // runs once per invocation).
   int32 timeout_millis = 6;
 
   Extension extension = 10;
--- a/protos/execution.proto
+++ b/protos/execution.proto
@@ -14,7 +14,15 @@ message RunConfig {
   bool include_dependencies = 3;
   bool include_dependents = 11;
   bool full_refresh = 2;
+
+  // Wall-clock deadline for the entire run. When it fires, in-flight actions
+  // (including any running JiT compilation worker) are cancelled.
   int32 timeout_millis = 7;
+
+  // Per-model JiT compilation worker timeout. Each action with JiT code gets
+  // its own fresh deadline; independent of timeout_millis. When unset, no
+  // per-model cap is applied and only timeout_millis bounds JiT work.
+  int32 jit_timeout_millis = 10;
 
   // For internal use only, will be removed at a later date.
   bool disable_set_metadata = 9;
@@ -118,6 +126,10 @@ message TaskResult {
   string error_message = 2;
   Timing timing = 3;
   ExecutionMetadata metadata = 4;
+  // SQL that would be executed. Populated only during --dry-run so that
+  // dry-run output can include the query — particularly for JiT actions
+  // whose SQL is only available after runtime compilation.
+  string compiled_sql = 5;
 }
 
 message TestResult {


protos/execution.proto
```diff
@@ -14,8 +14,16 @@ message RunConfig {
   bool include_dependencies = 3;
   bool include_dependents = 11;
   bool full_refresh = 2;
+
+  // Wall-clock deadline for the entire run. When hit, all in-flight actions are
+  // cancelled. Does not bound per-model JiT compilation; see jit_timeout_millis.
   int32 timeout_millis = 7;
 
+  // Per-model JiT compilation worker timeout. Each action with JiT code gets
```
ikholopov-omni, line 22: Why do we need it? Why not just use a generic timeout_millis? Scoping JiT stage of execution of timeout_millis looks like a broken contract to me.

protos/jit.proto
```diff
@@ -126,6 +126,8 @@ message JitCompilationRequest {
   repeated string file_paths = 6;
   // Current execution information for introspection.
   RunningExecutionData execution_data = 7;
+  // File name where the target is defined.
+  string file_name = 8;
```
ikholopov-omni: Why do we need it? Is it just for better error handling? (having it in PR description would be helpful)

protos/execution.proto
```diff
@@ -118,6 +128,7 @@ message TaskResult {
   string error_message = 2;
   Timing timing = 3;
   ExecutionMetadata metadata = 4;
+  string compiled_sql = 5;
```
ikholopov-omni, line 132: Why do we need it here? (I have some guess, but please add a comment explaining when it is populated).

protos/execution.proto
```diff
@@ -46,6 +54,8 @@ message ExecutionAction {
 
   string jit_code = 12;
 
+  bool disabled = 13;
```
ikholopov-omni: Why do we need this now? I don't think it was required before and don't understand why we need it now.

protos/core.proto
```diff
@@ -51,9 +51,16 @@ message CompileConfig {
...
-  // Override compilation timeout settings.
+  // Override top-level compilation timeout (the single compile worker that
+  // runs once per invocation). Does not bound per-model JiT compilation;
+  // see jit_timeout_millis.
   int32 timeout_millis = 6;
 
+  // Per-model JiT compilation worker timeout. Each model with JiT code gets
+  // its own fresh budget; this is per-model, not a shared budget across the
+  // compile.
+  int32 jit_timeout_millis = 13;
```
ikholopov-omni: Same as protos/execution.proto.

ikholopov-omni (changes requested):

protos/execution.proto
rafal-hawrylak: timeout_millis specifies the wall-clock deadline for the entire execution run (which can be hours). JiT compilation, however, is executed on a separate worker thread for each individual model immediately before its tasks run. If a single model's compilation hangs, allowing it to consume the entire timeout_millis would block the pipeline and starve other models. Each model needs its own small, independent compilation budget (defaulting to 60s) that resets per action.

protos/core.proto
rafal-hawrylak: Same reply as https://github.com/dataform-co/dataform/pull/2182#discussion_r3441900715

protos/jit.proto
rafal-hawrylak: The `file_name` property is required to enable correct resolution of relative module imports (via `require` / `import`) inside the sandboxed VM during JiT compilation. Without it, the VM cannot determine the correct directory context of the SQLX/JS source code to resolve relative paths to dependencies in the project structure.

protos/execution.proto
rafal-hawrylak: `compiled_sql` is needed to expose the dynamically compiled SQL during dry-runs (`--dry-run`). Because JiT actions are compiled right before execution, their generated SQL is only available at runtime. Populating `compiled_sql` in `TaskResult` allows dry-run executions to output the final compiled SQL to the user. I will add a code comment explaining that this is populated during dry-run executions.

protos/execution.proto
rafal-hawrylak: With the introduction of JiT, actions are compiled at execution time, meaning both enabled and disabled JiT actions initially have an empty `tasks` array. To prevent the runner from compiling and executing disabled JiT actions, the explicit `disabled` flag is required on the `ExecutionAction` proto so the runner can identify and skip them.

protos/execution.proto
ikholopov-omni: Ok. Why not only respect `timeout_millis` deadline and don't set per-jit task deadline at all - instead terminate them once an overarching `timeout_millis` is terminated?

protos/jit.proto
ikholopov-omni: Wait, why does JiT compilation would needs a source code? JiT only supports pure functions without any external references.

protos/execution.proto
ikholopov-omni: Understood. Please add an appropriate field description.

protos/execution.proto
ikholopov-omni: Can we somehow handle it in the execution action creation logic instead and just skip their creation in the first place?

ikholopov-omni (changes requested):
See comments

protos/jit.proto
rafal-hawrylak: Right, previously there was idea to support require to allow ie helpers. However for the parity for Dataform Cloud that has been removed. Removing this property.

protos/execution.proto
rafal-hawrylak: Added

protos/execution.proto
rafal-hawrylak: My idea here was to have separate budget for the workers so if among hundreds there is only one that hung or is long-running, all the rest would have a chance to be successfully executed. With global deadline this single action will eat the whole budget.

ikholopov-omni: I don't think I communicated my idea well. Timeouts shouldn't be budgets. They are not additive. Timeout should be a deadline. A point in time after which the whole run should ramp up the activity. If a single action is hanging - ok, others are still running in their execution context. If the general timeout is reached, we terminate whoever failed to complete execution.

protos/execution.proto
rafal-hawrylak: Yes, that is doable. It will be handled just as AoT is handled - empty tasks heuristic.

protos/execution.proto
rafal-hawrylak: Ok, a budget is not a fortunate word here - correct - timeouts should not add up.

But lets take another, bolder example. Lets say there is a separate timeout on connection to BQ or execution in BQ - which is a common practice. Such a timeout may coexist with the global timeout. Having dedicated timeouts usually allows more graceful custom subprocesses eviction.

ikholopov-omni: Ok, so summing up, you will do the following:
1) Have a global overarching timeout that covers the whole run. Once it is reached, whatever is running is terminated.
2) In addition, it is possible to set JiT action timeout. If JiT action reaches it, it is marked as failed, but can free up an execution slot for other queued actions (if there is a concurrency limit hit).

Is this description full & correct now? If yes, please update comments and remove unnecessary confusing statements or terms (like "Does not bound per-model JiT compilation" or "budget").

ikholopov-omni (changes requested):

protos/core.proto
ikholopov-omni: Why it is present in Compilation proto though? JiT timeout only makes sense at execution config.

protos/execution.proto
rafal-hawrylak: Yes, udpated

protos/core.proto
rafal-hawrylak: Correct, removed

ikholopov-omni (approved):

PR #2184, opened 2026-05-28
Integrate JiT compilation into CLI runtime

rafal-hawrylak:

### DIFF
--- a/cli/api/commands/run.ts
+++ b/cli/api/commands/run.ts
@@ -1,12 +1,16 @@
 import EventEmitter from "events";
 import Long from "long";
 
+import { JitCompileChildProcess } from "df/cli/api/commands/jit/compiler";
 import * as dbadapters from "df/cli/api/dbadapters";
 import { IBigQueryExecutionOptions } from "df/cli/api/dbadapters/bigquery";
+import { ExecutionSql } from "df/cli/api/dbadapters/execution_sql";
+import { DEFAULT_JIT_COMPILATION_TIMEOUT_MILLIS } from "df/cli/api/utils/constants";
 import { Flags } from "df/common/flags";
 import { retry } from "df/common/promises";
 import { deepClone, equals } from "df/common/protos";
 import { targetStringifier } from "df/core/targets";
+import { version } from "df/core/version";
 import { dataform } from "df/protos/ts";
 
 const CANCEL_EVENT = "jobCancel";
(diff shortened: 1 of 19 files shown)


cli/api/commands/run.ts
```diff
@@ -24,32 +28,70 @@ export interface IExecutedAction {
...
     graph,
-    executionOptions,
+    optionsOrResult,
     partiallyExecutedRunResult,
     runnerNotificationPeriodMillis
   ).execute();
 }
 
 export class Runner {
+  private static handleParamsOverloads(
```
ikholopov-omni: 1. Please don't insert static functions the class.
2. I don't think having it static vs having it as a standalone function brings any value.

cli/api/commands/run.ts
```diff
@@ -24,32 +28,70 @@ export interface IExecutedAction {
...
+    optionsOrResult: RunOptionsOrResult,
+    partiallyExecutedRunResult: dataform.IRunResult
+  ): { options: IExecutionOptions; runResult: dataform.IRunResult } {
+    if (optionsOrResult && "actions" in optionsOrResult) {
```
ikholopov-omni: Please only use "'key' in X"  checks only for types with signature '{[key: string]: ...}`. 

This will break on some JS compilers that use member-renaming optimizations.

Use `optionsOrResult.actions === undefined`

cli/api/commands/run.ts
```diff
@@ -63,22 +105,33 @@ export class Runner {
   private cancelled = false;
   private timeout: NodeJS.Timer;
   private timedOut = false;
+  // Populated when the runner cancels execution (timeout or external cancel).
+  // Surfaced as errorMessage on SKIPPED tasks so users can see why an action
+  // never ran or stopped mid-flight.
+  private skipReason: string = "";
```
ikholopov-omni, line 133: I don't think using raw strings for error handling is a good idea. Please use string literal types

cli/api/commands/run.ts
```diff
@@ -24,32 +28,70 @@ export interface IExecutedAction {
...
+  ): { options: IExecutionOptions; runResult: dataform.IRunResult } {
+    if (optionsOrResult && "actions" in optionsOrResult) {
+      return {
+        runResult: {
+          ...(optionsOrResult as dataform.IRunResult)
+        },
+        options: { projectDir: "." }
+      };
+    }
+    const options = { ...(optionsOrResult as IExecutionOptions) };
```
ikholopov-omni: Why not just assignment with assertion? Only to populate default?

cli/api/commands/run.ts
```diff
@@ -63,22 +105,33 @@ export class Runner {
...
+  // Surfaced as errorMessage on SKIPPED tasks so users can see why an action
+  // never ran or stopped mid-flight.
+  private skipReason: string = "";
   private executionTask: Promise<dataform.IRunResult>;
 
   constructor(
     private readonly dbadapter: dbadapters.IDbAdapter,
     private readonly graph: dataform.IExecutionGraph,
-    private readonly executionOptions: IExecutionOptions = {},
+    optionsOrResult: RunOptionsOrResult = {},
```
ikholopov-omni, line 153: I am not sure what is the motivation of complicating the code with mixed type resolution in class. Probably these two should be separate functions returning Runner instances.

cli/api/commands/run.ts
```diff
@@ -228,13 +284,24 @@ export class Runner {
...
           target: pendingAction.target,
           status: dataform.ActionResult.ExecutionStatus.SKIPPED,
-          tasks: pendingAction.tasks.map(() => ({
-            status: dataform.TaskResult.ExecutionStatus.SKIPPED
-          }))
+          // Surface at least one task carrying the skip reason; matches the
+          // shape of tasks the action would have produced. If the action has
+          // no built tasks (e.g. JiT-only), still emit one so the cause is
+          // visible in the JSON output.
+          tasks: (pendingAction.tasks && pendingAction.tasks.length > 0
```
ikholopov-omni: Please use ternary operators sparingly and only for simple statements. This is unnecessary complex code.

cli/api/commands/run.ts
```diff
@@ -466,6 +562,68 @@ export class Runner {
...
+    const jitResponse = await this.executionOptions.jitCompiler(
+      jitRequest,
+      this.executionOptions.projectDir,
+      this.dbadapter,
+      client,
+      timeoutMillis,
+      this.getBigQueryExecutionOptions(action)
+    );
+
+    if (jitResponse.table) {
```
ikholopov-omni, line 653: This if else branch sequence should be a function returning tasks.

cli/vm/jit_worker.ts
```diff
@@ -5,21 +5,42 @@ import { NodeVM } from "vm2";
...
-  if (res.type === "rpc_response") {
-    const callback = pendingRpcCallbacks.get(res.correlationId);
-    if (callback) {
-      pendingRpcCallbacks.delete(res.correlationId);
-      if (res.error) {
-        callback(res.error, null);
-      } else {
-        callback(null, res.response);
+
+// Guard against double-initialization in some environments (e.g. Bazel)
```
ikholopov-omni, line 9: Can we please avoid these hacks and keep initialization idempotent, please?

cli/index.ts
```diff
@@ -410,6 +428,7 @@ export function runCli() {
         ],
         processFn: async argv => {
           const projectDir = argv[projectDirMustExistOption.name];
+          const logger = new Logger(!argv[jsonOutputOption.name]);
```
ikholopov-omni, line 431: Please don't mix refactoring with new functionality introduction.

ikholopov-omni (changes requested):

cli/api/commands/run.ts
rafal-hawrylak: Ok, extracted to createTasksFromJitResponse

cli/index.ts
rafal-hawrylak: Extracted to https://github.com/dataform-co/dataform/pull/2210

cli/vm/jit_worker.ts
rafal-hawrylak: Will address it here: https://github.com/dataform-co/dataform/pull/2211. Not addressed yet.

PR #2185, opened 2026-05-28
JiT test suite

rafal-hawrylak:

### DIFF
--- a/cli/BUILD
+++ b/cli/BUILD
@@ -65,8 +65,13 @@ ts_test_suite(
         "index_project_test.ts",
         "index_compile_test.ts",
         "index_run_e2e_test.ts",
+        "tests/jit/index_jit_main_test.ts",
+        "tests/jit/index_jit_advanced_test.ts",
+        "tests/jit/index_jit_dependency_test.ts",
+        "tests/jit/index_jit_runtime_test.ts",
         "util_test.ts",
         "tests/jit/jit_build_test.ts",
+        "tests/jit/jit_run_test.ts",
     ],
     data = [
         ":node_modules",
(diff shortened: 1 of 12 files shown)


cli/api/commands/run.ts
```diff
@@ -519,6 +565,86 @@ export class Runner {
...
+    });
+
+    const jitResponse = await this.executionOptions.jitCompiler(
+      jitRequest,
+      this.executionOptions.projectDir,
+      this.dbadapter,
+      client,
+      this.graph.runConfig?.jitTimeoutMillis || undefined,
+      this.getBigQueryExecutionOptions(action),
+      handleCancel => this.eEmitter.on(CANCEL_EVENT, handleCancel)
```
apilaskowski: When  JitCompileChildProcess.compile  finishes (whether via  resolve ,  reject , or worker  terminate() ), the  CANCEL_EVENT  ( jobCancel ) listener on  this.eEmitter  is never unregistered ( this.eEmitter.off(CANCEL_EVENT, handleCancel)  is not called).

We probably should handle this.

cli/api/commands/run.ts
```diff
@@ -476,6 +518,10 @@ export class Runner {
     };
     parentAction.tasks.push(taskResult);
     this.notifyListeners();
+    if (options.bigquery?.dryRun) {
+      taskResult.compiledSql = task.statement;
+
```
apilaskowski: nit: Trailing enter here.

cli/api/commands/run.ts
rafal-hawrylak: Good catch, added a fix. The original (non-JiT) handler management seems to have the very same issue. Eventually it should be fixed too.

ikholopov-omni (approved):

PR #2190, opened 2026-05-31
Publishing Dataform security patches

dataform-team:
Updating NPM package version to 3.0.59


apilaskowski:
Is this change doing anything with security? The only change here is version (which was already updated to that version).

PR #2191, opened 2026-06-01
Fix caller-file error (#2177) and enforce Core/CLI version match

### LINKED ISSUE #2177: Dataform Compile Failure — “Unable to find valid caller file” Error
Hi,

We are currently encountering a Dataform compile failure in CI with the following error appearing across many unrelated .sqlx files:

Error: Unable to find valid caller file; please report this issue.

This is followed by:

Compiled 0 action(s).

and the compile step fails with:

dataform compile --timeout=10m

The issue does not appear model-specific, as it occurs across multiple unrelated domains/files.

I did some investigation and found what appears to be a related Dataform issue/fix:
https://github.com/dataform-co/dataform/commit/1bdeca596afa5e15fd64266321e2410cc03a47df

Just wondering whether this is a known issue? 

Thanks.

rafal-hawrylak:
vm2 3.11.3 strips file paths from V8 CallSite objects inside the sandbox, so `getCallerFile()` in @dataform/core can no longer resolve the current file from the stack. Track it via a host-side stack exposed through `__df_enter`/`__df_exit`/`__df_current` sandbox helpers and read it through a getter on `global.__dataform_current_file`.

For @dataform/core <= 3.0.56 (no global fallback in getCallerFile), patch the bundle text at load time to consult the global. Gated on the installed Core version so newer bundles are never touched.

Add a CLI/Core compatibility check: reject installed Core whose major differs from the CLI's, or whose minor is below the CLI's minor. The error tells the user the exact `dataformCoreVersion:` line to set in `workflow_settings.yaml`. Skipped when the CLI version can't be determined (unbundled local dev).

Mirror the same wrapper pattern in `testing/run_core.ts`.

Add `compile rejects @dataform/core with incompatible version` test in `cli/index_compile_test.ts`.

### DIFF
--- a/cli/index_compile_test.ts
+++ b/cli/index_compile_test.ts
@@ -65,6 +65,71 @@ suite("compile command", ({ afterEach }) => {
     );
   });
 
+  test("compile rejects @dataform/core with incompatible version", async () => {
+    const projectDir = tmpDirFixture.createNewTmpDir();
+    // dataformCoreVersion in workflow_settings.yaml triggers the stateless
+    // install path (compile.ts copies to a tmp dir and runs `npm i`), so the
+    // test exercises the same flow real users hit. 2.9.0 is the latest 2.x on
+    // the registry; its major (2) is incompatible with the current CLI (3.x).
+    fs.writeFileSync(
+      path.join(projectDir, "workflow_settings.yaml"),
+      dumpYaml(
+        dataform.WorkflowSettings.create({
+          defaultProject: "dataform",
+          dataformCoreVersion: "2.9.0"
+        })
+      )
+    );
+
+    // npm needs a writable cache; ~/.npm is read-only in the bazel sandbox.
+    const npmCacheDir = tmpDirFixture.createNewTmpDir();
+    const stderr = (
+      await getProcessResult(
+        execFile(nodePath, [cliEntryPointPath, "compile", projectDir], {
+          env: { ...process.env, NPM_CONFIG_CACHE: npmCacheDir }
+        })
+      )
+    ).stderr;
+    expect(stderr).contains("@dataform/core 2.9.0 is not compatible with @dataform/cli");
+    expect(stderr).contains("matching major.minor");
+    expect(stderr).contains("Set `dataformCoreVersion:");
+  });
+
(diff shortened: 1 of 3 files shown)


cli/vm/compile.ts
```diff
@@ -9,18 +9,44 @@ import { dataform } from "df/protos/ts";
...
       "Could not find a recent installed version of @dataform/core in the project. Check that " +
         "either `dataformCoreVersion` is specified in `workflow_settings.yaml`, or " +
         "`@dataform/core` is specified in `package.json`. If using `package.json`, then run " +
         "`dataform install`."
     );
   }
+  const corePackageJsonPath = path.join(
+    compileConfig.projectDir, "node_modules", "@dataform", "core", "package.json"
+  );
+  const dataformCoreVersion: string =
```
kolina: could we do what we were doing doing previously [here](https://github.com/dataform-co/dataform/blob/26a2a8112d7fc419843732953df62f98bdcdc138/cli/vm/compile.ts#L73)

because I see a couple of problems with this approach:
* a user may have `workflow_settings.yaml` without `package.json`
* they may have something like `latest` in their `package.json`

cli/index_compile_test.ts
```diff
@@ -65,6 +65,38 @@ suite("compile command", ({ afterEach }) => {
     );
   });
 
+  test("compile rejects @dataform/core with incompatible version", async () => {
```
kolina, line 68: I'd suggest to write an integration test using `workflow_settings.yaml` with stateless compilation (that won't be using the current version of code for Dataform core)

kolina (changes requested):

cli/vm/compile.ts
rafal-hawrylak: Done. Version now read via require("@dataform/core").version inside the vm, same path as the compiler.

cli/index_compile_test.ts
rafal-hawrylak: Done. Test now writes workflow_settings.yaml with dataformCoreVersion: 2.9.0 and invokes the CLI end-to-end, exercising the stateless install path.
Also added a success-path test that compiles a real @dataform/core@3.0.50 project end-to-end to verify the caller-file shim.

kolina (approved):

PR #2195, opened 2026-06-02
Support incremental_strategy parameter and new insert_overwrite strategy

SuchodolskiEdvin:
- updated proto with new parameters
- added new tests
- added validation for chosen incremental_strategies
- added new insert_overwrite strategy logic


cli/api/dbadapters/execution_sql.ts
```diff
@@ -470,6 +454,84 @@ when matched then
...
+    query: string,
+    partitionBy: string,
+    updatePartitionFilter: string
+  ): string {
+    const uniqueId = this.uniqueIdGenerator();
+    const stagingTableUnqualified = `staging_table_temp_${uniqueId}`;
+    const backtickedColumns = columns.map(column => `\`${column}\``);
+    const resolveTargetTable = this.resolveTarget(target);
+
+    return `CREATE OR REPLACE TEMP TABLE \`${stagingTableUnqualified}\` AS (
```
kolina, line 486: can you split separate SQL statements into granular `Task.statement` calls?

cli/api/dbadapters/execution_sql.ts
```diff
@@ -470,6 +454,84 @@ when matched then
...
+      SELECT DISTINCT ${partitionBy}
+      FROM \`${stagingTableUnqualified}\`
+      WHERE ${partitionBy} IS NOT NULL
+    )
+  );
+
+  MERGE ${resolveTargetTable} T
+  USING \`${stagingTableUnqualified}\` S
+  ON FALSE
+  WHEN NOT MATCHED BY SOURCE AND ${partitionBy} IN UNNEST(partitions_for_replacement) ${updatePartitionFilter ? `and T.${updatePartitionFilter}` : ""} THEN
```
kolina: Such code `T.${updatePartitionFilter}` will only work when `updatePartitionFilter` has exactly one expression?

cli/api/dbadapters/execution_sql.ts
```diff
@@ -470,6 +454,84 @@ when matched then
...
+          columns,
+          incrementalQuery,
+          table.bigquery && table.bigquery.partitionBy,
+          table.bigquery && table.bigquery.updatePartitionFilter
+        );
+      case dataform.IncrementalStrategy.MERGE:
+        return this.mergeInto(
+          table.target,
+          columns,
+          incrementalQuery,
```
kolina: Can you deduplicate this with a call below?

core/actions/incremental_table.ts
```diff
@@ -217,6 +217,30 @@ export class IncrementalTable extends ActionBuilder<dataform.Table> {
...
+          this.session.compileError(
+            new Error("incrementalStrategy 'merge' requires 'uniqueKey' to be set."),
+            config.filename,
+            this.proto.target
+          );
+        }
+        break;
+      default:
+        break;
+    }
```
kolina: let's move it into sub-function for readability

core/actions/incremental_table_test.ts
```diff
@@ -987,4 +992,128 @@ select \${incremental()} as is_incremental`
...
+        }
+        SELECT 1`
+      );
+
+      const result = runMainInVm(coreExecutionRequestFromPath(projectDir));
+
+      expect(result.compile.compiledGraph.graphErrors.compilationErrors).deep.equals([]);
+      expect(result.compile.compiledGraph.tables[0].incrementalStrategy).equals(
+        dataform.IncrementalStrategy.INSERT_OVERWRITE
+      );
```
kolina: let's check the whole object here (same below)

protos/configs.proto
```diff
@@ -485,6 +497,10 @@ message ActionConfig {
     // If unset, the value from workflow_settings.yaml is used. If neither is set, default BigQuery behavior applies.
     // Dataform CLI only (GCP Dataform support pending).
     string reservation = 27;
+
+    // Optional. The incremental strategy to use when updating the table.
+    // Defaults to MERGE if uniqueKey is configured, or APPEND otherwise.
```
kolina: I wouldn't include it into config protos (because it's run-time strategy and not related to compiled DAG generation)

kolina (changes requested):

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Yes, you are correct. It is a known limitation that T.${updatePartitionFilter} only works for simple expressions (and fails on multi-expression SQL). Current implementation is designed to match the existing behavior of the standard MERGE strategy to maintain consistency between the two strategies for now. The fix of using updatePartitionFilter with several expression will be introduced in a separate PR.

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: I believe we can't split this SQL because we are using TEMP table. We could in case we would use persistent table.

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Fixed.

core/actions/incremental_table.ts
SuchodolskiEdvin: Done.

core/actions/incremental_table_test.ts
SuchodolskiEdvin: Fixed.

protos/configs.proto
SuchodolskiEdvin: Fixed.

core/actions/incremental_table_test.ts
```diff
@@ -987,4 +992,178 @@ select \${incremental()} as is_incremental`
...
+        `config {
+          type: "incremental",
+          incrementalStrategy: "insert_overwrite"
+        }
+        SELECT 1`
+      );
+
+      const result = runMainInVm(coreExecutionRequestFromPath(projectDir));
+
+      expect(result.compile.compiledGraph.graphErrors.compilationErrors.length).greaterThan(0);
```
Tuseeq1, line 1092: I think this should be checked to be exactly equal to 1 to avoid any other uncaught errors the code change might cause.

core/actions/incremental_table_test.ts
```diff
@@ -987,4 +992,178 @@ select \${incremental()} as is_incremental`
...
+        `config {
+          type: "incremental",
+          incrementalStrategy: "merge"
+        }
+        SELECT 1`
+      );
+
+      const result = runMainInVm(coreExecutionRequestFromPath(projectDir));
+
+      expect(result.compile.compiledGraph.graphErrors.compilationErrors.length).greaterThan(0);
```
Tuseeq1, line 1165: Same here.

core/actions/incremental_table.ts
```diff
@@ -736,6 +739,63 @@ export class IncrementalTable extends ActionBuilder<dataform.Table> {
...
+        throw new Error(`IncrementalStrategy value "${incrementalStrategy}" is not supported`);
+    }
+  }
+
+  private checkIncrementalStrategyRequirements(config: dataform.ActionConfig.IIncrementalTableConfig) {
+    switch (this.proto.incrementalStrategy) {
+      case dataform.IncrementalStrategy.INSERT_OVERWRITE:
+        if (!this.proto.bigquery || !this.proto.bigquery.partitionBy) {
+          this.session.compileError(
+            new Error("incrementalStrategy 'insert_overwrite' requires 'partitionBy' to be set."),
```
Tuseeq1: Nit: Inconsistent style. in some places we have capital first letter and in others we don't. We also use "." in some places and miss it in others. for example on line 759.

cli/api/dbadapters/execution_sql.ts
```diff
@@ -470,6 +454,77 @@ when matched then
 when not matched then
   insert (${backtickedColumns.join(",")}) values (${backtickedColumns.join(",")})`;
   }
+
+  private insertOverwrite(
+    target: dataform.ITarget,
+    columns: string[],
+    query: string,
+    partitionBy: string,
```
Tuseeq1: What happens if the table.bigquery or table.bigquery.partitionBy is not provided? will it be an empty string or 'null'?

cli/api/dbadapters/execution_sql.ts
```diff
@@ -470,6 +454,77 @@ when matched then
...
+    return `CREATE OR REPLACE TEMP TABLE \`${stagingTableUnqualified}\` AS (
+  ${query}
+);
+
+BEGIN
+  DECLARE partitions_for_replacement DEFAULT (
+    ARRAY(
+      SELECT DISTINCT ${partitionBy}
+      FROM \`${stagingTableUnqualified}\`
+      WHERE ${partitionBy} IS NOT NULL
```
Tuseeq1, line 495: What will this statement look like if we have empty partitionBy? also will it have `` around the column name somehow? Do we have an example for final sql that will be generated.

cli/api/dbadapters/execution_sql.ts
Tuseeq1: I think what Nick means here is to split the sql creation into statements not the sql itself.

cli/api/goldens/insert_overwrite_ignore.sql
```diff
@@ -0,0 +1,23 @@
...
+  MERGE `project-id.dataset-id.incremental_on_schema_change` T
+  USING `staging_table_temp_test_uuid` S
+  ON FALSE
+  WHEN NOT MATCHED BY SOURCE AND DATE(ts) IN UNNEST(partitions_for_replacement) and T.ts >= '2024-01-01' THEN
+    DELETE
+  WHEN NOT MATCHED BY TARGET THEN
+    INSERT (`id`,`field1`) VALUES (`id`,`field1`);
+END;
+
+DROP TABLE IF EXISTS `staging_table_temp_test_uuid`
```
Tuseeq1, line 25: If I read the execution_sql.ts file correctly, there should be a ";" at the end no? same in `insert_overwrite_extend.sql`.

cli/api/execution_sql_test.ts
```diff
@@ -86,4 +86,34 @@ suite("ExecutionSql with 'onSchemaChange'", () => {
     const expectedSql = fs.readFileSync("cli/api/goldens/on_schema_change_ignore.sql", "utf8");
     expect(procedureSql).to.equal(expectedSql.trim());
   });
+
+  test("generates INSERT_OVERWRITE script for IGNORE strategy", () => {
+    const table = {
+      ...baseTable,
+      incrementalStrategy: dataform.IncrementalStrategy.INSERT_OVERWRITE,
+      bigquery: {
+        partitionBy: "DATE(ts)",
```
Tuseeq1, line 95: Do we have an example column that is not a function like Date. My concern is that we a re biased toward handling this method. so it its just a normal column name we might not handle it correctly.

Tuseeq1 (changes requested):

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: If table.bigquery or table.bigquery.partitionBy is not provided or it is empty string - insertOverwrite() won't be reached. There is [validation](https://github.com/dataform-co/dataform/pull/2195/changes#diff-9153376f4ea7432c08759d8bf2bab426d0606566b8631fc46dc01beeaf05b936R778) of partitionBy parameter.

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: If partitionBy would contain column wrapped in function (like Date(ts)) adding backticks cause BigQuery to treat the entire expression as a literal column name.

Yes, we have golden files covering the generated SQL for insertOverwrite in cli/api/goldens/insert_overwrite*.sql

cli/api/dbadapters/execution_sql.ts
Tuseeq1: the validation link doesnt lead anywhere so I cannot check that.

cli/api/dbadapters/execution_sql.ts
Tuseeq1: this is question I have. So if PartitionBy is not a function like Date and is just a column name lets say id. This will not add backticks to it. Will the query run in that case.

core/actions/incremental_table_test.ts
SuchodolskiEdvin: Fixed.

core/actions/incremental_table_test.ts
SuchodolskiEdvin: Fixed.

core/actions/incremental_table.ts
SuchodolskiEdvin: Fixed.

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: It's strange, because it's working for me. This is link to the code change in scope of this PR. Please check core/actions/incremental_table.ts:checkIncrementalStrategyRequirements() (LINE 778)

cli/api/goldens/insert_overwrite_ignore.sql
SuchodolskiEdvin: Good catch. Fixed.

cli/api/execution_sql_test.ts
SuchodolskiEdvin: I added additional test for normal column.

cli/api/goldens/insert_overwrite_ignore.sql
SuchodolskiEdvin: Reverted this change. concatenateQueries() helper function (in cli/api/dbadapters/tasks.ts) explicitly trims trailing whitespace and strips the final semicolon from every individual statement before joining them together with. Because of this stripping logic, the final sql that Dataform generates don't have a trailing semicolon at the very end.

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: @kolina can you clarify what exactly do you mean?

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Yes, the query will run successfully, assuming the column isn't a reserved BigQuery keyword (like 'order'). If it is a keyword, the user can still provide the column name wrapped in backticks directly in their config. Just FYI: this matches the existing behavior of the MERGE strategy, which also injects uniqueKey columns without automatically adding backticks.

cli/api/dbadapters/execution_sql.ts
Tuseeq1: thanks for adding the fll path.

Tuseeq1 (approved):

cli/api/dbadapters/execution_sql.ts
kolina: I mean splitting statements. If it's not possible, fine.

cli/api/dbadapters/execution_sql.ts
kolina: I think it's better to implement it correctly from the start. IIRC current behaviour for `MERGE` may require an opt-in flag to avoid breaking existing code, so I'd avoid such compatibility issues.

kolina (changes requested):

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Fixed.

cli/api/BUILD
```diff
@@ -66,7 +66,7 @@ ts_test_suite(
     data = [
         ":node_modules",
         "//packages/@dataform/core:package_tar",
-        "//test_credentials:bigquery.json",
+        # "//test_credentials:bigquery.json",
```
kolina: Please merge https://github.com/dataform-co/dataform/commit/3478e82a73aae1fc52361fb67e88f937018a58af into your PR and uncomment

cli/api/dbadapters/execution_sql.ts
```diff
@@ -451,25 +443,104 @@ DROP TABLE IF EXISTS ${emptyTempTableName};
...
+
+    switch (table.incrementalStrategy) {
+      case dataform.IncrementalStrategy.INSERT_OVERWRITE:
+        return this.insertOverwrite(
+          table.target,
+          columns,
+          incrementalQuery,
+          table.bigquery && table.bigquery.partitionBy,
+          table.bigquery && table.bigquery.updatePartitionFilter,
+          table.bigquery && table.bigquery.incrementalPredicates
```
kolina: nit: just pass `table.bigquery` here to reduce amount of arguments passed to this function (same below)

cli/api/dbadapters/execution_sql.ts
```diff
@@ -451,25 +443,104 @@ DROP TABLE IF EXISTS ${emptyTempTableName};
...
 
-    private mergeInto(
+  private mergeInto(
     target: dataform.ITarget,
     columns: string[],
     query: string,
     uniqueKey: string[],
-    updatePartitionFilter: string
+    updatePartitionFilter: string,
+    incrementalPredicates: string[]
```
kolina: do we check that these parameters are mutually exclusive and are not passed at the same time?

cli/api/dbadapters/execution_sql.ts
kolina, line 1: Can we add a golden test with `incrementalPredicates` set? one should be enough (you can change one existing to use it)

kolina (changes requested):

cli/api/BUILD
SuchodolskiEdvin: Done.

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Fixed.

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Good catch! I hadn't originally considered it, but I've updated the PR to enforce mutual exclusivity and throw a compilation error.

cli/api/dbadapters/execution_sql.ts
SuchodolskiEdvin: Done.

docs/reference/configs.md
```diff
@@ -134,6 +134,7 @@ Action config defines the contents of `actions.yaml` configuration files.
...
 | requirePartitionFilter       | [bool](#bool)                                                                                                                      |          | Declares whether the partitioned table requires a WHERE clause predicate filter that filters the partitioning column.                                                                                                           |
 | updatePartitionFilter        | [string](#string)                                                                                                                  |          | SQL-based filter for when incremental updates are applied.                                                                                                                                                                      |
```
SuchodolskiEdvin, line 136: @kolina  Should I mark this field as Deprecated?

cli/api/execution_sql_test.ts
```diff
@@ -86,4 +86,34 @@ suite("ExecutionSql with 'onSchemaChange'", () => {
...
+        incrementalPredicates: ["ts >= '2024-01-01'"]
```
kolina: Let's have use table aliases here to showcase how to use it

core/actions/incremental_table.ts
```diff
@@ -736,6 +742,80 @@ export class IncrementalTable extends ActionBuilder<dataform.Table> {
...
+  private checkMutuallyExclusivePredicates(config: dataform.ActionConfig.IIncrementalTableConfig) {
+    if (
+      this.proto.bigquery &&
+      this.proto.bigquery.updatePartitionFilter &&
+      this.proto.bigquery.incrementalPredicates &&
+      this.proto.bigquery.incrementalPredicates.length > 0
+    ) {
+      this.session.compileError(
+        new Error(
+          `IncrementalPredicates and updatePartitionFilter cannot be both set. Use only incrementalPredicates.`
```
kolina: `incrementalPredicates` at the start of sentence?

cli/api/BUILD
kolina: Hm, but it's still commented (so if tests are passing, it may not be needed and can be removed)

docs/reference/configs.md
kolina: You'll need to add support for this field in SQL generation in GCP code before deprecating the old field.

kolina (approved):

docs/reference/configs.md
SuchodolskiEdvin: Good point.

cli/api/execution_sql_test.ts
SuchodolskiEdvin: Good point. Updated.

core/actions/incremental_table.ts
SuchodolskiEdvin: Fixed.

PR #2199, opened 2026-06-07
Ignore trailing positional arguments in the global flag parser

### LINKED ISSUE #2198: CLI crashes (Arg neither flag name nor flag value) when a positional argument follows a flag
## Summary

The global flag parser in `common/flags/index.ts` runs in a static initializer at **module-load time** and throws `Arg neither flag name nor flag value: <arg>` whenever a positional argument appears *after* a `--flag` in `process.argv`. Because it runs before — and independently of — yargs, this crashes the entire CLI for otherwise-valid invocations where the user simply puts a positional after an option.

## Reproduction

```bash
# Works — positional before the flag:
dataform run my_project --schema-suffix dev

# Crashes — same arguments, positional after the flag:
dataform run --schema-suffix dev my_project
```

Output:

```
Error: Arg neither flag name nor flag value: my_project
    at .../common/flags/index.ts:71
    at Module._compile (node:internal/modules/cjs/loader)
    ...
```

(Any command + any flag-before-positional combination triggers it, since the parser scans `process.argv` at import time regardless of the command.)

## Expected behavior

Both argument orderings should work. yargs already parses options and positionals in any order; the lightweight `common/flags` parser shouldn't reject a valid ordering — and certainly shouldn't crash the process at load time.

## Root cause

`Flags.parsedArgv` is built in a static initializer that scans `process.argv`. Once it sees any `--flag` it sets `flagsStarted = true`; a later non-flag token that isn't consumed as a flag value hits:

```ts
} else if (flagsStarted) {
  throw new Error(`Arg neither flag name nor flag value: ${splitArg}`);
}
```

This treats positionals-after-flags as fatal. But this flag system coexists with yargs and only needs to extract `--flag value` pairs; positionals (the command, project dir, etc.) are yargs's responsibility and should be ignored, not fatal. The throw also fires at import time, so it takes down the CLI before any command handler runs.

## Proposed fix

Make the parser lenient — ignore tokens that are neither a flag nor a flag value, instead of throwing:

```ts
let currentFlagName = "";
for (const splitArg of splitArgv) {
  if (splitArg.startsWith("--")) {
    currentFlagName = splitArg.slice(2);
    parsedArgv[currentFlagName] = "";
  } else if (currentFlagName) {
    parsedArgv[currentFlagName] = splitArg;
    currentFlagName = "";
  }
  // Non-flag args (command, positionals) are ignored — yargs parses those.
}
```

(Extracting the loop into a small pure function, e.g. `parseArgvFlags(argv)`, also makes it unit-testable.)

## Environment

- Verified present in `common/flags/index.ts` on `main`.
- Happy to open a PR with the fix and a unit test if that's useful.

aicayzer:
## Summary

`common/flags/index.ts` parses `process.argv` in a static initialiser that runs at module load, before yargs. Once it had seen a `--flag`, any following token that was neither another flag nor a flag value made it throw `Arg neither flag name nor flag value: <arg>` — crashing the entire CLI for argument orderings that yargs happily accepts, e.g. `dataform run --some-flag value my_project`.

## What changed

- Make the parser ignore any token that is neither a flag nor a flag value instead of throwing. Positional arguments (the command and its operands) are left to yargs, exactly as they already were when they appeared *before* the first flag.
- Extract the parsing loop into an exported `parseArgv(argv)` function so it can be unit-tested in isolation (it previously only ran against the real `process.argv`).
- Add `common/flags/index_test.ts` covering flag/value pairs, `--flag=value`, `--no-` prefixes, multiple flags, no flags, and positional arguments both before and after flags (the regression case).

## Why

Fixes https://github.com/dataform-co/dataform/issues/2198

### DIFF
--- a/common/flags/BUILD
+++ b/common/flags/BUILD
@@ -1,16 +1,38 @@
 package(default_visibility = ["//visibility:public"])
 
 load("//tools:ts_library.bzl", "ts_library")
+load("@df//testing:index.bzl", "ts_test_suite")
 
 ts_library(
     name = "flags",
     srcs = glob(
         ["*.ts"],
+        exclude = ["*_test.ts", "*.spec.ts"],
     ),
     deps = [
         "//:modules-fix",
         "@npm//@types/long",
         "@npm//@types/node",
+        "@npm//long",
+    ],
+)
+
+ts_test_suite(
+    name = "tests",
+    srcs = glob(["*_test.ts"]),
+    data = [
+        "@npm//source-map-support",
+    ],
+    templated_args = [
+        "--node_options=--require=source-map-support/register",
+        "--bazel_patch_module_resolver",
+    ],
+    deps = [
+        ":flags",
+        "@df//testing",
+        "@npm//@types/chai",
+        "@npm//@types/node",
+        "@npm//chai",
     ],
 )
 
--- a/common/flags/index.ts
+++ b/common/flags/index.ts
@@ -39,45 +39,46 @@ export class Flags {
     Flags.parsedArgv[flagName] = value;
   }
 
-  private static readonly parsedArgv = (() => {
-    const parsedArgv: { [flagName: string]: string } = {};
-
-    const splitArgv = [];
-    for (let arg of process.argv) {
-      // TODO: This is a temporary hack to be backwards-compatible with yargs behaviour, where
-      // to switch off a boolean flag, it requires a 'no-' prefix in front of the flag name.
-      if (arg.startsWith("--no-")) {
-        arg = `--${arg.slice(5)}=false`;
-      }
-      if (arg.startsWith("--") && arg.includes("=")) {
-        splitArgv.push(arg.slice(0, arg.indexOf("=")));
-        splitArgv.push(arg.slice(arg.indexOf("=") + 1));
-      } else {
-        splitArgv.push(arg);
-      }
-    }
-
-    let flagsStarted = false;
-    let currentFlagName = "";
-    for (const splitArg of splitArgv) {
-      if (splitArg.startsWith("--")) {
-        flagsStarted = true;
-        currentFlagName = splitArg.slice(2);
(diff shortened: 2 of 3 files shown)


kolina (changes requested):
Your test seems to fail (I suppose you'll need to add it to `BUILD` dependencies)

```
Step #1: ================================================================================
Step #1:     at Module._compile (node:internal/modules/cjs/loader:1761:14)
Step #1:     at Object.<anonymous> (/builder/home/.cache/bazel/_bazel_root/eab0d61a99b6696edb3d2aff87b585e8/sandbox/linux-sandbox/31/execroot/df/bazel-out/k8-py2-fastbuild/bin/common/common/flags/index.ts:1:1)
Step #1:     at require (node:internal/modules/helpers:152:16)
Step #1:     at Module.require (node:internal/modules/cjs/loader:1504:12)
Step #1:     at wrapModuleLoad (node:internal/modules/cjs/loader:245:24)
Step #1:     at TracingChannel.traceSync (node:diagnostics_channel:328:14)
Step #1:     at Module._load (node:internal/modules/cjs/loader:1227:37)
Step #1:     at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1064:22)
Step #1:     at defaultResolveImpl (node:internal/modules/cjs/loader:1059:19)
Step #1:     at Module.module.constructor._resolveFilename (/builder/home/.cache/bazel/_bazel_root/eab0d61a99b6696edb3d2aff87b585e8/sandbox/linux-sandbox/31/execroot/df/bazel-out/k8-py2-fastbuild/bin/common/flags/index_test.sh.runfiles/df/common/flags/index_test_require_patch.js:480:17)
Step #1: Error: Cannot find module 'long'. Please verify that the package.json has a valid "main" entry
```

aicayzer:
Thanks @kolina.Added `@npm//long` to the `flags` library deps (matching `common/strings`). `//common/...` green locally.

kolina (approved):

PR #2200, opened 2026-06-09
Add optional original_file_path to Assertion protos

rafal-hawrylak:
Lets callers record a source declaration path distinct from the existing `file_name`/`filename`. When the new field is unset, consumers continue to use `file_name` as today.

Backwards-compatible: additive, optional.


ikholopov-omni (changes requested):
I don't think we have a good enough justification for adding such a thing. The semantics here are too murky and require a broader discussion.

rafal-hawrylak:
Confirmed, no need for this change.

PR #2206, opened 2026-06-16
Bazel js toolchain rewrite

ikholopov-omni:
- Migrate to maintained aspect-rules for JS toolchain
- Use pnpm instead of yarn, remove no-longer required scripts for switching registry (as pnpm-lock doesn't hard-code the domain, it always respects .npmrc config)
- Keep the same bundling layout - ESM for CLI, CommonJS for Core
- Updated scripts to use updated tooling
- Complete migration to BzlMod, cleanup WORKSPACE


packages/rollup.config.js
```diff
@@ -37,9 +35,63 @@ const checkImports = imports => {
...
+      }
+
+      if (source.startsWith("df/") || source.startsWith("packages/")) {
+        const relPath = source.startsWith("df/") ? source.slice(3) : source;
+
+        const candidate = path.resolve(
+          process.cwd(),
+          "../../..",
+          "bazel-out",
+          "k8-fastbuild-py2",
```
rafal-hawrylak: This may fail on other platforms than linux

rafal-hawrylak (changes requested):

packages/rollup.config.js
ikholopov-omni: Reworked to not use platform specific paths, like "k8" ones. Unfortunately, the current set of bazel tools are not fully functional on Windows platform. That said, bazel-5 toolset was also non-functional on Windows platform.

rafal-hawrylak: In other words - is there is a regression here?

ikholopov-omni: No, there is no regression for Windows platform. It didn't work before and still doesn't work.

rafal-hawrylak (approved):

packages/@dataform/core/webpack.config.js
```diff
@@ -4,36 +4,35 @@ const webpack = require("webpack");
...
-  }
+  const binDir = process.cwd().endsWith("bin")
+    ? process.cwd()
+    : path.resolve(process.cwd(), process.env.BAZEL_BINDIR || ".");
 
   const config = {
-    mode: argv.mode || "development",
-    target: 'node',
-    entry: [path.resolve(runfilesDir, workspaceName, "packages/@dataform/core/index")],
+    mode: argv.mode || "production",
```
kolina, line 12: is changing default here needed?

packages/rollup.config.js
```diff
@@ -1,15 +1,30 @@
...
+    dir = parent;
+  }
+  return path.resolve(dir, process.env.BAZEL_BINDIR);
+}
 
 function convertToRegex(pattern) {
   if (pattern instanceof RegExp) {
     return pattern;
   }
-  // If it's a string, turn it into a regex, by escaping any regex characters in the string.
```
kolina, line 7: AI loves to remove comments :)

kolina (approved):
It makes sense to manually check that it doesn't break GCP compilation

PR #2210, opened 2026-06-19
Refactor Runner parameter overloads and resolve review comments

rafal-hawrylak:
Pre follow up of https://github.com/dataform-co/dataform/pull/2184

### DIFF
--- a/cli/api/commands/run.ts
+++ b/cli/api/commands/run.ts
@@ -24,6 +24,7 @@ export interface IExecutedAction {
 }
 
 export interface IExecutionOptions {
+  projectDir?: string;
   bigquery?: {
     jobPrefix?: string;
     actionRetryLimit?: number;
@@ -32,10 +33,12 @@ export interface IExecutionOptions {
   };
 }
 
+export type CancelReason = "timeout" | "user" | "cancellation";
+
 export function run(
   dbadapter: dbadapters.IDbAdapter,
   graph: dataform.IExecutionGraph,
-  executionOptions?: IExecutionOptions,
+  executionOptions: IExecutionOptions = {},
   partiallyExecutedRunResult: dataform.IRunResult = {},
   runnerNotificationPeriodMillis: number = flags.runnerNotificationPeriodMillis.get()
 ): Runner {
@@ -49,6 +52,24 @@ export function run(
 }
 
 export class Runner {
+  public static create(
+    dbadapter: dbadapters.IDbAdapter,
+    graph: dataform.IExecutionGraph,
+    options: IExecutionOptions = {},
+    runnerNotificationPeriodMillis: number = flags.runnerNotificationPeriodMillis.get()
+  ): Runner {
+    return new Runner(dbadapter, graph, options, {}, runnerNotificationPeriodMillis);
+  }
+
+  public static resume(
+    dbadapter: dbadapters.IDbAdapter,
+    graph: dataform.IExecutionGraph,
+    runResult: dataform.IRunResult,
+    runnerNotificationPeriodMillis: number = flags.runnerNotificationPeriodMillis.get()
+  ): Runner {
+    return new Runner(dbadapter, graph, {}, runResult, runnerNotificationPeriodMillis);
+  }
+
   private readonly warehouseStateByTarget: Map<string, dataform.ITableMetadata>;
 
   private readonly allActionTargets: Set<string>;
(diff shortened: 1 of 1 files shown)


cli/api/commands/run.ts
```diff
@@ -24,6 +24,7 @@ export interface IExecutedAction {
 }
 
 export interface IExecutionOptions {
+  projectDir?: string;
```
apilaskowski, line 27: Is this `projectDir` used downstream in latter PRs?

apilaskowski: Looking at this PR I can see that we are adding this field here, and then we are setting it (under some conditions to "."), but I am not sure how it is used. Possibly it is needed latter. 

I can guess what this field does in general by the name, but is it needed?

cli/api/commands/run.ts
```diff
@@ -32,23 +33,72 @@ export interface IExecutionOptions {
...
     graph,
-    executionOptions,
-    partiallyExecutedRunResult,
+    options,
+    runResult,
     runnerNotificationPeriodMillis
   ).execute();
 }
 
+function handleParamsOverloads(
```
apilaskowski: This is not obvious to me why we are handling this that way. Can we put some information on what this function is supposed to do?

cli/api/commands/run.ts
```diff
@@ -32,23 +33,72 @@ export interface IExecutionOptions {
...
+  optionsOrResult: RunOptionsOrResult,
+  partiallyExecutedRunResult: dataform.IRunResult
+): { options: IExecutionOptions; runResult: dataform.IRunResult } {
+  if (optionsOrResult && (optionsOrResult as dataform.IRunResult).actions !== undefined) {
```
apilaskowski: If an IRunResult object is passed with actions omitted or undefined, it will inadvertently be misclassified as IExecutionOptions, right? Is that expected?

cli/api/commands/run.ts
```diff
@@ -63,22 +113,40 @@ export class Runner {
   private cancelled = false;
   private timeout: NodeJS.Timer;
   private timedOut = false;
+  private skipReason: string = "";
   private executionTask: Promise<dataform.IRunResult>;
+  private readonly executionOptions: IExecutionOptions;
 
+  constructor(
```
apilaskowski: Since we want to use `create` and `resume` from now on to increase readability, what about making those constructors private (or protected) to enforce usage of those methods? wdyt?

cli/api/commands/run.ts
```diff
@@ -131,7 +199,16 @@ export class Runner {
     this.stopped = true;
   }
 
-  public cancel() {
+  public cancel(reason: CancelReason = "user") {
```
apilaskowski, line 158: I would suggest preparing explicit map of possible reasons to avoid mistakes, and to make it more readable.

cli/api/commands/run.ts
```diff
@@ -131,7 +199,16 @@ export class Runner {
...
+    if (!this.skipReason) {
+      if (reason === "timeout") {
+        this.skipReason = `Run timed out after ${this.graph.runConfig.timeoutMillis / 1000} seconds`;
```
apilaskowski: Is it possible that `graph.runConfig.timeoutMillis` is not defined? I am wondering if this might cause NaN to be displayed.

apilaskowski (changes requested):

cli/api/commands/run.ts
rafal-hawrylak: Yes, this is needed in the next PR - https://github.com/dataform-co/dataform/pull/2211. It is passed as the second arg to jitCompiler (via JitCompileChildProcess.compile) — it's the resolution root the worker needs to locate the caller's @dataform/core install. Plumbing it in here is the way to reduce the size of the next PR and still keep the reasonable amount of PRs.

cli/api/commands/run.ts
rafal-hawrylak: Adding a doc block

cli/api/commands/run.ts
rafal-hawrylak: Yes, this is needed in the next PR — https://github.com/dataform-co/dataform/pull/2211. It is passed as the second arg to `jitCompiler` (via `JitCompileChildProcess.compile`) — it is the resolution root the worker needs to locate the caller's `@dataform/core` install. Plumbing it in here is the way to reduce the size of the next PR and still keep the reasonable amount of PRs.

cli/api/commands/run.ts
rafal-hawrylak: After another refactor I removed this function entirely.

cli/api/commands/run.ts
rafal-hawrylak: Force-pushed a follow-up taking the alternative approach: delete the union entirely (per ikholopov-omni's guidance on #2184).

cli/api/commands/run.ts
rafal-hawrylak: Same - force-pushed a follow-up taking the alternative approach: delete the union entirely (per ikholopov-omni's guidance on #2184).

cli/api/commands/run.ts
rafal-hawrylak: Used a switch over the CancelReason union instead of a Reason→string map. The map form loses type-safe exhaustiveness (a new CancelReason variant would fall through silently), and here it also can't work cleanly because the "timeout" message needs runtime state (graph.runConfig.timeoutMillis). The switch on the string-literal union gives compile-time exhaustiveness and keeps the runtime read local.

cli/api/commands/run.ts
rafal-hawrylak: Good catch — added a ?? 0 guard: if timeoutMillis is undefined the message falls back to "Run timed out" instead of "Run timed out after NaN seconds". In practice this branch only fires when cancel("timeout") is called, which is only wired to the setTimeout in execute() that itself requires timeoutMillis to be truthy — so the guard is defensive rather than reachable today.

apilaskowski:
- Makes some sense to include `projectDir` here in preparation for #2211 (in perfect world I would prefer to have it separately).
- Removing the parameter union and `handleParamsOverloads` in favor of `Runner.create` and `Runner.resume` makes the instantiation API significantly cleaner.
- The `switch` statement for `CancelReason` exhaustiveness and the fallback guard for missing `timeoutMillis` look great.
LGTM! 🚀

apilaskowski (approved):

ikholopov-omni (approved):

PR #2211, opened 2026-06-19
Integrate JiT compilation into CLI runtime

rafal-hawrylak:
https://github.com/dataform-co/dataform/pull/2184 with extracted refactorings.

### DIFF
--- a/cli/api/BUILD
+++ b/cli/api/BUILD
@@ -1,6 +1,6 @@
+load("//tools:ts_library.bzl", "ts_library")
 load("//testing:index.bzl", "ts_test_suite")
 load("//tools:node_modules.bzl", "node_modules")
-load("//tools:ts_library.bzl", "ts_library")
 
 package(default_visibility = ["//visibility:public"])
 
--- a/cli/api/BUILD
+++ b/cli/api/BUILD
@@ -1,6 +1,6 @@
-load("//tools:ts_library.bzl", "ts_library")
 load("//testing:index.bzl", "ts_test_suite")
 load("//tools:node_modules.bzl", "node_modules")
+load("//tools:ts_library.bzl", "ts_library")
 
 package(default_visibility = ["//visibility:public"])
 
--- a/cli/api/commands/base_worker.ts
+++ b/cli/api/commands/base_worker.ts
@@ -4,9 +4,10 @@ export abstract class BaseWorker<TResponse, TMessage = any> {
   protected constructor(private readonly loaderPath: string) {}
 
   protected async runWorker(
-    timeoutMillis: number,
+    timeoutMillis: number | undefined,
     onBoot: (child: ChildProcess) => void,
-    onMessage: (message: TMessage, child: ChildProcess, resolve: (res: TResponse) => void, reject: (err: Error) => void) => void
+    onMessage: (message: TMessage, child: ChildProcess, resolve: (res: TResponse) => void, reject: (err: Error) => void) => void,
+    onCancel?: (cancel: () => void) => void
   ): Promise<TResponse> {
     const forkScript = this.resolveScript();
     const child = fork(forkScript, [], {
@@ -22,20 +23,28 @@ export abstract class BaseWorker<TResponse, TMessage = any> {
           return;
         }
         completed = true;
-        clearTimeout(timeout);
+        if (timeout !== undefined) {
+          clearTimeout(timeout);
+        }
         child.kill("SIGKILL");
         fn();
       };
 
-      const timeout = setTimeout(() => {
-        terminate(() =>
-          reject(new Error(
-            `Compilation timed out after ${timeoutMillis / 1000} seconds. ` +
-            `To allow more time, re-run with a longer --timeout ` +
-            `(e.g. --timeout=2m, --timeout=1h).`
-          ))
-        );
-      }, timeoutMillis);
+      const timeout = timeoutMillis !== undefined
+        ? setTimeout(() => {
+            terminate(() =>
+              reject(new Error(
+                `Compilation timed out after ${timeoutMillis / 1000} seconds. ` +
+                `To allow more time, re-run with a longer --timeout ` +
+                `(e.g. --timeout=2m, --timeout=1h).`
+              ))
+            );
+          }, timeoutMillis)
+        : undefined;
+
+      onCancel?.(() =>
+        terminate(() => reject(new Error("Run cancelled while worker was in flight.")))
+      );
 
       child.on("message", (message: any) => {
         if (message.type === "worker_booted") {
--- a/cli/api/commands/base_worker.ts
+++ b/cli/api/commands/base_worker.ts
@@ -4,7 +4,7 @@ export abstract class BaseWorker<TResponse, TMessage = any> {
   protected constructor(private readonly loaderPath: string) {}
 
   protected async runWorker(
-    timeoutMillis: number | undefined,
+    timeoutMillis: number,
     onBoot: (child: ChildProcess) => void,
     onMessage: (message: TMessage, child: ChildProcess, resolve: (res: TResponse) => void, reject: (err: Error) => void) => void,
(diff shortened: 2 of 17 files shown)


rafal-hawrylak:
Need to address: https://github.com/dataform-co/dataform/pull/2184#discussion_r3351928109

rafal-hawrylak:
> Need to address: [#2184 (comment)](https://github.com/dataform-co/dataform/pull/2184#discussion_r3351928109)

Solved in:
- https://github.com/dataform-co/dataform/pull/2211/changes/bf38a435c1c4c4642d1048ab8e93ed259dabab02
- https://github.com/dataform-co/dataform/pull/2211/changes/10c4a99ace63aae23f5dee54c9d3d4f9cf9bd05b

cli/api/commands/jit/compiler.ts
```diff
@@ -90,7 +93,7 @@ export class JitCompileChildProcess extends BaseWorker<
       child.send({
         type: "rpc_response",
         correlationId: message.correlationId,
-        response
+        response: Array.from(response)
```
ikholopov-omni: Why are we doing this?

cli/api/commands/base_worker.ts
```diff
@@ -22,20 +23,28 @@ export abstract class BaseWorker<TResponse, TMessage = any> {
           return;
         }
         completed = true;
-        clearTimeout(timeout);
+        if (timeout !== undefined) {
```
ikholopov-omni: Why run this check? clearTimeout on undefined is a noop

cli/api/commands/build.ts
```diff
@@ -108,11 +109,16 @@ export class Builder {
   private toPartialExecutionAction(
     action: dataform.ITable | dataform.IOperation | dataform.IAssertion
   ) {
-    return dataform.ExecutionAction.create({
+    const jitCode = (action as any).jitCode;
```
ikholopov-omni: Why are we any casting here?

cli/api/commands/base_worker.ts
```diff
@@ -22,20 +23,28 @@ export abstract class BaseWorker<TResponse, TMessage = any> {
...
-      const timeout = setTimeout(() => {
-        terminate(() =>
-          reject(new Error(
-            `Compilation timed out after ${timeoutMillis / 1000} seconds. ` +
-            `To allow more time, re-run with a longer --timeout ` +
-            `(e.g. --timeout=2m, --timeout=1h).`
-          ))
-        );
-      }, timeoutMillis);
+      const timeout = timeoutMillis !== undefined
```
ikholopov-omni: Under what circumstances would we set no timeout at all?

cli/api/commands/run.ts
```diff
@@ -381,7 +414,20 @@ export class Runner {
     actionResult.timing = timer.current();
     this.notifyListeners();
 
-    await this.dbadapter.withClientLock(async client => {
+    try {
```
ikholopov-omni: try-if-try-catch-endif--catch
is a pretty long chain.

Can we refactor is as
```
if (...) {
  try {
  ...
  } catch()
}

try {
...
```

Not sure if second try is even required here...

cli/api/commands/run.ts
```diff
@@ -519,6 +565,86 @@ export class Runner {
...
+  private createTasksFromJitResponse(
+    action: dataform.IExecutionAction,
+    jitResponse: dataform.IJitCompilationResponse
+  ): dataform.IExecutionTask[] {
+    if (jitResponse.table) {
+      const table = dataform.Table.create({
+        ...action,
+        ...jitResponse.table,
+        enumType:
+          action.tableType === "view"
```
ikholopov-omni: Please don't chains more complex then a single ternary operator. Most of people (including myself) won't be able with certainty say what's the order precedence here without consulting JS spec.

cli/api/utils/constants.ts
```diff
@@ -1 +1,3 @@
-export const DEFAULT_COMPILATION_TIMEOUT_MILLIS = 60000;
+// Top-level compilation worker timeout (single dataform compile worker that
+// runs once per invocation).
+export const DEFAULT_COMPILATION_TIMEOUT_MILLIS = 300000;
```
ikholopov-omni: Why are we reducing it?

ikholopov-omni (changes requested):

cli/api/commands/base_worker.ts
rafal-hawrylak: True, removed

cli/api/commands/build.ts
rafal-hawrylak: Right, no need for that

cli/api/commands/jit/compiler.ts
rafal-hawrylak: child.send uses JSON serialization by default (fork isn't started with serialization: "advanced") and JSON turns a Uint8Array into {"0":n,"1":n,...}. On the receiving side (cli/vm/jit_worker.ts), Buffer.from(res.response) then throws ERR_INVALID_ARG_TYPE because that object shape isn't a valid Buffer input. Verified empirically. I've added an inline comment documenting the constraint.

cli/api/commands/run.ts
rafal-hawrylak: Fixed

cli/api/commands/run.ts
rafal-hawrylak: True

cli/api/utils/constants.ts
rafal-hawrylak: It was increased not decreased. But we can keep it here unchanged and revisit when needed.

cli/api/commands/base_worker.ts
rafal-hawrylak: Right, reverted the default behaviour

cli/api/BUILD
```diff
@@ -1,6 +1,6 @@
+load("//tools:ts_library.bzl", "ts_library")
```
apilaskowski: Is linter happy with this order?

testing/test.ts
```diff
@@ -26,7 +26,7 @@ export function test(
 }
 
 export class Test {
-  public static readonly DEFAULT_TIMEOUT_MILLIS = 30000;
+  public static readonly DEFAULT_TIMEOUT_MILLIS = 300000;
```
apilaskowski: Just making sure. Is this intentional that all tests will have timeout of 5 minutes now?

cli/api/commands/run.ts
```diff
@@ -381,7 +414,20 @@ export class Runner {
...
 
-    await this.dbadapter.withClientLock(async client => {
+    if (action.jitCode) {
+      try {
+        await this.compileJitAction(action, this.dbadapter);
+      } catch (e) {
+        actionResult.status = dataform.ActionResult.ExecutionStatus.FAILED;
+        actionResult.tasks.push({
+          status: dataform.TaskResult.ExecutionStatus.FAILED,
+          errorMessage: `JiT compilation error: ${e.message}`
```
apilaskowski: Should we use `e?.message || String(e)` instead? This is just in case error thrown does not have `message` field.

cli/api/BUILD
rafal-hawrylak: It was but changed it anyway

cli/api/commands/run.ts
rafal-hawrylak: Done

testing/test.ts
rafal-hawrylak: Restored it. Initially there was an idea to increase the default timeout (for non-tests). The idea was rejected, non-test code reverted but the timeout in test was overlooked.

apilaskowski (approved):
Lets also wait for review by @ikholopov-omni

ikholopov-omni (approved):

PR #2212, opened 2026-06-21
feat(cli): node selection for dataform compile

### LINKED ISSUE #2203: Node selection for `dataform compile` (compile/output a single model)
`dataform compile` always emits the **entire** compiled graph. There is no way to ask
the CLI for the compiled SQL of a single model (or a subset of models) the way
`--actions` / `--tags` already work for `dataform run`.

I'd like `dataform compile` (and `dataform compile --json`) to accept the existing
selection flags so I can get the compiled SQL for just the action(s) I care about, e.g.:

```bash
dataform compile --actions my_model --json
dataform compile --tags daily --include_deps --json
```

This mirrors dbt's `dbt compile --select my_model`, which is a feature I rely on for
quickly inspecting the rendered SQL of one model without scrolling through the whole graph
or post-processing the JSON with `jq`.

Looking at code in this repo i understand that the compiler must evaluate the whole project because `ref()` resolution and dependency graph require every action to be registered. So this request is not about **partial compilation** its purely about filtering the compile ouput to the selected actions (similar to what prune does for run) its an UX improvement in CLI.

## Possible Implementation

The filtering logic seems to already exists and is reused  `prune()` in
[`cli/api/commands/prune.ts`](https://github.com/dataform-co/dataform/blob/main/cli/api/commands/prune.ts) takes a `CompiledGraph` + a
`RunConfig` (`actions`, `tags`, `includeDependencies`, `includeDependents`) and returns a
filtered `CompiledGraph`. `dataform run`/`build` already call it.

The change is to make the `compile` command optionally apply `prune()` before printing:

## Open question i have

I would like to work on this but have some questions to discuss

1. When a selector is present, should compilation
  errors from *unselected* actions still fail the command / be printed? I lean toward keeping
  graph-level errors as-is (the whole project still had to compile) but suppressing per-action
  errors unrelated to the selection

2. f the selector matches nothing, should `compile` warn and
  exit non-zero (like a "no actions" case) or just emit an empty graph?

alandrebraga:
## Summary

Adds node selection to `dataform compile`, mirroring `dbt compile --select`. `compile` now accepts the same selection flags as `run`/`build` — `--actions`, `--tags`, `--include-deps`, `--include-dependents` — and filters the printed graph to just the selected action(s):

```bash
dataform compile --actions my_model --json
dataform compile --tags daily --include-deps --json
```

Resolves #2203.

## How it works

This is **output filtering, not partial compilation**. The whole project must
still compile (`ref()` resolution and the dependency graph require every action
registered), so we compile fully and then prune the result before printing —
reusing the existing `prune()` that `run`/`build` already use, plus the shared
yargs option definitions. No proto changes.

- A clean graph is pruned only when a selector (`--actions`/`--tags`) is present.
- On compile errors, the full graph + errors print unchanged (graph-level errors kept as-is).
- A selector matching nothing emits an empty graph and exits 0, matching `prune`'s run-side behavior.
- `--include-deps` / `--include-dependents` keep their existing validation (only valid alongside `--actions`/`--tags`).

## Tests

New `compile node selection` suite in `cli/index_compile_test.ts` over an
`upstream -> midstream -> downstream` project, covering: full graph (no selector),
`--actions`, `--include-deps`, `--include-dependents`, `--tags`, empty-match
exits 0, and the no-selector validation error. `//cli:index_compile_test` passes.

## Credit

Implementation by @ihistand (Ivan Histand), proposed in SQLAnvil/sqlanvil#27 in
response to this issue. Thank you!

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -5,7 +5,7 @@ import parseDuration from "parse-duration";
 import * as path from "path";
 import yargs from "yargs";
 
-import { build, compile, credentials, init, install, run, test } from "df/cli/api";
+import { build, compile, credentials, init, install, prune, run, test } from "df/cli/api";
 import { CREDENTIALS_FILENAME } from "df/cli/api/commands/credentials";
 import { BigQueryDbAdapter } from "df/cli/api/dbadapters/bigquery";
 import { prettyJsonStringify } from "df/cli/api/utils";
@@ -393,6 +393,10 @@ export function runCli() {
           dotOutputOption,
           timeoutOption,
           quietCompileOption,
+          actionsOption,
+          tagsOption,
+          includeDepsOption,
+          includeDependentsOption,
           {
             name: verboseOptionName,
             option: {
@@ -429,7 +433,24 @@ export function runCli() {
               timeoutMillis: argv[timeoutOption.name] || undefined,
               verbose: argv[verboseOptionName] || false
             });
-            printCompiledGraph(compiledGraph, outputType, argv[quietCompileOption.name]);
+
+            // The whole project must compile (ref() resolution needs every action
+            // registered), but the printed output can be filtered to the selected
+            // action(s) -- mirroring how `run`/`build` prune the graph. We only prune
+            // a clean graph; if compilation produced errors we print the full graph
+            // plus the errors, keeping graph-level errors as-is.
+            const hasSelector =
+              argv[actionsOption.name]?.length > 0 || argv[tagsOption.name]?.length > 0;
+            const outputGraph =
+              hasSelector && !compiledGraphHasErrors(compiledGraph)
+                ? prune(compiledGraph, {
+                    actions: argv[actionsOption.name],
+                    tags: argv[tagsOption.name],
+                    includeDependencies: argv[includeDepsOption.name],
+                    includeDependents: argv[includeDependentsOption.name]
+                  })
+                : compiledGraph;
+            printCompiledGraph(outputGraph, outputType, argv[quietCompileOption.name]);
             if (compiledGraphHasErrors(compiledGraph)) {
               print("");
               printCompiledGraphErrors(compiledGraph.graphErrors, argv[quietCompileOption.name]);
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -133,6 +133,65 @@ const includeDependentsOption: INamedOption<yargs.Options> = {
   }
 };
 
+// `compile` reuses the same prune() filtering as run/build, but these flags only
+// filter the *printed output* -- the whole project still compiles. The `output-`
+// prefix makes that distinction explicit (see PR #2212).
+const outputActionsOption: INamedOption<yargs.Options> = {
+  name: "output-actions",
+  option: {
+    describe:
+      "A list of action names or patterns to filter the compiled output to. Can include '*' wildcards.",
+    type: "array",
+    coerce: (rawActions: string[] | null) => rawActions.map(actions => actions.split(",")).flat()
+  }
+};
+
+const outputTagsOption: INamedOption<yargs.Options> = {
+  name: "output-tags",
+  option: {
+    describe: "A list of tags to filter the compiled output to.",
+    type: "array",
+    coerce: (rawTags: string[] | null) => rawTags.map(tags => tags.split(",")).flat()
+  }
+};
+
+const outputIncludeDepsOption: INamedOption<yargs.Options> = {
+  name: "output-include-deps",
+  option: {
+    describe: "If set, dependencies of the selected actions are also included in the output.",
+    type: "boolean"
+  },
+  check: (argv: yargs.Arguments) => {
+    if (
+      argv[outputIncludeDepsOption.name] &&
+      !(argv[outputActionsOption.name] || argv[outputTagsOption.name])
+    ) {
+      throw new Error(
+        `The --${outputIncludeDepsOption.name} flag should only be supplied along with --${outputActionsOption.name} or --${outputTagsOption.name}.`
+      );
+    }
+  }
+};
+
+const outputIncludeDependentsOption: INamedOption<yargs.Options> = {
+  name: "output-include-dependents",
+  option: {
+    describe:
+      "If set, dependents (downstream) of the selected actions are also included in the output.",
+    type: "boolean"
+  },
+  check: (argv: yargs.Arguments) => {
+    if (
+      argv[outputIncludeDependentsOption.name] &&
+      !(argv[outputActionsOption.name] || argv[outputTagsOption.name])
+    ) {
+      throw new Error(
+        `The --${outputIncludeDependentsOption.name} flag should only be supplied along with --${outputActionsOption.name} or --${outputTagsOption.name}.`
+      );
+    }
+  }
+};
+
 const credentialsOption: INamedOption<yargs.Options> = {
   name: "credentials",
   option: {
@@ -393,10 +452,10 @@ export function runCli() {
           dotOutputOption,
           timeoutOption,
           quietCompileOption,
-          actionsOption,
-          tagsOption,
-          includeDepsOption,
-          includeDependentsOption,
+          outputActionsOption,
+          outputTagsOption,
+          outputIncludeDepsOption,
+          outputIncludeDependentsOption,
           {
             name: verboseOptionName,
             option: {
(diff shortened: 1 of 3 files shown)


kolina (changes requested):
I think it makes sense to add some prefix to these options for `compile` to make it more clear that these options are different from not compiling them. I'd suggest `output-{...}` format, feel free to suggest something else if you want.

andrebmallmann:
> I think it makes sense to add some prefix to these options for `compile` to make it more clear that these options are different from not compiling them. I'd suggest `output-{...}` format, feel free to suggest something else if you want.

Thanks for the review! Just to check I understood correctly: the concern is that for compile these flags work as filters rather than changing what actually gets compiled, so the unprefixed names could lead to confusion, is that it?

This version of the solution leans towards keeping parity with commands like run and build, and from a DX perspective I personally like that more. Do you think a clear distinction between filtering vs. compiling outweighs that consistency?

If we go the prefix way, which do you think is better  --output-* or --filter-*? "Filter" reads a bit closer to the actual behavior for me, but I'm happy with either.

And with a prefix, would --include-deps / --include-dependents become --output-include-deps / --output-include-dependents?

kolina:
Sorry for the late reply from my side.

> Thanks for the review! Just to check I understood correctly: the concern is that for compile these flags work as filters rather than changing what actually gets compiled, so the unprefixed names could lead to confusion, is that it?

Yes, it may lead to confusion. Also if we decide to implement actual filtering of compiled nodes, it may be hard to pick filters for them in the future.

> This version of the solution leans towards keeping parity with commands like run and build, and from a DX perspective I personally like that more. Do you think a clear distinction between filtering vs. compiling outweighs that consistency?

In my opinion, yes.

> If we go the prefix way, which do you think is better --output-* or --filter-*? "Filter" reads a bit closer to the actual behavior for me, but I'm happy with either.

`output-` may be more clear to me, but I don't have a strong opinion here.

> And with a prefix, would --include-deps / --include-dependents become --output-include-deps / --output-include-dependents?

Yeah, I think so.

andrebmallmann:
Hello @kolina

Sorry for late reply on my side, rough weeks. 

I've pushed the rename we agreed on the compile selection flags now carry the output- prefix to make it explicit that they filter the printed output rather than change what gets compiled.

run/build keep the unprefixed flags, so there's no change there. Tests updated accordingly, and I verified it end-to-end against examples/stackoverflow_reporter

One thing I ran into while testing that I'd like your take on. Since this reuses prune(), the filtering only applies to tables/operations/assertions — prune() leaves declarations and the flat targets manifest untouched. So compile --output-actions user_stats --json correctly returns just the user_stats table, but the output still lists all declarations and all targets from the full graph:

Here we might have two options:

Two options:
1. Leave as-is  to keep consistent with run/build, less code; the actual SQL is correctly isolated in tables.
2. Trim declarations/targets in the compile output path only (not touching shared prune()) for a tighter single-model output.

I think for this PR we could close with (1) and if the option 2 is good for you i can start work in another PR.

Compiled example using my version

```
$CLI compile examples/stackoverflow_reporter --output-actions user_stats --json
```

```json
{
    "tables": [
        {
            "type": "table",
            "target": {
                "schema": "reporting",
                "name": "user_stats",
                "database": "dataform-open-source"
            },
            "query": "\n\nselect\n  stg_users.user_id,\n  stg_users.age,\n  stg_users.creation_date,\n  stg_users.user_tenure,\n  count(distinct stg_badges.badge_id) as badge_count,\n  count(distinct posts_all.post_id) as questions_and_answer_count,\n  count(distinct if(type=\"question\", posts_all.post_id, null)) as question_count,\n  count(distinct if(type=\"answer\", posts_all.post_id, null)) as answer_count,\n  max(stg_badges.award_timestamp) as last_badge_received_at,\n  max(posts_all.created_at) as last_posted_at,\n  max(if(type=\"question\", posts_all.created_at, null)) as last_question_posted_at,\n  max(if(type=\"answer\", posts_all.created_at, null)) as last_answer_posted_at\nfrom\n  `dataform-open-source.staging.stg_users` as stg_users\n  left join `dataform-open-source.staging.stg_badges` as stg_badges\n    on stg_users.user_id = stg_badges.user_id\n  left join `dataform-open-source.reporting.posts_combined` as posts_all\n    on stg_users.user_id = posts_all.owner_user_id\ngroup by\n  1,2,3,4",
            "disabled": false,
            "fileName": "definitions/reporting/user_stats.sqlx",
            "tags": [
                "daily"
            ],
            "actionDescriptor": {
                "description": "Create a summary table for all users including statistics on questions, answeers and badges received."
            },
            "dependencyTargets": [
                {
                    "schema": "staging",
                    "name": "stg_users",
                    "database": "dataform-open-source"
                },
                {
                    "schema": "staging",
                    "name": "stg_badges",
                    "database": "dataform-open-source"
                },
                {
                    "schema": "reporting",
                    "name": "posts_combined",
                    "database": "dataform-open-source"
                }
            ],
            "hermeticity": "NON_HERMETIC",
            "canonicalTarget": {
                "schema": "reporting",
                "name": "user_stats",
                "database": "dataform-open-source"
            },
            "enumType": "TABLE"
        }
    ],
    "operations": [],
    "assertions": [],
    "declarations": [
        {
            "target": {
                "schema": "stackoverflow",
                "name": "badges",
                "database": "bigquery-public-data"
            },
            "actionDescriptor": {
                "description": "raw badges table"
            },
            "fileName": "definitions/sources/badges.sqlx",
            "canonicalTarget": {
                "schema": "stackoverflow",
                "name": "badges",
                "database": "bigquery-public-data"
            }
        },
        {
            "target": {
                "schema": "stackoverflow",
                "name": "posts_answers",
                "database": "bigquery-public-data"
            },
            "actionDescriptor": {
                "description": "raw posts_answers table"
            },
            "fileName": "definitions/sources/posts_answers.sqlx",
            "canonicalTarget": {
                "schema": "stackoverflow",
                "name": "posts_answers",
                "database": "bigquery-public-data"
            }
        },
        {
            "target": {
                "schema": "stackoverflow",
                "name": "posts_questions",
                "database": "bigquery-public-data"
            },
            "actionDescriptor": {
                "description": "raw posts_questions table"
            },
            "fileName": "definitions/sources/posts_questions.sqlx",
            "canonicalTarget": {
                "schema": "stackoverflow",
                "name": "posts_questions",
                "database": "bigquery-public-data"
            }
        },
        {
            "target": {
                "schema": "stackoverflow",
                "name": "users",
                "database": "bigquery-public-data"
            },
            "actionDescriptor": {
                "description": "raw users table"
            },
            "fileName": "definitions/sources/users.sqlx",
            "canonicalTarget": {
                "schema": "stackoverflow",
                "name": "users",
                "database": "bigquery-public-data"
            }
        }
    ],
    "tests": [],
    "notebooks": [],
    "dataPreparations": [],
    "targets": [
        {
            "schema": "dataform_stackoverflow_reporter_example_assertions",
            "name": "reporting_posts_combined_assertions_uniqueKey_0",
            "database": "dataform-open-source"
        },
        {
            "schema": "reporting",
            "name": "posts_combined",
            "database": "dataform-open-source"
        },
        {
            "schema": "dataform_stackoverflow_reporter_example_assertions",
            "name": "reporting_user_stats_assertions_uniqueKey_0",
            "database": "dataform-open-source"
        },
        {
            "schema": "dataform_stackoverflow_reporter_example_assertions",
            "name": "reporting_user_stats_assertions_rowConditions",
            "database": "dataform-open-source"
        },
        {
            "schema": "reporting",
            "name": "user_stats",
            "database": "dataform-open-source"
        },
        {
            "schema": "stackoverflow",
            "name": "badges",
            "database": "bigquery-public-data"
        },
        {
            "schema": "stackoverflow",
            "name": "posts_answers",
            "database": "bigquery-public-data"
        },
        {
            "schema": "stackoverflow",
            "name": "posts_questions",
            "database": "bigquery-public-data"
        },
        {
            "schema": "stackoverflow",
            "name": "users",
            "database": "bigquery-public-data"
        },
        {
            "schema": "staging",
            "name": "stg_badges",
            "database": "dataform-open-source"
        },
        {
            "schema": "staging",
            "name": "stg_posts_answers",
            "database": "dataform-open-source"
        },
        {
            "schema": "staging",
            "name": "stg_posts_questions",
            "database": "dataform-open-source"
        },
        {
            "schema": "staging",
            "name": "stg_users",
            "database": "dataform-open-source"
        }
    ],
    "projectConfig": {
        "warehouse": "bigquery",
        "defaultSchema": "dataform_stackoverflow_reporter_example",
        "assertionSchema": "dataform_stackoverflow_reporter_example_assertions",
        "defaultDatabase": "dataform-open-source",
        "defaultLocation": "us"
    },
    "graphErrors": {},
    "dataformCoreVersion": "3.0.59",
    "jitData": {}
}
```

cli/index.ts
```diff
@@ -133,6 +133,65 @@ const includeDependentsOption: INamedOption<yargs.Options> = {
   }
 };
 
+// `compile` reuses the same prune() filtering as run/build, but these flags only
```
kolina, line 136: it'd be great if we can reuse `coerce` or `check` functions to not duplicate logic in 2 places

kolina (changes requested):
> One thing I ran into while testing that I'd like your take on. Since this reuses prune(), the filtering only applies to tables/operations/assertions — prune() leaves declarations and the flat targets manifest untouched. So compile --output-actions user_stats --json correctly returns just the user_stats table, but the output still lists all declarations and all targets from the full graph:

> Here we might have two options:

> Two options:

> Leave as-is to keep consistent with run/build, less code; the actual SQL is correctly isolated in tables.
Trim declarations/targets in the compile output path only (not touching shared prune()) for a tighter single-model output.
I think for this PR we could close with (1) and if the option 2 is good for you i can start work in another PR.

Yeah, option 1 is fine with me!

cli/index.ts
alandrebraga: Done! splitCommas for the coerce, now shared by actions, tags, output-actions and output-tags.
requiresSelection for the check, now shared by include-deps, include-dependents, output-include-deps and output-include-dependent

andrebmallmann:
@kolina 

One thing i hit during the review of coerce and i want to make it clear. 

The JSON I posted earlier is misleading. With --output-actions user_stats the tables list is
correctly just the one table, but targets still listed all 13 actions. prune() spreads the rest
of the graph through untouched, so targets was never filtered. Harmless for run and build since
nothing reads it, but compile prints it. Fixed in 52c2021a, tables and targets now agree.

kolina (approved):

PR #2214, opened 2026-06-29
Escape newlines in sqlString so multiline rowConditions produce valid SQL

### LINKED ISSUE #2201: Multiline `rowConditions` entries lead to invalid SQL
Given the following config:

```
config {
    type: 'table',
    assertions: {
        rowConditions: [
            `validCheck`,
            `
            invalidCheck
            `,
        ]
    }
}
```

Dataform compiles assertions into:

```sql
SELECT
  'validCheck' AS failing_row_condition,
  *
FROM `project.dataset.test`
WHERE NOT (validCheck)
UNION ALL
SELECT
  '
            invalidCheck
            ' AS failing_row_condition,
  *
FROM `project.dataset.test`
WHERE NOT (
            invalidCheck
            )
```

The generated SQL is invalid because BigQuery single-quoted string literals cannot span multiple lines. As a result, any multiline rowConditions entry causes a syntax error during execution.

While this is a somewhat synthetic example, more complex expressions can be easier to read and maintain when formatted across multiple lines.

As a workaround, I currently normalize whitespace before passing the condition:

```
config {
    type: 'table',
    assertions: {
        rowConditions: [
            `validCheck`,
            `
            invalidCheck
            `.replace(/\s+/g, " "),
        ]
    }
}
```

**Possible solutions:**

1. Generate triple-quoted string literals
  ```sql
  SELECT
  '''
            invalidCheck
            ''' AS failing_row_condition
  ```
2. Escape or normalize newlines before embedding
  ```sql
  SELECT
  '\n            invalidCheck\n            ' AS failing_row_condition
  ```

ihistand:
Fixes #2201.

## Problem

A multiline `rowConditions` entry compiles to invalid SQL:

```
config {
  type: "table",
  assertions: {
    rowConditions: [`
      invalidCheck
    `]
  }
}
```

`rowConditionsAssertion` embeds the condition twice: once raw inside `WHERE NOT (...)` (fine — multiline is valid there), and once as a string-literal label via `sqlString(...)`. `sqlString` only escaped backslashes and single quotes, so a raw newline survived into the single-quoted `failing_row_condition` literal — which BigQuery rejects, since single-quoted literals can't span multiple lines.

## Fix

The bug belongs in `sqlString`, not the assertion template — any caller that quotes a multiline string hits it. `sqlString` now also escapes newlines/carriage-returns, turning a raw newline into the two-char `\n` escape (which parses back to a newline, keeping the literal single-line):

```ts
return `'${stringContents
  .replace(/\\/g, "\\\\")
  .replace(/'/g, "\\'")
  .replace(/\n/g, "\\n")
  .replace(/\r/g, "\\r")}'`;
```

Ordering matters: backslash-doubling runs first, otherwise the `\n`/`\r` introduced here would themselves get re-escaped into `\\n`/`\\r` and print a literal backslash-n.

The raw `WHERE NOT (...)` clause is untouched — the multiline expression stays as-authored; only the label literal is normalized:

```sql
SELECT
  'a > 0\n  AND b > 0' AS failing_row_condition,
  *
FROM `project.dataset.test`
WHERE NOT (a > 0
  AND b > 0)
```

**Why escape over triple-quoting** (the issue's other suggestion): escaping reuses the quote-escaping path already in `sqlString`, so one code path handles quotes, backslashes, and newlines uniformly. Triple-quoting would need its own rules for embedded `'''`.

## Test

Added a regression test in `core/actions/assertion_test.ts` that compiles a table with a multiline `rowConditions` entry and asserts the `failing_row_condition` literal stays single-line (escaped `\n`) while the `WHERE NOT (...)` clause keeps the raw, multiline expression.

This was raised in #2201 and a maintainer invited a PR. Happy to adjust style or placement.

### DIFF
--- a/core/actions/assertion_test.ts
+++ b/core/actions/assertion_test.ts
@@ -306,6 +306,38 @@ actions:
     );
   });
 
+  test(`multiline rowConditions compile to single-line failing_row_condition literals`, () => {
+    // Regression test for #2201: a multiline rowConditions entry must not produce a
+    // multiline single-quoted string literal, which BigQuery rejects.
+    const projectDir = tmpDirFixture.createNewTmpDir();
+    fs.writeFileSync(path.join(projectDir, "workflow_settings.yaml"), VALID_WORKFLOW_SETTINGS_YAML);
+    fs.mkdirSync(path.join(projectDir, "definitions"));
+    fs.writeFileSync(
+      path.join(projectDir, "definitions/test.sqlx"),
+      `config {
+        type: "table",
+        assertions: {
+          rowConditions: ["a > 0\n  AND b > 0"]
+        }
+      }
+      SELECT 1 AS a, 1 AS b`
+    );
+
+    const result = runMainInVm(coreExecutionRequestFromPath(projectDir));
+
+    expect(result.compile.compiledGraph.graphErrors.compilationErrors).deep.equals([]);
+    const rowConditionsAssertion = result.compile.compiledGraph.assertions.find(
+      assertion => assertion.target.name === "defaultDataset_test_assertions_rowConditions"
+    );
+    // The failing_row_condition string literal has its newline escaped, so the
+    // single-quoted literal stays on one line...
+    expect(rowConditionsAssertion.query).contains(
+      "'a > 0\\n  AND b > 0' AS failing_row_condition"
+    );
+    // ...while the WHERE NOT (...) clause keeps the raw, multiline expression.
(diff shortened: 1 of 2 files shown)


kolina (approved):
LGTM

The tests are failing due to expired BQ credentials, can you please merge https://github.com/dataform-co/dataform/commit/3478e82a73aae1fc52361fb67e88f937018a58af into your PR?

ihistand:
Sorry @kolina , I didn't realize that my fork was required to keep around for the PR and mistakenly deleted it. I've resubmitted at #2226

PR #2220, opened 2026-07-15
Preserve Governance Controls in Dataform.

mohamed-saad:
Adding flags to enable preserving governance controls such as Data Policies, and Data Governance Tags during the tables creations.

### DIFF
--- a/protos/core.proto
+++ b/protos/core.proto
@@ -38,6 +38,8 @@ message ProjectConfig {
 
   bool include_tests_in_compiled_graph = 22;
 
+  bool preserve_governance_controls = 24;
+
   reserved 3, 4, 6, 8, 10, 12, 13;
 }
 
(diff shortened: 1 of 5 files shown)


version.bzl
```diff
@@ -1 +1 @@
-DF_VERSION = "3.0.61"
+DF_VERSION = "3.0.62"
```
mfiglus: please move this change to a separate dedicated PR

protos/core.proto
```diff
@@ -111,6 +111,7 @@ message BigQueryOptions {
   FileFormat file_format = 10;
   string storage_uri = 11;
   repeated string incremental_predicates = 12;
+  bool preserve_governance_controls = 13;
```
mfiglus, line 116: is this meant to be available on the project level (workflow_settings.yaml)? if it is then I believe you should add this option to ProjectConfig https://github.com/dataform-co/dataform/blob/main/protos/core.proto#L15 and workflow_settings.ts https://github.com/dataform-co/dataform/blob/main/core/workflow_settings.ts#L98

mohamed-saad: Done

version.bzl
mohamed-saad: Done

mfiglus (approved):

PR #2221, opened 2026-07-15
Fix JiT assertion compilation to accept object result shape

rafal-hawrylak:
Assertion JiT code returning an object (rather than a bare string) had the whole object stored into JitAssertionResult.query — a string field — producing broken read-back and downstream encoding errors.

Mirror makeJitTableResult: unwrap .query from an object, keep the bare-string path unchanged. Widen JitAssertionResult to `string | { query: string }` to match.

Regression tests cover the bare-object shape and an object with additional (ignored) fields.

Integration of JiT assertions will come in new PR after https://github.com/dataform-co/dataform/pull/2211 lands.

### DIFF
--- a/core/jit_compiler.ts
+++ b/core/jit_compiler.ts
@@ -73,7 +73,9 @@ function jitCompileAssertion(
   const jctx: JitContext<IActionContext> = new SqlActionJitContext(
     adapter, request,
   );
-  return mainBody(jctx).then(query => dataform.JitAssertionResult.create({ query }));
+  return mainBody(jctx).then(result =>
+    typeof result === "string" ? { query: result } : result
+  );
 }
 
 function jitCompileIncrementalTable(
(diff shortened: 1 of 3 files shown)


core/jit_compiler.ts
```diff
@@ -29,6 +29,11 @@ function makeJitTableResult(result: JitTableResult): dataform.IJitTableResult {
   return dataform.JitTableResult.create(jitResult);
 }
 
+function makeJitAssertionResult(result: JitAssertionResult): dataform.IJitAssertionResult {
+  const query = typeof result === "string" ? result : result.query;
```
ikholopov-omni:
```
const assertionResult: IJitAssertionResult = typeof result === "string" ? { query: result } : result;
```
?

rafal-hawrylak: Ok, wrapped to oneliner again.

ikholopov-omni (approved):

apilaskowski (approved):

PR #2223, opened 2026-07-15
Assertion JiT integration

rafal-hawrylak:
Assertion actions were not integrated into the JiT compilation runtime, even though the compiler (core/jit_compiler.ts) already supports them. Two gaps in Runner:

- compileJitAction had no branch for action.type === "assertion", so assertions fell through to JIT_COMPILATION_TARGET_TYPE_UNSPECIFIED and the compiler rejected the request.
- createTasksFromJitResponse had no branch for jitResponse.assertion, so even a successful compile produced zero tasks and the action silently did nothing.

This PR wires both - assertions with jitCode now compile through the JiT worker and produce the same task set as their AoT counterparts (via executionSql.createAssertionTasks).

### DIFF
--- a/cli/BUILD
+++ b/cli/BUILD
@@ -65,8 +65,13 @@ ts_test_suite(
         "index_project_test.ts",
         "index_compile_test.ts",
         "index_run_e2e_test.ts",
+        "tests/jit/index_jit_main_test.ts",
+        "tests/jit/index_jit_advanced_test.ts",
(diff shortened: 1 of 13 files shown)


ikholopov-omni:
Is this PR rebased onto any other PRs? 

If yes - please send to review after those are merged.

ikholopov-omni (approved):

PR #1349, opened 2022-07-07
Add ignore js flag to format subcommand

andres-lowrie:
Hello

I figured this option may come in handy for users who have different linting/formatting rules for their javascript files but still want to leverage the `sqlx` formatter. (Which is the case for us)

I took a look at `/tests/sqlx/format.spec.ts` and the other files in general to see if there was somewhere I could add tests to but didn't find anything obvious. The file that I'm mentioning seems to specifically test the formatting output and not how the cli calls it. If we need to include tests for this addition, please point me in the right direction and I'll do the needful

PR #2227, opened 2026-07-18
Add ignore js flag to format subcommand

andres-lowrie:
Updated for latest changes (from #1349) 

Allows users to only format `.sqlx` files in case they use different formatters for js files.

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -225,6 +225,15 @@ const icebergOption: INamedOption<yargs.Options> = {
   },
 };
 
+const fmtIgnoreJsOptionName: INamedOption<yargs.Options> = {
+  name: "ignore-js-files",
+  option: {
+    describe: "If set, the formatter will not consider javascript files (.js)",
+    type: "boolean",
+    default: false,
+  },
+};
+
 const testConnectionOptionName = "test-connection";
 
 const watchOptionName = "watch";
@@ -718,7 +727,8 @@ export function runCli() {
           }
         ],
         processFn: async argv => {
-          let actions = ["{definitions,includes}/**/*.{js,sqlx}"];
+          const extensions = argv[fmtIgnoreJsOptionName.name] ? "*.sqlx" : "*.{js,sqlx}"
+          let actions = [`{definitions,includes}/**/${extensions}`];
           if (actionsOption.name in argv && argv[actionsOption.name].length > 0) {
             actions = argv[actionsOption.name];
           }
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -225,6 +225,15 @@ const icebergOption: INamedOption<yargs.Options> = {
   },
 };
 
+const fmtIgnoreJsOption: INamedOption<yargs.Options> = {
+  name: "ignore-js-files",
+  option: {
+    describe: "If set, the formatter will not consider javascript files (.js)",
+    type: "boolean",
+    default: false,
+  },
+};
+
 const testConnectionOptionName = "test-connection";
 
 const watchOptionName = "watch";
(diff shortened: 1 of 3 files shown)


cli/index.ts
apilaskowski, line 718: I think we should add `fmtIgnoreJsOptionName` to the options list for the format subcommand builder.

cli/index.ts
```diff
@@ -718,7 +727,8 @@ export function runCli() {
           }
         ],
         processFn: async argv => {
-          let actions = ["{definitions,includes}/**/*.{js,sqlx}"];
+          const extensions = argv[fmtIgnoreJsOptionName.name] ? "*.sqlx" : "*.{js,sqlx}"
```
apilaskowski, line 730: Missing semicolon?

cli/index.ts
```diff
@@ -225,6 +225,15 @@ const icebergOption: INamedOption<yargs.Options> = {
   },
 };
 
+const fmtIgnoreJsOptionName: INamedOption<yargs.Options> = {
```
apilaskowski, line 228: Lets drop Name suffix for consistency. Options generally have option suffix, while name is generally used for strings.

apilaskowski (changes requested):
Lets also add missing tests.

1. dataform format --help includes --ignore-js-files.
2. running dataform format --ignore-js-files formats .sqlx files without modifying .js files.

andres-lowrie:
> Lets also add missing tests.
> 
>     1. dataform format --help includes --ignore-js-files.
> 
>     2. running dataform format --ignore-js-files formats .sqlx files without modifying .js files.

hey so I'm not having any luck using `bazel` to show me the failing tests, for example if I add the following to `cli/index_help_test.ts`

```javascript
   test("shows help for 'format' command", async () => {
    const result = await getProcessResult(execFile(nodePath, [cliEntryPointPath, "help", "format"]));
    expect(result.exitCode).equals(0);
    const output = result.stdout;
    expect(output).to.include("Format the dataform project's files.");
    expect(output).to.include("--check");
    expect(output).to.include("Check if files are formatted correctly without modifying them.");
    expect(output).to.include("--actions");`
    expect(output).to.include("XXXXXXYYYYZZZZ");
    //                          ^
    //                          |
    //                          +--- should fail since this won't be included in the output
  });
```

and then I do 
```
`bazel` test ... --test_tag_filters=-integration //cli
```

I get back a whole bunch of build failures but in particular I get back these lines 

```
//cli:index_compile_test                                              NO STATUS
//cli:index_help_test                                                 NO STATUS
//cli:index_init_test                                                 NO STATUS
//cli:index_project_test                                              NO STATUS
//cli:index_run_e2e_test                                              NO STATUS
```

I would expect `//cli:index_help_test` to fail and the others to pass

:sweat_smile: I don't know anything about bazel so I'm just grepping and praying, reading the BUILD files in confusion and hoping for the best here 

the `contributing.md`  has `bazel test //cli:index_test` but that comes back with 

```
ERROR: no such target '//cli:index_test': target 'index_test' not declared in package 'cli' 
```

so I'm guessing bazel stuff changed but the docs didn't?

kolina:
It actually shows you the targets you can run, for example

```
bazel test //cli:index_help_test
```

or just

```
bazel test //cli:all
```

andres-lowrie:
okay let me give that a go when I get a moment

andres-lowrie:
Okay I think I got the tests setup right, I'm also adding some screenshots if that help at all. 

lmk if I missed anything.

**before**
before-fmt-run


**after**
after-fmt-run

apilaskowski (approved):

PR #2234, opened 2026-07-28
Support '*' wildcards in --actions selectors as documented (#2224)

### LINKED ISSUE #2224: --actions flag does not support * wildcards despite CLI help text claiming it does
The dataform run help text for --actions states:

▎ A list of action names or patterns to run. Can include '*' wildcards.

In practice, no wildcard pattern is matched — not even a bare *.

Steps to reproduce:

    dataform run --dry-run --actions "*"
    dataform run --dry-run --actions "mrd*"
    dataform run --dry-run --actions "*mrd_features*"
    
Expected: matches every/any action whose name contains the pattern.

Actual: Compiled successfully. followed by `No actions to run.` for every wildcard pattern tried, even though the target actions clearly exist in the compiled graph. Only an exact unqualified action name (e.g. --actions "mrd_features_inference") selects anything.

Root cause: `matchPatterns` in `core/utils.ts` does plain string equality only:

```ts
function matchPatterns(patterns: string[], values: string[]) {
  const fullyQualifiedActions: string[] = [];
  patterns.forEach(pattern => {
    if (pattern.includes(".")) {
      if (values.includes(pattern)) fullyQualifiedActions.push(pattern);
    } else {
      const matchingActions = values.filter(value => pattern === value.split(".").slice(-1)[0]);
      ...
    }
  });
  return fullyQualifiedActions;
}
```

There is no glob/regex handling of * anywhere in this function, so the documented wildcard behavior never executes.

Suggested fix: either implement glob matching in matchPatterns (e.g. via minimatch) to match the documented behavior, or update the `--actions` help text in `cli/index.ts` to remove the wildcard claim.

ihistand:
Fixes #2224 — following up on https://github.com/dataform-co/dataform/issues/2224#issuecomment-5079116804 ("Feel free to send a PR :)").

## Problem

The `--actions` help text for `run` and `compile` says patterns "can include '*' wildcards", but `matchPatterns` in `core/utils.ts` does plain string equality only, so no wildcard pattern ever matches — `dataform run --actions "*"` reports `No actions to run.` even when the compiled graph has actions.

## Fix

Wildcard patterns are compiled to an anchored `RegExp`:

- every non-`*` character matches literally (regex metacharacters are escaped, so `.` stays a literal dot);
- each `*` matches any run of characters (`mrd*` → `/^mrd.*$/`, `*features*` → `/^.*features.*$/`).

Scoping mirrors the existing exact-match branches: a pattern containing `.` matches against the fully-qualified action name; otherwise it matches against the unqualified last segment. Wildcard matches bypass the ambiguous-name error since selecting many actions is the intent. Exact (non-wildcard) selection behavior is completely unchanged.

## Tests

Adds a `matchPatterns` suite to `core/utils_test.ts` (the function was previously untested) covering: exact unqualified/qualified selection, the ambiguity error, bare `*`, prefix and substring wildcards, qualified wildcards (`schema.*`), no-match returning empty, and the literal-dot escaping edge case.

Understood that GCP Dataform's hosted actions filter doesn't use this implementation — this change only brings the open-source CLI in line with its own documented behavior.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0171FwKo8gRQQ35VoYDtSHNU

### DIFF
--- a/core/utils.ts
+++ b/core/utils.ts
@@ -26,13 +26,25 @@ type actionsWithDependencies =
 export const nativeRequire =
   typeof __webpack_require__ === "function" ? __non_webpack_require__ : require;
 
-// Turn a selector pattern containing '*' wildcards into an anchored RegExp.
-// Every character except '*' is matched literally (regex metacharacters are
-// escaped); each '*' matches any run of characters, so "mrd*" -> /^mrd.*$/ and
+// Turns a selector pattern into an anchored RegExp in which "*" is a wildcard and
+// everything else is literal text, e.g. "mrd*" -> /^mrd.*$/ and
 // "*features*" -> /^.*features.*$/.
+//
+// Splitting on "*" first is what keeps the rest simple: every "*" in a pattern is a
+// wildcard by definition, so the pieces between them are pure literal text and can be
+// escaped wholesale, then rejoined with ".*".
+//
+// The escape is needed because an action name is not regex-safe. Names are
+// dot-separated ("project.dataset.name"), and "." in a regex matches any character, so
+// without escaping "schema.*" would also select "schemaXtable" - see the "literal dot"
+// case in utils_test.ts. The set below is the usual list of JavaScript regex
+// metacharacters with one deliberate omission: "*", which is left out because the split
+// above has already consumed every "*", so none can reach here. ("]" and "\" carry
+// backslashes for the character class's own syntax; "-" and "/" are not metacharacters
+// outside a class and so need no escaping.)
 function globToRegExp(pattern: string): RegExp {
-  const escaped = pattern.replace(/[.*+?^${}()|[\]\\]/g, "\\$&").replace(/\\\*/g, ".*");
-  return new RegExp(`^${escaped}$`);
+  const escapeLiteral = (literal: string) => literal.replace(/[.+?^${}()|[\]\\]/g, "\\$&");
+  return new RegExp(`^${pattern.split("*").map(escapeLiteral).join(".*")}$`);
 }
 
 export function matchPatterns(patterns: string[], values: string[]) {
--- a/core/utils.ts
+++ b/core/utils.ts
@@ -26,10 +26,33 @@ type actionsWithDependencies =
 export const nativeRequire =
   typeof __webpack_require__ === "function" ? __non_webpack_require__ : require;
 
+// Turn a selector pattern containing '*' wildcards into an anchored RegExp.
+// Every character except '*' is matched literally (regex metacharacters are
+// escaped); each '*' matches any run of characters, so "mrd*" -> /^mrd.*$/ and
+// "*features*" -> /^.*features.*$/.
+function globToRegExp(pattern: string): RegExp {
+  const escaped = pattern.replace(/[.*+?^${}()|[\]\\]/g, "\\$&").replace(/\\\*/g, ".*");
+  return new RegExp(`^${escaped}$`);
+}
+
 export function matchPatterns(patterns: string[], values: string[]) {
   const fullyQualifiedActions: string[] = [];
   patterns.forEach(pattern => {
-    if (pattern.includes(".")) {
+    if (pattern.includes("*")) {
+      // Wildcard selector. A pattern that contains "." matches against the
+      // fully-qualified action name; otherwise it matches against the unqualified
+      // name (last segment), mirroring the exact-match branches below. Wildcards
+      // are expected to select many actions, so no ambiguity error applies here.
+      const regExp = globToRegExp(pattern);
+      const scope = pattern.includes(".")
+        ? values
+        : values.map(value => value.split(".").slice(-1)[0]);
+      values.forEach((value, i) => {
+        if (regExp.test(scope[i])) {
+          fullyQualifiedActions.push(value);
+        }
+      });
+    } else if (pattern.includes(".")) {
       if (values.includes(pattern)) {
         fullyQualifiedActions.push(pattern);
       }
--- a/core/utils_test.ts
+++ b/core/utils_test.ts
@@ -6,6 +6,7 @@ import {
   getEffectiveTableFolderSubpath,
   getFileFormatValueForIcebergTable,
   getStorageUriForIcebergTable,
+  matchPatterns,
   validateConnectionFormat,
   validateNoMixedCompilationMode,
   validateStorageUriFormat,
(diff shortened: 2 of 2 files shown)


apilaskowski:
PTAL at following error: `ERROR: /workspace/core/utils.ts:35:10 - Found non-literal argument to RegExp Constructor`

core/utils.ts
```diff
@@ -26,10 +26,33 @@ type actionsWithDependencies =
 export const nativeRequire =
   typeof __webpack_require__ === "function" ? __non_webpack_require__ : require;
 
+// Turn a selector pattern containing '*' wildcards into an anchored RegExp.
+// Every character except '*' is matched literally (regex metacharacters are
+// escaped); each '*' matches any run of characters, so "mrd*" -> /^mrd.*$/ and
+// "*features*" -> /^.*features.*$/.
+function globToRegExp(pattern: string): RegExp {
+  const escaped = pattern.replace(/[.*+?^${}()|[\]\\]/g, "\\$&").replace(/\\\*/g, ".*");
```
kolina, line 34: Can you simplify this logic?

core/utils.ts
```diff
@@ -26,10 +26,33 @@ type actionsWithDependencies =
...
 export function matchPatterns(patterns: string[], values: string[]) {
   const fullyQualifiedActions: string[] = [];
   patterns.forEach(pattern => {
-    if (pattern.includes(".")) {
+    if (pattern.includes("*")) {
+      // Wildcard selector. A pattern that contains "." matches against the
+      // fully-qualified action name; otherwise it matches against the unqualified
+      // name (last segment), mirroring the exact-match branches below. Wildcards
+      // are expected to select many actions, so no ambiguity error applies here.
+      const regExp = globToRegExp(pattern);
```
kolina, line 46: Shouldn't you first to split by components and then apply regexes?

kolina (changes requested):

core/utils.ts
ihistand: Agreed, that was doing too much at once — escaping every metacharacter including `*`, then
un-escaping `\*` back into `.*`. You had to read both replaces together to see what it did.

Rewritten to split on the wildcards, escape the literal parts, and rejoin:
```ts
function globToRegExp(pattern: string): RegExp {
  const escapeLiteral = (literal: string) => literal.replace(/[.+?^${}()|[\]\\]/g, "\\$&");
  return new RegExp(`^${pattern.split("*").map(escapeLiteral).join(".*")}$`);
}
```
Behaviour is unchanged — the existing `matchPatterns` tests pass, and I diffed the two
implementations exhaustively over patterns covering every regex metacharacter plus `**` and  `a**b`.

core/utils.ts
ihistand: > Happy to switch if that's the semantics you want, but it isn't behaviour-neutral, so I'd rather
> confirm the intent than guess.
>
> I implemented component-wise matching (split pattern and value on `.`, require the same number of
> parts, apply one regex per part) and diffed it against the current implementation. One case
> changes: **whether `*` may cross a dot.**
>
> The values here are `targetAsReadableString` output — `project.dataset.name`, or `dataset.name`
> when `defaultProject` isn't set. On the usual three-part name:
>
> | pattern | current | component-wise |
> |---|---|---|
> | `*`, `orders`, `orders*` | same | same |
> | `*.dataset.orders`, `*.dataset.*` | same | same |
> | **`*.orders`** | every `…orders` action | **nothing** |
>
> Component-wise is the conventional glob rule (`*` stops at the separator, like shell `*` and `/`),
> and I'm not against it. The one thing that gives me pause is that `--actions "*.orders"` reads like
> "the orders table in whichever dataset", and component-wise it selects nothing on a three-part
> name — the user has to know to write `*.*.orders`. Since wildcards deliberately don't raise the
> no-match/ambiguity error, that failure is silent. Letting `*` span dots avoids it, at the cost of
> being less strict.
>
> Which would you prefer? If component-wise, I'll push it with tests pinning the part-count
> behaviour explicitly.
>
> (Probably out of scope here, but worth separating out: `dataset.*` matches nothing under *either*
> scheme on a three-part name, because a pattern containing `.` is matched against the fully-qualified
> name. Exact `dataset.orders` behaves the same way today, so the wildcard branch is at least
> consistent with the existing rule — happy to look at that separately if it's worth changing.)

core/utils.ts
ihistand: (note that this change isn't pushed yet, awaiting response)

kolina: sounds ok, but let's comment the choice of characters and why first replace is needed

core/utils.ts
kolina: ok, after your explanation I'm ok with matching the whole target without splitting by components

kolina (changes requested):
Let's fix issues from https://github.com/dataform-co/dataform/pull/2234#issuecomment-5101832217

core/utils.ts
ihistand: // Turns a selector pattern into an anchored RegExp in which "*" is a wildcard and
// everything else is literal text, e.g. "mrd*" -> /^mrd.*$/ and
// "*features*" -> /^.*features.*$/.
//
// Splitting on "*" first is what keeps the rest simple: every "*" in a pattern is a
// wildcard by definition, so the pieces between them are pure literal text and can be
// escaped wholesale, then rejoined with ".*".
//
// The escape is needed because an action name is not regex-safe. Names are
// dot-separated ("project.dataset.name"), and "." in a regex matches any character, so
// without escaping "schema.*" would also select "schemaXtable" — see the "literal dot"
// case in utils_test.ts. The set below is the usual list of JavaScript regex
// metacharacters with one deliberate omission: "*", which is left out because the split
// above has already consumed every "*", so none can reach here. ("]" and "\" carry
// backslashes for the character class's own syntax; "-" and "/" are not metacharacters
// outside a class and so need no escaping.)
function globToRegExp(pattern: string): RegExp {
  const escapeLiteral = (literal: string) => literal.replace(/[.+?^${}()|[\]\\]/g, "\\$&");
  return new RegExp(`^${pattern.split("*").map(escapeLiteral).join(".*")}$`);
}

kolina (changes requested):
I don't have more comments, but https://github.com/dataform-co/dataform/pull/2234#issuecomment-5101832217 still needs to be fixed

PR #2235, opened 2026-07-28
feat(core): add lineage.enabled to WorkflowSettings and ProjectConfig

rafal-hawrylak:
Introduces `LineageSettings { bool enabled }` and threads it through `WorkflowSettings` (yaml opt-in) into `ProjectConfig` (compiled graph) so downstream Dataform runners can decide whether to emit OpenLineage events to Knowledge Catalog (datalineage.googleapis.com) after each action completes.

Config plumbing only — no emission logic is added in this change.

### DIFF
--- a/protos/configs.proto
+++ b/protos/configs.proto
@@ -71,6 +71,16 @@ message WorkflowSettings {
   // Optional. If set to true, governance controls (e.g. Data Policies) will be
   // preserved on BigQuery tables.
   bool preserve_governance_controls = 18;
+
+  // Optional. Configuration for Knowledge Catalog lineage emission.
+  LineageSettings lineage = 19;
+}
+
+// Configuration for Knowledge Catalog lineage emission.
+message LineageSettings {
+  // If true, downstream runners will emit OpenLineage events to Knowledge
+  // Catalog for actions executed from this project.
+  bool enabled = 1;
 }
 
 // Action configs defines the contents of `actions.yaml` configuration files.
--- a/protos/core.proto
+++ b/protos/core.proto
@@ -40,6 +40,8 @@ message ProjectConfig {
 
   bool preserve_governance_controls = 24;
 
+  optional bool lineage_enabled = 25;
+
   reserved 3, 4, 6, 8, 10, 12, 13;
 }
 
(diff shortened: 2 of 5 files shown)


protos/core.proto
SuchodolskiEdvin, line 5: Shouldn't we add 'import "configs.proto";' at the top of file? Standard `protoc` compilers might require it, just like 'extension.proto' is imported.

SuchodolskiEdvin (changes requested):

protos/core.proto
rafal-hawrylak: Moved LineageSettings into core.proto instead - to match DefaultIcebergConfig (#2104).

rafal-hawrylak: Also checked protoc and indeed it requires the import - good catch - added it.

protos/configs.proto
```diff
@@ -10,6 +10,7 @@ option go_package = "github.com/dataform-co/dataform/protos/dataform";
 
 import "google/protobuf/struct.proto";
 import "extension.proto";
+import "core.proto";
```
kolina: Let's not introduce dependencies between core and config protos, we just got rid of them recently :)

kolina (changes requested):

SuchodolskiEdvin (approved):

protos/configs.proto
rafal-hawrylak: Ok, kept it separated. I did not know this rule.

protos/core.proto
```diff
@@ -40,6 +40,8 @@ message ProjectConfig {
 
   bool preserve_governance_controls = 24;
 
+  bool lineage_enabled = 25;
```
kolina: I think this is one of use cases where we actually may need to have `optional` here if we want to differentiate between this field not set and being set to `false`

Otherwise it may complicate future changes of default behaviour

kolina (changes requested):

protos/core.proto
rafal-hawrylak: That is a good catch! Changed.

kolina (approved):

PR #2236, opened 2026-07-31
Fix incremental table compilation failing on metadata.extraProperties

fuzmish:
## Example

This fails to compile:
```js
config {
  type: "incremental",
  onSchemaChange: "IGNORE",
  metadata: {
    extraProperties: {
      myField: "example"
    }
  }
}

SELECT 1
// -> Unexpected property "myField", or property value type of "string" is incorrect.
```

## Cause

verifyObjectMatchesProto(protoType, object) in common/protos/index.ts:

```js
const proto = protoType.create(object);          // shallow copy, no type conversion
const protoCastObject = protoType.toObject(proto);
checkFields(object, protoCastObject);             // throws on typeof mismatch at any key
```

`create()` copies extraProperties as-is, without converting it to the Struct format.
`toObject()` then can't read it as a Struct, so `protoCastObject` ends up with an empty extraProperties.
`checkFields` finds `myField` present in the raw object but missing from the round-tripped one, and throws.

## Fix

Convert metadata.extraProperties into the Struct wire format in IncrementalTable.verifyConfig(), before calling verifyObjectMatchesProto, so both sides of the comparison already have matching shapes.
The conversion logic already existed; it is extracted into a named, exported function and incremental_table.ts call it directly.

## Test

- Added a test in core/actions/incremental_table_test.ts covering the example above.
- `bazel test $(bazel query 'kind(test, //core/...)')` passed.

Related: https://github.com/dataform-co/dataform/pull/2119

### DIFF
--- a/common/protos/index.ts
+++ b/common/protos/index.ts
@@ -32,8 +32,10 @@ const Struct = google.protobuf.Struct;
 
 // Save references to the original generated methods
 const originalVerify = Struct.verify;
+const originalFromObject = Struct.fromObject;
 
-export function verifyStruct(this: any, object: any) {
+// Monkey Patching Methods
+Struct.verify = function (object: any) {
   if (object && typeof object === "object" && !("fields" in object)) {
     const fields: { [key: string]: any } = {};
     for (const [k, v] of Object.entries(object)) {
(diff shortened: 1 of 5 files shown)


kolina (changes requested):
Does the same error happen for non-incremental tables or other types of actions? Will we need the same fix for them?

fuzmish:
@kolina 
c918fc7 I have added some tests.
I think this is specific to incremental models, since it is the only model type with an enum field (`onSchemaChange`) checked before `metadata` field in the generated protobuf verify method.

kolina (changes requested):
Ok, so after a closer look the  reason for this is that `protoType.verify` in `verifyObjectMatchesProto` stops early due to string value of enum field and doesn't monkey-path struct field.

I think the more robust fix is to replace `verify` with `fromObject` in `verifyObjectMatchesProto`, I've prototyped it here https://github.com/dataform-co/dataform/commit/6e8c3fa9c2539b3d0da967ddfa7871a3ed8f6042 and it seems to work.

If you're ok with this change, feel free to update your PR with it

fuzmish:
@kolina Thank you for your feedback. fa2b5c9 I have applied that.

kolina (approved):

PR #2243, opened 2026-08-17
Add proto surface for PropertyGraph action

rafal-hawrylak:
Introduces config- and compiled-form protos for BigQuery Property Graphs.

- protos/property_graph_config.proto (new): PropertyGraphConfig and children, plus DatasetDataSource / CatalogDataSource / GraphRef variants for data_source. GraphRef lets an entity or relationship reference another Dataform action by name (with optional schema and database) instead of a fully qualified BigQuery table.
- protos/configs.proto: import property_graph_config.proto and add PropertyGraphConfig to ActionConfig.action.
- protos/core.proto: PropertyGraph and children. Wire repeated PropertyGraph property_graphs = 16 on CompiledGraph.
- protos/BUILD: include property_graph_config.proto in dataform_proto.

Config and compiled types are decoupled. Only core/actions/property_graph.ts will bridge them.

### DIFF
--- a/protos/core.proto
+++ b/protos/core.proto
@@ -438,6 +438,55 @@ message MergeLoadMode {
   repeated string unique_key = 1;
 }
 
+message PropertyGraph {
+  Target target = 1;
+  Target canonical_target = 2;
+  repeated Target dependency_targets = 3;
+  string file_name = 4;
+  string description = 5;
+  repeated GraphEntity entities = 6;
+  repeated GraphRelationship relationships = 7;
+  string graph_body = 8;
+}
+
+message GraphEntity {
+  string name = 1;
+  Target data_source = 2;
+  repeated string keys = 3;
+  repeated GraphLabel labels = 4;
+}
+
+message GraphRelationship {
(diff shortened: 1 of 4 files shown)


protos/property_graph_config.proto
kolina: can you add comments? it applies to all fields you added in proto in this PR

protos/core.proto
```diff
@@ -438,6 +438,55 @@ message MergeLoadMode {
...
 
+message PropertyGraph {
+  Target target = 1;
+  Target canonical_target = 2;
+  repeated Target dependency_targets = 3;
+  string file_name = 4;
+  string description = 5;
+  repeated GraphEntity entities = 6;
+  repeated GraphRelationship relationships = 7;
+  string body_fragment = 8;
```
kolina: let's think about alternative name for `body_fragment`

kolina (approved):

protos/core.proto
rafal-hawrylak: Changed to graph_body

PR #2246, opened 2026-08-18
Add PropertyGraph action class

rafal-hawrylak:
Introduces core/actions/property_graph.ts implementing the PropertyGraph action for BigQuery Property Graph DDL generation. The class parses PropertyGraphConfig, applies normalization (scalar keys, string field shorthand, joinKeys array shorthand, default label synthesis, wildcard expansion), validates cross-entity references and emits the NODE TABLES / EDGE TABLES graph body.

Catalog-based data sources (BigLake/Iceberg 4-part references) are parsed but rejected at compile time until project-wide 4-part Target support lands.

Adds 26 direct-instantiation unit tests covering happy path, edge cases and error paths. Remaining tests requiring session-registered actions and runMainInVm are deferred.

Files:
- core/actions/property_graph.ts (new)
- core/actions/property_graph_test.ts (new)
- core/BUILD


core/actions/property_graph.ts
```diff
@@ -0,0 +1,480 @@
...
+  }
+  if (!endpointConfig.joinKeys) {
+    throw new Error(`${where}: must declare 'joinKeys'.`);
+  }
+  const relCols = endpointConfig.joinKeys.relationshipColumns || [];
+  const entityCols = endpointConfig.joinKeys.entityColumns || [];
+  if (relCols.length === 0) {
+    throw new Error(`${where}: 'joinKeys.relationshipColumns' must be non-empty.`);
+  }
+  if (entityCols.length > 0 && entityCols.length !== relCols.length) {
```
kolina, line 339: you see to check it in multiple places

core/actions/property_graph.ts
```diff
@@ -0,0 +1,480 @@
...
+        synonyms: field.synonyms || []
+      })
+    ),
+    importAll: !!wildcard?.importAll,
+    importExcept: wildcard?.except || []
+  });
+}
+
+function parseTablePath(path: string, where: string): dataform.Target {
+  const parts = path.split(".");
```
kolina, line 408: I suspect it's more complicated

for example `my-project.`my_dataset`.`table.view` may be a valid identifier

core/actions/property_graph_test.ts
```diff
@@ -0,0 +1,480 @@
...
+      entities: [
+        {
+          name: "Account",
+          dataSourceString: "proj.ds.Accounts",
+          keys: ["id"]
+        }
+      ]
+    });
+
+    expect(asPlainObject(compiled.target)).deep.equals({
```
kolina, line 35: could we just compare the whole result object in these tests?

core/actions/property_graph_test.ts
```diff
@@ -0,0 +1,480 @@
...
+    });
+    expect(compiled.fileName).equals("definitions/graph.yaml");
+    expect(compiled.entities).length(1);
+    expect(compiled.relationships).length(0);
+    expect(asPlainObject(compiled.entities[0].dataSource)).deep.equals({
+      database: "proj",
+      schema: "ds",
+      name: "Accounts"
+    });
+    expect(compiled.graphBody).contains("NODE TABLES");
```
kolina, line 53: I'd prefer to check the whole body to validate how different settings interact with each other

kolina (approved):

core/actions/property_graph.ts
rafal-hawrylak: I saw the comments after I clicked "merge" button. I'll address all those in the next PR.

rafal-hawrylak: Removed the pre-check as duplicate because the post-check is the one that catches all mismatches (including those introduced by resolveEndpointDefaults), while the pre-check only covers the both-provided case.

core/actions/property_graph_test.ts
rafal-hawrylak: done

core/actions/property_graph_test.ts
rafal-hawrylak: done

core/actions/property_graph.ts
rafal-hawrylak: Not clear what do you mean. Is it about backtick?
Is this the value you say is valid: my-project.my_dataset.table.view` ?
Or this one: my-project.my_dataset.table.view ?

core/actions/property_graph.ts
rafal-hawrylak: The follow up PR - https://github.com/dataform-co/dataform/pull/2249

PR #2248, opened 2026-08-18
feat(lineage): emit OpenLineage RunEvents from CLI to Knowledge Catalog Lineage

rafal-hawrylak:
Adds an opt-in lineage emitter that publishes OpenLineage
START/COMPLETE/FAIL/ABORT RunEvents to Knowledge Catalog Lineage
(datalineage.googleapis.com) for each action executed by
`dataform run`. Off by default. Enabled via `--emit-lineage`
on the CLI or `lineageEnabled: true` in workflow_settings.

PR 1 of a 3-PR series. Scope: emitter + payload builder +
factory + integration tests. Facets included: parent,
nominalTime, gcp_lineage, gcp_bq_pipelines_run.runType.
Fail-open: lineage failures never affect workflow status.

Excluded from this PR (later PRs): endpoint router with REP
fallback, gax retry, structured error classification,
externalQuery/errorMessage/sql/jobType/gcp_bq_pipelines_job
facets, staging endpoint override.

### DIFF
--- a/cli/api/BUILD
+++ b/cli/api/BUILD
@@ -10,6 +10,7 @@ ts_library(
         ["**/*.ts"],
         exclude = [
             "utils/**/*.*",
+            "lineage/testing/**/*.*",
             "**/*_test.ts",
         ],
     ),
@@ -27,6 +28,7 @@ ts_library(
         "//sqlx:lexer",
         "//cli/vm:vm",
         "@npm//@google-cloud/bigquery",
+        "@npm//@google-cloud/lineage",
         "@npm//@types/fs-extra",
         "@npm//@types/glob",
         "@npm//@types/js-beautify",
@@ -63,6 +65,10 @@ ts_test_suite(
         "commands/jit/rpc_test.ts",
         "dbadapters/bigquery_test.ts",
         "execution_sql_test.ts",
+        "lineage/emitter_factory_test.ts",
+        "lineage/emitter_test.ts",
+        "lineage/lineage_integration_test.ts",
+        "lineage/payload_builder_test.ts",
     ],
     data = [
         ":node_modules",
@@ -72,10 +78,12 @@ ts_test_suite(
     ] + glob(["goldens/**"]),
     deps = [
         ":api",
+        "//cli/api/lineage/testing",
         "//core",
         "//protos:ts",
         "//testing",
         "@npm//@google-cloud/bigquery",
+        "@npm//@google-cloud/lineage",
         "@npm//@types/chai",
         "@npm//@types/fs-extra",
         "@npm//@types/js-yaml",
@@ -84,6 +92,7 @@ ts_test_suite(
         "@npm//chai",
         "@npm//fs-extra",
         "@npm//js-yaml",
+        "@npm//long",
         "@npm//ts-mockito",
     ],
 )
--- a/cli/api/commands/run.ts
+++ b/cli/api/commands/run.ts
@@ -5,6 +5,7 @@ import { JitCompileChildProcess } from "df/cli/api/commands/jit/compiler";
 import * as dbadapters from "df/cli/api/dbadapters";
 import { IBigQueryExecutionOptions } from "df/cli/api/dbadapters/bigquery";
 import { ExecutionSql } from "df/cli/api/dbadapters/execution_sql";
+import { LineageEmitter } from "df/cli/api/lineage/emitter";
 import { Flags } from "df/common/flags";
 import { retry } from "df/common/promises";
 import { deepClone, equals } from "df/common/protos";
@@ -43,6 +44,7 @@ export interface IExecutionOptions {
     options?: IBigQueryExecutionOptions,
     onCancel?: (cancel: () => void) => void
   ) => Promise<dataform.IJitCompilationResponse>;
+  lineageEmitter?: LineageEmitter;
 }
 
 export type CancelReason = "timeout" | "user" | "cancellation";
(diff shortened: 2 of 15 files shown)


cli/api/commands/run.ts
apilaskowski, line 438: When an action starts execution, a START RunEvent is emitted (at 418). If JiT compilation throws an error at 424 the method returns early bypassing:
`actionResult.timing = timer.end();
    if (this.executionOptions.lineageEmitter) {
      this.executionOptions.lineageEmitter.emitForAction(action, actionResult);
    }`

As a result:
* The terminal FAIL event is never emitted for this action.
* The activeRunIds map retains a stale runId entry for this action key.

Suggestion:
Ensure that timer.end() and this.executionOptions.lineageEmitter.emitForAction(action, actionResult) are triggered for JiT compilation failures (e.g. by falling through to the final status/emit block or using a finally block)

cli/api/lineage/payload_builder.ts
```diff
@@ -0,0 +1,195 @@
...
+      _schemaURL: "https://openlineage.io/spec/facets/1-0-1/NominalTimeRunFacet.json",
+      nominalStartTime: new Date(
+        actionResult.timing?.startTimeMillis?.toNumber?.() || Date.now()
+      ).toISOString()
+    };
+    if (actionResult.timing?.endTimeMillis) {
+      nominalTime.nominalEndTime = new Date(
+        actionResult.timing.endTimeMillis.toNumber()
+      ).toISOString();
+    }
```
apilaskowski, line 90: * If endTimeMillis is a plain JavaScript number (e.g., in JSON deserialization or custom fixtures), calling .toNumber() throws:
      TypeError: actionResult.timing.endTimeMillis.toNumber is not a function.
* For startTimeMillis, if it is already a primitive number, startTimeMillis.toNumber?.() evaluates to undefined, which unexpectedly falls back to Date.now().

Suggestion - Extract a small helper:
```
    function toMillis(val?: Long | number): number | undefined {
      if (typeof val === "number") return val;
      if (val && typeof (val as Long).toNumber === "function") return (val as Long).toNumber();
      return undefined;
    }
```

cli/api/lineage/payload_builder.ts
```diff
@@ -0,0 +1,195 @@
...
+      namespace: "bigquery",
+      name: `${dep.database}.${dep.schema}.${dep.name}`
+    }));
+
+    const outputs = [
+      {
+        namespace: "bigquery",
+        name: `${action.target.database}.${action.target.schema}.${action.target.name}`
+      }
+    ];
```
apilaskowski, line 73: In `emitter.ts:123`, projectId defaults with action.target.database || this.credentials.projectId.
However, in payload_builder.ts, if action.target.database or dep.database is undefined/empty, table names become "undefined.schema.name".

Fallback to projectId when constructing table names:
    ```
const inputs = (action.dependencyTargets || []).map(dep => ({
      namespace: "bigquery",
      name: `${dep.database || projectId}.${dep.schema}.${dep.name}`
    }));

    const outputs = [
      {
        namespace: "bigquery",
        name: `${action.target.database || projectId}.${action.target.schema}.${action.target.name}`
      }
    ];
```

File: cli/api/lineage/payload_builder.ts [hunk 3]
```diff
@@ -0,0 +1,195 @@
...
+      .toLowerCase()
+      .replace(/[^a-z0-9]+/g, "-")
+      .replace(/^-+|-+$/g, "");
+    const base = sanitized || "workdir";
+    const shortHash = this.workdirHash.slice(0, 8);
+    return `${base}-${shortHash}`;
+  }
+}
+
+function generateUuid(): string {
```
apilaskowski: Node.js crypto module is already imported in payload_builder.ts. Replacing the Math.random generator with import { randomUUID } from "crypto" provides standard, cryptographically secure RFC 4122 v4 UUIDs.

apilaskowski (changes requested):
Overall LGTM, let me know about those small issues I found

cli/api/lineage/emitter.ts
```diff
@@ -0,0 +1,140 @@
...
+    actionResult: dataform.IActionResult
+  ): void {
+    if (this.emissionDisabledThisRun) {
+      return;
+    }
+
+    if (this.emitterOptions.dryRun) {
+      if (!this.dryRunSkipLogged) {
+        this.stderr.write(
+          "[lineage] Skipped lineage emission (dry-run mode; once-per-run): skip_reason=dry_run\n"
```
apilaskowski, line 80: Is this standard message?

cli/api/BUILD
```diff
@@ -10,6 +10,7 @@ ts_library(
         ["**/*.ts"],
         exclude = [
             "utils/**/*.*",
+            "lineage/testing/**/*.*",
             "**/*_test.ts",
```
apilaskowski, line 14: I wonder if we could merge those 2 lines

cli/index.ts
```diff
@@ -804,6 +823,9 @@ export function runCli() {
             runner.onChange(printExecutedGraph);
           }
           const runResult = await runner.result();
+          if (lineageEmitter) {
+            await lineageEmitter.drain(15000);
```
apilaskowski: Can we extract this number to global variable or sth? I see random number and I don't know what it represents here. I don't know what this drain does without checking, and this is just a number here.

cli/api/BUILD
rafal-hawrylak: mock_lineage_client.ts has no _test.ts suffix, so both **/*_test.ts and lineage/testing/**/*.* are needed.

cli/api/lineage/emitter.ts
rafal-hawrylak: New for this feature. Consistent across the 5 skip sites so tests can assert on the label.

apilaskowski: I ment that in that message we have "skipped .. dry-run ... skip_...dry_run", but this is likely intended.

apilaskowski (approved):

PR #2251, opened 2026-08-19
Wire PropertyGraph action into compile pipeline

rafal-hawrylak:
Wire PropertyGraph action into compile pipeline

Builds on top of #2246 (PropertyGraph action class) and #2249 (review
follow-up). The class parsed and rendered a graph body but was not yet
loaded from user projects or exposed to downstream code.

- Session loading: main.ts finds `graph.yaml` files under `definitions/`
  and registers a PropertyGraph action per file. At most one graph.yaml
  is accepted per project.
- Action index: actions/index.ts and session.ts expose PropertyGraph so
  downstream code treats it like tables and operations.
- Emitter extensions: description and synonyms on both entities and
  relationships, fieldWildcard normalization for `fields: { importAll:
  true, except: [...] }`, and an OPTIONS clause on the graph carrying
  the graph description.
- Proto: adds `is_default` to GraphLabel for the label wiring above.
- Coverage: property_graph_test.ts grows for the new normalization and
  emitter behavior, and main_test.ts adds session-level tests plus a
  propertyGraph case in the shared test-helper switch.

Older `dataformCoreVersion` values silently skip PropertyGraph files so
projects pinning an earlier core do not break at load time.

### DIFF
--- a/core/actions/property_graph.ts
+++ b/core/actions/property_graph.ts
@@ -113,22 +113,7 @@ export class PropertyGraph extends ActionBuilder<dataform.PropertyGraph> {
       throw new Error(`${where} must declare 'keys'.`);
     }
 
-    const rootFields = entityConfig.fields || [];
-    const rootWildcard = entityConfig.fieldWildcard;
-    const configuredLabels = entityConfig.labels || [];
-    if ((rootFields.length > 0 || rootWildcard) && configuredLabels.length > 0) {
-      throw new Error(
-        `${where} cannot combine root-level 'fields'/'fieldWildcard' with a 'labels' list.`
-      );
-    }
-
-    const labels = buildLabels(
-      entityName,
-      rootFields,
-      rootWildcard,
-      configuredLabels,
-      where
-    );
+    const labels = buildLabels(entityName, entityConfig, where);
 
     return dataform.GraphEntity.create({
       name: entityName,
@@ -159,16 +144,7 @@ export class PropertyGraph extends ActionBuilder<dataform.PropertyGraph> {
     const source = buildEndpoint(relConfig.source, `${where} source`);
     const destination = buildEndpoint(relConfig.destination, `${where} destination`);
 
-    const rootFields = relConfig.fields || [];
-    const rootWildcard = relConfig.fieldWildcard;
-    const configuredLabels = relConfig.labels || [];
-    if ((rootFields.length > 0 || rootWildcard) && configuredLabels.length > 0) {
-      throw new Error(
-        `${where} cannot combine root-level 'fields'/'fieldWildcard' with a 'labels' list.`
-      );
-    }
-
-    const labels = buildLabels(relName, rootFields, rootWildcard, configuredLabels, where);
+    const labels = buildLabels(relName, relConfig, where);
 
     return dataform.GraphRelationship.create({
       name: relName,
@@ -276,12 +252,16 @@ export class PropertyGraph extends ActionBuilder<dataform.PropertyGraph> {
   }
 
   private emitGraphBody(): string {
-    const nodeEntries = this.proto.entities.map(entity => renderNode(entity));
     const parts: string[] = [];
+    const nodeEntries = this.proto.entities.map(entity => renderNode(entity));
     parts.push(`NODE TABLES (\n  ${nodeEntries.join(",\n  ")}\n)`);
     if (this.proto.relationships.length > 0) {
       const edgeEntries = this.proto.relationships.map(rel => renderEdge(rel));
       parts.push(`EDGE TABLES (\n  ${edgeEntries.join(",\n  ")}\n)`);
+    }
+    const options = renderOptionsClause(this.proto.description, undefined);
+    if (options) {
+      parts.push(options);
     }
     return parts.join("\n");
   }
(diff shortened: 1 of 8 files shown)


core/actions/property_graph.ts
```diff
@@ -168,7 +170,15 @@ export class PropertyGraph extends ActionBuilder<dataform.PropertyGraph> {
...
 
-    const labels = buildLabels(relName, rootFields, rootWildcard, configuredLabels, where);
+    const labels = buildLabels(
+      relName,
+      relConfig.description,
+      relConfig.synonyms,
+      rootFields,
+      rootWildcard,
+      configuredLabels,
+      where
```
kolina: I'd rather not create functions with such a big amount of arguments and keep them at 5 arguments max. Can you figure out how to optimize it?

core/actions/property_graph.ts
```diff
@@ -465,10 +513,42 @@ function renderPropertiesClause(label: dataform.IGraphLabel): string {
...
+  if (description) {
+    parts.push(`description=${quoteString(description)}`);
+  }
+  if (synonyms && synonyms.length > 0) {
+    parts.push(`synonyms=[${synonyms.map(quoteString).join(", ")}]`);
+  }
+  return parts.length > 0 ? `OPTIONS(${parts.join(", ")})` : "";
+}
+
+function quoteString(value: string): string {
```
kolina: We already have such function in other places, let's reuse it instead of copy-pasting

core/actions/property_graph_test.ts
```diff
@@ -514,6 +514,201 @@ suite("property_graph", () => {
...
+          dataSourceString: "p.d.A",
+          keys: ["id"],
+          fields: { importAll: true }
+        }
+      ]
+    });
+
+    expect(compiled.entities[0].labels).length(1);
+    expect(compiled.entities[0].labels[0].importAll).equals(true);
+    expect(compiled.graphBody).contains("PROPERTIES ARE ALL COLUMNS");
```
kolina: I've asked in the previous PR and will ask again: can we just check the whole compiled object, not only specific fields?

Looking at current tests, it'll be hard to understand how the overall graph is supposed to look like for different inputs

Could we update all tests in such a manner?

core/main.ts
```diff
@@ -222,6 +223,69 @@ function loadActionConfigsFile(
   return dataform.ActionConfigs.fromObject(actionConfigsAsJson);
 }
 
+function loadPropertyGraphs(session: Session, filePaths: string[]) {
+  const graphPaths = filePaths
+    .filter(
+      path =>
+        path.startsWith(`definitions${Path.separator}`) &&
+        Path.basename(path) === "graph" &&
```
kolina, line 234: so if a user would like to have multiple graphs, they'd need to create separate directories for each one? it may not be very convenient

core/main.ts
```diff
@@ -222,6 +223,69 @@ function loadActionConfigsFile(
...
+    )
+    .sort();
+  if (graphPaths.length === 0) {
+    return;
+  }
+  if (graphPaths.length > 1) {
+    session.compileError(
+      new Error(
+        `At most one graph.yaml is allowed per project (found ${graphPaths.length}: ` +
+          `${graphPaths.join(", ")}).`
```
kolina: let's add that it may be changed in future versions

core/main.ts
```diff
@@ -222,6 +223,69 @@ function loadActionConfigsFile(
...
+    return;
+  }
+  try {
+    session.actions.push(new PropertyGraph(session, configAsJson, graphPath));
+  } catch (e) {
+    session.compileError(e, graphPath);
+  }
+}
+
+function snakeToCamelKeys(value: any): any {
```
kolina: could we put this function into existing place where we collect such utilities?

core/main_test.ts
```diff
@@ -2380,4 +2383,365 @@ select 1 as a`
...
+      expect(result.compile.compiledGraph.graphErrors.compilationErrors).deep.equals([]);
+      expect(result.compile.compiledGraph.propertyGraphs).to.have.lengthOf(1);
+      expect(asPlainObject(result.compile.compiledGraph.propertyGraphs[0].target)).deep.equals({
+        database: "defaultProject",
+        schema: "defaultDataset",
+        name: "SimpleGraph"
+      });
+      expect(result.compile.compiledGraph.propertyGraphs[0].graphBody).to.include(
+        "NODE TABLES"
+      );
```
kolina: same: can we verify the whole result so it'll be easier to see examples of generated graphs?

kolina (changes requested):

core/main.ts
rafal-hawrylak: Well, not really. The intention of current implementation is to:
- allow graph.yaml to be in definitions dir or any level of subdir - a bit of flexibility
- reject if more than single graph.yaml is found this way

So this is not an implementation supporting multiple graph.yaml just yet.

core/actions/property_graph_test.ts
rafal-hawrylak: Correct, will be more cautious next time.

core/actions/property_graph.ts
rafal-hawrylak: Done via JSON.stringify. I could't find any relevant existing util to drop in here

core/actions/property_graph.ts
rafal-hawrylak: Reused existing containers: dataform.IGraphEntityConfig and dataform.IGraphRelationshipConfig to wrap params

core/actions/property_graph_test.ts
```diff
@@ -514,6 +766,403 @@ suite("property_graph", () => {
...
+            ]
+          }
+        ],
+        graphBody:
+          "NODE TABLES (\n  `p.d.A` AS Account KEY (id) LABEL Account PROPERTIES (id)\n)"
+      })
+    );
+  });
+
+  test("named label description and synonyms are not emitted as inline LABEL OPTIONS", () => {
```
kolina: I don't see synonyms to be included into graph body in this test?

core/main.ts
kolina: ok, let's leave a comment to clarify that it may be changed in the future

kolina (approved):

core/actions/property_graph_test.ts
rafal-hawrylak: Fixed

core/main.ts
rafal-hawrylak: Sure, done

PR #2253, opened 2026-08-21
Follow-up to #2251: address PropertyGraph review comments

rafal-hawrylak:
- Rename `GraphRef` to `DataSourceRef` in `PropertyGraphConfig` so the type name matches its field shape (a target identifier not a graph reference).
- Restrict `description` and `synonyms` OPTIONS to the DEFAULT label. Non-DEFAULT labels that declare either now error out. A label named `DEFAULT` (case-insensitive) is treated as the default and only one DEFAULT label is allowed per entity or relationship.
- Keep graph-level `description` on the compiled proto but stop emitting it in `graphBody`. DDL propagation is planned.
- Add `disabled` field to `PropertyGraphConfig` and propagate it onto the compiled `PropertyGraph` proto matching the convention used by Table View Assertion and Operation actions.

### DIFF
--- a/core/actions/property_graph_test.ts
+++ b/core/actions/property_graph_test.ts
@@ -43,6 +43,8 @@ suite("property_graph", () => {
         target: graphTarget("Simple"),
         canonicalTarget: graphTarget("Simple"),
         fileName: "definitions/graph.yaml",
+        description: "",
+        disabled: false,
         entities: [
           {
             name: "Account",
@@ -68,6 +70,8 @@ suite("property_graph", () => {
         target: graphTarget("G"),
         canonicalTarget: graphTarget("G"),
         fileName: "definitions/graph.yaml",
+        description: "",
+        disabled: false,
         entities: [
           {
             name: "A",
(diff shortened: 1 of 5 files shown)


core/actions/property_graph_test.ts
```diff
@@ -1153,17 +1188,146 @@ suite("property_graph", () => {
...
+            }
+          ]
+        }
+      ]
+    });
+
+    expect(asPlainObject(compiled).entities[0].labels[0].isDefault).equals(true);
+    expect(asPlainObject(compiled).graphBody).equals(
+      "NODE TABLES (\n  `p.d.A` AS A KEY (id) DEFAULT LABEL PROPERTIES (id)\n)"
+    );
```
kolina, line 1308: check the whole compiled object for consistency?

kolina (approved):

core/actions/property_graph_test.ts
rafal-hawrylak: Done

PR #2255, opened 2026-08-24
Wire PropertyGraph into CLI build and DDL execution

rafal-hawrylak:
Follows the compile-pipeline wiring in the preceding commit. Adds the CLI-side plumbing that turns a compiled PropertyGraph action into a CREATE OR REPLACE PROPERTY GRAPH statement executed against BigQuery.

- Build wiring: cli/api/commands/build.ts includes PropertyGraph actions in the execution graph.
- DDL execution: cli/api/dbadapters/execution_sql.ts adds a code path for statements that produce a property graph so CREATE OR REPLACE PROPERTY GRAPH runs on the warehouse.
- Coverage: execution_sql_test.ts covers the new code path, and two golden `.sql` files pin the exact DDL for a minimal fingraph and an HR-style graph.

### DIFF
--- a/cli/api/commands/build.ts
+++ b/cli/api/commands/build.ts
@@ -62,7 +62,8 @@ export class Builder {
         )
       ),
       this.prunedGraph.operations.map(o => this.buildOperation(o)),
-      this.prunedGraph.assertions.map(a => this.buildAssertion(a))
+      this.prunedGraph.assertions.map(a => this.buildAssertion(a)),
+      this.prunedGraph.propertyGraphs.map(g => this.buildPropertyGraph(g))
     );
     return dataform.ExecutionGraph.create({
       projectConfig: this.prunedGraph.projectConfig,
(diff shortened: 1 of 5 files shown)


cli/api/commands/build.ts
```diff
@@ -106,17 +107,31 @@ export class Builder {
...
-    action: dataform.ITable | dataform.IOperation | dataform.IAssertion
+    action:
+      | dataform.ITable
+      | dataform.IOperation
+      | dataform.IAssertion
+      | dataform.IPropertyGraph
   ) {
     const jitCode = "jitCode" in action ? action.jitCode : undefined;
+    const actionDescriptor = "actionDescriptor" in action ? action.actionDescriptor : undefined;
+    const disabled = "disabled" in action ? action.disabled : false;
```
kolina: is this dynamic logic needed?

kolina (approved):

cli/api/commands/build.ts
rafal-hawrylak: Right, all four action types have it. Simplified.

PR #2258, opened 2026-08-25
Support ref() and DAG dependencies in PropertyGraph

rafal-hawrylak:
Extends PropertyGraph to accept `ref: [dataset, table]` in place of a bare table name for `nodeTables` and `edgeTables`. Refs resolve through the same `Session.resolve` path used by tables and operations so the graph action inherits the standard dependency edges.

The resolved dependencies land on the compiled action's `dependencyTargets` which makes the executor run source tables before the graph DDL. Ref resolution respects schema suffixes and database overrides so a graph declared with `ref: [authors]` in a suffixed workspace still targets the suffixed dataset.

Lock-in tests cover the resolved-DDL output for named refs, dataset-scoped refs, mixed literal-and-ref entries, and the error surface when a ref points at a nonexistent action.

### DIFF
--- a/core/actions/property_graph.ts
+++ b/core/actions/property_graph.ts
@@ -1,6 +1,5 @@
 import { verifyObjectMatchesProto, VerifyProtoErrorBehaviour } from "df/common/protos";
 import { ActionBuilder } from "df/core/actions";
-import { Declaration } from "df/core/actions/declaration";
 import { Session } from "df/core/session";
 import { checkAssertionsForDependency } from "df/core/utils";
 import { dataform } from "df/protos/ts";
@@ -122,21 +121,11 @@ export class PropertyGraph extends ActionBuilder<dataform.PropertyGraph> {
     if (!rawRef) {
       return true;
     }
-    const resolved = this.session.indexedActions.find(rawRef);
-    if (resolved.length !== 1) {
+    const target = this.session.resolveTarget(rawRef);
+    if (!target) {
       return false;
     }
-    const resolvedAction = resolved[0];
-    const target = resolvedAction.getTarget();
-    if (resolvedAction instanceof Declaration) {
-      entityOrRel.dataSource = target;
-    } else {
-      entityOrRel.dataSource = dataform.Target.create({
-        database: target.database && this.session.finalizeDatabase(target.database),
-        schema: this.session.finalizeSchema(target.schema),
-        name: this.session.finalizeName(target.name)
-      });
-    }
+    entityOrRel.dataSource = dataform.Target.create(target);
     return true;
   }
 
--- a/core/actions/property_graph.ts
+++ b/core/actions/property_graph.ts
@@ -1,15 +1,32 @@
 import { verifyObjectMatchesProto, VerifyProtoErrorBehaviour } from "df/common/protos";
 import { ActionBuilder } from "df/core/actions";
+import { Declaration } from "df/core/actions/declaration";
 import { Session } from "df/core/session";
+import { checkAssertionsForDependency } from "df/core/utils";
 import { dataform } from "df/protos/ts";
 
 const CATALOG_NOT_SUPPORTED_MESSAGE =
   "Catalog-based data sources (BigLake/Iceberg/external catalogs) are not yet supported.";
 
+function entityRefKey(name: string): string {
+  return `entity:${name}`;
+}
+
+function relationshipRefKey(name: string): string {
+  return `relationship:${name}`;
+}
+
 export class PropertyGraph extends ActionBuilder<dataform.PropertyGraph> {
   public session: Session;
+  public dependOnDependencyAssertions: boolean = false;
 
   private proto = dataform.PropertyGraph.create({ description: "", disabled: false });
+  // Populated during construction for entity/relationship configs that use a ref-style
+  // dataSource; drained in finalize() once all actions are indexed, since the target
+  // schema/database of a ref cannot be resolved until every action is known.
+  // Keys are produced by entityRefKey / relationshipRefKey.
+  private pendingRefs = new Map<string, dataform.ITarget>();
+  private dependencyKeys = new Set<string>();
 
   constructor(session?: Session, unverifiedConfig?: any, filename?: string) {
     super(session);
@@ -61,8 +78,6 @@ export class PropertyGraph extends ActionBuilder<dataform.PropertyGraph> {
     this.validateUniqueNames(config.name);
     this.resolveEndpointDefaults();
     this.validateEndpointReferences(config.name);
-
-    this.proto.graphBody = this.emitGraphBody();
   }
 
   public getFileName() {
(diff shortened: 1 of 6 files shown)


core/actions/property_graph.ts
```diff
@@ -74,20 +75,66 @@ export class PropertyGraph extends ActionBuilder<dataform.PropertyGraph> {
...
       this.proto,
       VerifyProtoErrorBehaviour.SUGGEST_REPORTING_TO_DATAFORM_TEAM
     );
+    return this.proto;
+  }
+
+  public finalize() {
+    let allResolved = true;
+    for (const entity of this.proto.entities) {
+      if (!this.resolveRefIntoDataSource(entity, `entity:${entity.name}`)) {
```
kolina: nit: do not hardcode logic of creating these keys like `entity:${entity.name}` throughout the code and create helpers for it

core/actions/property_graph.ts
```diff
@@ -10,6 +11,8 @@ export class PropertyGraph extends ActionBuilder<dataform.PropertyGraph> {
   public session: Session;
 
   private proto = dataform.PropertyGraph.create({ description: "", disabled: false });
+  private pendingRefs = new Map<string, dataform.ITarget>();
```
kolina, line 27: let's write a comment about how `pendingRefs` is used, I spend some time to understand it

core/actions/property_graph.ts
```diff
@@ -74,20 +75,66 @@ export class PropertyGraph extends ActionBuilder<dataform.PropertyGraph> {
...
+    const target = resolvedAction.getTarget();
+    if (resolvedAction instanceof Declaration) {
+      entityOrRel.dataSource = target;
+    } else {
+      entityOrRel.dataSource = dataform.Target.create({
+        database: target.database && this.session.finalizeDatabase(target.database),
+        schema: this.session.finalizeSchema(target.schema),
+        name: this.session.finalizeName(target.name)
+      });
+    }
```
kolina: this  logic repeats this a lot https://github.com/dataform-co/dataform/blob/b5a143595fa9f5f1624fba06a3bc952800f98f7a/core/session.ts#L250? could we use `session.resolve` method here?

core/actions/property_graph.ts
```diff
@@ -187,6 +235,26 @@ export class PropertyGraph extends ActionBuilder<dataform.PropertyGraph> {
...
+      const rawRef: dataform.ITarget = { name: ref.name };
+      if (ref.schema) {
+        rawRef.schema = ref.schema;
+      }
+      if (ref.database) {
+        rawRef.database = ref.database;
+      }
+      this.pendingRefs.set(pendingKey, rawRef);
+      const depKey = `${rawRef.database || ""}.${rawRef.schema || ""}.${rawRef.name}`;
+      if (!this.dependencyKeys.has(depKey)) {
```
kolina, line 261: I think this logic should respect https://docs.cloud.google.com/dataform/docs/dependencies#set-assertions-as-dependencies

https://github.com/dataform-co/dataform/blob/b5a143595fa9f5f1624fba06a3bc952800f98f7a/core/actions/table.ts#L370

core/main_test.ts
```diff
@@ -3078,5 +3873,219 @@ relationships:
         ]
       }));
     });
+
+    test("dataSourceString path matching a managed table target adds no dependency", () => {
```
kolina: do these tests need to be here? they don't seem to test e2e behaviour like new tests above

core/session.ts
```diff
@@ -545,6 +545,8 @@ export class Session {
 
     this.removeNonUniqueActionsFromCompiledGraph(compiledGraph);
 
+    this.finalizePropertyGraphs();
```
kolina, line 564: why do we need a separate callback for graph actions here? Regular SQLX actions also have to insert resolved ref in generated SQL body and I'm wondering if it's possible to have it using consistent interfaces inside session

kolina (changes requested):

core/actions/property_graph.ts
rafal-hawrylak: resolve() returns SQL string, we need target. I could still unify it introducing shared helper. Is this what you'd expect?

core/actions/property_graph.ts
rafal-hawrylak: done

core/actions/property_graph.ts
rafal-hawrylak: done

core/main_test.ts
rafal-hawrylak: Right, removed. Covered by property_graph_test.ts

core/session.ts
rafal-hawrylak: SQLX ref() runs inside the query closure, which the session doesn't evaluate until every action is indexed - so ref() sees a fully-populated index at call time. YAML has no closure, so we get the equivalent deferral by stashing raw refs in construction and draining them in finalize(). Could contextify graphBody as a Contextable to mirror SQLX, but that changes YAML from data into code.

core/actions/property_graph.ts
rafal-hawrylak: Done, but note graph.yaml has no surface for these flags today (DataSourceRef proto is just {name, schema, database}), so the plumbing is inert.
Would you like keep it that way for now? Revert it? Or add support through proto?

core/actions/property_graph.ts
```diff
@@ -1,15 +1,32 @@
...
+  return `entity:${name}`;
+}
+
+function relationshipRefKey(name: string): string {
+  return `relationship:${name}`;
+}
+
 export class PropertyGraph extends ActionBuilder<dataform.PropertyGraph> {
   public session: Session;
+  public dependOnDependencyAssertions: boolean = false;
```
kolina: There is no code that would set this variable to `true`. is it intended? Also we should add a test verifying this behaviour

core/actions/property_graph.ts
kolina: I suppose you can create `session.resolve` version that would return `Target` and then make the current version to call it?

core/actions/property_graph.ts
kolina: I suppose we'd need to add this flag to proto, yeah.

kolina (approved):
Not related to this PR: it'll be easier to review without force pushes, because with them it's harder to check the diff after my comments. If you'll be adding incremental commits to PR, it'll be possible to just check them.

If you want to avoid polluting commit history, you can do squash on merge.

core/actions/property_graph.ts
rafal-hawrylak: Ok, just to be double sure - you'd like to surface this to allow user configure it via graph.yaml file?

core/actions/property_graph.ts
rafal-hawrylak: Ok, done

core/actions/property_graph.ts
kolina: yes

PR #2259, opened 2026-08-26
feat(lineage): harden CLI lineage emission (endpoint routing, retry)

rafal-hawrylak:
feat(lineage): harden CLI lineage emission (endpoint routing, retry, structured errors)

PR 3 of 3. Adds production-hardening around the lineage emit path.

Endpoint routing (LineageEndpointRouter):
- REP-first policy: `datalineage.<loc>.rep.googleapis.com` by default.
- Per-location sticky fallback to `datalineage.googleapis.com` when the REP endpoint is DNS-unresolvable or returns HTTP 302 (region mismatch). Subsequent emits in the same location skip the REP endpoint.

Retry policy (LINEAGE_RETRY_CONFIG):
- Delegated to google-gax. Transient set follows AIP-194: DEADLINE_EXCEEDED, RESOURCE_EXHAUSTED, ABORTED, INTERNAL, UNAVAILABLE.
- Backoff 1s / 2s / 4s exponential, 2s per-attempt timeout, 15s total budget. Non-transient errors fall through immediately.

Structured error classification:
- On emit failure: single-line stderr `[lineage] Failed to emit ...: code=<NAME>(<n>) endpoint=<host> location=<loc> message=<err>` so telemetry can parse root cause without regex on prose.
- PERMISSION_DENIED / SERVICE_DISABLED / FAILED_PRECONDITION disable emission for the rest of the run and log a single `skip_reason=api_disabled` line with hints for both missing IAM and API-not-enabled (7 is ambiguous on Google Cloud). Dedup guard flips before the stderr write so concurrent in-flight failures collapse to one line.
- HTTP 302 region mismatch on a non-REP endpoint disables the run with `skip_reason=endpoint_region_mismatch`.

Client attribution:
- LineageClient receives `libName=dataform-cli` + `libVersion=<version>` so KC-side telemetry can attribute traffic to the Dataform CLI.

Test coverage:
- New `endpoint_router_test.ts`.
- `emitter_test.ts` expands with tests covering retry config shape, REP-fallback loop, PERMISSION_DENIED / SERVICE_DISABLED skip-and-dedup, non-DNS error propagation, and the structured `Failed to emit` line format.
- `lineage_integration_test.ts` extends Case 5 with the once-per-run `skip_reason=api_disabled` assertion.

### DIFF
--- a/cli/api/BUILD
+++ b/cli/api/BUILD
@@ -65,11 +65,7 @@ ts_test_suite(
         "commands/jit/rpc_test.ts",
         "dbadapters/bigquery_test.ts",
         "execution_sql_test.ts",
-        "lineage/emitter_factory_test.ts",
-        "lineage/emitter_test.ts",
-        "lineage/lineage_integration_test.ts",
-        "lineage/payload_builder_test.ts",
-    ],
+    ] + glob(["lineage/*_test.ts"]),
     data = [
         ":node_modules",
         "//packages/@dataform/core:package_tar",
--- a/cli/api/lineage/emitter.ts
+++ b/cli/api/lineage/emitter.ts
@@ -1,10 +1,13 @@
 import { LineageClient } from "@google-cloud/lineage";
 
+import {
+  GLOBAL_LINEAGE_ENDPOINT,
+  LineageEndpointRouter
+} from "df/cli/api/lineage/endpoint_router";
 import { LineagePayloadBuilder, toProtoStruct } from "df/cli/api/lineage/payload_builder";
 import { coerceAsError } from "df/common/errors/errors";
+import { version } from "df/core/version";
 import { dataform } from "df/protos/ts";
-
-const GLOBAL_LINEAGE_ENDPOINT = "datalineage.googleapis.com";
 
 export interface IEmitterOptions {
   lineageEnabled?: boolean;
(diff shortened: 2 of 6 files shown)


cli/api/BUILD
```diff
@@ -67,6 +67,7 @@ ts_test_suite(
         "execution_sql_test.ts",
         "lineage/emitter_factory_test.ts",
         "lineage/emitter_test.ts",
+        "lineage/endpoint_router_test.ts",
         "lineage/lineage_integration_test.ts",
         "lineage/payload_builder_test.ts",
```
apilaskowski: Can we collapse this part to sth like "lineage/*_test.ts"?

cli/api/lineage/emitter.ts
```diff
@@ -35,20 +40,80 @@ export function createLineageClientProvider(
         new LineageClient({
           projectId: targetProjectId,
           apiEndpoint: endpoint,
-          credentials: credentials.credentials && JSON.parse(credentials.credentials)
+          credentials: credentials.credentials && JSON.parse(credentials.credentials),
+          libName: LINEAGE_CLIENT_LIB_NAME,
+          libVersion: version
```
apilaskowski, line 45: Looking at those 2 lines:

```
libName: LINEAGE_CLIENT_LIB_NAME,
libVersion: version
```

I needed to spent too much time gathering that this is DFcli name and version. Mb we should go for different name of this variable `LINEAGE_CLIENT_LIB_NAME`?
On first glance I expected that this is lineage library version and was confused.

apilaskowski (approved):

cli/api/BUILD
rafal-hawrylak: Done

cli/api/lineage/emitter.ts
rafal-hawrylak: Renamed

PR #2260, opened 2026-08-27
Fix LegacyConfigConverter mutating caller-shared configs

rafal-hawrylak:
Shallow-clone .bigquery and .assertions before in-place coercions so `publish("t1", shared); publish("t2", shared);` no longer leaves t2/t3 with an empty bigquery block. Regression from #1780 (3.0.10).

core/actions/index.ts


kolina (approved):
Let's add a test to avoid regressions in the future

PR #2268, opened 2026-09-02
Add CLI runtime coverage and E2E integration test for PropertyGraph

rafal-hawrylak:
Wires PropertyGraph into the CLI runtime paths that were missing from the initial action plumbing:

- `dataform prune` includes PropertyGraph actions in tag and dependency filtering
- `dataform run` prints graph actions through the console progress reporter with the same status lines as tables
- `dataform run --dry-run` skips PROPERTY GRAPH DDL because BigQuery dry-run requires the referenced NODE and EDGE tables to physically exist which is not true on a fresh workspace

Adds an integration test at tests/integration/property_graph_project that compiles a three-table fixture (authors, books, wrote) plus one LibraryGraph over them, executes it against BigQuery, asserts all four actions succeed, verifies the graph is created after the source tables by timestamp, and confirms the graph appears in
INFORMATION_SCHEMA.PROPERTY_GRAPHS. The test drops the graph and dataset before and after the run since `cleanWarehouse` does not know about property graphs.


cli/api/commands/prune.ts
```diff
@@ -67,7 +75,7 @@ function computeIncludedActionNames(
   // Determine actions selected with --tag option and update applicable actions
   if (hasTagSelector) {
     allActions
-      .filter(action => action.tags.some(tag => runConfig.tags.includes(tag)))
+      .filter(action => "tags" in action && action.tags?.some(tag => runConfig.tags.includes(tag)))
```
kolina, line 78: could we actually support `tags` for graph actions? it makes sense to do anyway and we won't have to update this code

cli/api/commands/run.ts
```diff
@@ -390,6 +390,18 @@ export class Runner {
...
+  private shouldSkipPropertyGraphDryRun(action: dataform.IExecutionAction): boolean {
+    if (!this.executionOptions.bigquery?.dryRun) {
+      return false;
+    }
+    if (action.type !== "propertyGraph") {
+      return false;
+    }
+    return (action.dependencyTargets || []).some(dep =>
+      this.allActionTargets.has(targetStringifier.stringify(dep))
+    );
```
kolina: I'm not really sure about the source of this limitation to skip dry run for graph actions: for table actions dry run can also fail if its Dataform dependencies haven't been created yet

If we don't skip dry run for tables in such cases, I wouldn't skip them for graph either

tests/api/api.spec.ts
```diff
@@ -1376,6 +1376,151 @@ suite("@dataform/api", () => {
...
+            actions: [
+              withCompiledSql(successfulTableActionResult, TABLE_STATEMENT),
+              withCompiledSql(successfulGraphActionResult, GRAPH_STATEMENT)
+            ]
+          }).toJSON()
+        );
+        verify(mockedDbAdapter.execute(GRAPH_STATEMENT, anything())).once();
+      });
+
+      test("runs property graph in non-dry-run mode even when deps are in this run", async () => {
```
kolina: The test name here is a bit confusing: in non-dry run mode I expect it's an intuitive behaviour

tests/integration/property_graph_project/workflow_settings.yaml
```diff
@@ -0,0 +1,3 @@
+defaultProject: dataform-open-source
+defaultLocation: us
```
kolina: let's skip it in new project examples

tests/integration/property_graph_project/package.json
kolina, line 1: is this file needed?

kolina (approved):

cli/api/commands/prune.ts
rafal-hawrylak: Good idea - will file follow up soon.

cli/api/commands/run.ts
rafal-hawrylak: Removed — consistent with tables now

tests/api/api.spec.ts
rafal-hawrylak: Fixed

tests/integration/property_graph_project/package.json
rafal-hawrylak: Yes, when it is absent it triggers "dataformCoreVersion must be specified either in workflow_settings.yaml or via a package.json". And this is the pattern in integration tests, ie similar to tests/integration/bigquery_project.

tests/integration/property_graph_project/workflow_settings.yaml
rafal-hawrylak: Noted, dropped

PR #2272, opened 2026-09-03
Filter the stateless-install project copy using .gitignore

### LINKED ISSUE #2269: Stateless install copies entire project directory unfiltered, scaling compile time with unrelated files
Related to #2099 / #2100

## What's happening

When `dataformCoreVersion` is set in `workflow_settings.yaml`, `compile()` in [`cli/api/commands/compile.ts`](https://github.com/dataform-co/dataform/blob/main/cli/api/commands/compile.ts) copies the entire project directory to a temp dir before installing core and compiling:

```ts
fs.copySync(resolvedProjectPath, temporaryProjectPath);
```

No `filter` is passed, so `fs-extra`'s `copySync` copies everything in the project directory byte-for-byte — including directories that have nothing to do with the Dataform project itself (`.venv/`, `.git/`, unrelated build output, etc.), if they happen to live alongside `definitions/`/`includes/`. This cost scales with total files/bytes under the project root, not with the size of the actual Dataform project, and it's paid on every single `compile` invocation.

Nothing else in the compile pipeline mitigates this — the `.gitignore` written by `dataform init` is only for git hygiene and isn't consulted here, and the later `glob.sync("!(node_modules)/**/*.*", ...)` call that enumerates source files happens *after* this copy has already paid the I/O cost.

## Impact

On a repo where a Python `.venv` or similar directory sits next to the Dataform project (common in mixed-tooling repos, e.g. Dataform + a Python-based ELT/testing setup), this can turn a sub-second copy into many seconds, independent of the actual project's size — worsening the fixed overhead already reported in #2099.

## Suggested fix

Pass a `filter` to `copySync` that skips, at minimum, `node_modules` (already excluded from the later glob) and `.git`, and ideally respects a `.gitignore`/`.dataformignore` in the project root if present — similar in spirit to how most file-watching/bundling tools (webpack, esbuild, etc.) avoid walking ignored directories by default. This wouldn't change the stateless-install guarantee (core is still freshly installed every time) — it just stops copying files that are never going to be `require()`'d anyway.

Happy to put up a PR for this if a maintainer can confirm the preferred approach (hardcoded excludes vs. `.gitignore`-aware).

dmargol1:
Fixes #2269.

Per @kolina's comment on the issue ("Approach with `.gitignore` seems more effective"), this takes the `.gitignore`-aware route rather than a hardcoded exclude list.

## The problem

When `dataformCoreVersion` is set in `workflow_settings.yaml`, `compile()` copies the whole project directory to a temp dir before running `npm i` there:

```ts
fs.copySync(resolvedProjectPath, temporaryProjectPath);
```

No `filter`, so the copy's cost scales with everything under the project root rather than with the Dataform project itself. In a mixed-tooling repo, a Python `.venv` sitting next to `definitions/` gets copied byte-for-byte on every `compile`.

## The change

Adds `buildProjectCopyFilter()`, an `fs-extra` `copySync` filter built from two sources:

1. The project's own root `.gitignore`, if present. `dataform init` already writes one, but nothing in the compile pipeline consulted it until now.
2. A small always-ignored floor of `.git` and `node_modules`, checked independently of the `.gitignore` so a negation pattern in a user's `.gitignore` cannot override it.

Using the project's `.gitignore` avoids having to enumerate every ecosystem's junk directories (`.venv`, `target/`, `__pycache__/`, `vendor/`, `coverage/`, ...). A project's `.gitignore` already states what that project treats as disposable.

The stateless-install guarantee is unchanged: core is still freshly installed into the temp dir on every compile.

## Behavior change to weigh in review

A gitignored file is no longer copied, so it is also no longer compiled. A project that generates definitions into a gitignored path (say a script emitting `definitions/generated/*.sqlx`) would need to unignore that path. I think that trade-off is the right one, but it is the judgement call in this PR, so flagging it explicitly rather than burying it.

Two things soften it:

- Verbose mode now prints which exclusions are applied, so a user debugging a missing file has a thread to pull.
- A project with no `.gitignore` is unaffected beyond the `.git`/`node_modules` floor.

If you would rather this be opt-in, or keyed off a separate `.dataformignore` instead, say so and I will rework it.

## Notes on the implementation

- `lstatSync`, not `statSync`, classifies directories for trailing-slash patterns like `.venv/`. `statSync` follows symlinks and throws `ENOENT` on a dangling one, which is a perfectly copyable input. `lstat` also matches what `copySync` itself does by default, and matches real `.gitignore` semantics where a symlink is never itself a directory.
- Only the root `.gitignore` is read. Nested `.gitignore` files, `.git/info/exclude` and the user's global excludes are not consulted, so this under-excludes relative to `git status` rather than over-excluding. Documented in the code.
- A top-level `node_modules` cannot actually be present on this path, since `compile()` rejects the project before copying if it finds one. That floor entry therefore covers nested ones.
- No lockfile regeneration is needed. `ignore` is already resolved in `yarn.lock` at 5.3.2 as an existing transitive dependency, and the declared `^5.2.0` range matches that entry.

## Testing

Six new unit tests in `cli/api/commands/compile_copy_filter_test.ts`, covering the no-`.gitignore` case, `.gitignore` patterns alongside the floor, negation patterns against the floor, dangling symlinks, in-project paths beginning with two dots, and one integration-style test that runs a real `copySync` and asserts on the resulting directory.

All test targets in the package pass:

```
bazel test //cli/api/...
Executed 11 out of 11 tests: 11 tests pass.
```

`./scripts/lint` is clean.

### DIFF
--- a/cli/api/commands/compile.ts
+++ b/cli/api/commands/compile.ts
@@ -5,6 +5,7 @@ import * as tmp from "tmp";
 import { promisify } from "util";
 
 import { BaseWorker } from "df/cli/api/commands/base_worker";
+import { buildProjectCopyFilter } from "df/cli/api/commands/compile_copy_filter";
 import { MISSING_CORE_VERSION_ERROR } from "df/cli/api/commands/install";
 import { readConfigFromWorkflowSettings } from "df/cli/api/utils";
 import { DEFAULT_COMPILATION_TIMEOUT_MILLIS } from "df/cli/api/utils/constants";
@@ -52,9 +53,12 @@ export async function compile(
     if (compileConfig.verbose) {
       print(`Using isolated environment for @dataform/core@${workflowSettingsDataformCoreVersion}\n`);
       print(`Copying project to temporary directory: ${temporaryProjectPath}\n`);
+      print(`Excluding .git, node_modules, and paths matched by the project's .gitignore\n`);
     }
     const copyStartTime = performance.now();
-    fs.copySync(resolvedProjectPath, temporaryProjectPath);
+    fs.copySync(resolvedProjectPath, temporaryProjectPath, {
+      filter: buildProjectCopyFilter(resolvedProjectPath)
+    });
     if (compileConfig.verbose) {
       print(`Project copy completed in ${performance.now() - copyStartTime}ms\n`);
     }
--- a/cli/api/commands/compile_copy_filter.ts
+++ b/cli/api/commands/compile_copy_filter.ts
@@ -1,0 +1,74 @@
+import * as fs from "fs-extra";
+import ignore from "ignore";
+import * as path from "path";
+
+// Excluded whatever the project's .gitignore says. `.git` holds no Dataform project
+// files. A top-level `node_modules` can't be present at all here -- `compile()` rejects
+// the project before copying if it finds one -- so that entry covers nested ones, which
+// are likewise never part of a Dataform project.
+//
+// Checked independently of the `ignore` instance below, rather than seeded into it, so
+// that a project's .gitignore cannot override this floor: `ignore` lets later patterns
+// override earlier ones by design, so a `!node_modules` negation would otherwise
+// un-ignore it.
+const ALWAYS_IGNORED_NAMES = new Set([".git", "node_modules"]);
+
+/**
+ * Builds a filter for fs-extra's `copySync`, so the stateless-install copy in `compile()`
+ * skips files that can't be part of the Dataform project -- most commonly a large
+ * `.venv`, build-output or cache directory sitting alongside `definitions/`, whose size
+ * the copy would otherwise pay for.
+ *
+ * Exclusions come from the project's own `.gitignore` rather than from a hardcoded list
+ * of directory names: no fixed list covers every ecosystem's junk directories (`.venv`,
+ * `target/`, `__pycache__/`, `vendor/`, `coverage/`, ...), whereas a project's
+ * `.gitignore` already states exactly what that project treats as disposable, and
+ * `dataform init` writes one.
+ *
+ * Only the project root's `.gitignore` is read. Nested `.gitignore` files,
+ * `.git/info/exclude` and the user's global excludes file are not consulted, so a
+ * project relying on those has more copied than `git status` would suggest. A project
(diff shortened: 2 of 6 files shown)


rafal-hawrylak:
Please rebase and and solve conflicts

packages/@dataform/cli/BUILD
```diff
@@ -38,6 +38,7 @@ externals = [
     "fs-extra",
     "glob",
     "google-sql-syntax-ts",
+    "ignore",
```
rafal-hawrylak, line 41: yarn.lock update is missing

cli/api/commands/compile_copy_filter.ts
```diff
@@ -0,0 +1,74 @@
...
+ * `.git/info/exclude` and the user's global excludes file are not consulted, so a
+ * project relying on those has more copied than `git status` would suggest. A project
+ * with no `.gitignore` at all gets only the ALWAYS_IGNORED_NAMES floor.
+ *
+ * Note that a gitignored file is never copied, so it is also never compiled: a project
+ * that generates definitions into a gitignored path needs that path unignored.
+ */
+export function buildProjectCopyFilter(resolvedProjectPath: string): (src: string) => boolean {
+  const ig = ignore();
+  const gitignorePath = path.join(resolvedProjectPath, ".gitignore");
```
rafal-hawrylak, line 38: Supporting .gitignore by default is the right step for 99% of users. We could add a new support to a .dataformignore file (similar to .dockerignore / .eslintignore). If .dataformignore exists, it could either override or supplement .gitignore (allowing users to un-ignore paths specifically for Dataform).

cli/api/commands/compile.ts
```diff
@@ -52,9 +53,12 @@ export async function compile(
     if (compileConfig.verbose) {
       print(`Using isolated environment for @dataform/core@${workflowSettingsDataformCoreVersion}\n`);
       print(`Copying project to temporary directory: ${temporaryProjectPath}\n`);
+      print(`Excluding .git, node_modules, and paths matched by the project's .gitignore\n`);
```
rafal-hawrylak, line 56: If the project has no .gitignore, this log message still claims it is excluding paths from .gitignore. A minor conditional check (hasGitignore ? ... : ...) would be cleaner.

rafal-hawrylak (changes requested):

PR #2278, opened 2026-09-04
Support extensions on PropertyGraph

rafal-hawrylak:
Add repeated `GraphExtension {key, value}` to the PropertyGraphConfig sub-messages (entity, relationship, label, field) and the compiled counterparts. `renderOptionsClause` emits `extensions=[(k, v), ...]` alongside description and synonyms. Root-level entity and relationship extensions fold into the synthesized DEFAULT label, and non-default labels declaring extensions are rejected (existing DEFAULT-only rule). Covered by property_graph_test, execution_sql_test, a new golden and the integration fixture's INFORMATION_SCHEMA round-trip.


rafal-hawrylak:
Revoking till full BigQuery Graph capabilities of extensions support land.

PR #2282, opened 2026-09-05
Preserve explicit includeDependentAssertions=false on PropertyGraph refs

rafal-hawrylak:
Preserve explicit includeDependentAssertions=false on PropertyGraph refs (#2281)

The dataSourceRef branch used a truthy check when copying includeDependentAssertions onto the raw target. A user setting the flag to false was indistinguishable from not setting it, so checkAssertionsForDependency fell back to the graph-level dependOnDependencyAssertions and silently flipped the value to true.

Switch to hasOwnProperty to match configTargetToCompiledGraphTarget in utils.ts. Add a regression test covering the graph-level-true and ref-level-false combination.

Porting the changes from PR (#2281)

### DIFF
--- a/core/actions/property_graph.ts
+++ b/core/actions/property_graph.ts
@@ -256,7 +256,7 @@ export class PropertyGraph extends ActionBuilder<dataform.PropertyGraph> {
       if (ref.database) {
         rawRef.database = ref.database;
       }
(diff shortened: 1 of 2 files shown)


apilaskowski (approved):
If tests are passing I am ok with this change :)

PR #2285, opened 2026-09-07
Eliminate hardcoded project names for integration tests

Ford544:
Replace hardcoded references to the default dataform-open-source project with a constant in cli/index_test_base, set based on the contents of test_credentials/bigquery.json if exists, and the defaults (dataform-open-source, US location) otherwise.

Update the Integration Test section of contributing.md.

### DIFF
--- a/cli/index_test_base.ts
+++ b/cli/index_test_base.ts
@@ -4,10 +4,14 @@ import * as fs from "fs-extra";
 import { dump as dumpYaml, load as loadYaml } from "js-yaml";
 import * as path from "path";
 
+import { Logger } from "df/cli/console";
 import { version } from "df/core/version";
 import { dataform } from "df/protos/ts";
 import { corePackageTarPath, getProcessResult, nodePath, npmPath } from "df/testing";
 import { TmpDirFixture } from "df/testing/fixtures";
+
+const DEFAULT_PROJECT = "dataform-open-source";
+const DEFAULT_LOCATION = "US";
 
 const runfilesDir = process.env.RUNFILES;
 let workspaceName = "df";
@@ -16,6 +20,8 @@ if (!fs.existsSync(path.resolve(runfilesDir, "df"))) {
 }
 
 export const CREDENTIALS_PATH = path.resolve(runfilesDir, workspaceName, "test_credentials/bigquery.json");
+
+const logger = new Logger(true);
 
 function getCredentialsProjectId(): string {
   try {
@@ -28,7 +34,8 @@ function getCredentialsProjectId(): string {
   } catch (e) {
     // Fall back to default
   }
-  return "dataform-open-source";
+  logger.log(`Project name not specified; defaulting to ${DEFAULT_PROJECT}`);
+  return DEFAULT_PROJECT;
 }
 
 function getCredentialsLocation(): string {
@@ -42,12 +49,13 @@ function getCredentialsLocation(): string {
   } catch (e) {
     // Fall back to default
   }
-  return "US";
+  logger.log(`Location not specified; defaulting to ${DEFAULT_LOCATION}`);
+  return DEFAULT_LOCATION;
 }
 
-export const DEFAULT_DATABASE = getCredentialsProjectId();
-export const DEFAULT_LOCATION = getCredentialsLocation();
-export const DEFAULT_RESERVATION = `projects/${DEFAULT_DATABASE}/locations/${DEFAULT_LOCATION.toLowerCase()}/reservations/dataform-test`;
+export const INTEGRATION_TEST_PROJECT = getCredentialsProjectId();
+export const INTEGRATION_TEST_LOCATION = getCredentialsLocation();
+export const INTEGRATION_TEST_RESERVATION = `projects/${INTEGRATION_TEST_PROJECT}/locations/${INTEGRATION_TEST_LOCATION.toLowerCase()}/reservations/dataform-test`;
 
 export const cliEntryPointPath = "cli/node_modules/@dataform/cli/bundle.js";
 
@@ -59,7 +67,7 @@ export async function setupJitProject(
   const packageJsonPath = path.join(projectDir, "package.json");
 
   await getProcessResult(
-    execFile(nodePath, [cliEntryPointPath, "init", projectDir, DEFAULT_DATABASE, DEFAULT_LOCATION])
+    execFile(nodePath, [cliEntryPointPath, "init", projectDir, INTEGRATION_TEST_PROJECT, INTEGRATION_TEST_LOCATION])
   );
 
   const workflowSettingsPath = path.join(projectDir, "workflow_settings.yaml");
--- a/cli/index_test_base.ts
+++ b/cli/index_test_base.ts
@@ -9,10 +9,6 @@ import { dataform } from "df/protos/ts";
 import { corePackageTarPath, getProcessResult, nodePath, npmPath } from "df/testing";
 import { TmpDirFixture } from "df/testing/fixtures";
 
-export const DEFAULT_DATABASE = "dataform-open-source";
-export const DEFAULT_LOCATION = "US";
-export const DEFAULT_RESERVATION = "projects/dataform-open-source/locations/us/reservations/dataform-test";
-
 const runfilesDir = process.env.RUNFILES;
 let workspaceName = "df";
 if (!fs.existsSync(path.resolve(runfilesDir, "df"))) {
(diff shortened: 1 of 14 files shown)


cli/index_test_base.ts
```diff
@@ -21,6 +17,38 @@ if (!fs.existsSync(path.resolve(runfilesDir, "df"))) {
...
+function getCredentialsLocation(): string {
+  try {
+    if (fs.existsSync(CREDENTIALS_PATH)) {
+      const parsed = JSON.parse(fs.readFileSync(CREDENTIALS_PATH, "utf8"));
+      if (parsed?.location) {
+        return parsed.location;
+      }
+    }
+  } catch (e) {
+    // Fall back to default
```
apilaskowski, line 50: Can we log sth here, so user knows that we are using default here? (The same above).

cli/tests/jit/index_jit_dependency_test.ts
```diff
@@ -51,7 +52,7 @@ suite("JiT support dependencies", ({ afterEach }) => {
     expect(executedGraph.actions.some((a: any) => a.target.name === "table_a")).to.equal(true);
     const actionB = executedGraph.actions.find((a: any) => a.target.name === "table_b");
     expect(actionB).to.not.equal(undefined);
-    expect(actionB.tasks[0].compiledSql).to.include("SELECT '`dataform-open-source.dataform.table_a`' as ref_name");
+    expect(actionB.tasks[0].compiledSql).to.include(`SELECT '\`${DEFAULT_DATABASE}.dataform.table_a\`' as ref_name`);
```
apilaskowski: Should we also leave a note that those tables needs to be prepared beforehand?

tests/integration/bigquery.spec.ts
```diff
@@ -29,7 +30,7 @@ suite("@dataform/integration/bigquery", { parallel: true }, ({ before, after })
 
       // Drop schemas to make sure schema creation works.
       await dbadapter.execute(
-        "drop schema if exists `dataform-open-source.df_integration_test_project_e2e` cascade"
```
apilaskowski, line 32: I wonder if we should also extract this one?

tests/integration/bigquery.spec.ts
```diff
@@ -531,7 +532,7 @@ defaultDataset: df_integration_test_jit
       
       // Drop dataset to start fresh
       await dbadapter.execute(
-        "drop schema if exists `dataform-open-source.df_integration_test_jit` cascade"
+        `drop schema if exists \`${DEFAULT_DATABASE}.df_integration_test_jit\` cascade`
```
apilaskowski: About formatting of those strings, can we keep them with "" as outer quotes, and `` for inner ones? I am not adamant about this, but I don't like that we need to escape those here (and everywhere). 
Also, lets check what is the standard way of handling this in the codebase in general before committing to any.

contributing.md
```diff
@@ -44,21 +44,18 @@ bazel test //core/...
 
 To run the CLI integration test against your own GCP project:
 
-1. Comment out the following dependency in `cli/BUILD`:
+1. Overwrite the contents of `test_credentials/BUILD` with:
```
apilaskowski: Can we not prepare this beforehand, so we don't need this step here?

contributing.md
```diff
@@ -70,10 +67,10 @@ To run the CLI integration test against your own GCP project:
    }
    ```
 
-3. Run the test:
+3. Run the tests:
 
    ```bash
-   bazel test //cli:index_test
+   ./scripts/run_integration_tests
```
apilaskowski, line 74: I am not sure if 

```
bazel test //cli:index_test
```

is equivalent to 

```
#!/bin/bash
set -e

# Run tslint.
bazel run @nodejs//:yarn
bazel build @npm//tslint/bin:tslint && bazel-bin/external/npm/tslint/bin/tslint.sh --project .

# Run integration tests
bazel test ... \
  --test_tag_filters=integration \
  --build_tests_only \
  --test_env=USE_CLOUD_BUILD_NETWORK=true
```

Are we for sure executing the same tests after this change?
If not, lets prepare this as separate step in this readme.

cli/index_test_base.ts
```diff
@@ -21,6 +17,38 @@ if (!fs.existsSync(path.resolve(runfilesDir, "df"))) {
...
+  }
+  return "US";
+}
+
+export const DEFAULT_DATABASE = getCredentialsProjectId();
```
apilaskowski: This is not `default` most of the time now.
Lets change this name (I am voting to consider changing it to sth that includes `project` instead - I think that from perspective of GCP user this is more intuitive as a project).

@rafal-hawrylak wdyt?

cli/index_test_base.ts
```diff
@@ -21,6 +17,38 @@ if (!fs.existsSync(path.resolve(runfilesDir, "df"))) {
...
+export const DEFAULT_LOCATION = getCredentialsLocation();
```
apilaskowski: This is also not `default` anymore (most of the time).

cli/index_test_base.ts
```diff
@@ -21,6 +17,38 @@ if (!fs.existsSync(path.resolve(runfilesDir, "df"))) {
...
+export const DEFAULT_RESERVATION = `projects/${DEFAULT_DATABASE}/locations/${DEFAULT_LOCATION.toLowerCase()}/reservations/dataform-test`;
```
apilaskowski: same as above

apilaskowski (changes requested):

cli/tests/jit/index_jit_dependency_test.ts
Ford544: They don't need to be, it worked fine for me without setting anything up on the project. I believe this particular test does not even create anything since it uses the --dry-run flag

cli/index_test_base.ts
Ford544: Makes sense, I would also change it if @rafal-hawrylak  agrees

tests/integration/bigquery.spec.ts
Ford544: Not sure what do you mean here? The hardcoded dataform-open-source gets removed and replaced with the constant

contributing.md
Ford544: Currently, this file instructs bazel to decrypt the test_credentials/bigquery.json.enc file and use it as credentials - I believe we still want to preserve this workflow for those contributors who have access to the default project? But maybe we could use the unencrypted file if it exists, and the encrypted one otherwise, that should cover both workflows.

contributing.md
Ford544: It is not equivalent, but the original command does not work anyway (because the file //cli:index_test no longer exists, it's been split into multiple ones). So we either use the script, or point to one of the files which only includes some of the tests contained in the original //cli:index_test.

tests/integration/bigquery.spec.ts
apilaskowski: I ment this part after first constant `df_integration_test_assertions_project_e2e`, because on first glance I though it is more repetive acorss tests, but I see that there are not identical, so lets skip this one.

contributing.md
apilaskowski: I wanted to avoid copy-pasting step, which can be done automatically. 
Since this is replacing a file with static content we can figure out a way to skip this step and do it automatically somehow. 

We can live with this manual step, but if there is a way to automate this to be backward compatible then lets do it.

contributing.md
apilaskowski: Since we are updating this README (and this was apparently missed when split happened) lets mention both testing scripts.

So lets add an entry for `./scripts/run_tests` for unit tests. 
And lets clearly state which script tests what :)

tests/integration/bigquery.spec.ts
```diff
@@ -503,7 +504,7 @@ suite("@dataform/integration/bigquery", { parallel: true }, ({ before, after })
     fs.writeFileSync(
       path.join(projectDir, "workflow_settings.yaml"),
       `
-defaultProject: dataform-open-source
+defaultProject: ${DEFAULT_DATABASE}
 defaultLocation: US
```
rafal-hawrylak: Lets extract DEFAULT_LOCATION as well

cli/index_test_base.ts
rafal-hawrylak: Agreed

cli/index_test_base.ts
rafal-hawrylak: There is no default reservation living in custom database / location. What is the test outcome when user customizes those and run tests?

rafal-hawrylak: In other words `reservations/dataform-test` exists only in `dataform-open-source/locations/us`. By default developers won't have it in their databases and locations.

tests/integration/bigquery.spec.ts
```diff
@@ -7,6 +7,7 @@ import * as dfapi from "df/cli/api";
 import * as dbadapters from "df/cli/api/dbadapters";
 import { BigQueryDbAdapter } from "df/cli/api/dbadapters/bigquery";
 import { ExecutionSql } from "df/cli/api/dbadapters/execution_sql";
+import { DEFAULT_DATABASE } from "df/cli/index_test_base"
```
rafal-hawrylak: nit: missing semicolon

rafal-hawrylak (changes requested):

tests/integration/bigquery.spec.ts
Ford544: This is the only way unfortunately, as TS requires backticks for variable embedding, and BigQuery does not allow other types of quotes for identifiers.

cli/index_test_base.ts
Ford544: Added. Note that by default, the bazel test command in ./scripts/run_integration_tests will not print those messages to console, only to .log file - if that's not the desired behavior, we can add --test_output=all flag to the script

contributing.md
Ford544: Done. The BUILD script will pick up a bigquery.json file if provided, and use the bundled encrypted one otherwise

contributing.md
Ford544: Updated contributing.md

cli/index_test_base.ts
Ford544: Changed to INTEGRATION_TEST_PROJECT, INTEGRATION_TEST_LOCATION and INTEGRATION_TEST_RESERVATION, while DEFAULT_PROJECT and DEFAULT_LOCATION refer to the hardcoded fallbacks for when no bigquery.json is provided by the user

tests/integration/bigquery.spec.ts
Ford544: Done

cli/index_test_base.ts
Ford544: Currently, all the tests that use this do --dry-runs, so the reservation does not actually need to exist. But if there are plans to expand test coverage it might become a problem; I suppose we could then have to add a note to contributing.md, and maybe expose some option in the script to disable tests that require specific infrastructure on the project?

tests/integration/bigquery.spec.ts
Ford544: Fixed

apilaskowski (approved):

rafal-hawrylak (approved):

PR #2286, opened 2026-09-07
refactor(cli): decompose monolithic cli/index.ts into modular command modules

mbiernacik:
Splits the large `cli/index.ts` file into separate files for each command under `cli/commands/`.
Common options and project config flags were also moved into helper files. Now `cli/index.ts` only sets up the CLI and registers each command.
There are no behavior or API changes—all commands and options work exactly the same as before.


cli/commands/run_command.ts
```diff
@@ -0,0 +1,376 @@
...
+      "per-model cap is applied and only --execution-timeout bounds JiT work. " +
+      "Examples: '30s', '2m'.",
+    type: "string",
+    default: null,
+    coerce: (rawTimeoutString: string | null) =>
+      rawTimeoutString ? parseDuration(rawTimeoutString) : null
+  }
+};
+
+const jobPrefixOption: INamedOption<yargs.Options> = {
```
apilaskowski, line 115: While this PR is technically not about fixes, lets do those obvious ones while we are at it.

This `jobPrefixOption` is not properly passed to `options` and is not displayed when `help` flag is requested.

Lets just add it to list of other options :)

cli/project_config_options.ts
```diff
@@ -0,0 +1,166 @@
...
+    name: "assertion-schema",
+    option: {
+      describe: "Default assertion schema. If unset, the value from workflow_settings.yaml is used."
+    }
+  };
+
+  public static databaseSuffix: INamedOption<yargs.Options> = {
+    name: "database-suffix",
+    option: {
+      describe: "Default assertion schema. If unset, the value from workflow_settings.yaml is used."
```
apilaskowski: This is not an accurate description (copy-paste from one above).

This was also already here, but since I noticed it lets fix it here.

I suggest using `A suffix to be appended to output database names. If unset, the value from workflow_settings.yaml is used.`

cli/commands/run_command.ts
```diff
@@ -0,0 +1,376 @@
...
+  processFn: async argv => {
+    const isJsonOutput = argv[jsonOutputOption.name];
+    const logger = new Logger(!isJsonOutput);
+
+    if (isJsonOutput && !argv[dryRunOptionName]) {
+      printError(
+        `For execution, the --${jsonOutputOption.name} option is only supported if the ` +
+          `--${dryRunOptionName} option is enabled`
+      );
+      return;
```
apilaskowski: This should return numeric value. Lets return `1` here to adhere to promised return value (numeric).

apilaskowski (changes requested):
Overall LGTM from me, lets just fix those small things before submitting.

cli/common_options.ts
```diff
@@ -0,0 +1,106 @@
...
+};
+
+export const timeoutOption: INamedOption<yargs.Options> = {
+  name: "timeout",
+  option: {
+    describe: "Duration to allow project compilation to complete. Examples: '1s', '10m', etc.",
+    type: "string",
+    default: null,
+    coerce: (rawTimeoutString: string | null) =>
+      rawTimeoutString ? parseDuration(rawTimeoutString) : null
```
rafal-hawrylak: This already was there but it has multiple copies and could be extracted:
```
coerce: (rawTimeoutString: string | null) =>
      rawTimeoutString ? parseDuration(rawTimeoutString) : null
```

cli/common_options.ts
```diff
@@ -0,0 +1,106 @@
...
+export const actionsOption: INamedOption<yargs.Options> = {
+  name: "actions",
+  option: {
+    describe: "A list of action names or patterns to run. Can include '*' wildcards.",
+    type: "array",
+    coerce: splitCommas
+  }
+};
+
+export function getCredentialsPath(projectDir: string, credentialsPath: string) {
```
rafal-hawrylak: This wrapper is pointless.

cli/common_options.ts
```diff
@@ -0,0 +1,106 @@
...
+          argv[projectDirOption.name]
+        } does not appear to be a dataform directory (missing workflow_settings.yaml file).`
+      );
+    }
+  }
+};
+
+// Splits repeated and comma-separated values into a flat list, e.g.
+// `--actions a,b --actions c` -> ["a", "b", "c"].
+export const splitCommas = (raw: string[] | null) => raw.map(value => value.split(",")).flat();
```
rafal-hawrylak: Again - already existing but it will fail when null will be passed and null is allowed here

rafal-hawrylak (approved):
only pre-existing nits

cli/commands/run_command.ts
mbiernacik: Done, added to options.

cli/project_config_options.ts
mbiernacik: Done, updated description.

cli/commands/run_command.ts
mbiernacik: Done, now returns 1.

cli/common_options.ts
mbiernacik: Done, extracted to separate function.

cli/common_options.ts
mbiernacik: Agree, wrapper was removed and used actuallyResolve directly.

cli/common_options.ts
mbiernacik: Done, added null check.

apilaskowski (approved):

PR #2287, opened 2026-09-08
Refactor cli tests

Ford544:
* extracted common operations to functions in index_test_base
* extracted shared setup to beforeEach in some suites
* made suite structure more consistent

Note: this PR does not cover splitting larger test files into several seperate ones, this is planned but will be done in a seperate PR.

### DIFF
--- a/cli/index_help_test.ts
+++ b/cli/index_help_test.ts
@@ -1,12 +1,11 @@
 import { expect } from "chai";
-import { execFile } from "child_process";
 
-import { cliEntryPointPath } from "df/cli/index_test_base";
-import { getProcessResult, nodePath, suite, test } from "df/testing";
+import { runCli } from "df/cli/index_test_base";
+import { suite, test } from "df/testing";
 
 suite("help command", () => {
   test("shows global help with the help command", async () => {
-    const result = await getProcessResult(execFile(nodePath, [cliEntryPointPath, "help"]));
+    const result = await runCli("help");
     expect(result.exitCode).equals(0);
     const output = result.stdout;
     expect(output).to.include("dataform [command]");
@@ -20,7 +19,7 @@ suite("help command", () => {
   });
 
   test("shows help for 'init' command", async () => {
-    const result = await getProcessResult(execFile(nodePath, [cliEntryPointPath, "help", "init"]));
+    const result = await runCli("help", "init");
     expect(result.exitCode).equals(0);
     const output = result.stdout;
     expect(output).to.include("Create a new dataform project.");
@@ -29,7 +28,7 @@ suite("help command", () => {
   });
 
   test("shows help for 'install' command", async () => {
-    const result = await getProcessResult(execFile(nodePath, [cliEntryPointPath, "help", "install"]));
+    const result = await runCli("help", "install");
     expect(result.exitCode).equals(0);
     const output = result.stdout;
     expect(output).to.include("Install a project's NPM dependencies.");
@@ -37,7 +36,7 @@ suite("help command", () => {
   });
 
   test("shows help for 'init-creds' command", async () => {
-    const result = await getProcessResult(execFile(nodePath, [cliEntryPointPath, "help", "init-creds"]));
+    const result = await runCli("help", "init-creds");
     expect(result.exitCode).equals(0);
     const output = result.stdout;
     expect(output).to.include("Create a .df-credentials.json file for Dataform to use when accessing BigQuery.");
@@ -47,7 +46,7 @@ suite("help command", () => {
   });
 
   test("shows help for 'compile' command", async () => {
-    const result = await getProcessResult(execFile(nodePath, [cliEntryPointPath, "help", "compile"]));
+    const result = await runCli("help", "compile");
     expect(result.exitCode).equals(0);
     const output = result.stdout;
     expect(output).to.include("Compile the dataform project.");
@@ -57,7 +56,7 @@ suite("help command", () => {
   });
 
   test("shows help for 'test' command", async () => {
-    const result = await getProcessResult(execFile(nodePath, [cliEntryPointPath, "help", "test"]));
+    const result = await runCli("help", "test");
     expect(result.exitCode).equals(0);
     const output = result.stdout;
     expect(output).to.include("Run the dataform project's unit tests.");
@@ -69,7 +68,7 @@ suite("help command", () => {
   });
 
   test("shows help for 'run' command", async () => {
-    const result = await getProcessResult(execFile(nodePath, [cliEntryPointPath, "help", "run"]));
+    const result = await runCli("help", "run");
     expect(result.exitCode).equals(0);
     const output = result.stdout;
     expect(output).to.include("Run the dataform project.");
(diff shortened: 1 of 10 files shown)


cli/index_test_base.ts
```diff
@@ -96,10 +100,51 @@ export async function setupJitProject(
...
+export async function alterWorkflowSettings(projectDir: string, workflowSettingsOverrides: Partial<dataform.IWorkflowSettings>): Promise<void> {
+  const workflowSettingsPath = path.join(projectDir, "workflow_settings.yaml");
+  const workflowSettings = dataform.WorkflowSettings.create(
+    loadYaml(fs.readFileSync(workflowSettingsPath, "utf8"))
+  );
+  const workflowSettingsNew = dataform.WorkflowSettings.create({
+    ...workflowSettings,
+    ...workflowSettingsOverrides
+  });
+  fs.writeFileSync(workflowSettingsPath, dumpYaml(workflowSettingsNew));
```
apilaskowski: Direct YAML dumping of a protobufjs instance: dumpYaml(workflowSettingsNew) dumps the class instance directly without calling dataform.WorkflowSettings.toObject(..., { enums: String }). This can serialize protobuf internal properties or dump enum values as numeric integers instead of string names.

cli/index_test_base.ts
```diff
@@ -96,10 +100,51 @@ export async function setupJitProject(
...
+  fs.writeFileSync(fullPath, content);
+}
+
+export async function alterWorkflowSettings(projectDir: string, workflowSettingsOverrides: Partial<dataform.IWorkflowSettings>): Promise<void> {
+  const workflowSettingsPath = path.join(projectDir, "workflow_settings.yaml");
+  const workflowSettings = dataform.WorkflowSettings.create(
+    loadYaml(fs.readFileSync(workflowSettingsPath, "utf8"))
+  );
+  const workflowSettingsNew = dataform.WorkflowSettings.create({
+    ...workflowSettings,
```
apilaskowski: Object spread on proto instance: Spreading ...workflowSettings into dataform.WorkflowSettings.create(...) copies instance-level own properties rather than pure schema values. Spreading the plain object loaded from YAML is cleaner.

cli/index_test_base.ts
```diff
@@ -96,10 +100,51 @@ export async function setupJitProject(
...
+  );
+}
+
+export function writeDefinitionFile(projectDir: string, filename: string, content: string): void {
+  const fullPath = path.join(projectDir, "definitions", filename);
+  fs.ensureFileSync(fullPath);
+  fs.writeFileSync(fullPath, content);
+}
+
+export async function alterWorkflowSettings(projectDir: string, workflowSettingsOverrides: Partial<dataform.IWorkflowSettings>): Promise<void> {
```
apilaskowski: All file system calls are synchronous; no await is used

apilaskowski: Suggested fix accommodating mentioned issues: 

```
    export function alterWorkflowSettings(
      projectDir: string,
      workflowSettingsOverrides: Partial<dataform.IWorkflowSettings>
    ): void {
      const workflowSettingsPath = path.join(projectDir, "workflow_settings.yaml");
      const existingSettings = loadYaml(fs.readFileSync(workflowSettingsPath, "utf8")) as dataform.IWorkflowSettings;
      const workflowSettings = dataform.WorkflowSettings.create({
        ...existingSettings,
        ...workflowSettingsOverrides
      });
      fs.writeFileSync(
        workflowSettingsPath,
        dumpYaml(dataform.WorkflowSettings.toObject(workflowSettings, { enums: String }))
      );
    }
```

This also aligns with what is currently proposed in `setupProject`.

cli/index_help_test.ts
```diff
@@ -20,7 +19,7 @@ suite("help command", () => {
   });
 
   test("shows help for 'init' command", async () => {
-    const result = await getProcessResult(execFile(nodePath, [cliEntryPointPath, "help", "init"]));
+    const result = await runCli("help", "init");
```
apilaskowski: Here "init" is passed as `dir`, which is second parameter of `runCli`.

Either pass subcommands via options or provide an overload. I am not a fan of `argsOrDir` options, so lets figure out a way to solve this cleanly.

cli/index_test_base.ts
```diff
@@ -96,10 +100,51 @@ export async function setupJitProject(
...
+  dir?: string,
+  options: string[] = [],
+  execOptions?: ExecFileOptions
+): Promise<{
+  exitCode: number;
+  stdout: string;
+  stderr: string;
+}> {
+  const args = [cliEntryPointPath, cmd, ...(dir !== undefined ? [dir] : []), ...options];
+  return await getProcessResult(
```
apilaskowski: I think you can drop `await` here, and `return getProcessResult(...)` will be sufficient.

cli/index_run_e2e_test.ts
```diff
@@ -193,56 +160,24 @@ select 1 as \${dataform.projectConfig.vars.testVar2}
...
-          "install",
-          "--prefix",
-          projectDir,
-          "--cache",
-          npmCacheDir,
-          corePackageTarPath
-        ])
-      );
+    beforeEach("setup test project", async () => {
+      projectDir = tmpDirFixture.createNewTmpDir()
```
apilaskowski: Missing `;`

cli/index_run_e2e_test.ts
```diff
@@ -316,20 +239,19 @@ SELECT 1 as id
     };
 
     test("with --disable-assertions flag", async () => {
-      await setUpWorkflowSettings(false);
+      await alterWorkflowSettings(projectDir, { disableAssertions : false });
```
apilaskowski: Linter should have fixed that, but lets be consistent and not have spaces before colons.

cli/index_run_e2e_test.ts
```diff
@@ -365,21 +286,20 @@ SELECT 1 as id
     });
 
     test("with --job-labels flag", async () => {
-      await setUpWorkflowSettings(false);
+      await alterWorkflowSettings(projectDir, { disableAssertions : false });
```
apilaskowski: Also no space before colon.

apilaskowski (changes requested):

cli/index_test_base.ts
Ford544: Agreed, fix incorporated

cli/index_help_test.ts
Ford544: Good point. I think the best way will be to remove the dir parameter altogother and pass it as part of options, having it a special parameter unnecessarily constraints the function.

cli/index_test_base.ts
Ford544: Dropped

cli/index_run_e2e_test.ts
Ford544: Addressed this and the semicolon spaces

cli/index_test_base.ts
Ford544: Fixed

apilaskowski (approved):

PR #2301, opened 2026-09-11
Split core/main_test.ts

Ford544:
* Split core/main_test.ts into several smaller files
* Extract the writeDefinitionFile to testing dir


apilaskowski (approved):
Great refactoring. Splitting up main_test.ts significantly improves readability and test maintenance. Nice catch converting the filenames with multiple dots cause compilation errors block to a proper test(...).

The PR description mentions extracting writeDefinitionFile, but writeWorkflowSettingsFile was also added (which was very helpful in cleaning up workflow_settings.yaml boilerplate).

PR #2302, opened 2026-09-11
Split core/actions/table_test and core/actions/incremental_table_test

Ford544:
Split two large test files, core/actions/table_test.ts and core/actions/incremental_table_test.ts. The two files have very similar structure and were split in the same way.


apilaskowski (approved):
Verified that all tests are preserved, boilerplate eliminated, good split.

PR #2304, opened 2026-09-11
Refactor core/main_property_graphs_test.ts

Ford544:
The file core/main_property_graphs_test.ts contains a large number of very similar tests, which all execute the same workflows with different inputs and outputs. Storing the test configurations in a list and creating tests in a loop allows removing repeating boilerplate code and increasing readability.

### DIFF
--- a/core/main_property_graphs_test.ts
+++ b/core/main_property_graphs_test.ts
@@ -33,13 +33,42 @@ suite("property graphs", ({ afterEach }) => {
     ...extra
   });
 
-  test("valid graph.yaml compiles end-to-end", () => {
-    const projectDir = tmpDirFixture.createNewTmpDir();
-    writeWorkflowSettingsFile(projectDir, VALID_WORKFLOW_SETTINGS_YAML);
-    writeDefinitionFile(
-      projectDir,
-      "graph.yaml",
-      `
+  const missingRefTarget = {
+    schema: "defaultDataset",
+    name: "MissingRefGraph",
+    database: "defaultProject"
+  };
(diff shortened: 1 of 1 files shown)


core/main_property_graphs_test.ts
```diff
@@ -1761,7 +1573,77 @@ entities:
...
+    test(testParameters.testName, () => {
+      const projectDir = tmpDirFixture.createNewTmpDir();
+      writeWorkflowSettingsFile(projectDir, testParameters.workflowSettings);
+      testParameters.definitionFiles.forEach(file => {
+        writeDefinitionFile(projectDir, file.name, file.contents);
+      });
+
+      const result = runMainInVm(coreExecutionRequestFromPath(projectDir));
+
+      if (testParameters.expectedGraph) {
```
apilaskowski, line 1588: If a future contributor adds a new test case to the array and misspells the property (e.g. expectedPropertyGraph: or expectedGraphs:) or forgets an expectation property altogether, neither if block executes, and the test will silently pass without running any assertions.

Add an explicit check or throw an error if neither expectation is provided:
```
if (!testParameters.expectedGraph && !testParameters.expectedPropertyGraphs) {
      throw new Error(
        `Test case "${testParameters.testName}" must specify either expectedGraph or expectedPropertyGraphs`
      );
    }
```

apilaskowski (changes requested):

core/main_property_graphs_test.ts
Ford544: Added a check + explicit typing for the test cases

apilaskowski (approved):

PR #2312, opened 2026-09-16
Enforce prettier formatting standards on PRs

Ford544:
Add a formatting check on pull request creation to prevent unformatted files from being published to the repo.
* Add a format check to the lint script - will result in a failure if a .ts file that has been modified since the branching point does not meet prettier standards
* Add the lint script as a step before tests in the run_tests_on_cloudbuild script (invoked on PR publish)
* Add format script as a shortcut to run prettier on all files modified since the branching point
* Add a note on formatting to contributing.md

Tested using a draft PR (#2309) - verified that the check fails for commits with unformatted files and passes for commits with all modified files formatted.


scripts/lint
```diff
@@ -1,6 +1,14 @@
 #!/bin/bash
+set -e
 
 bazel run @nodejs//:yarn
```
apilaskowski: Not necessarily something to fix here, but worth noting now that this script runs in CI: this is an unfrozen `yarn install`, while `run_tests_on_cloudbuild` installs with `--frozen-lockfile` just above it. The same bare `yarn` is already in `run_tests` and `run_integration_tests`, so it's a pre-existing pattern rather than anything new.

What does change is that prettier becomes load-bearing in CI, and `package.json` has it as `^1.14.2` — a re-resolve can float it within 1.x, and 1.14 → 1.19 changed formatting defaults. That could give format failures on PRs unrelated to their contents that don't reproduce locally. Pinning prettier to an exact version would sidestep it.

scripts/format
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+bazel run @nodejs//:yarn
+
+# Run prettier on .ts files files changed since the branching point
+CHANGED_TS_FILES=$(git diff --name-only --diff-filter=ACMR "$(git merge-base origin/main HEAD)" -- '*.ts')
```
apilaskowski: `scripts/lint` picked up `set -e` but this one didn't, and I think the difference matters here.

If `git merge-base origin/main HEAD` fails — most likely because `origin/main` doesn't exist locally — the inner substitution yields an empty string, the outer `git diff` errors on the empty rev, `CHANGED_TS_FILES` comes back empty, the `if [ -n ... ]` guard below is false, and the script exits 0 having formatted nothing.

That's a slightly unfortunate failure mode for this particular script, since it's the one contributors are told to run: no output looks like success, so they'd push and then get rejected by CI — the loop this PR is closing. Something like:

```bash
set -euo pipefail
MERGE_BASE=$(git merge-base origin/main HEAD) || {
  echo "Could not determine merge base with origin/main. Try: git fetch origin main" >&2
  exit 1
}
```

Related, and worth considering for both scripts: `origin/main` is hardcoded. In the usual fork workflow `origin` is the contributor's own fork and `upstream` is this repo, so `origin/main` could be weeks or months stale, and both scripts would then operate on a much larger file set than the PR actually touches. CI handles this by fetching explicitly, but `scripts/format` never refreshes anything. Falling back might be enough:

```bash
BASE_REF=$(git rev-parse --verify --quiet upstream/main || echo origin/main)
```

Or just a line in contributing.md noting that `origin/main` needs to be current.

scripts/run_tests_on_cloudbuild
```diff
@@ -6,6 +6,11 @@ bazel run //tools/registry-tools:switch_registry -- $(pwd)/yarn.lock https://us-
 # Install from trusted registry, prohibit lockfile changes.
 bazel run @nodejs//:yarn install -- --frozen-lockfile
 
+# Run lint and format checks
+# Need origin/main to check which files to format
+git fetch origin main --unshallow 2>/dev/null || git fetch origin main
```
apilaskowski: `cloudbuild-test.yaml` has no explicit checkout step, so the git setup here is whatever the trigger gives us — which is an argument for this line not depending on it. Three concrete things:

1. **Make the refspec explicit.** This relies on `git fetch origin main` *opportunistically* updating `refs/remotes/origin/main`, which only happens when `remote.origin.fetch` is the standard wildcard. Naming the destination ref removes the question.
2. **Assert the ref exists.** Otherwise the first sign of trouble is `merge-base` failing inside `scripts/lint`, a file away from the actual cause.
3. **Don't blanket-silence stderr.** `2>/dev/null` hides auth and network failures alongside the expected "not a shallow repository" message, and the `||` then masks them a second time.

```bash
git fetch --unshallow origin +refs/heads/main:refs/remotes/origin/main 2>/dev/null \
  || git fetch origin +refs/heads/main:refs/remotes/origin/main
git rev-parse --verify --quiet origin/main >/dev/null \
  || { echo "origin/main unavailable; cannot compute changed files" >&2; exit 1; }
```

Separately: `--unshallow` pulls the full history on every PR build. A bounded `--depth` deep enough to reach the fork point would be cheaper, if you can pick a bound you trust.

scripts/lint
```diff
@@ -1,6 +1,14 @@
...
 
 # Run eslint.
 bazel build @npm//eslint/bin:eslint && bazel-bin/external/npm/eslint/bin/eslint.sh --rulesdir eslint-rules '**/*.ts'
+
+# Run prettier on .ts files changed against origin/main
+CHANGED_TS_FILES=$(git diff --name-only --diff-filter=ACMR "$(git merge-base origin/main HEAD)" -- '*.ts')
+
+if [ -n "$CHANGED_TS_FILES" ]; then
+  bazel build @npm//prettier/bin:prettier && bazel-bin/external/npm/prettier/bin/prettier.sh --ignore-path .eslintignore --check $CHANGED_TS_FILES
```
apilaskowski: Two small things here:

`$CHANGED_TS_FILES` is unquoted, so paths with spaces or glob characters would break (SC2086). `git diff -z ... | xargs -0` sidesteps it, and makes the `if` guard above unnecessary.

Also, `--ignore-path .eslintignore` couples the two tools together — since there's already a `.prettierrc`, a `.prettierignore` might be tidier.

contributing.md
```diff
@@ -76,12 +76,20 @@ To run the CLI integration test against your own GCP project:
...
 ./scripts/lint
 ```
 
+All .ts files are expected to be formatted using prettier. You can format all files modified in your current branch by using this script:
+
+```bash
+./scripts/format
+```
+
+Alternatively, you may want to configure prettier to run on save or as a pre-commit hook.
```
apilaskowski, line 91: Would it be worth adding a sentence here about what happens on files that predate prettier?

`prettier --check` works per file rather than per hunk, so a one-line change to a large unformatted file will fail the check and require reformatting the whole thing — turning a small fix into a several-hundred-line diff and making `git blame` less useful. That's inherent to the incremental approach and may well be an acceptable trade, I just think contributors will hit it and be surprised.

Lets also put it clearly into PR message.

scripts/format
```diff
@@ -0,0 +1,10 @@
...
+
+if [ -n "$CHANGED_TS_FILES" ]; then
+  bazel build @npm//prettier/bin:prettier && bazel-bin/external/npm/prettier/bin/prettier.sh --ignore-path .eslintignore --write $CHANGED_TS_FILES
+fi
```
apilaskowski: Btw, in those we sometimes DO have newlines at the end, and sometimes we do not, right? Lets keep it consistent.

apilaskowski (changes requested):

apilaskowski:
Also remember to clean PR (https://github.com/dataform-co/dataform/pull/2309) after we are done here to keep queue clean.

apilaskowski:
Thanks for doing this — enforcing formatting is one of those things that only gets harder to introduce the longer it's left, and I like that you've scoped it to changed files rather than reformatting the world in one go. Nice touch validating it against a draft PR too.

I've marked this as changes requested, though nothing here is fundamental. The two I'd most like addressed are the missing `set -e` in `scripts/format` — without it, a failed `merge-base` leaves the file list empty and the script exits 0 having formatted nothing — and the `origin/main` fetch in `run_tests_on_cloudbuild`. Both are inline with concrete suggestions.

Two smaller things that didn't fit on a line:

- **Duplicated merge-base logic.** The same three lines now live in both `scripts/lint` and `scripts/format`, differing only in `--check` vs `--write`. The comments above them have already drifted slightly ("changed against origin/main" vs "changed since the branching point"), which is usually the first sign of copy-paste divergence. Extracting a small shared helper, or making `format` a thin wrapper over `lint`, would keep them honest.
- **Prettier version.** `package.json` pins `"prettier": "^1.14.2"` — that's prettier 1.x from 2018, two majors behind. Not a blocker, and there's a reasonable argument for landing enforcement first and upgrading second. But worth knowing the eventual 2.x/3.x bump will reformat the whole codebase (arrow parens and trailing commas both changed), so it might be worth an issue to track. Pinning exactly rather than with a caret would also help with the unfrozen-install point I left inline.

Happy to be wrong on any of these — the overall direction is clearly right.

scripts/lint
Ford544: I added --frozen-lockfile to the yarn calls in run_tests, run_integration_tests, and lint. This should ensure no version drift vs CI. We can consider upgrading prettier in a seperate PR, and then pinning to a specific version just in case.

scripts/format
Ford544: Actually, the error message like "fatal: Not a valid object name origin/master
fatal: bad revision '' is already displayed if git merge-base fails so it's not exactly 'no output'. Adding set -e ensures that an error code is also emitted though, so it makes sense to add that too.

Using upstream by default and falling back to origin seems sensible.

scripts/run_tests_on_cloudbuild
Ford544: Added explicit refspec paths and throw error if origin/main is not verified otherwise. Also used if statement to check if repo is shallow to avoid --unshallow failing on complete repo; 2>/dev/null removed.

Setting any --depth introduces the risk of it not being deep enough; and if it's set to a very high number it's functionally no different from --unshallow. The repo is not so huge so I think it's alright to stick with --unshallow

scripts/lint
Ford544: Modified to use xargs. Added .prettierignore - for now it has the exact same contents as .eslintignore, but we may want to differentiate them in the future.

contributing.md
Ford544: If we are sticking with the incremental approach, do you think we should also advise contributors to split their pull requests into a pure formatting one and another for the actual change? This would be more work for the contributors but make it a bit easier to track changes later.

scripts/format
Ford544: Fixed

Ford544:
For the prettier version - for now I added --frozen-lockfile to yarn calls that will be run in CLI, which should prevent divergence. I can update prettier in the next PR, and pin the specific version then to ensure no accidental yarn.lock changes - for now it doesn't really matter since the currently installed vesion is the final one in 1.\*.\*, and the caret in specification will not permit anything >= 2.\*.\*

contributing.md
apilaskowski: I think that generally it is ok to have some formatting updates while we apply changes.

I would not recommend going into separate PRs just for formatting.

I also think that we will generate majority of movement at this point, so we likely will fight with that ourselves.

apilaskowski: The best approach would be to make it possible to fix it by chunk, not by file, but I am not sure if this is possible with current tooling.

Ford544: This might be possible in principle, as prettier has the range option to only format parts of a file. But this would probably require custom logic to get the appropriate ranges by parsing git diffs, as I don't see any built-in utility to do this, so it might get quite complicated.

Ford544: I added a note to contributing.md regarding this

scripts/format
apilaskowski: Subtle bash gotcha here (and in `scripts/lint`): even with `set -e` added at the top, this pipeline **still exits `0`** if `git merge-base` fails!

1. In bash, `set -e` ignores failures inside `$(...)` when used as an argument to another command (`git diff ... "$(git merge-base ...)"`), rather than in a standalone variable assignment.
2. `git diff` then fails with exit code `128` (`fatal: ambiguous argument ''`), but because it is on the left side of a pipe (`| xargs -0 -r ...`) without `set -o pipefail`, bash takes the exit status of `xargs -0 -r` — which receives 0 bytes on stdin and exits `0`.

Adding `set -eo pipefail` and assigning `MERGE_BASE` on its own line fixes this cleanly:
```bash
set -eo pipefail
...
MERGE_BASE=$(git merge-base "$BASE_REF" HEAD)
git diff -z --name-only --diff-filter=ACMR "$MERGE_BASE" -- '*.ts' \
  | xargs -0 -r bazel-bin/external/npm/prettier/bin/prettier.sh --ignore-path .prettierignore --write
```
(Also, consider adding `-- --frozen-lockfile` to `yarn` on line 4 here as well to match `scripts/lint`.)

If `scripts/format` takes `MODE="${1:---write}"` as its last argument, `scripts/lint` could also just call `./scripts/format --check` so we don't have to maintain this block in two places.

contributing.md
apilaskowski: Agreed — custom diff-to-range logic for prettier would be overkill here. Full-file formatting on touched legacy files is a totally reasonable trade-off (and if a massive legacy file reformat ever pollutes blame too much, we can always add the commit SHA to a `.git-blame-ignore-revs` file later). The note you added to `contributing.md` looks great!

apilaskowski (approved):

PR #2323, opened 2026-09-18
Handle GCP multi-universe

ybelleguic:
## What

Adds support for configuring the `universeDomain` used by the BigQuery and
data-lineage clients, so the Dataform CLI can connect to a non-default universe
(for example a Trusted Partner Cloud / TPC deployment) instead of the Google
Default Universe (`googleapis.com`).

## Why

`@google-cloud/bigquery` already accepts `universeDomain` on its constructor
options, but there was no way to set it from Dataform: the BigQuery client is
built in `createBigQueryClientProvider()` (`cli/api/dbadapters/bigquery.ts`)
purely from the credentials, and the credentials shape is defined by the
`BigQuery` proto. Since `.df-credentials.json` is strictly validated against
that proto (`verifyObjectMatchesProto` rejects unknown fields), users targeting
a non-default universe had no supported way to specify it.

## Changes

- `protos/profiles.proto`: add an optional `string universe_domain = 5;` field
  to the `BigQuery` credentials message.
- `cli/api/dbadapters/bigquery.ts`: forward `universeDomain` to
  `new BigQuery({ ... })` when set.
- `cli/api/lineage/emitter.ts`: forward `universeDomain` to
  `new LineageClient({ ... })` when set (google-gax `ClientOptions` supports it),
  so lineage emission targets the same universe.
- `cli/credentials.ts`: prompt for an optional universe domain during
  `dataform init-creds` (leaving it blank keeps the default).
- `contributing.md`: document the new optional field in the credentials schema.
- Tests: `cli/api/commands/credentials_test.ts` (new) covers credential parsing
  including rejection of unknown fields; `cli/api/dbadapters/bigquery_test.ts`
  verifies the value reaches the client and defaults to `googleapis.com` when
  unset.

The field is optional and empty by default, so existing credentials files and
callers are completely unaffected.

## Testing

- `bazel test //cli/api:commands/credentials_test //cli/api:dbadapters/bigquery_test` — pass.
- `./scripts/run_tests` (eslint + full Bazel suite) — pass.
- Manual: `./scripts/run run <project> --dry-run` with a `.df-credentials.json`
  containing `universeDomain` is accepted and the BigQuery request targets the
  derived endpoint (`https://bigquery.<universeDomain>/...`).


cli/api/lineage/emitter.ts
```diff
@@ -42,7 +42,8 @@ export function createLineageClientProvider(
           apiEndpoint: endpoint,
           credentials: credentials.credentials && JSON.parse(credentials.credentials),
           libName: DATAFORM_CLI_LIB_NAME,
-          libVersion: version
+          libVersion: version,
+          ...(credentials.universeDomain ? { universeDomain: credentials.universeDomain } : {})
```
apilaskowski: **Blocking:** this has no effect today, and it disables a safety check.

`LineageClient` resolves its host like this (`@google-cloud/lineage@2.1.1`, `build/src/v1/lineage_client.js:107-109`):

```js
this._universeDomain = opts?.universeDomain ?? opts?.universe_domain ?? envVar ?? 'googleapis.com';
this._servicePath    = 'datalineage.' + this._universeDomain;
const servicePath    = opts?.servicePath || opts?.apiEndpoint || this._servicePath;
```

`apiEndpoint` always wins, and the `endpoint` we pass in comes from `LineageEndpointRouter`, where both endpoints hardcode `googleapis.com`:

```ts
export const GLOBAL_LINEAGE_ENDPOINT = "datalineage.googleapis.com";
function regionalEndpointFor(location: string) { return `datalineage.${location}.rep.googleapis.com`; }
```

So on a TPC universe, lineage RPCs still target GDU hosts. Worse, gax validates the configured universe against the credential's universe before each call: today TPC creds + a GDU endpoint fail loudly, but after this change both sides report the TPC domain, the validation passes, and we silently keep dialing a GDU host with TPC credentials.

Two ways out, either works for me:

1. (preferred) Make `endpoint_router.ts` universe-aware — `datalineage.${universeDomain}` and `datalineage.${loc}.rep.${universeDomain}`, defaulting to `googleapis.com` — and thread the domain through, with coverage in `endpoint_router_test.ts`.
2. Drop this hunk, and instead skip lineage emission when `universeDomain` is set to anything other than `googleapis.com`, reusing the existing skip-reason mechanism (e.g. `skip_reason=non_default_universe`). Full support can then be a follow-up.

Either way, please add a test: `emitter_test.ts` already exists and is wired into the suite, but this hunk currently has no coverage at all.

One note if the option does stay here: the truthiness guard is genuinely required, because gax uses `??` and an empty string would set `_universeDomain = ""`. Worth a short comment so nobody "simplifies" it away later.

cli/credentials.ts
```diff
@@ -14,6 +14,11 @@ export function getBigQueryCredentials(): dataform.IBigQuery {
   if (locationIndex === 2) {
     location = question("Enter the location's region name (e.g. 'asia-south1'):");
   }
+  const universeDomain = question(
+    "Enter the universe domain to connect to, or leave blank to use the default " +
+      "('googleapis.com'). Set this only when targeting a non-default universe such as a " +
+      "Trusted Partner Cloud (TPC):"
+  ).trim();
```
apilaskowski: **Blocking (UX):** this puts an unconditional, three-line Trusted Partner Cloud question in front of every user who runs `init-creds`. Practically all of them are on GDU, so it's permanent friction for a niche field, and the prompt is far more verbose than anything else in this flow.

Preferences, in order:

1. Don't prompt at all — derive it. For the JSON-key path, read `universe_domain` from the key file (TPC service-account keys carry it); for ADC, `GOOGLE_CLOUD_UNIVERSE_DOMAIN` is already honoured by the auth layer. Anyone with an exotic setup can hand-edit `.df-credentials.json`, which your proto change alone makes possible.
2. If you do want it interactive, gate it behind the existing `selectionQuestion` pattern so the default path is unchanged.

Whatever we land here is also currently untested — `cli/index_init_test.ts` only covers the failure path.

cli/credentials.ts
```diff
@@ -40,6 +46,7 @@ export function getBigQueryCredentials(): dataform.IBigQuery {
   return {
     projectId: cloudCredentials.project_id,
     credentials: fs.readFileSync(cloudCredentialsPath, "utf8"),
-    location
+    location,
+    ...(universeDomain ? { universeDomain } : {})
```
apilaskowski: Related correctness issue on this path: if a user supplies a TPC service-account key but leaves the new prompt blank, we write no `universeDomain`. BigQuery then defaults its endpoint to `googleapis.com` while google-auth derives the key's own universe from the JSON, so they disagree and the call fails at query time with a fairly opaque error.

Suggest falling back to the key file:

```suggestion
    ...(universeDomain || cloudCredentials.universe_domain
      ? { universeDomain: universeDomain || cloudCredentials.universe_domain }
      : {})
```

cli/api/dbadapters/bigquery.ts
```diff
@@ -52,7 +52,8 @@ export function createBigQueryClientProvider(
           projectId,
           scopes: EXTRA_GOOGLE_SCOPES,
           location: credentials.location,
-          credentials: credentials.credentials && JSON.parse(credentials.credentials)
+          credentials: credentials.credentials && JSON.parse(credentials.credentials),
+          ...(credentials.universeDomain ? { universeDomain: credentials.universeDomain } : {})
```
apilaskowski: Non-blocking: this guard is redundant here. `@google-cloud/bigquery@8.3.0` already does `if (options.universeDomain)` before using the value (`build/src/bigquery.js:120`), so an empty string is ignored. (The equivalent guard in `emitter.ts` *is* necessary, because gax uses `??` — worth noting the asymmetry in a comment there rather than mirroring the spread here.)

Reads more cleanly as a plain property:

```suggestion
          credentials: credentials.credentials && JSON.parse(credentials.credentials),
          universeDomain: credentials.universeDomain || undefined
```

cli/api/commands/credentials_test.ts
```diff
@@ -0,0 +1,47 @@
...
+  });
+
+  test("read leaves universeDomain unset when omitted", () => {
+    const credentialsPath = writeCredentials({ projectId: "my-project", location: "US" });
+
+    const credentials = read(credentialsPath);
+
+    expect(credentials.universeDomain).to.satisfy(
+      (value: string) => value === "" || value === undefined
+    );
```
apilaskowski: This assertion is ambiguous where the behaviour isn't. `read()` calls `verifyObjectMatchesProto`, which returns `protoType.create(object)` — a protobufjs message — so an unset `string` field is deterministically `""` via the prototype default. As written, the test would keep passing if that ever flipped to `undefined`, which is exactly the distinction the truthiness guards in `bigquery.ts` and `emitter.ts` rely on.

```suggestion
    expect(credentials.universeDomain).to.equal("");
```

cli/api/dbadapters/bigquery_test.ts
```diff
@@ -146,4 +146,26 @@ suite("BigQueryDbAdapter", () => {
 
     await adapter.setMetadata(action);
   });
+
+  suite("createBigQueryClientProvider", () => {
```
apilaskowski: Nit: this suite doesn't exercise `BigQueryDbAdapter`, so nesting it inside `suite("BigQueryDbAdapter", ...)` is a little misleading. Could it be a sibling top-level suite instead?

protos/profiles.proto
```diff
@@ -11,6 +11,10 @@ message BigQuery {
   string credentials = 3;
   // Options are listed here: https://cloud.google.com/bigquery/docs/locations
   string location = 4;
+  // The universe domain to connect to (e.g. "googleapis.com"). Leave unset to
```
apilaskowski: Nit: `(e.g. "googleapis.com")` sitting next to "leave unset to use the default" is a bit self-contradictory — `googleapis.com` is precisely the value one should never set here. Using a non-default universe as the example would make the intent clearer.

contributing.md
```diff
@@ -57,6 +57,7 @@ To run the CLI integration test against your own GCP project:
    - `projectId`: your GCP project id
    - `credentials`: the entire content of your GCP service account key JSON file as a single string (you can generate it with `jq -Rsa < path/to/key.json`).
    - `location`: location to use in your project
+   - `universeDomain` (optional): the universe domain to connect to (e.g. `googleapis.com`). Leave unset to use the default Google Default Universe (GDU). Set this only when targeting a non-default universe such as a Trusted Partner Cloud (TPC).
```
apilaskowski: Thanks for documenting it. This file covers the integration-test credentials only, though — the user-facing `.df-credentials.json` reference lives in the Cloud docs. Is a docs update tracked somewhere? Without one, the field is effectively undiscoverable for exactly the users who need it.

cli/api/lineage/emitter.ts
apilaskowski, line 46: This doesn't take effect: `LineageClient` resolves `servicePath = opts.servicePath || opts.apiEndpoint || 'datalineage.' + universeDomain`, and we always pass `apiEndpoint` from `LineageEndpointRouter`, which hardcodes `googleapis.com`. So lineage keeps targeting GDU hosts.

It's also a bit worse than a no-op: gax compares the configured universe against the credential's before each call, so setting it here makes that check pass while we carry on dialing a GDU endpoint — a loud failure becomes a silent one.

Could the endpoint router become universe-aware, or this hunk drop out for now? Either way it needs coverage; `emitter_test.ts` doesn't touch it.

cli/credentials.ts
apilaskowski, line 21: Could we avoid prompting every user for this? Nearly all of them are on GDU, and the proto change by itself already lets the rare TPC user set the field by hand in `.df-credentials.json`.

If it does stay interactive, it's currently untested — `cli/index_init_test.ts` only covers the failure path.

cli/credentials.ts
apilaskowski, line 50: If someone supplies a TPC service-account key but leaves the prompt blank, we write no `universeDomain`: BigQuery then defaults to `googleapis.com` while google-auth derives the key's own universe, and they disagree at query time. Worth falling back to `cloudCredentials.universe_domain` here.

cli/api/dbadapters/bigquery.ts
apilaskowski, line 56: Nit: redundant here — `@google-cloud/bigquery` already falsy-checks `options.universeDomain` (`bigquery.js:120`), so `universeDomain: credentials.universeDomain || undefined` would do.

The guard in `emitter.ts` is needed though, since gax uses `??` — might be worth a comment there noting the asymmetry.

cli/api/commands/credentials_test.ts
apilaskowski, line 39: `read()` returns a protobufjs message, so an unset string is deterministically `""` — this can just be `expect(credentials.universeDomain).to.equal("")`. As written it would still pass if the value became `undefined`, which is the distinction the guards in `bigquery.ts` and `emitter.ts` depend on.

cli/api/dbadapters/bigquery_test.ts
apilaskowski, line 150: Nit: this suite doesn't exercise `BigQueryDbAdapter` — could it be a sibling top-level suite rather than nested inside it?

protos/profiles.proto
apilaskowski, line 14: Nit: `(e.g. "googleapis.com")` next to "leave unset to use the default" is a little contradictory, since that's the one value you'd never set here. A non-default universe would be a clearer example.

contributing.md
apilaskowski, line 60: This covers the integration-test credentials only — is a matching update to the user-facing Cloud docs tracked anywhere?

apilaskowski (changes requested):

apilaskowski:
Thanks for this — the BigQuery half looks good.

Two things I'd like to sort out before this might be considered:

1. The lineage hunk doesn't actually take effect, and it suppresses gax's universe-mismatch check.
2. `init-creds` now asks every user a TPC question.

One question: `EXTRA_GOOGLE_SCOPES` hardcodes the Drive scope — is that grantable in TPC? And did your manual test cover the ADC path as well as the JSON-key one?

(FYI the CLA check is failing, and the title is still the generated branch name.)

cli/credentials.ts
ybelleguic: Take into account in `2962ae0a097bfef4fbe5a73fe39b77aa22519b54`

cli/credentials.ts
ybelleguic: take into account in `8bf34e9b90d4525151d2b567bcc658e468bed3b7`

cli/api/dbadapters/bigquery.ts
ybelleguic: Take into account in `d99767c7`

PR #2326, opened 2026-09-18
Remove rimraf

Ford544:
Remove rimraf dependency, which is only used once and can be easily replaced by Node's fs library.


apilaskowski (approved):
Perfect!

PR #2333, opened 2026-09-21
Pin uuid version to 1.14.2

Ford544:
This resolves the dependabot issue https://github.com/dataform-co/dataform/security/dependabot/269 by pinning the uuid version to 14.0.2 in package.json resolutions. This is a workaround necessary because currently the newest version of @google-cloud/storage is dependant on the vulnerable version of uuid - the resolution forces yarn to use the up-to-date version instead. Ideally, this workaround will be removed once the issue is fixed in @google-cloud/storage.


apilaskowski:
Do we have a TODO or other method to remember to properly fix once @google-cloud/storage is good to go?

apilaskowski:
Please put this into some TODO or internal tracker.

apilaskowski (approved):

PR #2339, opened 2026-09-23
Sqlx comment parsing backslashes

Ford544:
Escape backslashes in sql comments in sqlx files.
Currently, the compiler escapes backticks (\`) and dollar signs ($) in sql comments to prevent them from being interpreted as the end of string/JS block by the JS runtime. However, these symbols can still remain unescaped if they are preceded by a backslash in the user code, for example:
-- \\\`
after escaping becomes:
-- \\\\\`
the backslash is escaped and the backtick remains unescaped,  which is likely to trigger a SyntaxError
With this change, all backslashes are escaped, so the result would instead be:
-- \\\\\\\`
with the backtick still escaped.

Additionally, tests were added covering this behavior.


core/compilers.ts
```diff
@@ -68,7 +63,8 @@ export function compile(code: string, path: string): string {
 }
 
 export function extractJsBlocks(code: string): { sql: string; js: string } {
```
mbiernacik, line 65: This function will be removed in #2340

core/compilers.ts
```diff
@@ -11,16 +11,14 @@ const CONTEXT_FUNCTIONS = [
   "when",
   "incremental",
   "schema",
-  "database"
+  "database",
```
rafal-hawrylak: Please drop formatting commit. Added full repo reformat: https://github.com/dataform-co/dataform/pull/2342

core/main_sqlx_chars_test.ts
```diff
@@ -115,7 +115,7 @@ quotes
     writeDefinitionFile(
       projectDir,
       "file.sqlx",
-      `config { type: "table" }` + sqlContents + `post_operations { ${sqlContents} }`
+      `config { type: "table" }` + sqlContents + `post_operations { ${sqlContents} }`,
     );
```
rafal-hawrylak: New test only exercises SQL comments in a top-level table body. The escaper change also affects comments emitted inside pre_operations { ... } / post_operations { ... } blocks (all routes go through SQL_STATEMENT_ESCAPERS.get(SQL_COMMENT) via escapeNode). Extend the test with one variant that puts the same \`` / ${comments inside apre_operationsblock, mirroring the shape of the neighboringsqlContents + `pre_operations { ${sqlContents} }`` tests in this file.

core/compilers.ts
```diff
@@ -267,16 +262,16 @@ function createEscapedStatements(nodes: Array<string | SyntaxTreeNode>) {
 const SQL_STATEMENT_ESCAPERS = new Map([
   [
     SyntaxTreeNodeType.SQL_COMMENT,
-    (str: string) => str.replace(/\\/g, "\\\\").replace(/`/g, "\\`").replace(/\${/g, "\\${")
+    (str: string) => str.replace(/\\/g, "\\\\").replace(/`/g, "\\`").replace(/\${/g, "\\${"),
```
rafal-hawrylak: I know this is not part of this change but I think we miss a test here. The chained .replace(/\\/g, "\\\\").replace(/\/g, "\").replace(/\${/g, "\\${") is correct only because backslash-doubling runs first (otherwise the \\\`` and \${produced by the later.replaces would themselves get double-escaped on a re-pass). Add a test input that contains a literal `inside the comment (e.g.-- foo \` bar) - the current test covers \${and bare `` but not `\\\ at the comment boundary, which is exactly the escape-order failure mode.

rafal-hawrylak (changes requested):

core/compilers.ts
Ford544: I might be misunderstanding, but the "backticks and dollar signs are escaped in sql comments" already includes the line

> -- Comment with backticks \`

which would fail (the backtick would be double escaped and thus left unescaped) if the order was reversed as you describe

(the actual line in the code has a backslash, but it's only because it's in a backtick-enclosed JS script, so the end result is a literal backtick as presented above)

core/compilers.ts
Ford544: Done. The build checks are going to be failing now; I will rebase and fix once https://github.com/dataform-co/dataform/pull/2342 is merged

core/main_sqlx_chars_test.ts
Ford544: Added a pre_ops check to the test

rafal-hawrylak (approved):

PR #2341, opened 2026-09-23
Remove unused git workflow

Ford544:
Remove the unused test github workflow. 
The workflow is manually disabled and has never been run according to logs; it's function is fullfilled by the CloudBuild-side workflows and the cloudbuild-test.yaml.
Removing it resolves the Code Scanning issue https://github.com/dataform-co/dataform/security/code-scanning/25


Ford544:
GraphQL check fails because it expects HitHubActions code but doesn't find any. Seems like resolving this failure will require changing GraphQL configuration in the repo settings.

rafal-hawrylak (approved):

